Issue #14, opened 2014-01-23
add ability to disable build make relative path


peter1000:
sometimes a package builder might want to set no rpath or a own one for internal usage.

it would be nice if one could have an option to disable the mk_relative rpath part.

build example something like
```
#CHECK: BUILD_NO_RPATH
if [ "${BUILD_NO_RPATH}" != "y" ]; then
    echo 'this recipes needs conda config: build_no_rpath: True'
    exit 1
fi

unset LD_RUN_PATH
./bootstrap.sh
LDFLAGS="-Wl,-rpath,${PREFIX}/usr" \
./configure --prefix=${PREFIX}
make
make install
```

peter1000:
closing my old issues - at the moment I do not use conda

PR #15, opened 2014-01-23
* add ability to disable build make relative path

peter1000:
https://github.com/conda/conda-build/issues/14

sometimes a package builder might want to set no rpath or a own one for internal usage.

it would be nice if one could have an option to disable the mk_relative rpath part.

build example something like
```
#CHECK: BUILD_NO_RPATH
if [ "${BUILD_NO_RPATH}" != "y" ]; then
    echo 'this recipes needs conda config: build_no_rpath: True'
    exit 1
fi

unset LD_RUN_PATH
./bootstrap.sh
LDFLAGS="-Wl,-rpath,${PREFIX}/usr" \
./configure --prefix=${PREFIX}
make
make install
```

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -3,3 +3,4 @@
   * initial release
   * includes conda-build, conda-convert, conda-index, conda-skeleton
   * depends on new conda version 3
+  * add ability to disable build make relative path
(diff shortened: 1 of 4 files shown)


peter1000:
closing my old issues - at the moment I do not use conda

PR #19, opened 2014-01-23
* add ability to install into a separate clean installdir instead of _build PREFIX

peter1000:
* add ability to install into a separate clean installdir instead of
    _build PREFIX where all the build required files are located.

as mentioned some times ago if one compile a package which requires itself (e.g. make, gcc, glibc, ect..) one ends up with an empty r nearly empty final conda package with only the info folder.

This is also understandable as files are only selected on a diff from the _build PREFIX folder.

In my opinion there are 3 options:
- leave it as it is and say that packages like make, gcc, glibc ect are not really the focus
- a better solution but not really fine: do a diff with time stamp check: so if a file gets installed into PREFIX it can still be included in the final conda pkg
- use a separate install folder: which I propose as a subfolder of PREFIX 

<i>e.g. usage: </i>
```
./configure --prefix=${INSTALLDIR}
make
make install
```
<i>or: </i>
```
/configure --prefix=whatever
make
make DESTDIR=${INSTALLDIR} install
```

<i>e.g. usage for python apps: </i>
```
    $PYTHON setup.py install --prefix=./ --root=${INSTALLDIR} || exit 1
```

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -3,3 +3,8 @@
   * initial release
   * includes conda-build, conda-convert, conda-index, conda-skeleton
   * depends on new conda version 3
+  * add ability to install into a separate clean installdir instead of
+    _build PREFIX
+    e.g. usage: ./configure --prefix=${INSTALLDIR}
+    for python apps:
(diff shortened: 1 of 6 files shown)


peter1000:
closing my old issues - at the moment I do not use conda

PR #42, opened 2014-02-14
update MACOSX_DEPLOYMENT_TARGET to 10.6

minrk:
10.5 can cause build problems, as the SDK is rarely present.

### DIFF
--- a/conda_build/environ.py
+++ b/conda_build/environ.py
@@ -53,7 +53,7 @@ def get_dict(m=None, prefix=build_prefix):
         d['CFLAGS'] = '-arch %(OSX_ARCH)s' % d
         d['CXXFLAGS'] = d['CFLAGS']
         d['LDFLAGS'] = d['CFLAGS']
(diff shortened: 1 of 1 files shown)


ilanschnell:
I have no problems merging this PR.

minrk:
Thanks!

minrk:
I guess conda-build isn't used to build Python, then? I see that the latest Python 3.5 package is still targeting 10.5, which is causing various headaches.

ccordoba12:
Not Python (yet). But we're in the process of moving to use only conda-build.

Pinging @ilanschnell so he's aware of this.

ilanschnell:
Yes, I'm aware of this.  Python 3.5 is still being build on Mac OS X 10.5 using your internal system (which is similar to conda-build).  I tried building it on 10.7 using conda-build, but ran into various little issues.  Therefore I decided it would be safest to build Python 3.5 like all other Python's before.  As Carlos said, we are in the process of moving your recipes to conda-build recipes.

PR #223, opened 2014-10-06
Trying to get the git describe build string to include the python version

asmeurer:
This doesn't work because this information is not known when the templating is
run.

### DIFF
--- a/conda_build/environ.py
+++ b/conda_build/environ.py
@@ -59,11 +59,12 @@ def get_git_build_info(m, src_dir):
     output = output.decode('utf-8')
     d['GIT_FULL_HASH'] = output
     # set up the build string
-    if not m:
-        build_id = d[key_name("NUMBER")]
-    else:
-        build_id = m.default_build_id(int(d[key_name("NUMBER")]))
     if key_name('NUMBER') in d and key_name('HASH') in d:
+        if not m:
+            build_id = d[key_name("NUMBER")]
+        else:
+            build_id = m.default_build_id(int(d[key_name("NUMBER")]))
+
         d['GIT_BUILD_STR'] = '{}_{}'.format(d[key_name('HASH')],
                                             build_id)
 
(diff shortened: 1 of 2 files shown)


dan-blanchard:
I've been running into this exact same issue. I'm glad to see you guys are working on it.

conda_build/environ.py
```diff
@@ -84,6 +88,9 @@ def get_dict(m=None, prefix=None):
     d['NPY_VER'] = get_npy_ver()
     d['SRC_DIR'] = source.get_dir()
 
+    if os.path.isdir(os.path.join(d['SRC_DIR'], '.git')):
+        d.update(**get_git_build_info(m, d['SRC_DIR']))
```
asmeurer: I guess this was an incorrect merge conflict resolution.

asmeurer:
This doesn't work.

I give up trying to fix the code flow to make it work. The Jinja templating needs to have access to the rest ot the Metadata object metadata to correctly add it to the GIT_BUILD_STR, but when it is run it doesn't (and can't, because that information can't even be read until after the meta.yaml is templated). 

The jinja2 templating is broken by design, and I honestly think we should just get rid of it, and replace it with some kind of variable system where the meta.yaml can be parsed and the variables replaced at a later time (possibly many times as more information is known throughout the build process).

ilanschnell:
I agree, we should get rid of the jinja2 templating.  It has been a source of confusion for a long time.  Maybe we can have special variables (such as `__GIT_BUILD_STR__`) in the meta.yaml file.

dan-blanchard:
Not that anybody asked for my opinion, but I completely agree.  The jinja2 system is really confusing for beginners, as jinja2 isn't even a requirement for conda-build.  I will say that if you do this though, it'd be nice if there was a special variable for the current date (formatted as a version like `2015.03.10`).  That's the only thing I've ever used the templating for other than `GIT_BUILD_STR`.

PR #226, opened 2014-10-17
Allow conda build to be run without arguments

srossross:
`conda build` == `conda build .`

I think this is a good default, like make and Makefile

cc @asmeurer @joelhullcio


tswicegood:
This has been addressed in #327.  Closing.

Issue #119, opened 2014-06-04
patch error with skeleton pypi on Windows 7 64-bit


asmeurer:
Reported on the conda mailing list. 

```

C:\gca\temp>conda skeleton pypi pymarc
Using url https://pypi.python.org/packages/source/p/pymarc/pymarc-3.0.1.tar.gz (131783) for pymarc.
Downloading pymarc (use --no-download to skip this step)
Unpacking pymarc...
done
working in c:\users\gaborca\appdata\local\temp\tmpvjyhclconda_skeleton_pymarc
Fetching package metadata: ....
Solving package specifications: .
Applying patch: u'c:\\users\\gaborca\\appdata\\local\\temp\\tmpvjyhclconda_skeleton_pymarc\\pypi-distutils.patch'
69 unparsed bytes left at the end of stream
file 1/1:        core.py
 hunk no.1 doesn't match source file at line 165
  expected:
  actual  :             if DEBUG:
source file is different - core.py
Traceback (most recent call last):
  File "C:\gca\program_files\Miniconda\Scripts\conda-skeleton-script.py", line 4, in <module>
    sys.exit(main())
  File "C:\gca\program_files\Miniconda\lib\site-packages\conda_build\main_skeleton.py", line 124, in main
    args.func(args, p)
  File "C:\gca\program_files\Miniconda\lib\site-packages\conda_build\main_skeleton.py", line 135, in execute
    pypi.main(args, parser)
  File "C:\gca\program_files\Miniconda\lib\site-packages\conda_build\pypi.py", line 314, in main
    run_setuppy(src_dir, tempdir, args)
  File "C:\gca\program_files\Miniconda\lib\site-packages\conda_build\pypi.py", line 485, in run_setuppy
    apply_patch(join(stdlib_dir, 'distutils'), patch)
  File "C:\gca\program_files\Miniconda\lib\site-packages\conda_build\source.py", line 200, in apply_patch
    check_call([patch, '-p0', '-i', path], cwd=src_dir)
  File "C:\gca\program_files\Miniconda\lib\subprocess.py", line 540, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '[u'C:\\gca\\program_files\\Miniconda\\Scripts\\patch.bat', u'-p0', u'-i', u'c:\\users\\gaborca\\appdata\\local\\temp\\tm
pvjyhclconda_skeleton_pymarc\\pypi-distutils.patch']' returned non-zero exit status -1

```

[...]

PR #236, opened 2014-11-05
Changed line numbers in patch of distutils

mcrot:
Currently on Windows I have the problem that 
```
  conda skeleton pypi <package name>
```
fails because there is a patch of disutils which cannot be applied. See #119.
It seems that adjusting the line specifications in the patch code fixes
this issue for me. After the modifcations the distuils patch can still be applied by Linux' `patch` tool.
What I don't understand is why the previous lines numbers worked with GNU patch.


msarahan:
Closing.  Patch on Windows is GnuWin32, patch version 2.5.9, and supports fuzz.

PR #246, opened 2014-11-13
Added git functions

srossross:
Hey, this POC of how to call `git_describe_tag` as a function so you don't need `source.gut_url` in your meta.yaml

This needs to be cleaned up but, I tested this on my mac with the following meta.yaml

```yaml
package:
  name: conda_test_package
  version: {{git_describe_tag(RECIPE_DIR)}}

build:
  number: 1

```

cc @tswicegood  @asmeurer

### DIFF
--- a/conda_build/jinja_context.py
+++ b/conda_build/jinja_context.py
@@ -9,8 +9,8 @@ import json
 import os
 
 from conda.compat import PY3
-from conda_build import environ
 from .environ import get_dict as get_environ
+from subprocess import check_output, CalledProcessError
 
 _setuptools_data = None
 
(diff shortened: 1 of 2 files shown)


jasongrout:
It seems that the recommended way to do this now is to use `environ.get('GIT_DESCRIBE_TAG')`, right?  Of course, it doesn't seem to work now (see #357), but it's the solution in the docs.  Should this issue be closed?

msarahan:
Current practice is just variable names.  Docs are at http://conda.pydata.org/docs/building/environment-vars.html, code is at https://github.com/conda/conda-build/blob/master/conda_build/environ.py#L104 and https://github.com/conda/conda-build/blob/master/conda_build/environ.py#L247 - works with both source:git_url and source:path.

Closing this issue - please reopen if it needs further discussion.

Issue #247, opened 2014-11-13
jinja2 errors not caught properly on Windows 32


tswicegood:
Full traceback: https://gist.github.com/tswicegood/0dcdeb2313748435d53f

It should display the error and stop.

Issue #137, opened 2014-07-17
Jinja2 missing requirement shouldn't show stack trace


bkreider:
It's very nice that conda-build catches the fact that this error is caused by not having Jinja2 installed, but it shouldn't also show the massive stacktrace.  (The long path is for the binary path re-writing)

```
bradleys-mbp-2:go bkreider$ conda build conda.recipe/
There was an error importing jinja2.
Please run `conda install jinja2` to enable jinja template support
Traceback (most recent call last):
  File "/opt/mincondamincondamincondamincondamincondaminicondaminicondaminicondaminicondaminiconda/bin/conda-build", line 5, in <module>
    sys.exit(main())
  File "/opt/mincondamincondamincondamincondamincondaminicondaminicondaminicondaminicondaminiconda/lib/python3.4/site-packages/conda_build/main_build.py", line 93, in main
    args.func(args, p)
  File "/opt/mincondamincondamincondamincondamincondaminicondaminicondaminicondaminicondaminiconda/lib/python3.4/site-packages/conda_build/main_build.py", line 196, in execute
    m = MetaData(recipe_dir)
  File "/opt/mincondamincondamincondamincondamincondaminicondaminicondaminicondaminicondaminiconda/lib/python3.4/site-packages/conda_build/metadata.py", line 187, in __init__
    self.meta = parse(get_contents(self.meta_path))
  File "/opt/mincondamincondamincondamincondamincondaminicondaminicondaminicondaminicondaminiconda/lib/python3.4/site-packages/conda_build/metadata.py", line 95, in parse
    res = yamlize(data)
  File "/opt/mincondamincondamincondamincondamincondaminicondaminicondaminicondaminicondaminiconda/lib/python3.4/site-packages/conda/utils.py", line 77, in __call__
    value = self.func(*args, **kw)
  File "/opt/mincondamincondamincondamincondamincondaminicondaminicondaminicondaminicondaminiconda/lib/python3.4/site-packages/conda_build/metadata.py", line 90, in yamlize
    return yaml.load(data)
  File "/opt/mincondamincondamincondamincondamincondaminicondaminicondaminicondaminicondaminiconda/lib/python3.4/site-packages/yaml/__init__.py", line 72, in load
    return loader.get_single_data()
  File "/opt/mincondamincondamincondamincondamincondaminicondaminicondaminicondaminicondaminiconda/lib/python3.4/site-packages/yaml/constructor.py", line 37, in get_single_data
    return self.construct_document(node)
[...]

PR #248, opened 2014-11-13
Halt build on YAML error without jinja2

tswicegood:
This fixes the issue described in both #247 and #137 by detecting parse errors related to "{{" inside a YAML file.  The major reason for those to be there is when doing template work.

### DIFF
--- a/conda_build/exceptions.py
+++ b/conda_build/exceptions.py
@@ -1,0 +1,46 @@
+import textwrap
+SEPARATOR = "-" * 70
+
+indent = lambda s: textwrap.fill(textwrap.dedent(s))
+
+
+class CondaBuildException(Exception):
+    pass
+
+
+class YamlParsingError(CondaBuildException):
+    pass
+
+
+class UnableToParse(YamlParsingError):
+    def __init__(self, original, *args, **kwargs):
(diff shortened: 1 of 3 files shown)


bkreider:
What's the status of this PR?  This solves two github issues that have already been closed.

conda_build/exceptions.py
```diff
@@ -0,0 +1,46 @@
+import textwrap
+SEPARATOR = "-" * 70
+
+indent = lambda s: textwrap.fill(textwrap.dedent(s))
+
+
+class CondaBuildException(Exception):
```
asmeurer: Conda catches RuntimeError and converts it into sys.exit. So any error that will be raised to the user that doesn't indicate a bug should subclass from RuntimeError.

tswicegood: @asmeurer these aren't caught by the user, it's handled inside `conda_build.main_build.main` (see the diff below).  FWIW, I thought this has been merged and released already.  I'm going to go ahead and merge it since changing from `Exception` to `RuntimeException` isn't a backwards incompatible change.  I'm -0 on the change, but if you want to do it, go for it.

I'd love to get this released.  Having a friendly error message instead of a stack trace would be a pretty big improvement.

bkreider:
Thanks!

Issue #259, opened 2014-12-01
Long files names in source tarballs/zipfiles cause issues on Windows


mwcraig:
In trying to build a conda recipe this weekend I ran into an issue with unpacking the tarball for boost on Windows. Some of the paths are longer than the default maximum number of characters [allowed by the windows API, 260 characters](http://msdn.microsoft.com/en-us/library/aa365247%28VS.85%29.aspx#maxpath).

While this isn't a conda-build issue per se it would be nice if it used the mechanism described at the link above (adding `\\?\` to the beginning of an absolute path) to work around this limit.

A PR to implement this is coming momentarily...

mwcraig:
Closing, see: https://github.com/conda/conda-build/pull/260#issuecomment-232545662

PR #260, opened 2014-12-01
Use long path names on windows

mwcraig:
This PR addresses #259 by appending the prefix needed on windows to allow paths with up to 32,767 characters.

### DIFF
--- a/conda_build/source.py
+++ b/conda_build/source.py
@@ -78,11 +78,15 @@ def unpack(meta, verbose=False):
     os.makedirs(WORK_DIR)
     if verbose:
         print("Extracting download")
+    if sys.platform.startswith('win'):
+        long_path_names = '\\\\?\\'
+    else:
+        long_path_names = ''
     if src_path.lower().endswith(('.tar.gz', '.tar.bz2', '.tgz', '.tar.xz',
-            '.tar', 'tar.z')):
-        tar_xf(src_path, WORK_DIR)
+                '.tar', 'tar.z')):
+        tar_xf(src_path, long_path_names + WORK_DIR)
     elif src_path.lower().endswith('.zip'):
-        unzip(src_path, WORK_DIR)
+        unzip(src_path, long_path_names + WORK_DIR)
     else:
         # In this case, the build script will need to deal with unpacking the source
         print("Warning: Unrecognized source format. Source file will be copied to the SRC_DIR")


msarahan:
@mwcraig any interest in revisiting this?  It seems like we might need to add this modification in a few more places, but it's definitely an improvement.

mwcraig:
Sure, can get it rebased Monday and also look for other places to make the change, if you want.

msarahan:
Sounds great.  I don't know all the places to look, unfortunately - anywhere files are being accessed, copied, linked, or removed.  I bet it will take us a few iterations to get them all, but this is going in the right direction.

msarahan:
See also https://github.com/conda/conda-build/issues/891 and https://github.com/conda/conda/pull/2452

mwcraig:
Thanks, looks like I won't get to this until tomorrow...

msarahan:
No problem.  This will be good whenever it happens.

mwcraig:
First rebase introduced a typo...fixed and squashed now.

mcg1969:
@mwcraig, I don't know what the status of this PR is, but note that we are experiencing some subtle and yet fatal issues with the use of UNC paths; see https://github.com/conda/conda/issues/2846

I don't know if this will affect conda-build's use of same, but we will need to be careful.

mwcraig:
@mcg1969, I haven't progressed much beyond compiling a list of places it looks likes paths would have to be hacked (screenshot of the browser tabs below :)). 


conda-build_metadata_py_at_e8e83de1eadb5a6f479fd2a716662d4df931e1c4_ _conda_conda-build

I did a tiny bit of reading to learn more about fixes of this issue at a lower level in python and came across this [issue on the python bug tracker](https://bugs.python.org/issue18199), which has made me a bit nervous about changing this broadly. 

At this point I'd lean towards changing this only in very limited circumstances. The relative scarcity of bug reports related to this seems to indicate it comes up infrequently. In this PR the change is made only when unpacking source tarballs, which shouldn't have a terrible effect outside of that.

mwcraig:
@msarahan -- looking back over this, I'm inclined to close without merging. I ran into the problem with boost, probably trying to build in a local VM.

boost is building fine on conda-forge + appveyor (whether due to a change in the boost package or something in conda build) so the original issue (#259) is moot.

Issue #266, opened 2014-12-05
conda-build failing to report git describe info to templated yaml file


ddale:
I'm working on a project that keeps its conda recipe in PROJROOT/conda.recipe. In `meta.yaml` file, I have:

```
package:
  name: hexrd
  version: {{ environ['GIT_DESCRIBE_TAG'][1:]+environ.get('GIT_DESCRIBE_NUMBER', '') }}

source:
  path: ../

```

When I try to build it, I get:

```
$ conda build conda.recipe/
An unexpected error has occurred, please consider sending the
following traceback to the conda GitHub issue tracker at:

    https://github.com/conda/conda-build/issues

Include the output of the command 'conda info' in your report.


Traceback (most recent call last):
  File "/opt/conda/bin/conda-build", line 5, in <module>
    sys.exit(main())
  File "/opt/conda/lib/python2.7/site-packages/conda_build/main_build.py", line 110, in main
    args_func(args, p)
  File "/opt/conda/lib/python2.7/site-packages/conda_build/main_build.py", line 312, in args_func
    args.func(args, p)
  File "/opt/conda/lib/python2.7/site-packages/conda_build/main_build.py", line 244, in execute
    m = MetaData(recipe_dir)
  File "/opt/conda/lib/python2.7/site-packages/conda_build/metadata.py", line 250, in __init__
    self.parse_again()
  File "/opt/conda/lib/python2.7/site-packages/conda_build/metadata.py", line 258, in parse_again
    self.meta = parse(get_contents(self.meta_path))
  File "/opt/conda/lib/python2.7/site-packages/conda_build/metadata.py", line 234, in get_contents
    contents = template.render(environment=env)
  File "/opt/conda/lib/python2.7/site-packages/jinja2/environment.py", line 969, in render
    return self.environment.handle_exception(exc_info, True)
  File "/opt/conda/lib/python2.7/site-packages/jinja2/environment.py", line 742, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/home/darren/Projects/hexrd/conda.recipe/meta.yaml", line 4, in top-level template code
    version: {{ environ['GIT_DESCRIBE_TAG'][1:]+environ.get('GIT_DESCRIBE_NUMBER', '') }}
jinja2.exceptions.UndefinedError: 'dict object' has no attribute 'GIT_DESCRIBE_TAG'
```

[...]

PR #268, opened 2014-12-07
report empty GIT_DESCRIBE env variables, fixes issue #266

ddale:
A `meta.yaml` file templated with `GIT_DESCRIBE` fails to parse the first time through, since the source is not available the first time it is parsed, and conda-build skips creating the environment variables if it doesn't find a `.git` directory in `SRC_DIR`. This pull request creates empty GIT_ environment variables if `.git` is not available, allowing `meta.yaml` to parse so the source can be provided for the second round of parsing (which already exists in conda-build).

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -312,7 +312,7 @@ def build(m, get_src=True, verbose=True, post=None):
 
         # Display the name only
         # Version number could be missing due to dependency on source info.
-        print("BUILD START:", m.dist())
+        print("BUILD START:", m.name())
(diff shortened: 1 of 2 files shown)


tswicegood:
Just getting a chance to dig into this.  These changes look reasonable, @ddale.  Would you mind merging `master` in kick off a build with the latest code so we can make sure everything passes?

tswicegood:
Actually -- I just opened up #305 to kick off the tests.  Assuming that passes I'll merge that.  Thanks, @ddale!

PR #275, opened 2014-12-12
scrubs all build prefixes

ddale:
This pull request scrubs unix-style build prefixes in windows conda packages.

It also explicitly lists all files containing a build prefix placeholder, and does not use the `/opt/anaconda1anaconda2anaconda3` default placeholder. The handling of `/opt/anaconda1anaconda2anaconda3` is somewhat anomolous, since it is not explicitly recorded in `has_prefix`, and it is a unix-style path that gets replaced with a windows style path at installation time.

I don't see any backward compatibility consequences for not producing packages with the old default placholder.

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -85,6 +85,7 @@ def have_prefix_files(files):
     alt_prefix = prefix.replace('\\', '/')
     alt_prefix_bytes = alt_prefix.encode('utf-8')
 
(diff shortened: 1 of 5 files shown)


msarahan:
@ddale I'm not sure this is still necessary.  If you know otherwise, do you have any interest in revisiting this?

Issue #270, opened 2014-12-10
conda-build ignores script tag in meta.yaml


meawoppl:
On my windows7 machine conda-build does not seem to acknowledge the existence of

``` yaml
build:
  script: python setup.py install
```

This is a bit confusing as it is used specifically in most simple examples.  

Workaround is to make "build.sh" and "bld.bat" files.  A follow-up question is there a reason that this name dissimilarity exists?

PR #286, opened 2014-12-25
support build:script for win32, fixes issue #270

ddale:
`build:script` defined in `meta.yaml` is currently only respected for non-win32. This pull request provides a fix.

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -347,17 +347,17 @@ def build(m, get_src=True, verbose=True, post=None):
             f.write(u'\n'.join(sorted(list(files1))))
             f.write(u'\n')
 
+        script = m.get_value('build/script', None)
+        if isinstance(script, list):
+            script = '\n'.join(script)
         if sys.platform == 'win32':
             import conda_build.windows as windows
-            windows.build(m)
+            windows.build(m, script=script)
         else:
             env = environ.get_dict(m)
             build_file = join(m.path, 'build.sh')
 
-            script = m.get_value('build/script', None)
             if script:
-                if isinstance(script, list):
-                    script = '\n'.join(script)
                 with open(build_file, 'w') as bf:
                     bf.write(script)
                 os.chmod(build_file, 0o766)
--- a/conda_build/windows.py
+++ b/conda_build/windows.py
@@ -90,7 +90,7 @@ def kill_processes():
             continue
 
 
-def build(m):
+def build(m, script=None):
     env = dict(os.environ)
     env.update(environ.get_dict(m))
 
@@ -101,6 +101,9 @@ def build(m):
 
     src_dir = source.get_dir()
     bld_bat = join(m.path, 'bld.bat')
+    if script:
+        with open(bld_bat, 'w') as bf:
+            bf.write(script)
     if exists(bld_bat):
         with open(bld_bat) as fi:
             data = fi.read()


asmeurer:
The binstar Windows build passed but I'm not sure if this was even tested.

conda_build/build.py
```diff
@@ -347,17 +347,17 @@ def build(m, get_src=True, verbose=True, post=None):
             f.write(u'\n'.join(sorted(list(files1))))
             f.write(u'\n')
 
+        script = m.get_value('build/script', None)
+        if isinstance(script, list):
+            script = '\n'.join(script)
```
srossross: In the win32 case of a list of commands. it may be good to do an error check for each line e.g.

```py
line = "(%s) || (exit /b 1)" % line
```

conda_build/build.py
```diff
@@ -347,17 +347,17 @@ def build(m, get_src=True, verbose=True, post=None):
             f.write(u'\n'.join(sorted(list(files1))))
             f.write(u'\n')
 
+        script = m.get_value('build/script', None)
```
srossross: Also, maybe win32 build should try to find a `"build/batch"` entry first.

conda_build/build.py
asmeurer: What does the /b do? I've never seen that.

srossross:
```
    /B        When used in a batch script, this option will exit 
              only the script (or subroutine) but not CMD.EXE
```

asmeurer: I don't really understand what the difference between using /b and not using /b would be here. I do agree it would be good to add `|| exit 1` here.

meawoppl:
This PR deserves some attention.  The CI error appears unrelated to any actual problems with the PR, and should be rerun.

tswicegood:
This PR seems reasonable to me.  @ddale can you merge `master` in to this to make sure the tests pass appropriately on TravisCI now?

Also, it would be really helpful if a test example was created.  We have windows tests running on binstar-build now (see https://github.com/conda/conda-build/blob/master/.binstar.yml).  That'll make it a lot easier to get merged as we'll have something reproducible to verify everything's executing correctly.

Thanks!!

asmeurer:
Travis tests are mostly irrelevant for this PR as it only affects Windows.

asmeurer:
Regarding the Windows builds, you have to push the commit up to this repo for it to be tested (I must have done that at some point since it's showing up in the GitHub CI list).

ddale:
Are there any remaining issues with this PR that need to be addressed?

asmeurer:
I suppose it would be a good idea to implement the `if errorlevel 1 exit 1` suggestion.

ovz:
Is this pull request on track for a merge?

asmeurer:
It looks like there are merge conflicts.

mcg1969:
Any word on the status of this, and when it will be updated on anaconda.org?

meawoppl:
Also issue #270

patricksnape:
There appears to be some logic [here](https://github.com/conda/conda-build/blob/fa67dac2dc6fec614bf5f51a435c57959fc2c657/conda_build/build.py#L472) that copes with this. Also, this appears to handle the same issue as #620.

msarahan:
fixed by https://github.com/conda/conda-build/pull/799 - closing.

PR #287, opened 2014-12-25
do not attempt to import os.readlink on win32

ddale:
Fixes a minor bug recently introduced on master that breaks conda-build on win32


asmeurer:
This was fixed by the very similar https://github.com/conda/conda-build/pull/298

PR #291, opened 2015-01-09
Fix for building dependencies

rmcgibbo:
[Here](https://github.com/omnia-md/conda-recipes/issues/82), we observed that when building a large collection of packages, some of the packages did not build correctly. In the log file, I see things like

```
BUILD START: mixtape-0.2.2-np18py27_1
Fetching package metadata: ...
Solving package specifications: 
Error: Could not find some dependencies for mdtraj: scripttest
```

The issue is that "mixtape" has a build-time dependency on "mdtraj", and "mdtraj" has a _runtime_ dependency on "scripttest", but _not_ a build time dependency on "scripttest". If the "scriptest" package does not exist yet, then This results in [conda/resolve.py](https://github.com/conda/conda/blob/master/conda/resolve.py#L311-L313) throwing the observed `Error: Could not find some dependencies for mdtraj: scripttest`.

If possible, `conda-build` should catch this error and try to build the "scriptest" package.

### DIFF
--- a/conda_build/main_build.py
+++ b/conda_build/main_build.py
@@ -272,7 +272,7 @@ def execute(args, parser):
                     build.build(m, verbose=not args.quiet, post=post)
                 except RuntimeError as e:
                     error_str = str(e)
-                    if error_str.startswith('No packages found'):
+                    if error_str.startswith('No packages found') or error_str.startswith('Could not find some'):
                         # Build dependency if recipe exists
                         dep_pkg = error_str.split(': ')[1]
                         # Handle package names that contain version deps.


rmcgibbo:
There is actually another way that this can break, which is harder to fix. It's described [here](https://github.com/omnia-md/conda-recipes/issues/90), but I figured I should link it here too.

Assume package A depends on package B, and both depend on python. Assume that a version of package B built for py27 exists in `~/miniconda/conda-bld/<platform>/`, and we run `CONDA_PY=33 conda build A`.

The desired behavior is that it builds a py33 version of package B. But actually it fails with the current version of `conda-build`.

The reason for the failure is that [this line](https://github.com/conda/conda-build/blob/master/conda_build/main_build.py#L272) gets an error from the SAT solver, and throws a SystemExit. The error is not a `RuntimeError`, nor does it start with `"No packages found"`, so the branch that exists to invoke `conda build A` never runns.

asmeurer:
Looks good to me. The other issue is known. conda build should probably just parse the hint for the common case of missing a package for a specific Python version. I hit that all the time.

asmeurer:
We can also change it to be `RuntimeError` in conda. The two are effectively equivalent because there is a global `except RuntimeError` that converts them to `SystemExit`. Or we can `except (RuntimeError, SystemExit)` here.

rmcgibbo:
> The other issue is known. conda build should probably just parse the hint for the common case of missing a package for a specific Python version. I hit that all the time.

My use case involves building a whole bunch of conda recipes at once. For this, the most straightforward thing is to do a topological sort on the recipes, and build them in order that way, instead of explicitly trapping unsatisfied dependency-related failures. Is that something you want to do in conda-build?

asmeurer:
What it does now is effectively a topological sort, except it builds out the graph as it goes.

PR #292, opened 2015-01-10
Allow specifying files to always include

tswicegood:
This removes files from the list list of initial files so they are included.  It's required for building things like conda that are going to replace the `$PREFIX/bin/conda` file inside the build environment with their current version.

See also: https://github.com/conda/conda/pull/1086


tswicegood:
This is required to use `conda build` to build `conda` and also opens up the possibility of using `conda build` to build `conda-build`.

PR #293, opened 2015-01-10
Add recipe for building conda-build with conda-build

tswicegood:
Looking to automate this process to build conda-build on binstar so internal users can subscribe to a dev channel and work with early releases before actual releases occur.


tswicegood:
Ok -- I've tested this build on Ubuntu on a 64 bit machine.  Going to bump to 1.10alpha.0 and merge to see if the binstar auto-build is working properly.

PR #299, opened 2015-01-14
Distinguish a binary file detected with a prefix

flub:
Before if detect_binary_files_with_prefix was used all binary files
would be shown as being detected as text files.  This is more
accurate and shows better what conda will be doing to the files.


asmeurer:
Ah, I thought there was something wrong with the output.

PR #301, opened 2015-01-14
Added path_ignore to meta.yaml for copying source files

bchapman:
Allows you to ignore a list of files/folders that should not be copied from the source path.

Ex meta.yaml

```
path: ../
path_ignore: ../build
```

You can also use a list of patterns to exclude:

```
path_ignore:
    - ../build
    - ../dist
    - ../extra*
```

I'm currently working in an environment where my conda recipes are stored in the same repo as my source code.
With this, I want to be able to build my conda package in this same folder, just like python setuptools does.
This creates a build/conda folder in the root of my repo.
Without the ability to ignore this build folder, when conda copies the source files, it will include the build folder, creating a recursive loop.

### DIFF
--- a/conda_build/metadata.py
+++ b/conda_build/metadata.py
@@ -100,6 +100,7 @@ def parse(data):
             raise RuntimeError("The %s field should be a dict, not %s" % (field, res[field].__class__.__name__))
     # ensure those are lists
     for field in ('source/patches',
+                  'source/path_ignore',
                   'build/entry_points',
                   'build/features', 'build/track_features',
                   'requirements/build', 'requirements/run',
@@ -178,7 +179,8 @@ def _git_clean(source_meta):
 # conda-docs/docs/source/build.rst
 FIELDS = {
     'package': ['name', 'version'],
-    'source': ['fn', 'url', 'md5', 'sha1', 'sha256', 'path',
+    'source': ['fn', 'url', 'md5', 'sha1', 'sha256',
+               'path', 'path_ignore',
                'git_url', 'git_tag', 'git_branch', 'git_rev',
                'hg_url', 'hg_tag',
                'svn_url', 'svn_rev', 'svn_ignore_externals',
--- a/conda_build/source.py
+++ b/conda_build/source.py
@@ -254,8 +254,6 @@ def provide(recipe_dir, meta, patch=True):
                         result.append(fn)
             return result
 
-        print("ignorePatterns: {0}".format(ignorePatterns)) # TESTING
-
         print("Copying %s to %s" % (abspath(join(recipe_dir, meta.get('path'))), WORK_DIR))
         copytree(abspath(join(recipe_dir, meta.get('path'))), WORK_DIR, ignore=ignore)
     else: # no source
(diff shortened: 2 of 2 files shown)


asmeurer:
I recommend using `git_url:../`, which will do a local clone of the repo, preventing you from getting anything that isn't tracked by git.

bchapman:
So I've been trying `git_url`, and it works great as far as only getting stuff tracked by git.
However, having to have a git commit for every minor dev change is becoming too much work.

Every published conda package should no doubt have an associated git commit, but when you are just developing the package, you shouldn't have to create a new commit for every change.

Python packages aren't a huge deal, because you can test most everything with just setuptools, then conda is just a small layer above setuptools.
However, when I'm building a tool that doesn't use setuptools or relies heavily on a long build, pre-link, post-link workflow with build dependencies stored in a conda-recipe, it's a lot of overhead to commit every change to git just for dev testing.

I'm up for suggestions though, so if you have any other ideas, I'll try them.

dan-blanchard:
I would love to see this merged.  I'm in a situation where there's a bunch of data in repository for a private Python package that is only needed for running experiments and things, and is purposefully not part of the egg.  It all ends up in the `recipe` directory in the conda package though, because we have a `conda.yaml` file in the root directory and use `git_url: ./`.

dan-blanchard:
I've bumped into another package where I've needed this.  When trying to build a recipe for [invoke](https://pypi.python.org/pypi/invoke/0.10.1), I run into the issue that since they've vendorized Python 2 and Python 3 versions of the same library, conda build will not work with Python 2 because it fails to byte compile the Python 3 code.  Being able to have specify:

```yaml
path_ignore: invoke/vendor/yaml3 [py2k]
path_ignore: invoke/vendor/yaml2 [py3k]
```

would be enormously helpful.

asmeurer:
Can't you just delete these things in the meta.yaml?

dan-blanchard:
@asmeurer True, I could.  It just seems like it might be a common problem.  I'll do that for now though.

desilinguist:
This would be very helpful to me too. I am using `conda.yaml` to build packages and there currently seems to be no way to exclude large files that should *not* be included in the final conda package.

jasongrout:
A short comment: (1) deleting these things in build.sh means you also have to delete them in bld.bat - and keeping that exclude information in two different places makes it more prone to inconsistency, and (2) having the information in the structured yaml file, in general, seems cleaner than doing it in the wild west atmosphere of the build files.

groutr:
Is this PR still alive?

mausch:
+1

neevor:
+1

msarahan:
If anyone is willing to adopt this PR and fix the merge conflicts, I'd be happy to take a look.

I think the overall issue is that "best practices" for conda recipes generally means putting your recipe in a folder by itself.  This PR will not change that best practice - just make it easier to ignore that best practice.

@dan-blanchard had an actually useful purpose for this PR, which makes it much more compelling.

dan-blanchard:
I tried to resolve the conflicts in #1949

rs2:
:+1:

msarahan:
Closing as inactive. Happy to reopen if anyone wants to pick this one up. You'll need to start by answering my questions at https://github.com/conda/conda-build/pull/1949#issuecomment-420327147

PR #303, opened 2015-01-16
Cran

asmeurer:
This is a `conda skeleton cran` command.


asmeurer:
I've been using this for a week without any changes so to me that indicates that it's ready to go.

PR #307, opened 2015-01-21
Fix #304

### LINKED ISSUE #304: requirements.txt should not override requirements in meta.yaml
The change in #190 blew up in my face today because I have a package that I want to be installable both via `pip` and via `conda`.  Calling the pip requirements file `requirements.txt` has been standard practice for some time, and I had no idea conda-build was reading it now.

Anyway, what happened was that I had `beautifulsoup4` in my `requirements.txt`, which is the name of the package on PyPI, whereas the conda package is called `beautiful-soup`.  The build test kept failing saying it couldn't find a conda package called `beautifulsoup4` and I was perplexed, because my requirements section in my `meta.yaml` clearly reads:

```yaml
requirements:
  build:
    - python
    - regex
    - setuptools
    - beautiful-soup
    - six

  run:
    - python
    - regex
    - setuptools
    - beautiful-soup
    - six
```

but if you look in the `index.json` that gets generated, you see:

```python
{
  ...
  "depends": [
    "beautifulsoup4",
    "regex",
    "six"
  ],
  ...
}
```

Overriding what people actually put in `meta.yaml` seems highly undesirable to me, and if you want to allow it at all, it should be optional (and not on by default).

dan-blanchard:
Do not override `self.meta['requirements']['run']` if it's already set.

### DIFF
--- a/conda_build/metadata.py
+++ b/conda_build/metadata.py
@@ -258,7 +258,7 @@ class MetaData(object):
             return
         self.meta = parse(get_contents(self.meta_path))
 
-        if isfile(self.requirements_path):
+        if isfile(self.requirements_path) and not self.meta['requirements']['run']:
             self.meta.setdefault('requirements', {})
             run_requirements = specs_from_url(self.requirements_path)
             self.meta['requirements']['run'] = run_requirements


dan-blanchard:
@asmeurer Thanks for merging so quickly!

PR #311, opened 2015-01-27
Introduce build/use_destdir option in meta.yaml

flub:
This introduces a build/use_destdir option to meta.yaml.  When this
option is set to true files should be installed into $DESTDIR
instead of $PREFIX.  Inside $DESTDIR the $PREFIX should still be
respected, e.g. $DESTDIR/$PREFIX/bin/my_binary is a valid file.

This is useful in case one needs to build a package which
build-depends on itself.  In this case the normal build system would
not detect the new files added.

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -336,6 +336,14 @@ def build(m, get_src=True, verbose=True, post=None):
     :type post: bool or None. None means run the whole build. True means run
     post only. False means stop just before the post.
     '''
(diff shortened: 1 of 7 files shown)


flub:
After some discussion on the mailing list is was decided that this adds a little too much complexity while simply:
```
conda create --yes --quiet -p reqs gcc
export PATH=$(pwd)/reqs/bin:$PATH
```
at the start of build.sh does achieve this much simpler and appears to be good enough.

PR #312, opened 2015-01-27
Use @rpath in the install name of shared libraries on OS X

### LINKED ISSUE #279: On OSX, library install name should contain @rpath
In order to make is easier for user to link executable (not packaged with conda) with library
installed with conda, library install names shoud contain `@rpath`.

Current situtation
==================

From what I understand looking at `pkgs/hdf5-1.8.13-1/lib/libhdf5.dylib` and
`pkgs/hdf5-1.8.13-1/bin/h5dump` here is the current situation, sorry if
I missunderstand something, or miss other source of informations:

On OS X, for a conda pakage `foo` containing a library `libfoo.dylib` and an executable `xfoo`,
`conda install foo` yields to something like (actually, the the binary files are just
extracted from the package, but it doesn't matter for the example):

    $ ROOT=$PWD
    $ mkdir -p $ROOT/miniconda/pkgs/foo/bin $ROOT/miniconda/pkgs/foo/lib
    
    $ cd $ROOT/miniconda/pkgs/foo/lib
    $ echo 'int foo(int a, int b){return a+b;}' > foo.c
    $ clang foo.c -dynamiclib -install_name libfoo.dylib -o libfoo.dylib
    
    $ cd $ROOT/miniconda/pkgs/foo/bin
    $ echo 'int foo(int,int); int main(){return foo(2,4);}' > xfoo.c
    $ clang xfoo.c -L../lib -lfoo  -o xfoo
    $ install_name_tool -change libfoo.dylib '@loader_path/../lib/libfoo.dylib' xfoo
    
Now, the directory libfoo.dylib and xfoo moved for example as:

    $ mkdir -p $ROOT/miniconda/envs/envname/lib $ROOT/miniconda/envs/envname/bin
    $ ln $ROOT/miniconda/pkgs/foo/lib/libfoo.dylib $ROOT/miniconda/envs/envname/lib/libfoo.dylib
    $ ln $ROOT/miniconda/pkgs/foo/bin/xfoo $ROOT/miniconda/envs/envname/bin/xfoo
    
safely, because in `xfoo`, the install name of the dependent shared library `foo `in relative to where `xfoo` is.
`foo` and `xfoo` are rellocatables.

Problem
=======

A user developping an executable `yfoo` needing the `foo` library will want to reuse the library `libfoo.dylib`
provided by conda, due to the easiness and flexibility of the installation with conda.

However, this require to run `install_name_tool -change` on the created executable:

    $ mkdir $ROOT/perso ; cd $ROOT/perso
    $ echo 'int foo(int,int); int main(){return foo(2,4);}' > yfoo.c
    $ clang yfoo.c -L$ROOT/miniconda/envs/envname/lib -lfoo  -o yfoo
    $ install_name_tool -change libfoo.dylib "$ROOT/miniconda/envs/envname/lib/libfoo.dylib" yfoo
   
Proposal
========

The library install name should contain `@rpath`, so `conda install foo` yields to something like this:

    $ mkdir -p $ROOT/miniconda_bis/pkgs/foo/bin $ROOT/miniconda_bis/pkgs/foo/lib
    
    $ cd $ROOT/miniconda_bis/pkgs/foo/lib
    $ echo 'int foo(int a, int b){return a+b;}' > foo.c
    $ clang foo.c -dynamiclib -install_name '@rpath/libfoo.dylib' -o libfoo.dylib
    
    $ cd $ROOT/miniconda_bis/pkgs/foo/bin
    $ echo 'int foo(int,int); int main(){return foo(2,4);}' > xfoo.c
    $ clang xfoo.c -L../lib -lfoo  -o xfoo
    $ install_name_tool -change '@rpath/libfoo.dylib' '@loader_path/../lib/libfoo.dylib' xfoo
    
    $ mkdir -p $ROOT/miniconda_bis/envs/envname/lib $ROOT/miniconda_bis/envs/envname/bin
    $ ln $ROOT/miniconda_bis/pkgs/foo/lib/libfoo.dylib $ROOT/miniconda_bis/envs/envname/lib/libfoo.dylib
    $ ln $ROOT/miniconda_bis/pkgs/foo/bin/xfoo $ROOT/miniconda_bis/envs/envname/bin/xfoo
    
So that the user can links with `libfoo.dylib` using rpath:

    $ mkdir $ROOT/perso_bis ; cd $ROOT/perso_bis
    $ echo 'int foo(int,int); int main(){return foo(2,4);}' > yfoo.c
    $ clang yfoo.c -L$ROOT/miniconda_bis/envs/envname/lib -Wl,-rpath,$ROOT/miniconda_bis/envs/envname/lib -lfoo -o yfoo

asmeurer:
@rpath will, by convention, point to the lib directory of the environment,
using @loader_path (a future commit will add this to all shared libraries
automatically).

The install name is copied into depending libraries at link time. This allows
libraries to refer back to dependent libraries in a relocatable way. As long
as some library that has been loaded has an LC_RPATH pointing to the lib
directory, the loader will be able to find the dependencies.

The situation works very similarly to the way it does on Linux. The difference
is that in Linux, libraries are loaded by their file name only, whereas on OS
X, they can use arbitrary paths from the install name. This is better, because
it lets us use only a single rpath (the lib directory) to point to
everything.

This makes it possible to build packages that link against libraries from a
conda package without conda build, and have them work without calling
install_name_tool, using special compiler flags, or setting
DYLD_LIBRARY_PATH.

This fixes #279.


asmeurer:
This sets the install name of shared libraries to be `@rpath/relative/path/to/library.dylib`, where `@rpath` should be replaced with `$PREFIX/lib` to give the path to the library.  It also adds the `LC_RPATH` to the libraries like `@loader_path/relative/path/to/lib`. (e.g., for `$PREFIX/lib/R/lib/libR.dylib`, the install name is `@rpath/R/lib/libR.dylib` and the `LC_RPATH` is set to `@loader_path/../../`). 

The install name ends up getting pulled in at link time, and as long as the loader can find one of the `LC_RPATH`s in the dependency chain, it finds everything.

I manually applied the changes from the pull request to libpython.dylib, libhdf5.dylib, and libhdf5_hl.dylib, and was able to pip install h5py and run it (without the change, with the current conda python and hdf5 packages, pip install h5py leads to "Library not loaded: libhdf5.9.dylib"). I got similar results with R packages

I believe this change should make developing against conda packages "just work", without using install_name_tool, compiler flags, or DYLD_LIBRARY_PATH. In the worst case you may need to add -rpath to your compiler flags if it can't find any of the rpaths at load time.

The reason it "just works" for Python or R packages is that inside of Python or R, the libpython or libR have already been loaded, and so the `LC_RPATH`s inside of them will be search by the loader, which will point them to the lib directory and the dependent library it is trying to find.

asmeurer:
Note, to see the install name of a shared library on OS X, look at the first indented line of `otool -L`. The dependent install names are the other indented lines. To see `LC_RPATH` (if it is set), use `otool -l path/to/library.dylib | tail`.

dfroger:
Changes look great to fix #279. Thanks. I'll use 8b882bb next week to build conda package on OS X, and will report problems if any.

asmeurer:
FYI, there is a known issue with this PR, which is that the code [here](https://github.com/conda/conda-build/pull/312/files#diff-60400bbbdb51124a0a8a06ca17ecec5cR153) makes the invalid assumption that libraries will be next to one another. I think the only way to fix it is to write some code that searches the prefix for a given library (since the original install name with the full path is replaced with just the file name).

ccordoba12:
@asmeurer will this address the issues with having multiple Qt's on Mac too?

asmeurer:
I don't know what causes that issue.

asmeurer:
I want to merge this. @ilanschnell can you take a look? The main things to look at are `post.py` and `macho.py` (the other things are additions to `conda skeleton cran` and `conda inspect`, which are not that important).

ilanschnell:
Looks fine.

PR #319, opened 2015-02-02
Add support for adding a README

### LINKED ISSUE #318: Conda-build should allow inclusion of a readme in the metadata
Binstar.org supports uploading a README, however conda does not allow for including a README. 
Can we add a `readme` tag in the `meta.yaml` to include a `README.*` in the info?

asmeurer:
Fixes #318.

@srossross take a look


ilanschnell:
Ah, so the readme file you specify in the `meta.yaml` is put into the `info` directory.

Issue #322, opened 2015-02-07
Documentation around GIT_DESCRIBE environment variables 


mrocklin:
Hi all, 

I've tried naively following the instructions surrounding the use of `git describe` environment variables found here http://conda.pydata.org/docs/build.html . My solution is here https://github.com/mrocklin/into/blob/conda-build-recipe/conda.recipe/meta.yaml

However when I run the following:

    ~/workspace/into/$ conda build conda.recipe

I get the following error:

```
  File "/home/mrocklin/Software/anaconda/lib/python2.7/site-packages/conda_build/metadata.py", line 311, in build_number
    return int(self.get_value('build/number', 0))
TypeError: int() argument must be a string or a number, not 'NoneType'
```

In a small attempt to locate the issue I inspect what's happening in `get_git_build_info`. I find oddly that the `git_dir` and `src_dir` doesn't point to my repository, but rather to something in `conda-bld`

```
> /home/mrocklin/Software/anaconda/lib/python2.7/site-packages/conda_build-1.10.0_117_g807392d_dirty-py2.7.egg/conda_build/environ.py(37)get_git_build_info()
-> if os.path.isdir(git_dir):
(Pdb) print git_dir
/home/mrocklin/Software/anaconda/conda-bld/work/.git
```

Probably I'm doing something dumb.  But it's unclear to me what that dumb thing is.

melund:
I get the same error. Maybe something is broken in the newest version of conda-build... 

Downgrading to conda-build 1.10.0 solves the problem for me.

asmeurer:
There are two issues at play here:

The GIT_* environment variables are set to empty strings when the source is not a git repository. This behavior is wrong given the `environ.get()` code that we recommend putting in people's meta.yaml. They should not be set at all, so that the second argument of the get is used, rather than being set to None (an empty value in a dictionary in yaml is implicitly null). This is easy to fix, and I will do so now. The fix for this should make your recipe work. 

The second issue is that the source directory in the `get_git_build_info` function is not a git repository yet when it is run, because it hasn't been cloned yet. @tswicegood I know we had issues like this before. Do you remember what the status of this is?

tacaswell:
[...]

PR #325, opened 2015-02-10
Remove the work directory earlier in the build

asmeurer:
See #322. The templating of the meta.yaml might pull in false information from
previous work directories.

Note that I left the removal in the provide() function because it
can be called from elsewhere.

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -322,7 +322,7 @@ def build(m, get_src=True, verbose=True, post=None):
         rm_rf(config.short_build_prefix)
         rm_rf(config.long_build_prefix)
         print("Removing old work directory")
-        rm_rf(config.WORK_DIR)
+        rm_rf(source.WORK_DIR)
 
         if (m.get_value('build/detect_binary_files_with_prefix')
             or m.binary_has_prefix_files()):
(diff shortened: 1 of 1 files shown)


tswicegood:
This is going to make builds potentially slower.  One advantage to this directory sticking around is that only new data has to be pushed.  Rather than force the directory removal, what about naming the directory something unique to the project.  Say `os.path.join(work_dir, slugified_pkg_name)`?  Might be more work than its worth, but that's what I thought was happening.

asmeurer:
I don't follow you. The work directory is already always removed, just in a different place (when the source is retrieved).

tswicegood:
It might be that my understanding of what was happening is flawed.  As I understood it with the directory still being there a pull was done instead of a clone.  I don't remember if that was an assumption or if I based that on code.  I thought I had seen code to that effect, but might be mistaken.

asmeurer:
I see. No, that doesn't happen. See https://github.com/conda/conda-build/blob/72c4e8610692c0d92d91f7c0d465ee14b1f95a7c/conda_build/source.py#L246. I suppose could special-case git and do a `git clean -df; git clean -Xdf; git checkout`. I'm not really sure it would be that much faster. I do often see removing the work and build directories take a few seconds (usually more the build directory than the work directory), though that's typically only on my VMs with slow filesystems. But even then it's never been something that I've felt was slow enough to warrant trying to avoid.

tswicegood:
Fair enough.  :+1: on this.

PR #327, opened 2015-02-13
Default to conda.recipe when calling conda build

tswicegood:


tswicegood:
@asmeurer @ilanschnell any problems with this?  I get tired of having to include `./conda.recipe/` all the time. :-)

ilanschnell:
Having the conda recipe in the source tree itself is an exception.  Making this exception the default behavior will make it appear as if we encourage this.

tswicegood:
@ilanschnell I think we should encourage.  The behavior I've seen across the organization (conda after I requested it, bokeh, wakari, and binstar) is that the recipe is shipped with the source code.  I agree there are lots of cases where recipe directories such as [conda-recipes](https://github.com/conda/conda-recipes) or [conda-javascript-recipes](https://github.com/conda/conda-javascript-recipes) are valid, but that's not the norm for someone actively developing one particular conda package.

asmeurer:
This conflicts with https://github.com/conda/conda-build/pull/226. The fact that both exist indicates that there are at least two prevailing opinions on what the "default" conda build behavior should be.

ilanschnell:
@tswicegood there is a distinction between developing a "package" and a "conda recipe".  What exactly do you mean when you say "conda package"?

tswicegood:
@asmeurer doh -- I forgot about simple `conda.yaml` based recipes.  I think that's mostly what @srossross uses.  I've modified this to check for all of them.

@ilanschnell when I'm working on `conda-env`, I'm working on the the code that becomes that package.  As part of that testing, I want to be able to build a package from that code using the recipe.  That's different than making a package for io.js or numpy.  Those are packages that are created from outside the normal development workflow for those code bases.  I know you create a lot of those, but as far as I can tell you're one of the few people that works on those types of packages as the majority of their interaction with conda.

asmeurer:
I'm -1 (but not super strongly) to allowing `conda build` with no additional arguments to do something.

ilanschnell:
From the Zen of Python, by Tim Peters:
```
Explicit is better than implicit.
```

asmeurer:
> as far as I can tell you're one of the few people that works on those types of packages as the majority of their interaction with conda.

This is not true. There are a lot of "conda-recipes" type repos out there in the wild. Examples: https://github.com/SciTools/conda-recipes-scitools, https://github.com/ioos/conda-recipes, https://github.com/omnia-md/conda-recipes, https://github.com/TheClimateCorporation/conda-recipes. Just take a look at https://github.com/search?utf8=%E2%9C%93&q=conda-recipes.

ccordoba12:
I'm +1 on this change. I don't see any harm with it and as @tswicegood says, it seems really helpful for people with conda recipes as part of their source code.

My only suggestions are:

1. Add documentation about it to `conda build --help`

2. Improve the `RuntimeError` because it's too terse right now.

msarahan:
I think the spirit of this PR is captured in #908, which recursively looks for meta.yaml or conda.yaml, and complains if it finds more than one.

You still have to provide a starting directory as an argument to conda build, but it can be a folder further up the hierarchy.

I'm closing this for now.  If I have misunderstood the intent here, please reopen or leave a message.

PR #335, opened 2015-02-27
Try setting sudo: false in the .travis.yml

asmeurer:
I'm not sure if we need sudo, but if we don't, this should make the builds use
docker, so they will start faster.


asmeurer:
Seems to work.

PR #336, opened 2015-02-27
Correctly fix egg dirs that are part of the package

asmeurer:
The logic that was intended to skip egg dirs that were not part of the package
was incorrect and ended up skipping all egg dirs.


asmeurer:
This worked in my tests. @dhirschfeld if you still have issues please let me know. Maybe there are some additional problems, but I'm pretty sure this at least fixes one issue.

PR #337, opened 2015-03-03
Switch to using MD5 for comparison in conda index

dan-blanchard:
It used to rely on exact `mtime` matches, which was problematic for distributed file systems like GlusterFS.

I considered switching just converting the times to ints to get around precision issues, but since we already calculate `MD5`, I thought that might be less controversial.

### DIFF
--- a/conda_build/index.py
+++ b/conda_build/index.py
@@ -9,7 +9,7 @@ import bz2
 import sys
 import json
 import tarfile
-from os.path import join
+from os.path import join, getmtime
 
 from conda_build.utils import file_info
 from conda.compat import PY3
@@ -17,6 +17,7 @@ from conda.utils import md5_file
 
 
 def read_index_tar(tar_path):
+    """ Returns the index.json dict inside the given package tarball. """
     with tarfile.open(tar_path) as t:
         try:
             return json.loads(t.extractfile('info/index.json').read().decode('utf-8'))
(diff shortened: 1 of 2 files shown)


asmeurer:
I am +1. @ilanschnell?

asmeurer:
@ilanschnell pointed out that this would be much slower, especially for large repos. Let's just make this a command line flag.

asmeurer:
There is also `--force`. I don't know how it would compare in speed to what you have here.

ilanschnell:
Yes, the `--force` redoes the whole index, ie. calculating the md5's and extracting the packages to get the index.json.  To me it seems at as soon as you start reading all the files, it will take *very* long, such that you would normally run it overnight anyway.

dan-blanchard:
That's true, this would probably be quite a bit slower.  How would you feel about going with my alternative idea of reducing the required precision of the modification time matches (either via `int()` or an explicit precision option)?

dan-blanchard:
For our repository this is so much faster than the way things were before (where a lot of things ended up getting re-indexed because of millisecond timestamp differences), that I didn't even think about this being slower than how things are actually supposed to work. :smile:

asmeurer:
I think the best solution is to add this feature but to only enable it with some `conda index --check-md5` flag.

dan-blanchard:
I've added the flag.  Do we want the short version to be `-c` or `-m`? I made it `-c` because I usually just use the first letter of the longer version, but I know `-c` means `--channel` for most of the commands.

ilanschnell:
`-c` is fine with me.

dan-blanchard:
Okay, then everything in the pull request is up to date.  I also added some docstrings to the functions in index.py because I'm a pedant.

ilanschnell:
Thanks for your work, Dan.

Issue #178, opened 2014-08-19
Issue with tests when building a package for Python


asmeurer:
If you build a package for Python, it wants to run the tests using Python, which means that it tries to install the version of Python that conda is installed into into the test environment. If this differs from the Python version of the package, it fails because the packages conflict.

Issue #343, opened 2015-03-11
Windows is using the wrong Python version at test time


patricksnape:
# Environment Information

```
Current conda install:

             platform : win-32
        conda version : 3.9.1
  conda-build version : 1.11.0
       python version : 2.7.8.final.0
     requests version : 2.5.3
     root environment : C:\Anaconda\32  (writable)
  default environment : C:\Anaconda\32
     envs directories : C:\Anaconda\32\envs
        package cache : C:\Anaconda\32\pkgs
         channel URLs : https://conda.binstar.org/menpo/win-32/
                        https://conda.binstar.org/menpo/noarch/
                        http://repo.continuum.io/pkgs/free/win-32/
                        http://repo.continuum.io/pkgs/free/noarch/
                        http://repo.continuum.io/pkgs/pro/win-32/
                        http://repo.continuum.io/pkgs/pro/noarch/
          config file : C:\Users\pts08\.condarc
    is foreign system : False
```
## Packages Installed In Root

```
binstar                   0.10.1                   py27_2
clyent                    0.3.2                    py27_0
conda                     3.9.1                    py27_0
conda-build               1.11.0                   py27_0
conda-env                 2.1.3                    py27_0
dateutil                  2.1                      py27_2
jinja2                    2.7.3                    py27_1
markupsafe                0.23                     py27_0
menuinst                  1.0.4                    py27_0
psutil                    2.1.1                    py27_0
pycosat                   0.6.1                    py27_0
python                    2.7.8                         0
pytz                      2014.9                   py27_0
pyyaml                    3.11                     py27_0
requests                  2.5.3                    py27_0
setuptools                14.0                     py27_0
six                       1.9.0                    py27_0
```
# Replicate

In a conda build script, try running an arbitrary Python command:

```
test:
  commands:
  - python -c "import sys; print(sys.executable)"
```

which prints the wrong Python executable path (it will print the root Python path). Previously, I had this working, so this is a regression in behaviour.
# Example issue

We are seeing this on our [Appveyor builds](https://ci.appveyor.com/project/jabooth/menpo/build/1.0.392/job/skab8naf41m0m754#L1477) when trying to run our Nose tests. This is because Nose is not installed in the root environment, and the wrong Python interpreter is being picked up. This was not happening until very recently (last couple of days).

asmeurer:
This was probably a regression in conda-build 1.11.0, which was released on the 5th.

asmeurer:
When I create a test recipe with 

```
package:
[...]

PR #344, opened 2015-03-11
fix for issue #178

yannpaul:
This is a fix for issue #178 

Reordered building of package list for test environment. While adding package to list, if python is included, then ignore adding python to list for packages that test for valid python module/package imports.

I've manually tested this using miniconda3 for a build of a package that requires python 2.

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -462,7 +462,7 @@ def test(m, verbose=True, channel_urls=(), override_channels=False):
     specs_include_python = False
     for spec in m.get_value('test/requires', []):
         specs.append(spec)
-        if spec.startswith('python '):
+        if spec.startswith('python ') or spec == 'python':
             specs_include_python = True
 
     if py_files and not specs_include_python:
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -458,15 +458,19 @@ def test(m, verbose=True, channel_urls=(), override_channels=False):
     rm_rf(config.test_prefix)
     specs = ['%s %s %s' % (m.name(), m.version(), m.build_id())]
 
-    if py_files:
+    # add packages listed in test/requires
+    specs_include_python = False
+    for spec in m.get_value('test/requires', []):
+        specs.append(spec)
+        if spec.startswith('python '):
+            specs_include_python = True
+
+    if py_files and not specs_include_python:
         # as the tests are run by python, we need to specify it
         specs += ['python %s*' % environ.get_py_ver()]
     if pl_files:
         # as the tests are run by perl, we need to specify it
         specs += ['perl %s*' % environ.get_perl_ver()]
-    # add packages listed in test/requires
-    for spec in m.get_value('test/requires', []):
-        specs.append(spec)
 
     create_env(config.test_prefix, specs, verbose=verbose,
         channel_urls=channel_urls, override_channels=override_channels)


yannpaul:
FYI I'm newish to github, so please feel free to comment on how I can improve my pull requests. Also are there any suggestions/recommendations for testing?

yannpaul:
I should note I also tested building package against python 3. It also ran smoothly.

asmeurer:
As far as the git/GitHub stuff goes, it looks fine. The only suggestion I have there is to add the text "fixes #178" in the commit message so that it automatically closes the issue when merged. But it isn't that big of a deal if you don't do it.

conda_build/build.py
```diff
@@ -458,15 +458,19 @@ def test(m, verbose=True, channel_urls=(), override_channels=False):
     rm_rf(config.test_prefix)
     specs = ['%s %s %s' % (m.name(), m.version(), m.build_id())]
 
-    if py_files:
+    # add packages listed in test/requires
+    specs_include_python = False
+    for spec in m.get_value('test/requires', []):
+        specs.append(spec)
+        if spec.startswith('python '):
```
asmeurer: Won't this be wrong if the spec is just `python` (without a version)?

asmeurer:
If you want to try adding a test the way we've been doing it is to have some test recipes in tests/test-recipes, which should not fail when built. We haven't been really great at testing things, though, so I won't fault you if you don't add a test here.

yannpaul:
This might be wrong in regards to `python` without a version... I will test it out, as that's the standard requirement for pypi skeleton packages right? Is this issue at all related to #343 ?

asmeurer:
I don't think it's related to that issue. 

Any package generated with skeleton isn't going to have python in the test requires, as test requirements already include the runtime requirements automatically, and every package generated by skeleton runtime requires python.

yannpaul:
It seemed overkill to add build test cases if I'm testing one small component. I manually tested having run requirements (python 2.7*) different from CONDA_PY (python 3.4), with test requirements of various flavors (none, python, python 2.7*, python 3.4*). And everything seem to behave as expected.

conda_build/build.py
```diff
@@ -458,15 +458,19 @@ def test(m, verbose=True, channel_urls=(), override_channels=False):
...
+
+    if py_files and not specs_include_python:
         # as the tests are run by python, we need to specify it
         specs += ['python %s*' % environ.get_py_ver()]
     if pl_files:
         # as the tests are run by perl, we need to specify it
         specs += ['perl %s*' % environ.get_perl_ver()]
-    # add packages listed in test/requires
-    for spec in m.get_value('test/requires', []):
-        specs.append(spec)
```
asmeurer: Why did you delete these lines?

conda_build/build.py
```diff
@@ -458,15 +458,19 @@ def test(m, verbose=True, channel_urls=(), override_channels=False):
     rm_rf(config.test_prefix)
     specs = ['%s %s %s' % (m.name(), m.version(), m.build_id())]
 
-    if py_files:
+    # add packages listed in test/requires
+    specs_include_python = False
+    for spec in m.get_value('test/requires', []):
+        specs.append(spec)
```
yannpaul: @asmeurer They were just moved here, taken care of before the `py_files` check. I just think the diff is not picking up on the move.

asmeurer: Oh I see now. Sorry.

patricksnape:
This is a pretty horrible blocker for our automated builds at the moment. Would it be possible to get a bugfix release of conda build including this?

asmeurer:
Does the workaround described at https://github.com/conda/conda-build/issues/178#issuecomment-52694641 work for you?

PR #345, opened 2015-03-11
Use the --force-rpath flag to patchelf

asmeurer:
See #111 for discussion. Without the flag, patchelf modifies RUNPATH rather
than RPATH.


asmeurer:
Note that the packages built in the Anaconda build system (i.e., all the packages in the official Continuum repos) are modified using chrpath, which changes the RPATH.

asmeurer:
I just rebuilt R on Linux with this (but no R packages). It seems to work.

PR #366, opened 2015-04-15
Change source of summary description to metadata

lennart0901:
Without this change summary contains `0` for me.


jakirkham:
@lennart0901 Thanks for this. Had the same problem and was wondering why.

Issue #369, opened 2015-04-16
building python 3.4 packages from a 2.7 conda environment with pyqt has issues.


tacaswell:
The packaging step of building a python 3.4 (and I think py3k generically) package which uses pyqt using a py2.7 install of conda has issues.  @ccordoba12 reports that this is not an issue running this build from a py3k version of conda  (this issue is created at his request).

See https://github.com/NSLS-II/conda-prescriptions/issues/35 for a non-mpl based example.

This can be seen trying to build matplotlib using the official recipe from conda-recipes

### conda recipe:  
```
21:59 $ git rev-parse HEAD 
380a535ec32563b0593f9991447e1ba6a99d7b96 
✔ ~/other_source/conda-recipes [master|⚑ 3] 
```

### Top of build trace

showing what version of things are being used in the build env

```

00:01 $ conda build matplotlib/ --python=3.4 
Removing old build directory 
Removing old work directory 
BUILD START: matplotlib-1.4.3-np19py34_0 
Fetching package metadata: ...... 
Solving package specifications: . 
The following NEW packages will be INSTALLED: 

   fontconfig:      2.11.1-3      
   freetype:        2.5.2-0       
   libpng:          1.5.13-1      
   libxml2:         2.9.0-0       
   nose:            1.3.6-py34_0  
   numpy:           1.9.2-py34_0  
   openssl:         1.0.1k-1      
   pip:             6.1.1-py34_0  
   pyparsing:       2.0.3-py34_0  
   pyqt:            4.11.3-py34_1 
   python:          3.4.3-0       
   python-dateutil: 2.4.2-py34_0  
   pytz:            2015.2-py34_0 
   qt:              4.8.6-1       
   readline:        6.2-2         
   setuptools:      15.0-py34_0   
   sip:             4.16.5-py34_0 
   six:             1.9.0-py34_0  
   sqlite:          3.8.4.1-1     
   system:          5.8-2         
   tk:              8.5.18-0      
   tornado:         4.1-py34_0    
   xz:              5.0.5-0       
   zlib:            1.2.8-0       

```

### End of build trace:
 snipping out the pages and pages of mpl build spam:
```
Using /home/tcaswell/python_env/_build/lib/python3.4/site-packages 
Finished processing dependencies for matplotlib==1.4.3 
+ rm -rf /home/tcaswell/.virtualenvs/_build/lib/python3.4/site-packages/PySide 
[...]

PR #370, opened 2015-04-16
Only compile missing pyc files for files that are part of the package

asmeurer:
Related to #369.


asmeurer:
Hmm, this might not actually do what it says it does. It now only checks files that are part of the package, but if it finds a single missing pyc file it calls compileall.py, which makes no such distinction.

msarahan:
Closing in favor of #718

PR #375, opened 2015-04-21
> 1 test command doesn't work on windows

mpharrigan:
A newline was missing


mpharrigan:
Do you have an estimate for when this fix might make it into a release?

asmeurer:
We just did a release so it may be a week or two.

Issue #376, opened 2015-04-21
Conda convert fails to convert pure python packages completely to new platform


jakirkham:
It appears `conda convert` does not completely transform a package. For instance, I took a pure python package built on mac and attempted to convert it (cwd is `conda-bld`).

     $ conda convert -p linux-64 -o . osx-64/package-XXX.tar.bz2

It places the package in the `linux-64` directory as it should; however, running the following upload commands results in one success and one failure.

    $ binstar upload osx-64/package-XXX.tar.bz2
    $ binstar upload linux-64/package-XXX.tar.bz2
    [Conflict] file osx-64/package-XXX.tar.bz2 already exists for package...

I have seen basically the same result when converting a pure python package on linux to mac and uploading. When inspecting the uncompressed archive of the converted package (i.e `linux-64`) I find that the `index.json` file contains the following text. I would think this should have been changed.

     "subdir": "osx-64"

I don't see anything else obvious that is causing the package to be recognized incorrectly.

PR #377, opened 2015-04-21
Change the subdirectory to be that of the final platform

jakirkham:
Fixes conda/conda-build#376.


jakirkham:
For the record, I tried creating a conda build package based on this fix, which I put here ( https://binstar.org/jakirkham/conda-build ). This fixed the issue and I was able to convert and upload packages to binstar with this fix under the correct platform.

Juanlu001:
I was bitten by this today, thanks! :+1:

PR #382, opened 2015-04-22
Update command docs

asmeurer:
@kerrywatson1


asmeurer:
This should not be merged until conda is released because of https://github.com/conda/conda/pull/1296.

PR #385, opened 2015-04-28
Versioneer 0.14

jakirkham:
Sorry, thought I had added this. Versioneer 0.14 generate PEP440 compatible versions for every commit not only releases. It explicitly states the number of commits since the last release and short SHA of the commit so as to be unique (hopefully collisions aside).


asmeurer:
I am +1. @ilanschnell do you have any comments? This changes the format of the versions

```
$conda build -V
conda-build 1.12.0+18.gb0e6ef3
```

ilanschnell:
Does the new version format have any advantages?

asmeurer:
It's PEP 440 compatible.

ilanschnell:
Yes, In that case I'm also +1, as we want to move to PEP 440.

jakirkham:
Please let me know if you guys have feedback.

asmeurer:
Let's merge it and see if there are any problems. If there aren't, let's do the same for conda.

PR #386, opened 2015-04-28
Fix regression in always_include_files introduced in 1.12.0

tswicegood:
Starting in version 1.12.0, this switched from including files if present to requiring those files to be present.  This breaks the conda-env build and any other build that used the original design.

Given that it is a backwards in-compatible change, this a break without bumping to version 2.0.0.  Reverting this change and changing to a simple error message.

Original Commit: f708dd0225b5292b644133afd190800e24e5e047

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,3 +1,7 @@
+2015-04-28   1.12.1:
+  * fix regression in always_include_files that causes build failure (#386)
+
+
 2015-04-10   1.12.0:
 --------------------
   * correctly fix egg directories that are part of the package
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -363,8 +363,8 @@ def build(m, get_src=True, verbose=True, post=None, channel_urls=(), override_ch
         files1 = prefix_files()
         for f in m.always_include_files():
             if f not in files1:
-                sys.exit("Error: File %s from always_include_files not found"
-                    % f)
+                print("Error: File %s from always_include_files not found" % f,
+                      file=sys.stderr)
         files1 = files1.difference(set(m.always_include_files()))
         # Save this for later
         with open(join(config.croot, 'prefix_files.txt'), 'w') as f:


tswicegood:
I've confirmed this against conda-env -- the build now passes.  Updating the CHANGELOG and waiting on TravisCI before merging.

asmeurer:
So I guess it's OK to just push stuff in (even if there's a discussion and disagreement about it), so long as there's a pull request. https://github.com/conda/conda-build/pull/387

tswicegood:
Wow.  This fixed a regression in conda-build that broke other builds.  It's not a theoretical discussion, [conda-env's build][1] has been broken since the build machines were updated to 1.12.0.  I fixed that while leaving an error message in to help alleviate your concern about possible typos and this is your response?

[1]: https://binstar.org/conda/conda-env/builds

asmeurer:
You didn't fix it. You created and merged a pull request reverted my change even though there was an open discussion and my [stated position](https://github.com/conda/conda-build/commit/f708dd0225b5292b644133afd190800e24e5e047#commitcomment-10946919) was to leave it as is. 

Let's discuss things in GitHub, meetings, etc., but let's not battle things out in the commit history and merge pull requests behind each other's backs when they are at lunch.

tswicegood:
I had no idea that you were at lunch so to attribute my merging it to an act "behind [your] back" is projecting intent on my actions that were not there.

Adding `sys.exit` to an existing code path breaks backwards compatibility.  This PR *does* fix the error in the builds referenced above.  I work under the assumption that if I made a change to any of the conda code that broke existing code in significant ways that it would be reverted and/or otherwise modified to remove that backwards incompatibility.

Without exceptionally good reasons, you can't say "change your code" because of an unintended consequence of changes that are made.  For context, putting a developer's system security at risk or side effects that harm your system would qualify as an exceptionally good reason to me; protecting someone from fat-fingering a filename is not.

PR #387, opened 2015-04-28
always_include_files should error on files that aren't there

asmeurer:
This is to help avoid typos. If there are files that aren't always there, you
can filter them out with selectors.

See the discussion at
https://github.com/conda/conda-build/commit/f708dd0225b5292b644133afd190800e24e5e047


tswicegood:
@asmeurer please either:

* Update the CHANGELOG with a note about this change marked as version v2.0.0, or
* Revert this change and allow #386 to fix this without breaking builds

PR #400, opened 2015-05-11
Test binaries passed to `setup` via `scripts` keyword argument using `--help`

jakirkham:
Just included everything passed to `scripts` in the `--help` tests. This will help improve testing for Python packages not leveraging the keyword arguments added by `bdist_conda`. Also, it applies a reasonable on all other Python packages (namely having basic help for command line tools).

As this is something `argparse` pretty much handles (or the deprecated `optparse`), this seems like a reasonable constraint.

### DIFF
--- a/bdist_conda.py
+++ b/bdist_conda.py
@@ -214,14 +214,15 @@ class bdist_conda(install):
                 if not isinstance(entry_points, dict):
                     raise DistutilsGetoptError("ERROR: Could not add entry points. They were:\n" + entry_points)
                 else:
+                    rs = entry_points.get('scripts', [])
                     cs = entry_points.get('console_scripts', [])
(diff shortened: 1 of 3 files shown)


jakirkham:
This appears to be passing on my branch with Travis and my own local testing. However, Travis' web interface has been down so I have been unable to look at their logs more closely.

jakirkham:
Any thoughts on this @asmeurer? Open to suggestions as always.

asmeurer:
Not all scripts support --help but it's good enough practice that I'm in favor of doing this.

jakirkham:
Cool. I was trying to think if there were ways I might improve on this. Even though you have merged this, if you do have more thoughts, I am willing to contribute changes to improve this check.

PR #403, opened 2015-05-12
skeleton pypi: add --manual-url argument

hargup:
Currently, when conda-skeleton has more than one sources to choose from, it asks
the users for the decision. The patch changes this behaviour to download from
the source with the least source size. This is because this former choice can
only add to the confusion of a _normal_ user. An informed user can override the
behaviour with the newly added `--manual-url` command line argument.

Signed-off-by: Harsh Gupta <gupta.harsh96@gmail.com>

### DIFF
--- a/conda_build/main_skeleton.py
+++ b/conda_build/main_skeleton.py
@@ -94,6 +94,14 @@ def main():
         choices=['2.6', '2.7', '3.3', '3.4'],
         )
 
+    pypi.add_argument(
+        "--manual-url",
+        action='store_true',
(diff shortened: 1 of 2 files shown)


hargup:
ping @asmeurer @ilanschnell

ilanschnell:
Looks reasonable, I would say.  Since @asmeurer wrote this module, I'd like him to comment before merging.

asmeurer:
This looks good. I don't recall seeing source URLs that conda-build couldn't handle.

PR #405, opened 2015-05-13
[ENH] Add --version-compare argument to conda-skeleton

### LINKED ISSUE #397: ENH: skeleton --version-compare for latest versions
a naieve ``--version-compare`` routine in ``skeleton`` would make comparing recipe version metadata with e.g. PyPI version metadata more simple

seeAlso: https://github.com/conda/conda-recipes/issues/291

potentially: UBY (usability), SEC (security) due to packaging latency

hargup:
This PR aims to make comparison of version of locally available recipe with version on PyPI simpler.

fixes #397

### DIFF
--- a/conda_build/pypi.py
+++ b/conda_build/pypi.py
@@ -264,7 +264,7 @@ def main(args, parser):
 
         if not is_url:
             dir_path = join(output_dir, package.lower())
-            if exists(dir_path):
+            if exists(dir_path) and not args.version_compare:
                 raise RuntimeError("directory already exists: %s" % dir_path)
         d = package_dicts.setdefault(package,
             {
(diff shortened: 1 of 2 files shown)


hargup:
/cc @asmeurer 
@westurner can you apply this patch and verify if it has the intended behaviour.

hargup:
Is this PR good to go?

asmeurer:
This needs to do case normalization (similar to what normal `skeleton pypi` does):

```
$conda skeleton pypi --version-compare nikola
Local recipe for nikola has version 7.3.1.1
Error: nikola 7.3.1.1 is not available on PyPI.
$conda skeleton pypi --version-compare Nikola
Local recipe for Nikola has version 7.3.1.1
Following new versions of Nikola are avaliable
7.4.1
7.4.0
```

conda_build/pypi.py
```diff
@@ -287,15 +287,39 @@ def main(args, parser):
...
+                m = MetaData(recipe_dir)
+                local_version = m.version()
+                print("Local recipe for %s has version %s" % (package,
+                      local_version))
+                if local_version not in versions:
+                    sys.exit("Error: %s %s is not available on PyPI."
+                             % (package, local_version))
+                else:
+                    new_versions = [ver for ver in versions
+                                    if ver > local_version]
```
asmeurer: Is this doing a string comparison?

asmeurer: You should use `conda.resolve.normalized_version`, which will return a version object that you can compare against. Alternately, the versions on PyPI are already in order, so just check where the current version is in the list and return the ones before it (or after it, I don't remember).

hargup:
@asmeurer I have addressed your comments in the new commit, please have a look.

hargup:
@asmeurer @ilanschnell ping

hargup:
@asmeurer @ilanschnell is this pull request good to go?

asmeurer:
@hargup in the future, please add new commits to pull requests rather than doing `git rebase` or `git commit --amend`. It makes it much easier to review if I can see what changes are new since the last time I looked.

PR #413, opened 2015-05-18
Fix issue where 'conda index' with old packages would create bad metadata

campos-ddc:
This issue is also described here:
https://groups.google.com/a/continuum.io/forum/#!topic/conda/UXX9E7LkVMA

If necessary, this could also issue warnings about old style packages, or only fix them when an option is given.


asmeurer:
Looks fine.

PR #416, opened 2015-05-22
More fixes to index format ('requires -> depends and 'python/numpy wildcards')

campos-ddc:
Another commits fixing some differences between generated index and the ones used by continnum.
I'm trying to download all of their packages and mirror then in a local channel, but it is proving to be more troublesome than expected =)

Anyway, here it goes:

* Removing old-sytle 'requires' in favor of 'depends'
* Adding wildcard to 'python X' and 'numpy Y' dependencies

All of this is to make sure that repodata.json created from this index matches
http://repo.continuum.io/pkgs/<LIC>/<PLAT>/repodata.json as closely as possible.

### DIFF
--- a/conda_build/main_build.py
+++ b/conda_build/main_build.py
@@ -255,6 +255,7 @@ def execute(args, parser):
             channel_urls=channel_urls,
             override_channels=args.override_channels)
 
+    already_built = []
     with Locked(config.croot):
         recipes = deque(args.recipe)
         while recipes:
@@ -294,7 +295,7 @@ def execute(args, parser):
             if args.check:
                 continue
             if args.skip_existing:
-                if m.pkg_fn() in index:
+                if m.pkg_fn() in index or m.pkg_fn() in already_built:
                     print("%s is already built, skipping." % m.dist())
                     continue
             if args.output:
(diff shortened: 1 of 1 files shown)


conda_build/index.py
```diff
@@ -99,16 +99,30 @@ def update_index(dir_path, verbose=False, force=False, check_md5=False):
...
             try:
                 del info[varname]
             except KeyError:
                 pass
 
+        # Convert old packages to new format (use 'depends' instead of 'requires'
         if 'requires' in info and 'depends' not in info:
             info['depends'] = info['requires']
 
+        # Remove old format 'requires'. This reproduces index used by
```
asmeurer: You didn't finish the sentence here.

conda_build/index.py
```diff
@@ -99,16 +99,30 @@ def update_index(dir_path, verbose=False, force=False, check_md5=False):
...
+        info['requires'] = []
+
+        # Fix non-wildcard Python/NumPy dependencies
```
asmeurer: conda index shouldn't be changing the metadata, only extracting it.

asmeurer: Maybe you meant for this whole thing to be indented under the `if 'requires' in info and 'depends' not in info:` line above?

campos-ddc: I'm not sure this change is valid anymore.

I gave up on trying to reproduce continnum's index for old packages, maybe
we should just support index for new packages and leave it at that, what do
you think?

If you think this pull request is still good, I'll gladly fix those
problems Monday.

PR #417, opened 2015-05-22
Try extracting .tar.xz files with tar before requesting unxz

rwest:
Default Mac OS (10.10) has no 'unxz' installed, but the 'tar' can 
read .tar.xz files on its own so unxz is not needed.

I haven't been able to test this on other platforms, but it seems to work for me on MacOS.

### DIFF
--- a/conda_build/utils.py
+++ b/conda_build/utils.py
@@ -68,6 +68,16 @@ uncompress is required to unarchive .z source files.
         subprocess.check_call([uncompress, '-f', tarball])
         tarball = tarball[:-2]
     if not PY3 and tarball.endswith('.tar.xz'):
+        # try extracting directly with tar,
+        # which may be able to cope with xz compression
+        tar = external.find_executable('tar')
+        if tar:
+            try:
+                subprocess.check_call([tar, '-x', '-C', dir_path, '-f', tarball])
+                return
+            except:
+                # try another approach, using 'unxz' first.
+                pass
         unxz = external.find_executable('unxz')
         if not unxz:
             sys.exit("""\


conda_build/utils.py
```diff
@@ -68,6 +68,15 @@ def tar_xf(tarball, dir_path, mode='r:*'):
         subprocess.check_call([uncompress, '-f', tarball])
         tarball = tarball[:-2]
     if not PY3 and tarball.endswith('.tar.xz'):
+        try:
+            # try extracting directly with tar,
+            # which may be able to cope with xz compression
+            tar = external.find_executable('tar')
+            subprocess.check_call([tar, '-x', '-C', dir_path, '-f', tarball])
+            return
+        except:
```
asmeurer: You should only catch the exception raised by subprocess (and also guard against `tar` being `None`).

rwest: thanks. about to push revision

conda_build/utils.py
```diff
@@ -68,6 +68,16 @@ def tar_xf(tarball, dir_path, mode='r:*'):
...
         tarball = tarball[:-2]
     if not PY3 and tarball.endswith('.tar.xz'):
+        # try extracting directly with tar,
+        # which may be able to cope with xz compression
+        tar = external.find_executable('tar')
+        if tar:
+            try:
+                subprocess.check_call([tar, '-x', '-C', dir_path, '-f', tarball])
+                return
+            except:
```
asmeurer: Please only catch the specific exception (`subprocess.CalledProcessError` I believe).

wojdyr:
additional benefit of using tar would be that now unxz-ed tarballs are left in `src_cache` increasing the cache size

msarahan:
@rwest, any interest in revisiting this?  I agree with @asmeurer that you need to limit the scope of your except.  Otherwise, this is a nice addition, and I'd like to merge it.

PR #421, opened 2015-05-27
fix import_tests: module name != package name

hargup:
conda-skeleton implicitly assumes that package name is same module name and adds an import test for that, it is not necessarily the case. For example the package dnspython installs a module named dns and the package factory_boy installs the package factory. The earier behavior was causing build failures for such packages.


hargup:
@asmeurer @ilanschnell please review.

PR #430, opened 2015-06-02
Signed packages

ilanschnell:
This PR allows creating signed conda packages to work in conjunction with https://github.com/conda/conda/pull/1343

### DIFF
--- a/bin/conda-sign
+++ b/bin/conda-sign
@@ -1,0 +1,5 @@
+#!/usr/bin/env python
+import sys
+from conda_build.main_sign import main
+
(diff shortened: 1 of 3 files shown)


asmeurer:
Does the Crypto library already handle the error messages and whatnot if you don't have any keys already?

ilanschnell:
@asmeurer you have to read the keys yourself and create a Crypto key object (using the string data from the key files).  So, "no".  Crypto does not handle missing keys for you.  But the code in this PR does, and also allows different (named) key pairs.

PR #431, opened 2015-06-03
conda-skeleton: Add tests_require

hargup:


conda_build/pypi.py
```diff
@@ -86,7 +86,8 @@
   # You can also put a file called run_test.py in the recipe that will be run
   # at test time.
 
-  # requires:
+  requires:
+    - {tests_require}
```
asmeurer: The way `INDENT` works, it already includes the leading spaces and `-`. I think this should be

```
  {requires_comment}requires:{tests_require}
```

where `requires_comment` is `# ` if there are no test requires and empty if there are some.

hargup: Thanks, fixed it.

asmeurer:
Is there an example package with test_requires that I can test this on?

hargup:
Yes it is the package `fabric`. I should have included that in the commit message.

hargup:
@asmeurer is it good to go?

hargup:
@asmeurer @ilanschnell

asmeurer:
These need to be run through `spec_from_line` so that they are converted to the proper conda format. Right now, `fabric` gives `fudge<1.0` but it should be `fudge <1.0` with the space.

hargup:
@asmeurer fixed that in latest commit.

PR #434, opened 2015-06-03
Remove special logic if the username on Windows is "builder"

asmeurer:
This is a remnant from the Anaconda build system. It does not belong in
conda-build.


tswicegood:
:+1:

PR #207, opened 2014-09-17
Added HG_* variables to environ when build

pigmej:
Added:

``HG_BRANCH`` - current branch
``HG_NUM_ID`` - commit num id
``HG_LATEST_TAG_DISTANCE`` commit distance from latest known tag (possibly follows ``GIT_DESCRIBE_NUMBER`` logic)
``HG_SHORT_ID`` - commit short id
``HG_BUILD_STR`` - similar to ``GIT_BUILD_STR`` so NUM + SHORT joined with `_`

Allows to create meta.yaml like:

```
package:
  name: package_name
  version: {{ environ.get('HG_LATEST_TAG', "current") }}

build:
  number: {{ environ.get('HG_NUM_ID', 0) }}

  string: {{ environ.get('HG_BUILD_STR', '') }}

source:
  hg_url: /tmp/repo_path

```

Idea at first reported by me there: https://github.com/conda/conda/issues/916

PR #441, opened 2015-06-05
pypi build recipe from hg/git repository

pigmej:
**treat my branch as proof of concept which works**

The main idea was to be able to build pypi recipes from a hg repository. The logic is quite similar those from url.

The idea is like:
- `conda skeleton pypy hg:{repo_path}`
- for hg there is archive process (to use only what's in repo not in possibly dirty directory)
- then it works exactly like `conda skeleton pypi {url}` (they share the same logic).

What might be done:
- use remote repo (to clone it straight to bld directory)
- lock tag/commit inside recipe.

Why ? Mostly because of easier integration with CI/CD tools.

Currently only HG is supported but code is ready to work with GIT (if's are already on place).

What do you think about this ?


asmeurer:
@hargup maybe you can review this

hargup:
sure, btw @pigmej there are merge conflicts you need to rebase your commits on the latest master.

pigmej:
@hargup I'm aware of this, that's why I wrote "treat my branch as proof of concept which works"

The thing is, does it make sense to do it this way ?

What about previous points:
* use remote repo (to clone it straight to bld directory)
* lock tag/commit inside recipe.

This also contains changes to HG_* variables (https://github.com/conda/conda-build/pull/207), it should be now more consistent with GIT_* variables (and those changes are also in this PR).

I can obviously rebase it as it is currently, just let me know.

pigmej:
@hargup I'm aware of this, that's why I wrote "treat my branch as proof of concept which works"

The thing is, does it make sense to do it this way ?

What about previous points:
* use remote repo (to clone it straight to bld directory)
* lock tag/commit inside recipe.

This also contains changes to HG_* variables (https://github.com/conda/conda-build/pull/207), it should be now more consistent with GIT_* variables (and those changes are also in this PR).

I can obviously rebase it as it is currently, just let me know.

hargup:
@pigmej I tried out your code with a clone of https://bitbucket.org/pypa/bandersnatch and it worked, the source also looked good. So your code should be good to go in, thanks! These changes are surely helpful.

The use of remote repo and building recipe from a particular tag or commit are also good to have feature.

pigmej:
@hargup Yeah I will review this in near future. Quick look to code bases, and I will probaby just rewrite it on top of current master.

And yeah, I'm posting this mostly because I don't want to let this patch die :)

msarahan:
@pigmej any interest in revisiting this?  Seems like a good idea to me.

PR #442, opened 2015-06-05
Ensure that lines are split based on newlines, not by whitespace

tswicegood:


tswicegood:
This fixes an issue I ran into where filenames had spaces in them.  The empty call to `split()` worked, but it should have used `splitlines()` since it's actually lines.

tswicegood:
Thanks, @asmeurer.  FWIW, there's a bunch of `.split('\n')` calls throughout the code that should probably be converted to `.splitlines()`.  Many, if not all, of them are on files or strings that have been created by conda code, so it should be ok, but I could see weird edge cases like this cropping up again.

PR #448, opened 2015-06-15
skeleton: remove --no-download option

hargup:
The `--no-download` option of `conda-skeleton pypi` is non-functional. Presently conda-skeleton derives important metadata from the packages setup.py file with this setup there is no way to create any recipe  without the downloading the package source.


asmeurer:
Apparently the option is completely ignored, so +1.

jakirkham:
Unfortunate that is option got removed. Though I guess it never worked.

PR #449, opened 2015-06-15
Refactor conda-skeleton pypi

hargup:
Most of the heavy lifting of `conda-skeleton pypi` is done by the `get_package_metadata` function, this monolithic block of code is huge (~250 lines) which makes it hard to debug this code and reuse its parts in other. This PR factors out a few blocks into their own functions.

### DIFF
--- a/conda_build/pypi.py
+++ b/conda_build/pypi.py
@@ -322,65 +322,16 @@ def main(args, parser):
                     print("Use --version to specify a different version.")
                 d['version'] = versions[0]
 
-        data = client.release_data(package, d['version']) if not is_url else None
-        urls = client.release_urls(package, d['version']) if not is_url else [package]
-        if not is_url and not args.all_urls:
-            # Try to find source urls
-            urls = [url for url in urls if url['python_version'] == 'source']
-        if not urls:
-            if 'download_url' in data:
-                urls = [defaultdict(str, {'url': data['download_url']})]
-                if not urls[0]['url']:
-                    # The package doesn't have a url, or maybe it only has a wheel.
-                    sys.exit("Error: Could not build recipe for %s. "
-                        "Could not find any valid urls." % package)
-                U = parse_url(urls[0]['url'])
-                if not U.path:
-                    sys.exit("Error: Could not parse url for %s: %s" %
-                        (package, U))
-                urls[0]['filename'] = U.path.rsplit('/')[-1]
-                fragment = U.fragment or ''
-                if fragment.startswith('md5='):
-                    d['usemd5'] = ''
-                    d['md5'] = fragment[len('md5='):]
-                else:
-                    d['usemd5'] = '#'
-            else:
-                sys.exit("Error: No source urls found for %s" % package)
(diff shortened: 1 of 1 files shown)


conda_build/pypi.py
```diff
@@ -454,236 +454,188 @@ def version_compare(args, package, versions):
...
-    import yaml
     print("Downloading %s (use --no-download to skip this step)" % package)
-    tempdir = mkdtemp('conda_skeleton_' + d['filename'])
 
     [output_dir] = args.output_dir
 
-    if not isdir(SRC_CACHE):
-        makedirs(SRC_CACHE)
-
-    try:
```
hargup: @asmeurer This `try` block was used only to `rm_rf` the temporary directory in the `finally` block at [ln685](https://github.com/conda/conda-build/pull/449/files#diff-06936d397c05412b6dd38c5393523f72L685). This block has now been moved to the `get_pkginfo` function. Removing the `try` reduced the indentation of the whole code by one letter and this is leading to this shitty diff. To get a more useful diff use the `--historgram` algorithm, unfortunately github's diffs don't have option to change the diff algorithm used.

asmeurer:
Adding `?w=0` to the end of the GitHub url makes it ignore whitespace changes (https://github.com/conda/conda-build/pull/449/files?w=0).

conda_build/pypi.py
```diff
@@ -419,6 +370,67 @@ def main(args, parser):
...
+        pypiurl = package
+        U = parse_url(package)
+        if U.fragment and U.fragment.startswith('md5='):
+            md5 = U.fragment[len('md5='):]
+        else:
+            md5 = ''
+        # TODO: 'package' won't work with unpack()
+        filename = U.path.rsplit('/', 1)[-1] or 'package'
+
+    return (data, pypiurl, filename, md5)
```
asmeurer: Maybe this should return a dictionary. Then you could just run `d.update(get_download_data)`.

hargup:
> Adding ?w=0 to the end of the GitHub url makes it ignore whitespace changes (https://github.com/conda/conda-build/pull/449/files?w=0).

Cool, I googled it up found this cheat sheet few more interesting features: https://github.com/tiimgreen/github-cheat-sheet#readme :smile:

conda_build/pypi.py
hargup: Won't be as convenient as simply running `d.update(get_download_data)` because not everything returned is a part of dictionary `d`. See https://github.com/conda/conda-build/pull/449/files#diff-06936d397c05412b6dd38c5393523f72R328

asmeurer: What is `data`? Can we add it to `d`?

hargup: This `data` is the pypi release data for the package.

hargup:
@asmeurer There were merge conflicts I've fixed them. Is it good to go? Overall I find the code of skeleton hard to read and maintain, I'll try to simplify it further if I find time next week.

asmeurer:
Yeah, it's grown a bit unwieldy. If you can simplify it, that would be great.

PR #450, opened 2015-06-15
pipbuild: fix make_recipe when version is None

hargup:


hargup:
This part of the code doesn't get executed often because other methods to create the recipe takes precedence. To test this PR apply the following patch and run `pipbuild` with `linecache2`

```patch
diff --git a/conda_build/main_pipbuild.py b/conda_build/main_pipbuild.py
index eb69b39..004c93f 100644
--- a/conda_build/main_pipbuild.py
+++ b/conda_build/main_pipbuild.py
@@ -294,15 +294,9 @@ def make_recipe(package, version):
 
 
 def build_package(package, version=None):
-    if conda_package_exists(package):
-        return 0
     if ' ' in package:
         package, version = package.split(' ')
-    try:
-        directory = build_recipe(package, version=version)
-        dependencies = convert_recipe(directory, package)
-    except RuntimeError:
-        directory, dependencies = make_recipe(package, version)
+    directory, dependencies = make_recipe(package, version)
 
     try:
         print("package = %s" % package)
```

hargup:
@ilanschnell @asmeurer

PR #451, opened 2015-06-15
[WIP] Allow downloading of patches from github.

groutr:
This PR adds a feature to conda to allow downloading pull requests from github as patches which will then be applied to the source tree.

There are a couple of approaches
1) Download patch to a temporary file and apply to source tree
2) Download patch to a regular file in recipe directory and apply to source tree.  This would make the builds more reproducible in case the PR went away (in which case, the patch should be re-examined).

TODO:
- [ ] Add error checking code.  This assume that the correct github url has been given.  If there is an error, it will most likely die.
- [ ] Add tests (when idea is finalized).
- [ ] Start discussion on how to best implement this.

### DIFF
--- a/conda_build/source.py
+++ b/conda_build/source.py
@@ -2,7 +2,6 @@ from __future__ import absolute_import, division, print_function
 
 import os
 import sys
-import urlparse
 from os.path import join, isdir, isfile, abspath, expanduser
 from shutil import copytree, ignore_patterns, copy2
 from subprocess import check_call, Popen, PIPE
--- a/conda_build/source.py
+++ b/conda_build/source.py
@@ -2,6 +2,7 @@ from __future__ import absolute_import, division, print_function
 
 import os
 import sys
+import urlparse
 from os.path import join, isdir, isfile, abspath, expanduser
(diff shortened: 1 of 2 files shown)


conda_build/source.py
```diff
@@ -219,10 +219,20 @@ def parse_bool(s):
...
+    patch -p1 patches.
+
+    All local patches (in the recipe directory) are assumed to be
+    patch -p0 patches
+    """
     print('Applying patch: %r' % path)
     if not isfile(path):
         sys.exit('Error: no such patch: %s' % path)
-
+
```
hargup: There are trailing whitespaces.

rmcgibbo:
Really missing this feature. It would be very convenient to have.

msarahan:
@groutr - any interest in revisiting this one?

groutr:
@msarahan I can revisit this in my spare time.  Have we already implemented the changes in https://github.com/conda/conda-build/pull/1006?

msarahan:
Nope.  Not sure where that stands.  Ilan asked that we make a separate package instead.

beeankha:

PR #475, opened 2015-07-06
Added --post-link option to develop command

sandhujasmine:
implemented --post-link option for develop command
- looks in package directory for *.so and ensures they link against
  libraries in conda environment
- useful for packages that include compiled/cython code that we wish to build in develop mode.
- after ./setup.py develop, user can invoke conda develop --post-link to setup links correctly.


sandhujasmine:
There's an unnecessary for loop that needs cleaning up. I'll submit a new PR.

PR #477, opened 2015-07-07
Improvement of noarch_python.py docs

meawoppl:
The way that python noarch packages work is completely different from standard packages.  AFAIK, the source is the only documentation of this, so I thought it should have comments and docstrings.


meawoppl:
_link.py deserves some similar love as well.

ilanschnell:
Thanks for the improvements.

meawoppl:
Question for ya, is there a reason that windows can not build noarch packages?

PR #480, opened 2015-07-09
Add parameter --msvc {9.0, 10.0} to choose MSVC version

campos-ddc:
This PR adds an option to choose `--msvc` either `9.0` or `10.0`. This option is only available in windows.

I also made some changes to the way `conda-build` finds and configures MSVC, the main change being that it will now try to use `Microsoft SDK 7.1` if it is available (this is needed because `Visual Studio 2010 Express` fails silently when trying to configure an `amd64` environment through `vcvarsall.bat`

--

This might be a controversial one, since pretty much nobody else uses MSVC 2010 for Python 2.7

We do this in my company because we have a lot of C++ code linked with Python libraries, and our code needs newer versions of MSVC. Unfortunately, that means that we have to pretty much build Python and all external again, since the default compiler is always MSVC 2008

### DIFF
--- a/conda_build/config.py
+++ b/conda_build/config.py
@@ -23,6 +23,7 @@ class Config(object):
         '')).replace('.', ''))
     CONDA_NPY = int(os.getenv('CONDA_NPY', '19').replace('.', ''))
     CONDA_R = os.getenv("CONDA_R", "3.2.0")
+    CONDA_MSVC = os.getenv("CONDA_MSVC")
 
     PY3K = int(bool(CONDA_PY >= 30))
 
--- a/conda_build/main_build.py
+++ b/conda_build/main_build.py
@@ -126,6 +126,24 @@ different sets of packages."""
         multiple times to build against multiple versions.""",
         metavar="R_VER",
     )
+
+    if sys.platform == 'win32':
+        p.add_argument(
+            '--msvc',
+            action="append",
+            help="""Set the Microsoft Visual Studio Compiler version used by
+            conda build. Can be passed multiple times to build against multiple
+            versions.
+
+            By default, MSVC 9.0 (2008) is used for Python 2.*, and MSVC 10.0 (2010) for Python 3.*
+
+            Please keep in mind that Python 2.* and all extensions are compiled
+            with MSVC 2008 by default. If you wish to build anything with MSVC 2010,
+            you will have to build all packages (including python) with the same version.""",
+            choices=['9.0', '10.0'],
+            metavar="MSVC_VER",
+        )
+
     add_parser_channels(p)
     p.set_defaults(func=execute)
 
@@ -215,14 +233,19 @@ def execute(args, parser):
         'perl': None,
         'R': None,
         }
+    if sys.platform == 'win32':
+        all_versions['msvc'] = None
+
     conda_version = {
         'python': 'CONDA_PY',
         'numpy': 'CONDA_NPY',
         'perl': 'CONDA_PERL',
         'R': 'CONDA_R',
         }
+    if sys.platform == 'win32':
+        conda_version['msvc'] = 'CONDA_MSVC'
 
-    for lang in ['python', 'numpy', 'perl', 'R']:
+    for lang in all_versions.keys():
         versions = getattr(args, lang)
         if not versions:
             continue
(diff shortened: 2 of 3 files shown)


asmeurer:
I wonder if it would be better to put this in the recipe metadata.

campos-ddc:
Not sure, I can use the same recipe for either 2008 or 2010 as is.

On the other side, sometimes I forget to pass the parameter and get
screwed...

ccordoba12:
Python 3.5 is going to use VS 2015, so we'd need to add a new value when it comes up.

I agree with @campos-ddc that it's better to have this functionality as part of conda-build. By the way, thanks for working on this, it's really, really useful :-)

campos-ddc:
@ccordoba12 should be relatively easy, assuming vs2015 keeps using some sort of 'set env' bat.

bkreider:
>> I wonder if it would be better to put this in the recipe metadata. 

> Not sure, I can use the same recipe for either 2008 or 2010 as is.

Maybe it makes sense in either place?  Any recipe that requires a certain version could specify it in the metadata to make sure it is used.  Set the default in the meta.yaml and override on the cmd-line.

campos-ddc:
@bkreider I think this is the kind of stuff that should be as explicit as possible, since mixing compilers leads to bad times.

Perhaps in your suggestion, if the recipe mandates msvc2010, but you try to use --msvc=8.0, and error should be raised.

In general, I think that this option should be as explicit as possible and I wish there was a way to ensure that all packages in an environment were build with same compiler version (in Windows)

ccordoba12:
> I wish there was a way to ensure that all packages in an environment were build with same compiler version (in Windows)

In that case, the compiler info should be added to meta.yaml for sure.

campos-ddc:
> In that case, the compiler info should be added to meta.yaml for sure.

That still means that we'd need separate recipes for each compiler version. I'd rather have the command line option set some additional info after the build is done.

asmeurer:
I agree that this belongs as functionality of conda-build, since it does set the msvc environment variables. I'm just think it would be better to add a flag to meta.yaml rather than a command line flag, so that the build is more reproducible. Is it common to want to build against multiple msvc compilers? Wouldn't you want to also change some flag in the package as well if you do that (like a feature or at least the build string)?

campos-ddc:
I don't need to build against multiple compilers, so I haven't really thought about that.
If that were the case, indeed I think I'd have to change the recipe anyways, to indicate different build strings. I also agree that it would make things more reproducible, and that's good.

My problem here is, I have to build everything with 2010 + Python 2.7.. if this is only available through modifying recipes, I can no longer just use them from conda-recipes (nor can I contribute my recipes, since almost nobody uses 2010).

What about a configuration in `.condarc`, or any global configuration file for `conda build`?

asmeurer:
I think you may want to do that anyway. I think the best way to  manage that would be to create a `msvc10` [feature](http://conda.pydata.org/docs/building/meta-yaml.html#id1) to separate out the msvc 2010 Python 2.7 and Python 2.7 packages from the 2008 ones. You will also want to use a separate build string. That way, it is easy both for you and for conda to distinguish your packages from the normal ones. I would take the recipes from conda-recipes as a starting point, but you'll want to modify them and keep them separate.

campos-ddc:
Well, I'll probably figure out a way to use the same recipes in our fork (even if I just add some lines to `meta.yaml` just before building).

What do you guys want me to do with this pull-request? Change this to an option in `meta.yaml` and remove the command line option?

asmeurer:
Yes, please do that. There are also merge conflicts with master.

ccordoba12:
@campos-ddc, do you plan to finish this? If not, I'd like to finish your work because I'm needing it for mine :-)

msarahan:
FYI, you can already sort of set this using the --python flag:

--python="2.7" should use 2008
--python="3.x" with x < 5 should use 2010
--python="3.x" with x >= 5 should use 2015

Note: last one may not be in latest conda-build release.

ccordoba12:
Hey @msarahan, yes, I saw your PR :-) But I need finer control over the Visual Studio version used to build extensions.

For example, I needed to use VS 2012 for a Python 2.7 extension of a project I'm working on. And that led me to hack conda-build sources to make it work.

This PR (with some changes of mine) would avoid that.

msarahan:
Cool, definitely a worthwhile addition, then.  One thing that sometimes has come up is how to specify a path to compiler, rather than a particular compiler by name - maybe worth adding here?  If arg matches expected versions, use that version, else try to interpret as path (to vcvarsall.bat?)

campos-ddc:
Hey @ccordoba12, I had plans to finish this with the suggested changes, but it is not a high priority since I have something working for me right now.

If you want to finish this pull request, that'd be great :)

ccordoba12:
@msarahan, I think the compiler's path can be extracted from the registry (I read about it in another conda or conda-build issue this weekend :-). But that's work for another PR :-)

@campos-ddc, ok, I'll take your PR and make my changes on top of it :-)

groutr:
Referencing this for the record's completeness.
https://github.com/conda/conda-build/pull/600

PR #481, opened 2015-07-10
Fix error in conda convert (linux to win package) 

melund:
when package has no hard py-version dependency.

Conda convert only worked correctly when python dependency in the recipe was given as:
python X.X*

Any other specification (python >=X.X) or no version specification would prevent convert convert from translating the file structure "lib/pythonX:X/..." to "Lib/.." on windows.

We don't really need the version number when converting from linux to windows. The python version number is not part of the file structure on windows. It is only really needed if we convert the other way, because we need to know if the files should go into lib/python2.7/... or lib/python3.4/...

### DIFF
--- a/conda_build/convert.py
+++ b/conda_build/convert.py
@@ -161,7 +161,7 @@ path_mapping_bat_proxy = [
     ]
 
 path_mapping_unix_windows = [
-    (r'lib/python{pyver}/', r'Lib/'),
+    (r'lib/python(\d.\d)/', r'Lib/'),
     # Handle entry points already ending in .py. This is OK because these are
     # parsed in order. Only concern is if there are both script and script.py,
     # which seems unlikely
@@ -174,7 +174,7 @@ path_mapping_windows_unix = [
     (r'Scripts/', r'bin/'), # Not supported right now anyway
     ]
 
-pyver_re = re.compile(r'python\s+(\d.\d)')
+pyver_re = re.compile(r'python(?:(?:\s+[<>=]*)(\d.\d))?')
 
 def get_pure_py_file_map(t, platform):
     info = json.loads(t.extractfile('info/index.json').read().decode('utf-8'))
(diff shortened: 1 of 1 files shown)


asmeurer:
I think if a recipe has a version restriction on Python that isn't the default, that means it doesn't depend on the lib/pythonX.Y directory (or else the package was built incorrectly).

conda_build/convert.py
```diff
@@ -201,14 +201,21 @@ def get_pure_py_file_map(t, platform):
...
+    elif pythons[0].group(1):
         pyver = pythons[0].group(1)
 
         mapping = [(re.compile(i[0].format(pyver=pyver)),
             i[1].format(pyver=pyver)) for i in mapping]
+    else:
+        # No python version dependency was specified
+        # Only a problem when converting from windows to unix, since 
+        # the python version is part of the folder structure on unix. 
+        if source_type == 'win' and dest_type == 'unix':
```
asmeurer: Really if `dest_type == 'unix'` and `source_type` is anything. As I said, if a Unix conda package depends on more than one major Python version and has files in lib/pythonX.Y it is probably incorrect (unless the package author really wanted the stuff to only work for the one Python version).

melund:
@asmeurer, Thanks for looking at this. The problem was converting from unix to windows. I am sure that this pull request doesn't fix everything. 

I am not familiar with how python organizes the files on unix. Are they always placed under lib/PythonX,Y/ ? 

Is the file structure different if the python package doesn't depend on a specific version?

asmeurer:
Unix is always lib/pythonX.Y, yes. The file structure is the same. A package not depending on a specific version would have to meant that it doesn't use that directory. Python's `sys.path` only has paths that have the major Python version in them.

msarahan:
Sorry for the delay here @melund.  I think I grok this, and agree that it is a good thing.  Merging for the 2.0 release.

yoavram:
Hi, I have a problem with this fix - the regular expression `pyver_re` catches non Python dependencies, specifically, any package with a name starting with `python` such as [python-dateutil](https://pypi.python.org/pypi/python-dateutil).

My suggestion is to change the regex by moving the `\s+` outside of the `(?:...)`:

```py
pyver_re = re.compile(r'python\s+(?:(?:[<>=]*)(\d.\d))?')
```

msarahan:
Sure, good catch.  Please go ahead and open a PR, so that you get credit for this improvement.

PR #482, opened 2015-07-11
bugfix: symlinks not made relative if build_prefix contains a symlink

mjuric:
This PR fixes an issue where absolute symlinks in the built binaries are not recognized as such and made relative if conda's build environment path contains a symlink.

This is because the path to which the tested symlink is pointing to is resolved with `os.path.realpath()` (which expands _all_ symlinks that are in it), but `config.build_prefix` isn't. I've encountered this bug on a system where `/home` is a symlink to `/nsf/home` (and I've had conda installed in my home directory).

### DIFF
--- a/conda_build/post.py
+++ b/conda_build/post.py
@@ -347,12 +347,13 @@ def check_symlinks(files):
     if readlink is False:
         return  # Not on Unix system
     msgs = []
+    real_build_prefix = realpath(config.build_prefix)
     for f in files:
-        path = join(config.build_prefix, f)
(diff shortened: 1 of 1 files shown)


mjuric:
Hi all -- is there any way I can help getting this bug fix reviewed/merged (it's fairly small -- three lines)?

It's biting us hard on our primary development machine, so I'd love to see it in the next `conda-build` release.

asmeurer:
Looks good, and the existing symlink tests pass.

mjuric:
Thanks!

Issue #428, opened 2015-06-02
Allow simple regex with `always_include_files`


jakirkham:
At present, it appears an exact filename must be supplied here and if they are not found a failure occurs. However, this is a problem if a library is kept via this method (e.g. `dylib` on mac and `so` on linux). Alternatively, if I want to save a library and don't care about the version, this would be very helpful, as well.

asmeurer:
You can use selectors to handle that case (see for instance https://github.com/conda/conda-recipes/blob/master/libgcc/meta.yaml). Matching a glob wouldn't be a bad idea, though.

jakirkham:
Thanks for the suggestion. I did try selectors and they worked fine.

However, I think regex could help with make `meta.yml` a bit more terse. For example, if I have `lib/libquadmath.so.0.0.0`, it follows that I will want `lib/libquadmath.so.0` and `lib/libquadmath.so`.

ilanschnell:
With #484 being merged, I think we can close this issue.

PR #484, opened 2015-07-11
add regex to always included files

loriab:
A possible fix to #428

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -14,6 +14,7 @@ import sys
 import tarfile
 from os.path import exists, isdir, isfile, islink, join
 import fnmatch
+import re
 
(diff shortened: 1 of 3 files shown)


ilanschnell:
Thanks for your contribution.

loriab:
You're welcome. Thanks for showing me the ropes.

asmeurer:
I'd rather match a glob than a regex. I'll send a PR.

asmeurer:
https://github.com/conda/conda-build/pull/499

loriab:
Glob it is. Ilan preferred regex, so that's the why of the original formulation. I think this is out of my purview now, so just let me know if I actually need to do anything.

PR #490, opened 2015-07-16
Added conda develop --post-link

sandhujasmine:
added --post-link option for develop command
    - looks in package directory for *.so and ensures they link against
      libraries in conda environment
    - build_prefix points to current conda environment
    - 'path' of executable (*.so) is not in build_prefix

### DIFF
--- a/conda_build/main_develop.py
+++ b/conda_build/main_develop.py
@@ -14,7 +14,7 @@ import fnmatch
 from conda.cli.common import add_parser_prefix, get_prefix
 from conda.cli.conda_argparse import ArgumentParser
 from conda_build.main_build import args_func
-from conda_build.post import mk_relative_osx, mk_relative_linux
+from conda_build.post import mk_relative_osx
 
 from conda.install import linked
 
@@ -84,7 +84,7 @@ def relink_sharedobjects(pkg_path):
     for b_file in bin_files:
         if sys.platform.startswith('linux'):
             # mk_relative_linux(f, rpaths=m.get_value('build/rpaths', ['lib']))
-            raise NotImplementedError("setting rpath on linux is incomplete")
+            raise NotImplementedError("unclear what to do for this on linux")
         elif sys.platform == 'darwin':
             mk_relative_osx(b_file, develop=True)
 
--- a/conda_build/main_develop.py
+++ b/conda_build/main_develop.py
@@ -90,6 +90,16 @@ def relink_sharedobjects(pkg_path):
             mk_relative_osx(b_file, develop=True)
 
 
+def write_to_conda_pth(sp_dir, pkg_path):
+    '''
+    site-packages directory for current environment
+    append pkg_path to conda.pth
+    '''
+    with open(join(sp_dir, 'conda.pth'), 'a') as f:
+        f.write(pkg_path + '\n')
+        print("added " + pkg_path)
+
+
 def execute(args, parser):
     prefix = get_prefix(args)
     if not isdir(prefix):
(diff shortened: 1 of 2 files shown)


asmeurer:
The code looks good. Do you have an example package that I can test it against?

conda_build/main_develop.py
```diff
@@ -32,6 +36,12 @@ def main():
         nargs='+',
         help="Path to the source directory."
     )
+    p.add_argument(
+                   '-l', '--post-link',
```
asmeurer: Any reason to not do this by default

sandhujasmine: You can test against: https://github.com/sandhujasmine/CythonExample

One reason to not do this by default is more flexibility. In this case since conda develop doesn't yet build the extensions, user can do ./setup.py develop, then invoke --post-link. It gives the flexibility to use --post-link with setup commands.

Once we add a --build_ext command to conda develop, it should by default do the post-link after building extensions.

asmeurer: Oh I see the logic you've implemented now. I am thinking of the use-case where someone just types `conda develop` on a source directory and it just works. The way it is now, you have to have two steps, `conda develop` and `conda develop --post-link`. 

Would it make sense to make this enabled by default, and have `--no-post-link` and `--no-pth` flags to turn the different things off?

sandhujasmine: Currently, for conda develop it should just work as you suggested.
If --post-link is explicitly given, then it only does the relinking, so:

if not args.post_link:
    write_to_conda_pth()

# next step is to do the relinking for shared objects
relink_sharedobjects()

I just tested it on my example, by doing the following in CythonExample/:
./setup.py build_ext --inplace
conda develop .

The conda develop appends to conda.pth, then it does the relinking.

asmeurer: Ah, I see. In that case, the option name should probably just be called `--no-pth`.

sandhujasmine: aah - ok. So --no-pth-file would mean, don't append to conda.pth file. I can update.

sandhujasmine:
changed --post-link option to --no-pth-file. Left the short option as '-l'

asmeurer:
Why did you leave the short letter? --no-pth-file doesn't even contain an l.

sandhujasmine:
well, I liked having a shorter version as well. I understand the -l isn't in there, but it signifies 'linking' and I couldn't use '-n' or '-p'. I could do -np, but I thought the '-l' for link made more sense - at least to me, but maybe cause I'm used thinking of it as --post-link

sandhujasmine:
changed short letter option from -l to -npf

asmeurer:
I tested this and it looks like it works. 

A thought. Right now, the LC_RPATH is written relative to the source, like

```
$otool -l polygon/cy_polygon.so | tail
 datasize 72
Load command 13
      cmd LC_DATA_IN_CODE
  cmdsize 16
  dataoff 21336
 datasize 0
Load command 14
          cmd LC_RPATH
      cmdsize 64
         path @loader_path/../../../../anaconda/envs/test/lib/ (offset 12)
```

Probably it would be better to write an absolute path in this case. That way, the source can be moved around and still work. We normally use relative paths because we don't want to bake in the name of the environment, but that's not the case for conda develop. 

At any rate, I'm going to merge what is here.

sandhujasmine:
I thought about the relative vs absolute path as well and I agree there is no need for relative path in develop mode.

In develop mode, a relative path is certainly not very readable since source code could be anywhere in conda env. Though if you move the source code around, you should just build again in develop mode; but absolute is easier to read/understand.

I was thinking of adding a boolean rel_path=True or False that gets passed from mk_relative_osx() all the way to the callback function, osx_ch_link(). It can default to True and get set to False by main_develop. osx_ch_link() can do the appropriate thing when relinking based on rel_path value.

asmeurer:
Readability is probably not a huge issue (most people won't be looking at the LC_RPATH), but if it's just a simple fix to make it so that the source can be moved around, we should. The rel_path flag sounds like a good plan. To be clear, relative paths *should* be used when linking two different libraries in the same source together (in that case, an absolute path would prevent the source from being moved around), but not when linking against something in the conda environment.

asmeurer:
I think specifically, on the LC_RPATH should be absolute. A better flag name would be `rpath_rel_path` or something.

PR #512, opened 2015-07-29
develop options --build_ext, --clean_build_ext

sandhujasmine:
1. Added options to conda develop to build extensions
    `conda develop --build_ext pkg_path`
    Looks for setup.py in pkg_path, then invokes python setup.py build_ext --inplace
    No exception handling here, so if this fails then user should be alerted since their setup.py is broken. If setup.py not found, then simply exit.

2. Added option to clean and build extensions
    `conda develop --clean_build_ext pkg_path`
    This will invoke python setup.py clean && python setup.py build_ext --inplace

3. Both options above will fix the linkages in extensions after building so they first search in current conda environment's lib/. for dependencies.

4. Also updated write_to_conda_pth() so a path is only appended to conda.pth if the path does not already exist in the file.

### DIFF
--- a/conda_build/main_develop.py
+++ b/conda_build/main_develop.py
@@ -209,16 +209,17 @@ Error: environment does not exist: %s
 
     for path in args.source:
         pkg_path = abspath(expanduser(path))
-        setup_py = get_setup_py(pkg_path)
 
-        if args.clean:
-            clean(setup_py)
-            if not args.build_ext:
-                sys.exit()
+        if args.clean or args.build_ext:
+            setup_py = get_setup_py(pkg_path)
+            if args.clean:
+                clean(setup_py)
+                if not args.build_ext:
+                    sys.exit()
 
-        # build extensions before adding to conda.pth
-        if args.build_ext:
-            build_ext(setup_py)
+            # build extensions before adding to conda.pth
+            if args.build_ext:
+                build_ext(setup_py)
 
         if not args.no_pth_file:
             write_to_conda_pth(sp_dir, pkg_path)
--- a/conda_build/main_develop.py
+++ b/conda_build/main_develop.py
@@ -7,7 +7,7 @@
 from __future__ import absolute_import, division, print_function
 
 import sys
-from os.path import join, isdir, abspath, expanduser
+from os.path import join, isdir, abspath, expanduser, exists
 from os import walk
 import fnmatch
 
@@ -15,6 +15,7 @@ from conda.cli.common import add_parser_prefix, get_prefix
 from conda.cli.conda_argparse import ArgumentParser
 from conda_build.main_build import args_func
 from conda_build.post import mk_relative_osx
+from conda_build.utils import _check_call
 
 from conda.install import linked
 
@@ -42,6 +43,14 @@ This works by creating a conda.pth file in site-packages."""
                    help=("Relink compiled extension dependencies against "
                          "libraries found in current conda env. "
                          "Do not add source to conda.pth."))
+    p.add_argument(
+                   '-b', '--build_ext',
+                   action='store_true',
+                   help=("Invoke clean, then build extensions inplace: "
+                         "python setup.py clean && "
+                         "python setup.py build_ext --inplace; "
+                         "add to conda.pth; relink runtime libraries to "
+                         "environment's lib/."))
     add_parser_prefix(p)
     p.set_defaults(func=execute)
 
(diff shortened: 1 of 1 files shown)


conda_build/main_develop.py
```diff
@@ -42,6 +43,21 @@ def main():
...
                          "Do not add source to conda.pth."))
+    p.add_argument(
+                   '-b', '--build_ext',
+                   action='store_true',
+                   help=("Build extensions inplace, invoking: "
+                         "python setup.py build_ext --inplace; "
+                         "add to conda.pth; relink runtime libraries to "
+                         "environment's lib/."))
+    p.add_argument(
+                   '-c', '--clean_build_ext',
```
asmeurer: Maybe just separate `--clean` into a separate argument and run it like `conda develop --clean --build_ext`

asmeurer:
Just that one comment. The rest looks good (although I haven't tested it yet).

conda_build/main_develop.py
sandhujasmine: That's what I started with but wanted to keep --clean and --build_ext together just to be explicit that conda clean and conda develop clean are very different

I'll need to support following which will then just do ./setup.py clean
`conda develop --clean pkg_path`

asmeurer: Looking at the code that should be simple, no? You just need to split [this](https://github.com/conda/conda-build/pull/512/files#diff-8ebb521eb386d609915cfbd151033d13R176) block into a separate function. 

It's only a few extra characters to type `conda develop --clean --build_ext` vs. `conda develop --clean_build_ext` and it seems much cleaner (no pun intended).

sandhujasmine: Looks like I need to delve a bit more into argparse so I can allow:
`conda develop --clean --build_ext pkg_path'

but not allow:
`conda develop --clean pkg_path`

Again, this is just me wanting to be explicit that clean in develop mode applies to build_ext and should not be confused with `conda clean`. I'll look into this and update

asmeurer: Why do you want to disallow `conda develop --clean pkg_path`? I think it should work. 

At any rate, argparse doesn't let you specify this kind of logic. You'd have to check it yourself.

sandhujasmine: It's redundant, will invoke setup.py clean then exit. And it has different definition of clean than standard conda so was trying to be explicit. But is probably easier to write code to allow clean by itself rather than write custom code to parse. I can update and it is documented clearly

conda_build/main_develop.py
```diff
@@ -94,14 +110,81 @@ def relink_sharedobjects(pkg_path, build_prefix):
...
+    setup_py = get_setup_py(pkg_path)
+
+    if clean_:
+        # first call setup.py clean
+        cmd = ['python', setup_py, 'clean']
+        _check_call(cmd)
+        print("Completed: " + " ".join(cmd))
+        print("===============================================")
+
+    # next call setup.py develop
```
asmeurer: Is this comment wrong?

sandhujasmine: yep - it's needs updating.

conda_build/main_develop.py
sandhujasmine: Here's the problem. Currently, the logic for develop is as follows:
$ conda develop <options> path

if build_ext is in options: then invoke build
if np_pth_file is in options: then skip adding the path to conda.pth

At the end of logic based on input options, look for *.so in path and ensure they use conda environment for loading libraries at runtime.

If user tries any of the following:
$ conda develop path
$ conda develop --clean --build_ext path
$ conda develop --build_ext path

It's fine to add path to conda.pth unless --no-pth-file is given, and relink shared objects in path. However, if user does:
$ conda develop --clean path

it should only invoke $ python setup.py clean. It is seems unclean to add a path to conda.pth or try to relink shared objects when clean is invoked. I don't think the default behavior of conda develop that adds to conda.pth and relinks shared objects should not apply for --clean by itself. It doesn't break anything but doesn't seem right to do anything other than clean when conda develop --clean path is given.

sandhujasmine:
removed --clean_build_ext and added --clean

sandhujasmine:
Only look for setup.py if invoked with --clean or --build_ext commands since they require setup.py. If setup.py is not found, then simply exit.

Previous behavior for conda develop path was to append the path to conda.pth, and relink shared objects regardless of setup.py's existence and this now works the same way.

PR #519, opened 2015-08-02
Rename binstar -> anaconda

ilanschnell:
This PR is similar to https://github.com/conda/conda/pull/1458.  Also, only externally visible changes are made, in particular the option `--no-binstar-upload` is renamed to `--no-anaconda-upload` (where the first is also being accepted for now.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -16,20 +16,20 @@ Installation
 Building Your Own Packages
 --------------------------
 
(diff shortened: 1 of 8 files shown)


asmeurer:
What about config.binstar_upload

ilanschnell:
`config.binstar_upload` is internal conda code, which we decided to leave unchanged for now.  Only end user facing changes are being focused on right now.

Issue #211, opened 2014-09-20
Failure after TEST START on Python 2.7


bashtage:
Everything looks ok, but it tries (and fails) to clean up a binary in `_build`, while not running the test.

It is worth noting that this is only broken on Python 2.7 - Python 3.3 and 3.4 work as expected, run the tests suit, and upload to binstar.

```
BUILD END: arch-1.0-np18py27_0
TEST START: arch-1.0-np18py27_0
An unexpected error has occurred, please consider sending the
following traceback to the conda GitHub issue tracker at:

    https://github.com/conda/conda-build/issues

Include the output of the command 'conda info' in your report.


Traceback (most recent call last):
  File "C:\ANACONDA\Scripts\conda-build-script.py", line 4, in <module>
    sys.exit(main())
  File "C:\ANACONDA\lib\site-packages\conda_build\main_build.py", line 111, in main
    args_func(args, p)
  File "C:\ANACONDA\lib\site-packages\conda_build\main_build.py", line 309, in args_func
    args.func(args, p)
  File "C:\ANACONDA\lib\site-packages\conda_build\main_build.py", line 297, in execute
    build.test(m, verbose=not args.quiet)
  File "C:\ANACONDA\lib\site-packages\conda_build\build.py", line 397, in test
    rm_rf(config.build_prefix)
  File "C:\ANACONDA\lib\site-packages\conda\install.py", line 162, in rm_rf
    shutil.rmtree(path)
  File "C:\ANACONDA\lib\shutil.py", line 247, in rmtree
    rmtree(fullname, ignore_errors, onerror)
  File "C:\ANACONDA\lib\shutil.py", line 247, in rmtree
    rmtree(fullname, ignore_errors, onerror)
  File "C:\ANACONDA\lib\shutil.py", line 247, in rmtree
    rmtree(fullname, ignore_errors, onerror)
  File "C:\ANACONDA\lib\shutil.py", line 252, in rmtree
    onerror(os.remove, fullname, sys.exc_info())
  File "C:\ANACONDA\lib\shutil.py", line 250, in rmtree
    os.remove(fullname)
WindowsError: [Error 5] Access is denied: u'C:\\ANACONDA\\envs\\_build\\Lib\\site-packages\\markupsafe\\_speedups.pyd'
```

PR #521, opened 2015-08-03
Use trash on windows when deleting environments

rmcgibbo:
@asmeurer, this also fixes the issue in #520. I watched the .trash folder, and it does get removed.

### DIFF
--- a/tests/test-recipes/metadata/jinja2/meta.yaml
+++ b/tests/test-recipes/metadata/jinja2/meta.yaml
@@ -1,0 +1,14 @@
+package:
+  name: conda-build-test-jinja2-in-recipe
+  version: 1.0
+
+build:
+  number: 0
+
+requirements:
+    build:
+        - jinja2
+
+test:
+  commands:
+    - echo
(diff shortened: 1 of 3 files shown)


tests/test-recipes/metadata/jinja2/meta.yaml
```diff
@@ -0,0 +1,14 @@
+package:
+  name: conda-build-test-jinja2-in-recipe
+  version: 1.0
+
+build:
+  number: 0
+
+requirements:
+    build:
+        - jinja2
```
asmeurer: You should add python here so that it builds against the same Python version as conda build.

asmeurer:
I would like to get the trash stuff in rm_rf itself, but this will work for now.

asmeurer:
I went ahead and just fixed the python thing myself.

rmcgibbo:
Sweet.

Korijn:
I just submitted issue #524 yesterday which seems closely related to #520. Do you think this PR will fix that issue as well? When will this PR be included in a release so that I can test?

asmeurer:
Yes I think this will fix that as well.

jasongrout:
Thanks for getting this in.  I was pulling my hair out trying to debug the problem with _speedups.pyd.

jasongrout:
It looks like this might be the real solution to https://github.com/conda/conda-build/issues/211 too.

PR #533, opened 2015-08-08
set field to {} when None

ilanschnell:
This small change allows sections in `meta.yaml` to be empty.  This makes it easier when having selectors.  For example:
```
build:
  number: 1    [win]
```
On Linux this will get preprocessed to:
```
build:
```
Which this PR now allows.


ilanschnell:
Thanks

PR #536, opened 2015-08-10
Correctly remove egg directories on Windows

asmeurer:
Depends on conda/conda#1513.


ilanschnell:
Conda 3.17.0 was just released which includes #1513

PR #537, opened 2015-08-11
.git can be a file

### LINKED ISSUE #529: GIT_DESCRIBE_TAG isn't set when the .git is a file
Sometimes the .git directory is actually a file.  This happens, for example, when you have a git submodule.  This causes this test to fail incorrectly: https://github.com/conda/conda-build/blob/master/conda_build/environ.py#L36

Unfortunately, the .git file might point to a real git repo in a relative URL, so just moving the directory to the conda build directory can break it being a git repo.

asmeurer:
Fixes #529.

### DIFF
--- a/conda_build/environ.py
+++ b/conda_build/environ.py
@@ -33,7 +33,7 @@ def get_git_build_info(src_dir):
     env = os.environ.copy()
     d = {}
     git_dir = join(src_dir, '.git')
-    if os.path.isdir(git_dir):
+    if os.path.exists(git_dir):
         env['GIT_DIR'] = git_dir
     else:
         return d


jasongrout:
In the case of submodules (that I saw), the .git file uses a *relative* path.  Copying the directory to a work directory breaks the relative path, of course.  So things are still broken in that respect.  I'm not sure we should be mucking about with the contents of a .git file, but if it's appropriate, perhaps changing the path to an absolute path would be the thing to do.

jasongrout:
However, this is an improvement for when the .git file uses an absolute path.  Thanks!

PR #545, opened 2015-08-15
Add ability to add license file in info/license.txt

ilanschnell:
Using (in `meta.yaml`) the field `about/license_file`, allows adding a license file to conda package under `info/license.txt`.  Possible values are:
  * `True`: will search the root of the source tree for the following files (in that order): `LICENSE`, `LICENSE.txt`, `license`, 'license.txt`
  * `<filename>`: you can specify the filename of the license file explicitly.
  * `Flase`: meaning no license file will be included under `info/license.txt`, which is the default.

For example, if the source tree has `LICENSE` in its root,
```
build:
  license_file: True
```
will add this file under `info/license.txt` in in the conda package.

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -150,6 +150,20 @@ def create_info_files(m, files, include_recipe=True):
             else:
                 shutil.copy(src_path, dst_path)
 
+    license_file =  m.get_value('build/license_file')
+    if license_file:
+        filenames = 'LICENSE', 'LICENSE.txt', 'license.txt'
+        if license_file is True:
+            for fn in filenames:
+                src = join(source.get_dir(), fn)
+                if isfile(src):
+                    break
+            else:
+                sys.exit("Error: could not locate license file")
+        else:
+            src = join(source.get_dir(), license_file)
+        shutil.copy(src, join(config.info_dir, 'license.txt'))
+
     readme = m.get_value('about/readme')
     if readme:
         src = join(source.get_dir(), readme)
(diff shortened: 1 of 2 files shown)


jasongrout:
Great, thanks!  Can we add something to the docs of the package format, as well as the docs of the meta.yaml and the examples?

jasongrout:
I wrote up some of the documentation at https://github.com/conda/conda-docs/pull/182.  Looking at the meta.yaml, I think maybe it makes more sense to have the license_file option be in the about section, next to the about/license key.  What do you think?

ilanschnell:
You are right, there is already `about/readme`, which is very similar.  I'll change it.

jasongrout:
There's a readme?  Great! It's not documented in http://conda.pydata.org/docs/building/meta-yaml.html#about-section.

conda_build/build.py
```diff
@@ -150,6 +150,22 @@ def create_info_files(m, files, include_recipe=True):
             else:
                 shutil.copy(src_path, dst_path)
 
+    license_file = m.get_value('about/license_file')
+    if license_file:
+        filenames = 'LICENSE', 'LICENSE.txt', 'license', 'license.txt'
+        if license_file is True:
+            for fn in filenames:
+                src = join(source.get_dir(), fn)
+                if isfile(src):
```
jasongrout: Can we print a comment to stdout here about which file we are going to use for the license file?

PR #559, opened 2015-08-28
add license_family to index.json

ilanschnell:
Adding (optionally)
```
about:
  license_family: MIT
```
will add the `license_family` in the `info/index.json` file of the conda package.


asmeurer:
We should add the list of accepted families.

ilanschnell:
I've added the check for accepted families.

PR #568, opened 2015-08-31
Make config.PY3K a property

### LINKED ISSUE #565: "conda build --python x.y" does not work as advertised
Using the ``CONDA_PY`` environment variable works as expected. If it is set to ``34``, conda builds a Python 3.4 package using Visual Studio 2010. 

However, ``conda build --python 3.4`` does not produce the same result. Conda picks Visual Studio 2008, because ``config.PY3`` is incorrectly set to 0 and ``msvc_env_cmd()`` relies on this constant to determine the correct version of VS: 

  https://github.com/conda/conda-build/blob/master/conda_build/windows.py#L64

asmeurer:
That way it gets updated properly when CONDA_PY gets updated.

Fixes #565.

### DIFF
--- a/conda_build/config.py
+++ b/conda_build/config.py
@@ -24,7 +24,9 @@ class Config(object):
     CONDA_NPY = int(os.getenv('CONDA_NPY', '0').replace('.', '')) or None
     CONDA_R = os.getenv("CONDA_R", "3.2.1")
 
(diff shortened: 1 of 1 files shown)


asmeurer:
It was also making the `PY3K` build environment variable be wrong. Just tested it and that is fixed. The original issue should be fixed as well.

PR #572, opened 2015-09-01
Fixed crash when using `--skip-existing` while conda-bld is empty

frol:
The following command crashes if `$CONDA_ROOT/conda-bld/$PLATFORM` does not exist.

```
$ conda build --skip-existing conda.recipe
```

### DIFF
--- a/conda_build/main_build.py
+++ b/conda_build/main_build.py
@@ -255,6 +255,7 @@ def execute(args, parser):
     import shutil
     import tarfile
     import tempfile
+    from os import makedirs
     from os.path import abspath, isdir, isfile
 
     from conda.lock import Locked
@@ -308,8 +309,9 @@ def execute(args, parser):
                     (conda_version[lang], version))
 
     if args.skip_existing:
-        if exists(config.bldpkgs_dir):
-            update_index(config.bldpkgs_dir)
+        if not isdir(config.bldpkgs_dir):
+            makedirs(config.bldpkgs_dir)
+        update_index(config.bldpkgs_dir)
         index = build.get_build_index(clear_cache=True,
             channel_urls=channel_urls,
             override_channels=args.override_channels)
--- a/conda_build/main_build.py
+++ b/conda_build/main_build.py
@@ -308,7 +308,8 @@ def execute(args, parser):
                     (conda_version[lang], version))
 
     if args.skip_existing:
-        update_index(config.bldpkgs_dir)
+        if exists(config.bldpkgs_dir):
+            update_index(config.bldpkgs_dir)
         index = build.get_build_index(clear_cache=True,
             channel_urls=channel_urls,
             override_channels=args.override_channels)


conda_build/main_build.py
```diff
@@ -255,6 +255,7 @@ def execute(args, parser):
     import shutil
     import tarfile
     import tempfile
+    from os import makedirs
```
frol: I have no idea why main_build.py has so weird imports, but I have just followed the style.

conda_build/main_build.py
```diff
@@ -308,6 +309,8 @@ def execute(args, parser):
                     (conda_version[lang], version))
 
     if args.skip_existing:
+        if not isdir(config.bldpkgs_dir):
+            makedirs(config.bldpkgs_dir)
```
frol: Conda code is a mess, you have to copy things over and over. There are two other places with the same code:
https://github.com/conda/conda-build/blob/master/conda_build/build.py#L273
https://github.com/conda/conda-build/blob/master/conda_build/build.py#L486 (I have no idea if makedirs is necessary there)

frol: Also, check `config.bldpkgs_dir` implementation: https://github.com/conda/conda-build/blob/master/conda_build/config.py#L109 (This causes the issues when `conda-bld/noarch` path is created, but `get_index` tries to reach `conda-bld/noarch` and `conda-bld/$PLATFORM` always, but fails only if `$PLATFORM` is not found: https://github.com/conda/conda/blob/master/conda/fetch.py#L124)

conda_build/main_build.py
asmeurer: You mean the fact that it's inside the function? There's probably not a good reason for it.

conda_build/main_build.py
asmeurer: Should we move the makedirs call into `config`?

asmeurer:
What's here looks good. Merging.

conda_build/main_build.py
frol: I don't see a point in importing only specific functions from `os`. I would rather prefer to write `os.path.exists()` instead of `exists()` because the latter seems to me like a local function. Thus, I would vote for `import os` instead of a bunch of `from os.path import ...`

conda_build/main_build.py
frol: @asmeurer Moving `makedirs` into `config` will look awkward. I would probably a several changes:
1. add a new function something along the lines `update_local_index` that will check `noarch` and `$PLATFORM` build indecies.
2. update conda's fetch_repodata so it is not hardcoded against `'noarch'`, but rather has an option to silently (but with debug logging) skip non-existent repos, and in the worst case log a warning instead of rasing an exception if a given repo is unreachable.

conda_build/main_build.py
asmeurer: We do that a lot in conda, because os and os.path functions are used so ubiquitously. I know that this differs from common usage where only the module is imported. This way makes for much less verbose code that makes use of many os and os.path functions.

conda_build/main_build.py
asmeurer: +1 to both of those.

PR #578, opened 2015-09-09
allow jinja2 templates to be located in current conda environment

ukoethe:
As requested, this is an update of PR #575. (I nonetheless keep #575 open for now so that we can continue our conversation there.)


ukoethe:
@asmeurer : May I remind you of this little PR?

ukoethe:
Would someone from the core team please care to comment on this PR? It is very critical for my work. Thank you very much! @msarahan @ilanschnell 

To clarify the PR, I repeat its __rationale__ here:

Jinja templates are a powerful way to configure `meta.yaml`. Right now, `conda build` searches for these templates in two locations: `<conda_root>/Lib/site-packages/conda_build/templates` (for site-specific configurations) and `$RECIPE_DIR` (for package-specific configurations). I propose to add the current environment (as specified in the environment variable `$CONDA_ENV_PATH`) as an additional search location to support environment-specific configurations.

__Use case:__ Suppose one wants to build a set of packages with exactly identical toolset settings (same compiler options, tool versions etc.). This can be achieved by performing the build in a suitably prepared conda environment. However, conda does not yet provide a portable query method for information about the current environment within `meta.yaml`. The present PR adds this capability. For example, let's say that a jinja-template `toolset-info.yaml` is placed into the environment's root directory upon installation of some toolset package. This template can now be loaded in `meta.yaml` via
```yaml
{% import '$CONDA_ENV_PATH/toolset-info.yaml' as toolset %}
```
and later be accessed via variables like `{{ toolset.name }}` or `{{ toolset.version }}` etc.

ukoethe:
I just noticed that the environment variable `CONDA_ENV_PATH` was renamed into `CONDA_DEFAULT_ENV` recently and updated the PR accordingly.

ilanschnell:
There is no variable `CONDA_ENV_PATH`, only `CONDA_ENVS_PATH` (the environments location directory), which is separate from `CONDA_DEFAULT_ENV`.  None of these have been renamed.

ukoethe:
> There is no variable CONDA_ENV_PATH, only CONDA_ENVS_PATH (the environments location directory), which is separate from CONDA_DEFAULT_ENV. None of these have been renamed.

Strange. It worked for me, and I'm quite certain that I never defined `CONDA_ENV_PATH` myself. Anyway, in the latest commit I use `CONDA_DEFAULT_ENV` instead.

msarahan:
+1 from me.  The travis build failure is unrelated.  Missing OpenSSL package that we haven't built on python 3.3 yet.

msarahan:
Discussed with @ilanschnell and @groutr - we think this is a very nice addition.  Thanks!

ukoethe:
Thanks for merging, @msarahan! See https://github.com/conda/conda/issues/1589#issuecomment-150154106 for a usage example.

PR #582, opened 2015-09-10
Setup for Python 3.5; --force32 option

msarahan:
This also includes a fix to normalize line endings to UNIX style when Gnuwin patch.exe is used.


asmeurer:
Seems legit.

PR #588, opened 2015-09-13
Fix 'imports' tests when package python is not the same as conda-root python

### LINKED ISSUE #587: test:imports doesn't work if conda-root python version doesn't match the package's required python
I'm using the python-2.7 version of Miniconda.  IIUC, I should still be able to use conda-build to generate python-3 packages.  However, I get a failure for this test case:

```yaml
package:
 name: my-py3-library
 version: 0.1

requirements:
 build:
  - python 3.5*

 run:
  - python 3.5*

test:
 imports:
  - sys
```

```
$ conda build my-py3-library 
Removing old work directory
BUILD START: my-py3-library-0.1-py35_0
Solving package specifications: .

    openssl:    1.0.1k-1     
    pip:        7.1.2-py35_0 
    python:     3.5.0-0      
    readline:   6.2-2        
    setuptools: 18.1-py35_0  
    sqlite:     3.8.4.1-1    
    tk:         8.5.18-0     
    wheel:      0.24.0-py35_0
    xz:         5.0.5-0      
    zlib:       1.2.8-0      

Linking packages ...
[      COMPLETE      ]|###################################################################################################################################################| 100%
Removing old work directory
Copying /Users/bergs/Documents/workspace/ilastik-build-conda/my-py3-library to /miniconda/conda-bld/work
Package: my-py3-library-0.1-py35_0
source tree in: /miniconda/conda-bld/work
number of files: 0
Fixing permissions
Fixing permissions
BUILD END: my-py3-library-0.1-py35_0
TEST START: my-py3-library-0.1-py35_0
Solving package specifications: ..
Error: Unsatisfiable package specifications.
Generating hint: 
[      COMPLETE      ]|###################################################################################################################################################| 100%

Hint: the following packages conflict with each other:
  - my-py3-library 0.1 py35_0
  - python 2.7*

Use 'conda info my-py3-library' etc. to see the dependencies for each package.
```

The `python 2.7*` spec requirement comes from [this line](https://github.com/conda/conda-build/blob/274cd1218722bab72acfa462edaf96427b554f04/conda_build/build.py#L537).

That function expects the python spec to be included in the tests section explicitly, even though it was already included in the `requirements:build` section.

The test case above can be fixed by extending the `test:` section:

```yaml
test:
 imports:
  - sys

 requires:
  - python 3.5*
```

...but IIUC, `conda-build` shouldn't need that redundant spec.

A proposed fix for this can be found in ~~PR #588~~ PR #590.

stuarteberg:
Before auto-adding python to the test requirements, also check `requirements/run` in addition to `test/requires`
Fixes #587.

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -523,23 +523,24 @@ def test(m, verbose=True, channel_urls=(), override_channels=False):
     else:
         rm_rf(config.build_prefix)
         rm_rf(config.test_prefix)
-    specs = ['%s %s %s' % (m.name(), m.version(), m.build_id())]
+    test_specs = ['%s %s %s' % (m.name(), m.version(), m.build_id())]
 
     # add packages listed in test/requires
-    specs_include_python = False
-    for spec in m.get_value('test/requires', []):
-        specs.append(spec)
-        if spec.startswith('python ') or spec == 'python':
-            specs_include_python = True
+    test_specs += m.get_value('test/requires', [])
+
+    # Do we need to add the python spec ourselves?
+    test_and_run_specs = test_specs + m.get_value('requirements/run', [])
+    python_specs = list(filter(lambda spec: spec.startswith('python ') or spec == 'python', test_and_run_specs))
+    specs_include_python = len(python_specs) > 0
 
     if py_files and not specs_include_python:
         # as the tests are run by python, we need to specify it
-        specs += ['python %s*' % environ.get_py_ver()]
+        test_specs += ['python %s*' % environ.get_py_ver()]
     if pl_files:
         # as the tests are run by perl, we need to specify it
-        specs += ['perl %s*' % environ.get_perl_ver()]
+        test_specs += ['perl %s*' % environ.get_perl_ver()]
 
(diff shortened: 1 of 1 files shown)


stuarteberg:
I guess this PR is an improvement, but it's still not perfect: If `python` isn't explicitly listed as a `run:` dependency, but merely implicitly included by some other `run:` dependency, the error will still occur.  Instead of checking `meta.yaml` for `python`, it would be better to check the fully resolved list of `run:` dependencies.  But for now, this PR is still an improvement.

stuarteberg:
BTW, it isn't obvious to me why we would ever want python to be automatically added to the spec list.  If the recipe lists `imports:` in the `test:` section, doesn't that imply that it should have included python as a `run:` dependency already?  If not, that's a bug in the recipe, right?

stuarteberg:
>BTW, it isn't obvious to me why we would ever want python to be automatically added to the spec list.

In light of this observation, I implemented an alternative PR: #589

stuarteberg:
Aaaand, nevermind there's a simpler fix: #590.  Closing this PR.

PR #589, opened 2015-09-13
Never auto-add spec for 'python' to the test environment.

stuarteberg:
Recipes are required to list it in either `requirements:run` or `test:requires`.

This is an alternative 'fix' for #587 (instead of PR #588).  But I'm probably missing something -- I don't understand why `python` was being auto-added to the test env in the first place (see [comments](https://github.com/conda/conda-build/pull/588#issuecomment-139900742) in #588.)

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -526,15 +526,8 @@ def test(m, verbose=True, channel_urls=(), override_channels=False):
     specs = ['%s %s %s' % (m.name(), m.version(), m.build_id())]
 
     # add packages listed in test/requires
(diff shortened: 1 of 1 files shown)


stuarteberg:
>I don't understand why python was being auto-added to the test env in the first place

The failing Travis test helped me understand -- `run_test.py` is sometimes used even for non-python packages.  (Python is, after all, a nicer scripting language than `sh`.)

The easy fix for this is not to require a *specific* version of python in the spec list: See #590.  Closing this PR.

PR #596, opened 2015-09-16
Add uninstall option for develop with unit tests

sandhujasmine:
Include option to uninstall a package path that was previously added via
conda develop. If path is not found in conda.pth, it prints a message
to inform that package was not installed via conda develop.

Added a test module for unit testing some functions of develop command

### DIFF
--- a/tests/test_main_develop.py
+++ b/tests/test_main_develop.py
@@ -1,0 +1,111 @@
+'''
+Simple tests for testing functions in main_develop module
+'''
+import os
+import shutil
+from os.path import dirname, join, exists
+
+from conda_build.main_develop import uninstall, write_to_conda_pth
+
+import pytest
+
+
+@pytest.fixture(scope="session")
+def sp_dir(request):
+    '''
+    create site-packges/ directory in same place where test is located. This
+    is where tests look conda.pth file. It is a session scoped fixture and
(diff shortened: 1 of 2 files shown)


tests/test_main_develop.py
```diff
@@ -0,0 +1,111 @@
...
+    '''
+    pth = join(sp_dir, 'conda.pth')
+    if exists(pth):
+        os.remove(pth)
+
+    return pth
+
+
+# Note: following list is data used for testing - do not change it
+_path_in_dev_mode = ['/Users/jsandhu/Documents/projects/CythonExample',
```
asmeurer: Will these tests work outside of your machine?

sandhujasmine: Yes! That was the idea - they are simple but useful. The CI should pass these tests.

asmeurer:
Does the test actually write to the filesystem? If so, where?

sandhujasmine:
It creates a site-packages/ directory inside tests, so tests/site-packages and all it is testing is that conda.pth gets created in here that it doesn't write duplicate paths and that it removes paths when uninstall is invoked. In the end the finalizer deletes site-packges from tests

asmeurer:
I see now. Looks good.

PR #600, opened 2015-09-23
Force MSVC version

groutr:
This allows a recipe to force which version of MSVC it builds with.

### DIFF
--- a/conda_build/windows.py
+++ b/conda_build/windows.py
@@ -58,13 +58,16 @@ def msvc_env_cmd(override=None):
         program_files = os.environ['ProgramFiles(x86)']
     else:
         program_files = os.environ['ProgramFiles']
-
-    if config.PY3K and config.use_MSVC2015 or override == 'msvc14':
+        
+    if config.PY3K and config.use_MSVC2015:
         version = '14.0'
-    elif config.PY3K or override == 'msvc10':
+    elif config.PY3K:
         version = '10.0'
     else:
         version = '9.0'
+
+    if override not None:
+        version = override
 
     vcvarsall = os.path.join(program_files,
                              r'Microsoft Visual Studio {version}'.format(version=version),
--- a/conda_build/windows.py
+++ b/conda_build/windows.py
@@ -53,7 +53,7 @@ def fix_staged_scripts():
         os.remove(join(scripts_dir, fn))
 
 
-def msvc_env_cmd():
+def msvc_env_cmd(override=None):
     if 'ProgramFiles(x86)' in os.environ:
         program_files = os.environ['ProgramFiles(x86)']
     else:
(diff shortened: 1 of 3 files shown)


ccordoba12:
Thanks @groutr and sorry for not creating this PR before :-) Got busy with other stuff.

I have a couple of comments:

1. I'd like to pass any Visual Studio version to conda-build. But if I understand things correctly, your PR would only let me override between `msvc10` and `msvc14`.
2. You need to set

        DISTUTILS_USE_SDK=1
        MSSdk=1

    in the environment, to override `distutils` Visual Studio detection and use the compiler you want for Python extensions.

conda_build/windows.py
```diff
@@ -61,19 +61,18 @@ def msvc_env_cmd():
...
         else:
-            vcvarsall = os.path.join(program_files,
-                                     r'Microsoft Visual Studio 10.0'
-                                     r'\VC\vcvarsall.bat')
+            vcvarsall = vcvarsall.format(version='10.0')
     else:
-        vcvarsall = os.path.join(program_files,
-                                 r'Microsoft Visual Studio 9.0'
-                                 r'\VC\vcvarsall.bat')
+        vcvarsall = vcvarsall.format(version='9.0')
```
ccordoba12: According to my comments, this block could be rewritten as:

```python
vs_versions = {'msvc9': '9.0', 'msvc10': '10.0', 'msvc11': '11.0',
               'msvc12': '12.0', 'msvc14': '14.0'}
if override:
    vcvarsall = vcvarsall.format(version=vs_versions[override])
    os.environ['DISTUTILS_USE_SDK'] = '1'
    os.environ['MSSdk'] = '1'
elif config.PY3K:
    if config.use_MSVC2015:
        vcvarsall = vcvarsall.format(version='14.0')
    else:
        vcvarsall = vcvarsall.format(version='10.0')
else:
    vcvarsall = vcvarsall.format(version='9.0')
```

ccordoba12:
Pinging @msarahan and @asmeurer. Those who are working on Windows packages really need this functionality :-)

conda_build/windows.py
msarahan: +1, this is cleaner.

msarahan:
+1 - nice work.  Please make edits according to @ccordoba12 suggestions, but otherwise ready to go.

groutr:
@ccordoba12, changing it so you set msvc_compiler like `msvc_compiler: 10.0`

groutr:
I will test the latest changes in my windows vm either tonight or tomorrow.

ccordoba12:
+1 from me. It seems ok :-)

groutr:
merging.

asmeurer:
Can you make a pull request to conda-docs that adds this to the docs. And you should note that this is only for advanced users who know what they are doing, since for most people, changing the compiler from the default (at least for Python extensions) will break things.

PR #601, opened 2015-09-23
Move path prepending to function for uniformity

msarahan:
The most important thing about this PR is adding Library/bin to PATH for Windows at test time.  It was there at build time already.  This makes build and test phases append bin paths as necessary to the appropriate prefix.


asmeurer:
@msarahan this PR is a mess now. Why do you keep force pushing?

msarahan:
Sorry, just trying to fix some mistakes and keep things clean.  I'll clean these up.

asmeurer:
Well clearly rebasing does the opposite of "keeping thing clean". I would recommend against it. 

Sorry, but seeing force pushing make me nervous, especially when you're pushing directly to origin.

msarahan:
OK.  I'll avoid it in the future, but I'd like to clean up this mess.

I will always keep things on my branches, FWIW - but this is a bad habit I picked up from scikit-image.

msarahan:
@groutr This takes care of packages built with Conda (for me).  Please look this over and see if it is what you need.  This is orthogonal to setting PATH for users with installers.  That's a separate PR.

groutr:
@msarahan, I'll review this tomorrow.

conda_build/scripts.py
```diff
@@ -58,3 +58,11 @@ def create_entry_points(items):
         os.mkdir(bin_dir)
     for cmd, module, func in iter_entry_points(items):
         create_entry_point(join(bin_dir, cmd), module, func)
+
+
+def prepend_bin_path(env, prefix):
+    env['PATH'] = prefix + os.pathsep + env['PATH']
```
ilanschnell: We should not prepend prefix itself, unless on Windows

msarahan: This is the way it was done before, just moved into a function.  See https://github.com/conda/conda-build/pull/601/files#diff-ac55d744935b1f37a3369258b88e8f1fL548

If you want it changed, then let's go ahead and change it, though...

msarahan: Never mind, I see that this was a subtle difference between https://github.com/conda/conda-build/pull/601/files#diff-ac55d744935b1f37a3369258b88e8f1fL548 and had no equivalent in conda_build/environ.py

PR #604, opened 2015-09-24
[WIP] Git shallow clones

groutr:
Using a git shallow clone can save a lot of bandwidth in some cases when cloning large repositories.

### DIFF
--- a/conda_build/metadata.py
+++ b/conda_build/metadata.py
@@ -224,7 +224,7 @@ def _git_clean(source_meta):
 FIELDS = {
     'package': ['name', 'version'],
     'source': ['fn', 'url', 'md5', 'sha1', 'sha256', 'path',
-               'git_url', 'git_tag', 'git_branch', 'git_rev',
+               'git_url', 'git_tag', 'git_branch', 'git_rev', 'git_depth',
                'hg_url', 'hg_tag',
                'svn_url', 'svn_rev', 'svn_ignore_externals',
                'patches'],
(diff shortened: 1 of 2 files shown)


jakirkham:
:+1:

pitrou:
Looking at the git clone man page, you may want to add the `--no-single-branch` option?
Also, is there a reason this option should not be enabled by default?

Last question :-) I see mention of a "git cache" in the source, but I cannot find a `git_cache` directory here. Where is it supposed to be located?

groutr:
I haven't considered that option.  I'll look at it soon.

I haven't thought about making this default.  I'm not even sure what a 
good default would be.  Git has no mechanism for fetching back to a 
tag.  A lot of conda builds rely on tags, which if we only did a shallow 
clone would probably not fetch back far enough, especially for active 
projects.

The git_cache is stored in the $PREFIX/conda-bld directory.

jakirkham:
@pitrou, I can think of a few reasons one might not want to have this enabled by default.

One easy example would be if you determine your version information from git. If you don't include enough commits, you might not reach a tag and your version information will be indeterminate or based off of a commit alone, which would be problematic.

Personally, I would prefer this be an option rather than the default.

groutr:
@jakirkham, I think I agree with you not having this the default.

kalefranz:
LGTM

jakirkham:
Realized this didn't get documented. So adding a brief example in PR ( https://github.com/conda/conda-docs/pull/565 ).

PR #606, opened 2015-09-24
Fix: 'conda build' crashes when git output contains non-printable characters

ukoethe:
The fix is easy: when writing to `sys.stdout`, strings should be encoded with `sys.stdout.encoding`.


ilanschnell:
Thanks for the fix.  Makes sense to me.

asmeurer:
I hope it works.

jakirkham:
Ran into an issue due, in part, to this when building a docker container. Further explanation and a fix here ( https://github.com/conda/conda-build/pull/614 ).

PR #613, opened 2015-10-01
BLD: add jinja2 to the run requires

tacaswell:
This bites me every single time I set up a new computer

### DIFF
--- a/conda_build.recipe/meta.yaml
+++ b/conda_build.recipe/meta.yaml
@@ -19,6 +19,7 @@ requirements:
   run:
     - python
     - conda
+    - jinja2
     - patchelf      [linux]
 
 test:


jakirkham:
Fixes https://github.com/conda/conda-build/issues/655

msarahan:
Thanks for this.  This is oddly enough not the right place to fix it, because I don't actually know when this recipe gets used.  The right place to fix it is in conda-recipes, and in our internal build system recipe.  I will see to it that we fix it in those places.

msarahan:
I take that back.  There is no conda-build on conda-recipes.  The internal conda-build package appears to have jinja2 as a runtime requirement.  Our git history doesn't go back very far due to a rename (moving internal build recipes to conda recipes).  I can't tell when it was added.  I can tell you that conda-build-1.18.1 **should** have jinja2 as a runtime requirement.  If it doesn't, then let us know.  We need to bump the build number and build out, so that this is included.

I am going to merge this PR.

tacaswell:
Thanks,  1.18.1 does have jinja2 as a runtime dep.

ccordoba12:
+1000 from me too! One less thing to worry about :-)

PR #614, opened 2015-10-02
Fix `None` encoding bug

jakirkham:
Turns out `sys.stdout.encoding` can be `None`, as well. This normally happens when `stdout` is redirected. This fixes that bug and simplifies the logic around that region to hopefully avoid its recurrence.

For the curious ones, here's a bit more on why this happens ( http://www.macfreek.nl/memory/Encoding_of_Python_stdout ).

This encoding problem can be reproduced by a trivial docker container. Building this container will print `None`.

    FROM python:2

    RUN python -c "from sys import stdout as so; print(so.encoding)"

### DIFF
--- a/conda_build/source.py
+++ b/conda_build/source.py
@@ -152,9 +152,10 @@ def git_info(fo=None):
                 ('git status', True)]:
         p = Popen(cmd.split(), stdout=PIPE, stderr=PIPE, cwd=WORK_DIR, env=env)
         stdout, stderr = p.communicate()
-        encoding = locale.getpreferredencoding() or 'utf-8'
+        encoding = locale.getpreferredencoding()
         if not fo:
-            encoding = sys.stdout.encoding or 'utf-8'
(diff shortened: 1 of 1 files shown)


msarahan:
Thanks!

asmeurer:
I've got a feeling this isn't the last we'll see of this bug.

msarahan:
Yes, I have already encountered it in a few other packages.

jakirkham:
I am similarly worried. Hence the linked article.

It might be worth monkey patching `stdout` and `stderr` somewhere commonly accessed as they suggest. Example for Python 2 below. 

    sys.stdout = codecs.getwriter('utf-8')(sys.stdout, 'strict')

Python 3 would be like so.

    sys.stdout = codecs.getwriter('utf-8')(sys.stdout.buffer, 'strict')

This would help guarantee that there is a specific encoding for these streams no matter what.

PR #615, opened 2015-10-02
Update noarch_python.py

scw:
Fix awkward wording.


ilanschnell:
I don't see how the wording is awkward.  You just changed "currently not" to "not currently".

scw:
For most English grammar, negation follows the [auxillary or special verb](https://en.wikipedia.org/wiki/English_grammar#Negation) used to connect the action, but it's fine to keep, just thought it improved readability this way.

PR #620, opened 2015-10-05
FIX: build/script section of meta.yaml on Windows

gabrielcnr:
The section `build/script` of the `meta.yaml` was being ignored completely on Windows. No errors/warning messages were given to the user, the build process simply stopped there.

I'm proposing to create the `bld.bat` with the commands contained on that section and proceed with the building process normally.

### DIFF
--- a/conda_build/windows.py
+++ b/conda_build/windows.py
@@ -118,6 +118,15 @@ def build(m):
 
     src_dir = source.get_dir()
     bld_bat = join(m.path, 'bld.bat')
+
+    if not exists(bld_bat):
+        script = m.get_value('build/script', None)
+        if script:
+            if isinstance(script, list):
+                script = '\n'.join(script)
+            with open(bld_bat, 'w') as bf:
+                bf.write(script)
+
     if exists(bld_bat):
         with open(bld_bat) as fi:
             data = fi.read()


msarahan:
Thanks for this.  This is an improvement.  Might it be even better to append any commands in here to any existing bld.bat?  Or do you find it more intuitive to simply not require creation of bld.bat, but rather migrate towards commands in Python?  I don't have a strong opinion, but will raise this at Continuum.

gabrielcnr:
I think this is a valid point, if you take into consideration how the "test" section works, if you have a `run_test.py` inside your conda recipe's dir, the contents of that file will be appended to the `run_test.py` that is generated by conda-build on that phase. (More info: `conda_build/create_test.py`).

I tried to minimize the impact of the change, but I understand that this could be inconsistent with other parts of the framework. What do you think? I'm happy to make further changes to check if `bld.bat` already exists, then append the commands listed inside `build/script` section to that file.

Would that work OK with all those GOTO usually present in batch scripting?

--Gabe

gabrielcnr:
any updates?

jakirkham:
Fixes https://github.com/conda/conda-build/issues/682

jakirkham:
It would be nice if the solution here is consistent with whatever is already done on OS X and Linux. I believe the proposal is, but may need to take a second look.

patricksnape:
There seems to be some logic [here](https://github.com/conda/conda-build/blob/fa67dac2dc6fec614bf5f51a435c57959fc2c657/conda_build/build.py#L472) about this. From what I can see that logic seems to match the Unix logic?

msarahan:
fixed by https://github.com/conda/conda-build/pull/799 - closing.

PR #630, opened 2015-10-08
Add text to alexTest

alexG53090:
Made a new file called alexTest and added some text.


ilanschnell:
Thanks for the PR, but we don't need this feature.  Keep up the great work!

PR #640, opened 2015-10-17
This is probably a typo.

ericdill:
The check here has no effect. It is probably a typo, though I do not
know the code base well enough to be sure.


msarahan:
I can't imagine this is anything but a typo.  Thanks for catching it.

PR #654, opened 2015-10-28
Drop extra mention of `jinja2`

jakirkham:

PR #660, opened 2015-11-02
environ: Fix NPY_VER for versions >= 1.10 (Should be '1.10', not '1.1.0')

stuarteberg:
Fixes a tiny bug in `environ.py`.  The `NPY_VER` environment variable was incorrect for numpy versions >= 1.10 (`NPY_VER`was evaluating to `1.1.0` instead of `1.10`).

### DIFF
--- a/conda_build/environ.py
+++ b/conda_build/environ.py
@@ -21,7 +21,13 @@ def get_py_ver():
     return '.'.join(str(config.CONDA_PY))
 
 def get_npy_ver():
-    return '.'.join(str(config.CONDA_NPY or ''))
+    if config.CONDA_NPY:
+        # Convert int -> string, e.g.
+        #   17 -> '1.7'
+        #   110 -> '1.10'
+        conda_npy = str(config.CONDA_NPY)
+        return conda_npy[0] + '.' + conda_npy[1:]
+    return ''
 
 def get_stdlib_dir():
     return join(config.build_prefix, 'Lib' if sys.platform == 'win32' else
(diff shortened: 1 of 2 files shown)


stuarteberg:
OK, I added a second little commit here. If the user never specified a `--numpy` version in the first place, then we should leave `NPY_VER` out of `environ` entirely (don't just add an empty string).

I want to use `NPY_VER` in `meta.yaml` (via the template mechanism).  But I would like to see an error if `NPY_VER` doesn't exist.  For example:

```
requirements:
  build:
    - numpy {{ environ.get('NPY_VER', 'NPY_VER-UNDEFINED') }}*
```

This way, if I forget to provide `--numpy=1.10`, then I get a nice error message:

```
$ conda build my-package
Removing old work directory
BUILD START: gala-0.3.0-py27_2
Error: No packages found in current osx-64 channels matching: numpy NPY_VER-UNDEFINED*
```

But without this commit, `NPY_VER` appears valid (it's just empty), so I might not notice that something went wrong.

stuarteberg:
Travis was complaining, so I added one more commit just to make `pyflakes` happy.  Now Travis is passing.  I don't think that failure was related to my changes, though...

msarahan:
Sorry about the Travis thing.  I appreciate your fix.  I intend to put in a better one - remove the unused import of jinja2 completely.  This was broken in https://github.com/conda/conda-build/pull/654 - your comment line at least ensures that future people don't remove this line again from not understanding why it is there.

PR #661, opened 2015-11-04
metadata: Create jinja2 environment with 'strict' mode for undefined values

stuarteberg:
For conda recipes, I think it is safer/desirable to use `jinja2` in "strict" mode, which raises an error if the template attempts to use a variable that isn't defined.

For instance, this recipe contains a typo in the `run` requirement:

```
requirements:
  build:
    - python {{PY_VER}}*
  run
    - python {{PY_VEER}}*
```

Without strict mode, the typo is ignored, and the template simply evaluates to nothing. (Not good.)

Here's another example:

```
requirements:
  build:
    - numpy {{NPY_VER}}*
```

No typos there, but if the user forgets to provide the `--numpy=` setting to `conda build`, then `NPY_VER` will be undefined.  Ideally, an error should be emitted in this case, but you won't see any error without strict mode.

This PR has the potential to break existing recipes, if they somehow rely template variables that might be undefined.  But my guess is that any such recipes were somehow broken anyway.

If anyone can think of a legitimate use-case for permitting undefined template variables in conda recipes, then don't accept this PR, obviously.

Edit: Here are the docs for jinja's strictness settings:
http://jinja.pocoo.org/docs/dev/api/#undefined-types

### DIFF
--- a/conda_build/metadata.py
+++ b/conda_build/metadata.py
@@ -312,8 +312,10 @@ def get_contents(meta_path):
 
     template = env.get_or_select_template(filename)
 
-    contents = template.render(environment=env)
-    return contents
+    try:
+        return template.render(environment=env)
+    except jinja2.TemplateError as ex:
+        sys.exit("Error: Failed to parse jinja template in {}:\n{}".format(meta_path, ex.message))
 
 
 def handle_config_version(ms, ver):
--- a/conda_build/metadata.py
+++ b/conda_build/metadata.py
@@ -306,7 +306,7 @@ def get_contents(meta_path):
         env_loader = jinja2.FileSystemLoader(conda_env_path)
         loaders.append(jinja2.PrefixLoader({'$CONDA_DEFAULT_ENV': env_loader}))
 
-    env = jinja2.Environment(loader=jinja2.ChoiceLoader(loaders))
+    env = jinja2.Environment(loader=jinja2.ChoiceLoader(loaders), undefined=jinja2.StrictUndefined)
     env.globals.update(ns_cfg())
     env.globals.update(context_processor())
 


groutr:
Sounds like a good idea to me.  @msarahan, any thoughts?

msarahan:
+0.98

My only concern is providing helpful error output when people have these template problems. Do we know what the messages look like? Are they readily interpretable? 

If not, can we add some error handling to this PR to make them readable? I would also be OK with documenting the errors that show up,  just to give people hints on where to look for a problem. 

If so, please merge.

stuarteberg:
>My only concern is providing helpful error output when people have these template problems. Do we know what the messages look like?

That's a great point.  At the moment, it's clear what the problem is, but they are too severe-looking.  They tell the user to file a new issue, which is not desirable:

```
$ conda build tifffile

  File "/miniconda/bin/conda-build", line 5, in <module>
    sys.exit(main())
  File "/miniconda/lib/python2.7/site-packages/conda_build/main_build.py", line 190, in main
    args_func(args, p)
  File "/miniconda/lib/python2.7/site-packages/conda_build/main_build.py", line 468, in args_func
    args.func(args, p)
  File "/miniconda/lib/python2.7/site-packages/conda_build/main_build.py", line 347, in execute
    m = MetaData(recipe_dir)
  File "/miniconda/lib/python2.7/site-packages/conda_build/metadata.py", line 349, in __init__
    self.parse_again()
  File "/miniconda/lib/python2.7/site-packages/conda_build/metadata.py", line 357, in parse_again
    self.meta = parse(get_contents(self.meta_path))
  File "/miniconda/lib/python2.7/site-packages/conda_build/metadata.py", line 304, in get_contents
    contents = template.render(environment=env)
  File "/miniconda/lib/python2.7/site-packages/jinja2/environment.py", line 989, in render
    return self.environment.handle_exception(exc_info, True)
  File "/miniconda/lib/python2.7/site-packages/jinja2/environment.py", line 754, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/Users/bergs/Documents/workspace/ilastik-build-conda/tifffile/meta.yaml", line 17, in top-level template code
    string: np{{CONDA_NPY}}py{{CONDA_PY}}_{{build_num}}
jinja2.exceptions.UndefinedError: 'CONDA_NPY' is undefined
```

I'll see what I can do about that, and report back.

tacaswell:
:+1: on this from a user point of view

stuarteberg:
OK, now the error message is clearer:

```
$ conda build my-package
Error: Failed to parse jinja template in /path/to/recipes/my-package/meta.yaml:
'FOOBAR' is undefined
```
I wish I could offer more information (i.e. the line number within `meta.yaml` that is causing trouble), but even in `jinja2`'s own traceback the line numbers seem to be inaccurate.  Looks like that's a known issue: https://github.com/mitsuhiko/jinja2/issues/276

Fortunately, `meta.yaml` files tend to be pretty simple, so it shouldn't be hard to locate the error on your own.

msarahan:
Fantastic.  Thanks for your work on this.  Merging.

stuarteberg:
Oh crap....  Remind me again: which road is the one that's "paved with good intentions"?

I've discovered a problem with this: essentially, I broke `GIT_` variables in `jinja2` templates, because they aren't defined at first.  I've proposed a fix for this in #663.

PR #663, opened 2015-11-05
Allow GIT_* variables in jinja templates

stuarteberg:
(This fixes the issue I [mentioned](https://github.com/conda/conda-build/pull/661#issuecomment-153950535) in thread #661, after it was merged.)

Now that #661 is merged, templates in `meta.yaml` are not permitted to use undefined variables.  But I have discovered a special case that's now broken.

The problem is related to the fact that `conda-build` actually parses `meta.yaml` **twice**.  Before the first pass, variables like `GIT_DESCRIBE_HASH` are undefined.  Then, in the first pass, variables from the `source:` section are read, so they can be added to the `environ`.  So, by the time we run the [second pass](https://github.com/conda/conda-build/blob/84bb9a6d738ec3328be33c4d5442c25897155974/conda_build/build.py#L395),  `GIT_DESCRIBE_HASH` *might* have a definition (if the recipe actually uses git), so the recipe should be allowed to use it.

In case that wasn't clear, here's a summary of what conda-build does:

- run jinja on `meta.yaml`
- parse `meta.yaml` (first pass)
- now that we know `source/git_tag`, etc., Update `environ['GIT...']` 
- re-run jinja on `meta.yaml`, with updated `environ`
- parse `meta.yaml` again (second pass)

In earlier versions of `conda-build`, the `GIT_` variables were always initialized to `''`.  But that led to problems like #322.  It would also permit the user to accidentally use a `GIT_` variable in a recipe that doesn't even use git (not good).  Hence, the default values for these were removed in 417a0743853249aa261a7c792f0a5eb46d95286d.

This PR implements a workaround: We should provide 'fake' values (empty strings) for `GIT_` variables during the first pass (only), just so `jinja2` doesn't complain about them. But don't add them during the second pass -- they will already be defined in `environ`, and if they aren't, then the recipe shouldn't be using them.

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -392,7 +392,7 @@ def build(m, get_src=True, verbose=True, post=None, channel_urls=(),
             source.provide(m.path, m.get_section('source'))
             # Parse our metadata again because we did not initialize the source
             # information before.
-            m.parse_again()
+            m.parse_again(provide_empty_git_variables=False)
 
         print("Package:", m.dist())
 
--- a/conda_build/environ.py
+++ b/conda_build/environ.py
@@ -36,6 +36,13 @@ def get_stdlib_dir():
 def get_sp_dir():
     return join(get_stdlib_dir(), 'site-packages')
 
+def get_git_build_info_variable_names():
+    return ["GIT_DESCRIBE_TAG",
+            "GIT_DESCRIBE_NUMBER",
+            "GIT_DESCRIBE_HASH",
+            "GIT_BUILD_STR",
+            "GIT_FULL_HASH"]
+
 def get_git_build_info(src_dir):
     env = os.environ.copy()
     d = {}
@@ -70,6 +77,8 @@ def get_git_build_info(src_dir):
         d['GIT_BUILD_STR'] = '{}_{}'.format(d[key_name('NUMBER')],
                                             d[key_name('HASH')])
 
+    for key in get_git_build_info_variable_names():
+        assert key in d
     return d
 
 def get_dict(m=None, prefix=None):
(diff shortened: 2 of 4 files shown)


msarahan:
Thanks for catching this so quickly.  Are there any other variables that might become "first-pass defaults?"  Is it worthwhile to generalize this more now, rather than down the road?  I don't know what else might be there, but perhaps changing the references from only GIT_* stuff to something like "first_pass_defaults" might make the intention more clear to future people?

I think you're actually in a better position to say than us: you have used the jinja2 templating feature probably more than we have.  If you think making it more general is a good idea, please do so.  Otherwise, I'll merge.

stuarteberg:
>Are there any other variables that might become "first-pass defaults?"

I tried to think of this last night, but I didn't see any.  The comment before the second pass implies that only the `source` section adds new variables, and the only variables from there are `GIT_*`.  I suppose conda could add analogous variables for `hg` and `svn` at some point, but they don't exist right now.

>Is it worthwhile to generalize this more now, rather than down the road?  

I suppose an alternative to the fix in this PR would be to disable jinja's "strict" mode for the first pass, and then re-enable it for the second pass.  The annoying thing is that there is not guaranteed to be a second pass at all.  (If the recipe deals only with local files, then there's no need for a `source` section at all.)  Still, that solution would be simpler by far.

>I don't know what else might be there, but perhaps changing the references from only GIT_* stuff to something like "first_pass_defaults" might make the intention more clear to future people?

Agree.  If we stick with this approach, then renaming the variables makes sense.  I suppose I could also drop a few comments here and there...  I'll push some updates later today.

stuarteberg:
OK, I think I've finally licked this one.  Closing this PR in favor of a better one: #666.

PR #665, opened 2015-11-05
conda_build/source.py: give symlinks=True to copytree()

pkgw:
This copies symbolic links as links. This should be more efficient, and I have
one pathological example of a SVN repository that contains a symlink to '.'
that results in a endlessly recursive copy that eventually throws an exception
from the path getting too long.

### DIFF
--- a/conda_build/source.py
+++ b/conda_build/source.py
@@ -221,7 +221,7 @@ def svn_source(meta):
         assert isdir(cache_repo)
 
     # now copy into work directory
-    copytree(cache_repo, WORK_DIR)
+    copytree(cache_repo, WORK_DIR, symlinks=True)
     return WORK_DIR
 
 


pkgw:
Ping?

msarahan:
Thanks for this, and thanks for the ping.  I'd like to merge this, but I'm much more familiar with Windows.  Can you identify a recipe for me in conda-recipes that has symlinks, so that I can rebuild that package with this PR and verify that things still work?  Sorry to put more burden on you, but I am very short of time.

pkgw:
My test case is my [casa-nrao-casacore](https://github.com/pkgw/conda-recipes/tree/master/recipes/casa-nrao-casacore) recipe. This patch only affects the initial unpacking of the source, so you could temporarily delete the dependencies on my other nonstandard packages in the `meta.yaml` and still test the patch's functionality. I'm pretty sure it won't work on Windows, since the whole point is that you can't approximate the repository's pathological symlink with a copy; but this software isn't buildable on Windows, and the patch definitively **does** take it from broken to working on Linux and OS X.

Although, ugh, that repo seems to have some kind of temporary issue with its SSL certificate, so to even get to the checkout you might need to manually run an SVN command on the URL in question and tell SVN to permanently remember the certificate. Also, be warned that the checkout is quite large.

msarahan:
Thanks for that clarification.  Unfortunately, I'm very bandwidth limited over the holidays.  If you can suggest any other smaller package (or even just create a simple one with exactly this error-causing condition), I'll take a look right away.  Otherwise, I need to wait until I return from backwoods Missouri (Jan 5).

pkgw:
OK. I am also very short of time and don't see a clever way to spelunk through public SVN repositories for ones containing weird symlinks, so I guess this will have to wait until you're back.

Alternatively this patch is literally flipping a single boolean argument to a standard Python library function, and I'm quite sure it won't cause any errors on Windows since the source tree won't contain symlinks either ([cf. implementation](https://hg.python.org/cpython/file/2.7/Lib/shutil.py#l185)), and it very concretely fixes my builds and make it so I no longer need to monkey-patch `conda-build` just to build my packages. So if it were my tree I'd be comfortable just merging the branch now. But I understand that there are processes to follow, etc.

msarahan:
I trust that this fixes your problem, I just need to be sure that it doesn't create new ones. Windows isn't relevant here, you're right. It is only relevant in so far as it is my expertise more than *nix, which unfortunately requires me to depend more on tests to ensure that I'm not messing anyone up by merging this.

I'll revisit this when I get back. @ilanschnell @groutr @ccordoba12 if any of you have time, please take a look.

ilanschnell:
Is the copy speed the only concern here?  From what I see, having `symlinks=False` is the safer approach.  @pkgw is is important to fix your pathological example?  Are there real world example for which this fix is important (besides speed)?

pkgw:
I cannot build the package I mentioned above — which is important to me in production work — without this fix, because there's a symlink in a directory that points to itself and so the copy code recurses infinitely and crashes.

ilanschnell:
As this fix, only affects SVN (which isn't widely used anymore), I think there is little risk of breaking peoples recipes.

pkgw:
Thanks!

PR #621, opened 2015-10-05
Add a method to shell out and execute a command through subprocess

alexandrucoman:
Hello,

This pull requests tries to remove some redundant blocks of code related to the `subprocess.Popen`.
I hope it will be useful. 

Thanks,
Alex

PR #666, opened 2015-11-05
metadata: Use jinja strict mode for second pass only

stuarteberg:
This PR addresses the problem described in #663, but in a cleaner and more general way.

The new solution is simple:

- Use non-strict mode for the first pass (permit undefined variables)
- Always perform a second pass (even if there's no `source` section in the recipe)
- Enforce strict mode in the second pass

BTW, I'm using a custom subclass of `jinja2.Undefined` so that I can be extremely UN-strict in the first pass: Things like `{{ MY_UNDEFINED_VAR[:2] }}` and `{{ float(MY_UNDEFINED_VAR) }}` are permitted, which would normally be rejected by jinja even in non-strict mode.

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -390,9 +390,11 @@ def build(m, get_src=True, verbose=True, post=None, channel_urls=(),
 
         if get_src:
             source.provide(m.path, m.get_section('source'))
-            # Parse our metadata again because we did not initialize the source
-            # information before.
-            m.parse_again()
+
+        # Parse our metadata again because we did not initialize the source
+        # information before.
+        # By now, all jinja variables should be defined, so don't permit undefined vars.
+        m.parse_again(permit_undefined_jinja=False)
 
         print("Package:", m.dist())
 
--- a/conda_build/metadata.py
+++ b/conda_build/metadata.py
@@ -275,11 +275,14 @@ def check_bad_chrs(s, field):
             sys.exit("Error: bad character '%s' in %s: %s" % (c, field, s))
 
 
-def get_contents(meta_path):
+def get_contents(meta_path, permit_undefined_jinja):
     '''
     Get the contents of the [meta.yaml|conda.yaml] file.
     If jinja is installed, then the template.render function is called
-    before standard conda macro processors
+    before standard conda macro processors.
+
+    permit_undefined_jinja: If True, *any* use of undefined jinja variables will
+                            evaluate to an emtpy string, without emitting an error.
     '''
     try:
         import jinja2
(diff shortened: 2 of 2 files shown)


msarahan:
Thanks, I like this solution.  Ping @groutr - I think we're ready to merge this, but would appreciate one more set of eyes.

jakirkham:
From our chat earlier, this sounds like the right solution. So, +1.

stuarteberg:
Pinging @groutr and @msarahan again:

Whoops, looks like this PR somehow fell off the table.  When you guys merged #661 (thanks again), a bug was introduced:  `GIT_` variables became unusable in `meta.yaml` files.  That bug is now present in conda-build 1.18.2  This PR fixes the bug.

frol:
@msarahan Haven't you noticed that #661 had a MAJOR regression yet?! PLEASE, merge this and release a new version!

/cc @groutr @asmeurer @ilanschnell

jakirkham:
@asmeurer is no longer at Continuum. So, maybe you shouldn't ping him, @frol. That being said, yes this PR needs some attention.

groutr:
I think we should add some tests for this PR before merging.

stuarteberg:
@frol:
>Haven't you noticed that #661 had a MAJOR regression yet?!

Whoa now, take a deep breath... ;-)

For now, you can workaround this issue by simply defining the necessary environment variables manually before invoking `conda build`.  You can give them nonsense values, since they'll be replaced during the metadata second-pass parse.  For example:

`GIT_FULL_HASH="" conda build mypackage`

@groutr:
>I think we should add some tests for this PR before merging.

I'll try to find time for that, but there will be some delay.  It's not immediately clear where to add the test, since (as far as I can tell), the test suite does not exercise the jinja template feature at all so far.  Also, since `jinja2` is an *optional* requirement of `conda-build`, the existing test suite doesn't even install it.  (Right?)  I guess I'll need to add `jinja2` to `.travis.yml` and/or `.binstar.yml` and/or `tests/install_miniconda.py`.  Is that right?  Anywhere else?

PS -- If anyone else wants to augment the test suite with general coverage of jinja templates, I'd be happy to merge those into this PR (or rebase this PR on top of their changes).

jakirkham:
> Also, since `jinja2` is an *optional* requirement of `conda-build`...

No longer optional. It is now a *required* dependency. See here ( https://github.com/conda/conda-build/pull/613 ).

jakirkham:
@msarahan, can confirm. ^

jakirkham:
@groutr, is there a test for building a git repo already? I thought I saw one, but wasn't able to find it in a quick search earlier. Maybe this would be a good place to start testing, @stuarteberg, (e.g. use the `GIT_` variables).

jakirkham:
@groutr, nevermind found the example.

@stuarteberg, created a simple test. Would like to see it fail on `conda-build` 1.18.2 just to see it is doing what we would expect. ( https://github.com/conda/conda-build/pull/691 )

stuarteberg:
>jinja2 is an optional requirement of conda-build...
>>No longer optional. It is now a required dependency.

YESSS!! Thanks for the tip.

>is there a test for building a git repo already?

Looks like it's here: [`conda-build/tests/test-recipes/metadata/source_git/`](https://github.com/conda/conda-build/tree/master/tests/test-recipes/metadata/source_git).

jakirkham:
@stuarteberg 

Sure.

Yep, that's what I'm using.

jakirkham:
So, I tried merging this in with my test and ran into an interesting problem. It appears that the value stored in `GIT_DESCRIBE`. In particular, it appears to store the latest tag available not the one checked out. See here ( https://travis-ci.org/jakirkham/conda-build/jobs/94013496#L1888-L1891 ).

stuarteberg:
>it appears to store the latest tag available not the one checked out.

Strange issue.  At first I wondered if somehow #667 might resolve this issue, but the more I think about it, the less likely that seems.

jakirkham:
I think this is the issue ( https://github.com/conda/conda-build/pull/621 ). Reverting it makes the versions check out correctly; however, it interestingly doesn't seem to be hitting the jinja bug, at present. Not sure why.

stuarteberg:
>I think this is the issue ( #621 ).

Weird.  That new code looked suspicious to me, so I stared at it for a while.  But I didn't see anything obviously wrong...

jakirkham:
FWIW, `jinja2` will now be included ( https://github.com/conda/conda-build/pull/692 ). Thanks to @msarahan.

jakirkham:
> Weird. That new code looked suspicious to me, so I stared at it for a while. But I didn't see anything obviously wrong...

I think it is just missing `cwd` arguments in multiple places or the default is wrong.

jakirkham:
I went through that PR ( https://github.com/conda/conda-build/pull/621 ) a few times and it was never quite obvious what was messed up by it. However, it touches a bunch of code in many different places in ways that look quite similar to the original code. I expect it will take quite a bit of time to find the issue. For now, I think the easiest solution is to revert and retry it later. As it will fail my current test, this will make it easier to get it fixed up and verify it is working again when the time comes.

jakirkham:
This is now included in an upstream PR with the tests ( https://github.com/conda/conda-build/pull/694 ).

frol:
Thank you guys!

stuarteberg:
Closing this, since it is superseded by #694.

Issue #676, opened 2015-11-15
conda build --test fails to find custom version


llawall:
During a 'conda build' get_build_metadata is called to read __conda_version__.txt. The same is not done when calling 'conda build --test' resulting in a failure to find the conda package that was built.

msarahan:
Closed from merging #677

PR #677, opened 2015-11-15
Read build metadata when testing package build

llawall:
get_build_metadata reads a custom version file which is necessary
for finding and installing a package when testing a package build.

#676


llawall:
Any thoughts on this? Are you happy to merge the code here or would you like it changed or refactored in some way?

msarahan:
Thanks for ironing out that inconsistency.

Issue #678, opened 2015-11-15
conda build --test should not move broken packages


llawall:
This is perhaps a matter of taste and is a change in intended behaviour though the original behaviour appears to be broken. 

Currently if tests fail to pass when conda build --test is run then the code attempts to move the package to the broken directory but fails as the source location is incorrect (tested on Windows only). The code should not move the package when running the tests independently of the build.

PR #679, opened 2015-11-15
Do not move packages to broken directory when running tests only

llawall:
Packages should be left wherever they were loaded from when running with conda build --test. 

#678


ilanschnell:
I agree, when just running
```
conda build --test
```
It should just test the package, and not move it anywhere (in case the test fails).
Packages should only end up in the `broken` directory, when the test initially fails (right after building it).

llawall:
Are you happy to merge the code here or would you like it changed or refactored in some way?

msarahan:
Thanks for the ping and for this PR.

PR #680, opened 2015-11-16
Add a few options to skeleton pypi

mwcraig:
This adds a few new options to `conda skeleton pypi`:

+ `--skeleton-setup-options` to allow options to be passed to `python setup.py` when it is called to determine dependencies (motivation/use case below).
+ `--recipe-setup-options` to allow options to be added to `python setup.py install` in `build.sh` and `bld.bat` in the recipe generated by skeleton (motivation/use case below).
+ `--pin-numpy` to add the proper build/run requirements to ensure the version of numpy gets pinned when the recipe is built. Without this there is no way, currently, to use skeleton to generate a recipe that pins the numpy version without manually editing the recipe afterwards.

Use case/motivation for the first two options: I maintain builds for several astropy affiliated packages, most of which use a helpers package, astropy-helpers, for some functionality needed at install. Most are also configured to check for bugfix updates on pypi, and that attempt to download raises an error when skeleton tries to construct a recipe.

There is an option to `setup.py` to turn off that check, but no way to specify it currently without editing `conda-build`.

The same option needs to be included in the build scripts, and it makes sense to me to be able to add it while generating the recipe.

Taking a step back, I want to be able, in the builder-bot I'm maintaining, to generate recipes on the fly and then build from them. 

**Edit 2/14/16`:** The first two options have been combined into a single option, `--setup-options`.

### DIFF
--- a/conda_build/main_skeleton.py
+++ b/conda_build/main_skeleton.py
@@ -159,6 +159,15 @@ Create recipe skeleton for packages hosted on the Python Packaging Index
         help="Creates recipe as noarch python"
     )
 
+    pypi.add_argument(
+        "--skeleton-setup-options",
+        action='append',
+        default=[],
+        help="Options to be passed to setup.py when generating recipe. "
+             "The characters '--' will be prepended to each option "
+             "automatically and should not be included."
+             "May be repeated to add several options.")
+
     cpan = repos.add_parser(
         "cpan",
         help="""
--- a/conda_build/main_skeleton.py
+++ b/conda_build/main_skeleton.py
@@ -168,6 +168,15 @@ Create recipe skeleton for packages hosted on the Python Packaging Index
              "automatically and should not be included."
              "May be repeated to add several options.")
 
+    pypi.add_argument(
+        "--recipe-setup-options",
+        action='append',
+        default=[],
+        help="Options to be added to setup.py install in the recipe. "
+             "The characters '--' will be prepended to each option "
+             "automatically and should not be included."
+             "May be repeated to add several options.")
+
     cpan = repos.add_parser(
         "cpan",
         help="""
--- a/conda_build/main_skeleton.py
+++ b/conda_build/main_skeleton.py
@@ -177,6 +177,13 @@ Create recipe skeleton for packages hosted on the Python Packaging Index
              "automatically and should not be included."
              "May be repeated to add several options.")
 
+    pypi.add_argument(
+        "--pin-numpy",
+        action='store_true',
+        help="Ensure that the generated recipe pins the version of numpy"
+             "to CONDA_NPY."
+    )
+
     cpan = repos.add_parser(
         "cpan",
         help="""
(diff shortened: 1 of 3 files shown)


stuarteberg:
FWIW, these options sound reasonable to me, especially `--pin-numpy`.  I have two questions:

1. Why would it be necessary to have two distinct options for `--skeleton-setup-options` vs. `--recipe-setup-options`?  It took me a bit of digging just to understand what the difference is.  Even *explaining* this distinction to the user essentially requires explaining an implementation detail of `conda-build`.  Expecting the user to understand it and then control over via from the command line seems like asking a bit much.  The explanation would go something like this:

>This feature lets you provide custom args to `setup.py` in the recipe build script.  Well, there are two different places these arguments get used.  You see, first conda does this hacky patch of `distutils`, then runs a `python setup.py install` as a pre-install step, and THEN runs the actual-install step.
>
>So... if you want to specify *different* custom setup args for the pre-install call to `python setup.py install` vs. the actual-install call `python setup.py install`, then you're in luck! That's the difference between `--skeleton-setup-options` and `--recipe-setup-options`.  (Although, as I'm writing this, I've already forgotten which one is which...)

Can't we get by with just one arg that is used in both places?

2. Your help text says: `The characters '--' will be prepended to each option`.  Wouldn't it be clearer to just forward the user's arguments into `setup.py` verbatim?  That would be more general, too (in case a package `setup.py` takes args that don't begin with `--`).

So, the usage would be:

```
conda skeleton pypi --skeleton-setup-args="--foo --bar -y hello world" yada-yada
```

mwcraig:
Thanks for the feedback, @stuarteberg!

1. I'll give it a try. For my application having one option would be fine. The other potential use for `--recipe-setup-options` is insert `--single-version-externally-managed --record=junk.txt` into the recipe to prevent downloading of dependencies during build, and am not sure what that does to skeleton. Will try it though.

But point taken that the distinction is a fine one.

2. Yes, you are right. Went back and forth a bit when writing the PR, mostly because it took me several attempts to remember that I had to use `--setup-args="--foo"` as ooposed to `--setup-args --foo`. That is an easy doc hint to add though.

mwcraig:
@stuarteberg -- updated, thanks for the suggestions!

mwcraig:
Not *positive*, but I don't think the test failure is related to these changes...

stuarteberg:
Looks good to me, for what that's worth... :-)

mwcraig:
Friendly bump...

mwcraig:
@msarahan -- I've added a couple of tests of the new functionality. One wrinkle: `--pin-numpy` adds `numpy x.x` to `run` but not `build`, consistent with the current version of `conda-build`. I think `conda-forge` is still running an older version of conda-build  ...so should I add it to both requirement sections even though it is unnecessary (though presumably harmless) in `build`?

msarahan:
Yeah, compatibility seems wise for now.

mwcraig:
Went the compatibility route and actually *pushed the tests* this time.

mwcraig:
I don't think the failures are related to my changes...

PR #681, opened 2015-11-17
normalize unicode so it can be encoded as ascii

groutr:
Conda skeleton would choke if there were unicode characters in cran metadata.


groutr:
Py35 tests are failed because of a failed subprocess call to svn.

Issue #349, opened 2015-03-13
Incorrect handling of unicode in git log files


jochym:
The binstar build service crashes on UTF git log entries (https://github.com/jochym/Elastic.git):

The output from `git log -n1` reads:

---

git log -n1
commit caf4906ee99fa816a0e14114815d42e7cf6f7d2d
Author: Paweł T. Jochym pawel.jochym@ifj.edu.pl
Date:   Mon Feb 9 00:53:34 2015 +0100

```
Build number.
```

---

The output from conda-build (1.11.0-py27_0):

---

Initialized empty Git repository in /opt/miniconda/conda-bld/git_cache/__github.com_jochym_Elastic.git/
Initialized empty Git repository in /opt/miniconda/conda-bld/work/.git/
==> git log -n1 <==

An unexpected error has occurred, please consider sending the
following traceback to the conda GitHub issue tracker at:

```
https://github.com/conda/conda-build/issues
```

Include the output of the command 'conda info' in your report.

Traceback (most recent call last):
  File "/opt/miniconda/bin/conda-build", line 5, in <module>
    sys.exit(main())
  File "/opt/miniconda/lib/python2.7/site-packages/conda_build/main_build.py", line 112, in main
    args_func(args, p)
  File "/opt/miniconda/lib/python2.7/site-packages/conda_build/main_build.py", line 321, in args_func
    args.func(args, p)
  File "/opt/miniconda/lib/python2.7/site-packages/conda_build/main_build.py", line 280, in execute
    channel_urls=channel_urls, override_channels=args.override_channels)
  File "/opt/miniconda/lib/python2.7/site-packages/conda_build/build.py", line 347, in build
    source.provide(m.path, m.get_section('source'))
  File "/opt/miniconda/lib/python2.7/site-packages/conda_build/source.py", line 250, in provide
    git_source(meta, recipe_dir)
  File "/opt/miniconda/lib/python2.7/site-packages/conda_build/source.py", line 134, in git_source
    git_info()
  File "/opt/miniconda/lib/python2.7/site-packages/conda_build/source.py", line 162, in git_info
    print(stdout + u'\n')
UnicodeEncodeError: 'ascii' codec can't encode character u'\u0142' in position 60: ordinal not in range(128)
command exited with status 1

asmeurer:
The workaround is to use Miniconda3.

jochym:
Not an option for me. Another option is to remove my first name from the git log message... ;)

asmeurer:
[...]

PR #688, opened 2015-11-26
"replace" unicode errors in git_info (hopefully fixing #349)

gdementen:
Trying to fix issue #349.

With this patch, I can successfully build on Windows using Python3 (it's broken otherwise). I tested the helper function using python2 but not a whole build.

### DIFF
--- a/conda_build/source.py
+++ b/conda_build/source.py
@@ -12,7 +12,7 @@ from conda.utils import hashsum_file
 
 from conda_build import external
 from conda_build.config import config
-from conda_build.utils import rm_rf, tar_xf, unzip
+from conda_build.utils import rm_rf, tar_xf, unzip, safe_print_unicode
 
 
 SRC_CACHE = join(config.croot, 'src_cache')
@@ -165,7 +165,7 @@ def git_info(fo=None):
             fo.write(stdout + u'\n')
         else:
             print(u'==> %s <==\n' % cmd)
-            print(stdout + u'\n')
+            safe_print_unicode(stdout + u'\n')
 
 
 def hg_source(meta):
--- a/conda_build/utils.py
+++ b/conda_build/utils.py
@@ -156,3 +156,25 @@ def comma_join(items):
     'a, b, and c'
     """
     return ' and '.join(items) if len(items) <= 2 else ', '.join(items[:-1]) + ', and ' + items[-1]
+
+
+def safe_print_unicode(*args, **kwargs):
+    """
+    prints unicode strings to stdout using configurable `errors` handler for
+    encoding errors
+
+    :param args: unicode strings to print to stdout
+    :param sep: separator (defaults to ' ')
+    :param end: ending character (defaults to '\n')
+    :param errors: error handler for encoding errors (defaults to 'replace')
+    """
+    sep = kwargs.pop('sep', u' ')
+    end = kwargs.pop('end', u'\n')
+    errors = kwargs.pop('errors', 'replace')
+    if PY3:
+        func = sys.stdout.buffer.write
+    else:
+        func = sys.stdout.write
+    line = sep.join(args) + end
+    encoding = sys.stdout.encoding or 'utf8'
+    func(line.encode(encoding, errors))


ccordoba12:
You're new function `safe_print_unicode` is giving an error in Python 2. Please see here

https://travis-ci.org/conda/conda-build/jobs/93384223#L954

gdementen:
The thing is, my new function did not give any error. It wasn't even called. In the line you quote, it is the "stdout + u'\n'" which fails.

What I think happens is that the tests which uses git_ifno uses the git repository it is running on (i.e. this PR) as a test, which breaks not because of new code I introduced but because of my name (which contains an accent -- so now you have a good test case ;-)), which fails on python2 even on Linux (apparently).

Now it fails on Python2 because utils.execute() does NOT .decode the strings it receives, so it returns python2 str and NOT unicode, which is then implicitly converted to unicode using an ASCII codec when trying to concatenate it with u"\n". See my commit above for one way to fix the problem. 

IMO, the ideal way would be to modify execute() to return unicode in all cases but I suspect this will be tricky to get right and I don't have any more time to spend on this. So if nobody is willing to go that "ideal" route, please merge this PR anyway because in the mean time conda-build is totally broken for many persons!

gdementen:
@ccordoba12 ping, just in case you are not subscribed to the issue :)

gdementen:
Also, with the second commit added I can build on Windows using both Python2 & Python3

ccordoba12:
@gdementen, thanks for the fix! I'll try to review it and merge it in a couple of days :-)

ccordoba12:
@gdementen, there are merge conflicts now. Please merge with master to solve them.

gdementen:
I rebased on master this morning... I could confirm it works on Python3 but had no time to do so on Python2. Hopefully the buildbot will tell us :)

stuarteberg:
The error you saw may have been due to #621, which was reverted yesterday via #694.  If you just try re-running your test case using the latest master (that is, *without* this PR), does the problem go away?  Your PR may not be needed after all.

gdementen:
@stuarteberg No, #621 was causing the failure of my initial PR (see @ccordoba12 comment), which my second commit fixed. But then I was asked to merge master (and rebased instead for a cleaner history -- removing in the process my second commit which is now indeed unnecessary). But the issue my PR fixes is that, in some cases, the stdout variable contains characters which the encoding of stdout cannot handle!

stuarteberg:
Ah.  I didn't read this thread carefully enough.  You've already covered the bases.  Shouldn't your new `safe_print_unicode()` function also be called from the [`git_source()`][1] function, too?

[1]: https://github.com/conda/conda-build/pull/688/files#diff-9fdc13ead8018ee54de2ac73289ec2e1R130

gdementen:
Probably, but I only care to fix the build issue I had.

gdementen:
On second thought, probably not, as a git revision hash cannot (AFAIK) contain non-ascii characters.

stuarteberg:
>On second thought, probably not, as a git revision hash cannot (AFAIK) contain non-ascii characters.

Ah, good point.  In that case, I don't see any other obvious places in the `conda-build` repo where `safe_print_unicode()` would be necessary.  This PR looks good-to-go, I think.

gdementen:
On Thu, Dec 3, 2015 at 8:51 PM, Stuart Berg <notifications@github.com>
wrote:

> On second thought, probably not, as a git revision hash cannot (AFAIK)
> contain non-ascii characters.
>
> Ah, good point. In that case, I don't see any other obvious places in the
> conda-build repo where safe_print_unicode() would be necessary. This PR
> looks good-to-go, I think.
>

In that case (no other place it would be necessary), you might want to
inline the function. It felt cleaner in a separate function, but I don't
really care either way as long as I can build packages ;-).

stuarteberg:
>It felt cleaner in a separate function

Agreed.  Keep it unchanged.

>but I don't really care either way as long as I can build packages

:-)  Yep, I think it should be merged, for what that's worth.  (I'm not a conda dev.)

groutr:
@gdementen, do you have any thoughts on how unicode is handled in 
https://github.com/conda/conda-build/blob/master/conda_build/cran.py#L574
It uses unicodedata.normalize to try and replace unicode characters with their ascii equivalent before encoding to ascii.

This is a valuable addition.  Thank you.  I think it might be a more flexible approach to have the utility function that we can pass a unicode string and it returns an ascii string.  In the case of cran.py, it writes these strings to a file (not stdout).  I'd be interested in hearing your opinion of this approach.

However, since these seem to be corner cases, I'm not opposed to merging the PR as-is (though I'm in favor of inlining if `safe_print_unicode` doesn't have a use elsewhere).

mcg1969:
This isn't working for me, I'm afraid. For one thing, the failure occurs before `safe_print_unicode` is even called,  in the `stdout + u'\n'` expression, as @gdementen pointed out above.

But suppose I just pass `stdout` alone to `safe_print_unicode`; I get this:
```
  File "/Users/mgrant/miniconda2/bin/conda-build", line 5, in <module>
    sys.exit(main())
  File "/Users/mgrant/miniconda2/lib/python2.7/site-packages/conda_build/main_build.py", line 190, in main
    args_func(args, p)
  File "/Users/mgrant/miniconda2/lib/python2.7/site-packages/conda_build/main_build.py", line 468, in args_func
    args.func(args, p)
  File "/Users/mgrant/miniconda2/lib/python2.7/site-packages/conda_build/main_build.py", line 391, in execute
    override_channels=args.override_channels, include_recipe=args.include_recipe)
  File "/Users/mgrant/miniconda2/lib/python2.7/site-packages/conda_build/build.py", line 392, in build
    source.provide(m.path, m.get_section('source'))
  File "/Users/mgrant/miniconda2/lib/python2.7/site-packages/conda_build/source.py", line 266, in provide
    git_source(meta, recipe_dir)
  File "/Users/mgrant/miniconda2/lib/python2.7/site-packages/conda_build/source.py", line 135, in git_source
    git_info()
  File "/Users/mgrant/miniconda2/lib/python2.7/site-packages/conda_build/source.py", line 160, in git_info
    safe_print_unicode(stdout) # + u'\n')
  File "/Users/mgrant/miniconda2/lib/python2.7/site-packages/conda_build/utils.py", line 257, in safe_print_unicode
    line = sep.join(args) + end
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 169: ordinal not in range(128)
```

gdementen:
Silly me. Tried to be clever and create a reusable function which mimics print() and got it wrong, while my initial quick and dirty code was correct. @mcg1969: thanks for testing this. I'll update the PR later tonight or tomorrow with a better fix.

gdementen:
@mcg1969 On second thought, my code is probably not buggy... It looks like the stdout variable you pass to safe_print_unicode is not unicode, which it should be at this point (after line https://github.com/conda/conda-build/pull/688/files#diff-9fdc13ead8018ee54de2ac73289ec2e1L159). How exactly did you test?

mcg1969:
You know, you are probably right. Thanks for your persistence, and I apologize for the trouble. I think I misapplied the patch.

mcg1969:
Indeed, I did patch my local copy incorrectly. This fixes my issue, and I recommend a marge. @msarahan

msarahan:
Excellent!  Merging.  Thanks @gdementen

mcg1969:
yes! Thanks @gdementen. I just re-ran my test directly from the merged version and all is well.

msarahan:
PS: @gdementen good luck to you and your wife.

gdementen:
@msarahan hehe, thanks! :)

PR #691, opened 2015-11-30
TEST: Adds a test using jinja2 environment variables

jakirkham:
Requires https://github.com/conda/conda-build/pull/692

This provides a simple test using environment variables. This would have worked on `conda-build` 1.18.1 (did when tested locally). It is expected to fail on `conda-build` 1.18.2.

### DIFF
--- a/tests/test-recipes/build_recipes.sh
+++ b/tests/test-recipes/build_recipes.sh
@@ -30,4 +30,7 @@ echo "$OUTPUT" | grep 'Error: Untracked file(s) ('\''conda-meta/nope'\'',)'
 ! OUTPUT=$(conda build --no-anaconda-upload recursive-build/ 2>&1)
 echo "$OUTPUT" | grep 'No packages found in current .* channels matching: recursive-build2 2\.0'
 
+! OUTPUT=$(conda build --no-anaconda-upload source_git_jinja2_oops/ 2>&1)
+echo "$OUTPUT" | grep '\''GIT_DSECRIBE_TAG'\'' is undefined'
+
 echo "TESTS PASSED"
--- a/tests/test-recipes/fail/source_git_jinja2_oops/bld.bat
+++ b/tests/test-recipes/fail/source_git_jinja2_oops/bld.bat
@@ -1,0 +1,18 @@
+if not exist .git exit 1
+git config core.fileMode false
+if errorlevel 1 exit 1
+git describe --tags --dirty
+if errorlevel 1 exit 1
+for /f "delims=" %%i in ('git describe') do set gitdesc=%%i
+if errorlevel 1 exit 1
+echo "%gitdesc%"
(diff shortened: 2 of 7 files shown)


stuarteberg:
This is a good start.  I think we need to make sure this test checks the following:

- `conda-build` doesn't choke on e.g. `GIT_DESCRIBE_TAG` during first pass
- ... but it *does* eventually choke on garbage variables (during the second pass)
- ... and verify that `GIT_DESCRIBE_TAG`, etc. have legitimate values (not empty strings) with an appropriate assertion in your test's `build.sh` file.  (E.g. compare `PKG_VERSION` to the manual output of `git describe --tags`)

jakirkham:
@groutr, an attempt at a test. Feel free to give feedback. At present, I don't think `jinja2` is present in all places. That problem should ultimately be solved by this ( https://github.com/conda/conda-build/pull/692 ), but isn't at present.

jakirkham:
> ... but it *does* eventually choke on garbage variables (during the second pass)

@stuarteberg, what does this mean?

jakirkham:
As far as I can tell, this is now showing the correct error. Namely that `GIT_DESCRIBE_TAG` is undefined on the first pass.

jakirkham:
Also, @msarahan, if you wouldn't mind double checking my Windows logic, that would be appreciated as I don't have a good test environment for it.

msarahan:
I will push for setting AppVeyor (back) up today.  I will also test this myself manually.

jakirkham:
Thanks @msarahan.

stuarteberg:
>>    ... but it does eventually choke on garbage variables (during the second pass)

>@stuarteberg, what does this mean?

What I mean is, suppose the user accidentally uses a bogus variable name (e.g. a typo like `GIT_FLUL_HASH`).  We must fail to build that recipe.  We are intentionally non-strict on the first pass, and very strict on the second pass.  I guess the right thing to do is add a test case to `tests/test-recipes/fail/`.  (Those recipes are all *expected* to fail.)

jakirkham:
Ah, thanks for clarifying. Sure, I will add this.

jakirkham:
@stuarteberg, I added a test in the event the git variable is incorrectly spelled however now I am seeing this [bug]( https://github.com/conda/conda-build/pull/666#issuecomment-160742923 ) here.

jakirkham:
@stuarteberg, I think the bug was introduced here ( https://github.com/conda/conda-build/pull/621 ).

jakirkham:
This appears to work; however, changes introduced in this PR ( https://github.com/conda/conda-build/pull/621 ) seem to break behavior with `git` tests and possibly other scenarios with `git`. At present, I recommend that it be reverted until a new non-breaking version can be established.

jakirkham:
This is now included in an upstream PR ( https://github.com/conda/conda-build/pull/694 ) with the fix applied.

PR #692, opened 2015-11-30
Include `jinja2` in CI environments.

jakirkham:


jakirkham:
@msarahan and @groutr, I think @stuarteberg is right. There are still a ton of place where `jinja2` doesn't seem to exist. Mind providing some help in identifying them and/or letting us move to a new version of `miniconda` as appropriate to make this simpler.

jakirkham:
Ok, this now install `jinja2` correctly. The fact that it is working correctly can be demonstrated by its use in this test ( https://github.com/conda/conda-build/pull/691 ).

msarahan:
@jakirkham what do you mean "letting us move to a new version of miniconda"?  For the purposes of testing this repo, or what?  I'd be totally OK with updating recipes to use miniconda "latest" - but we need to be careful about its installation path, since the python2 version used to install as miniconda, but now installs as miniconda2 by default.  Nothing an explicit setting can't fix.

jakirkham:
@msarahan, sorry, no longer relevant. Feel free to disregard as to works with the versions present.

jakirkham:
Also, this should be ready to merge barring any other comments.

msarahan:
Thanks!

PR #705, opened 2015-12-07
OpenBSD 5.x support

NewbiZ:
This is a followup on conda/conda#1891.

conda-build should use `/bin/sh` instead of `/bin/bash`, for reasons exposed in the above pull request. This allows the use of conda-build on *BSDs and some linux distributions which chose to distribute bash in e.g. `/usr/local/bin/bash`.

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -582,7 +582,7 @@ def test(m, verbose=True, channel_urls=(), override_channels=False):
         else:
             test_file = join(tmp_dir, 'run_test.sh')
             # TODO: Run the test/commands here instead of in run_test.py
-            cmd = ['/bin/bash', '-x', '-e', test_file]
+            cmd = ['/bin/sh', '-x', '-e', test_file]
             try:
                 subprocess.check_call(cmd, env=env, cwd=tmp_dir)
             except subprocess.CalledProcessError:
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -439,7 +439,7 @@ def build(m, get_src=True, verbose=True, post=None, channel_urls=(),
                 os.chmod(build_file, 0o766)
 
             if isfile(build_file):
-                cmd = ['/bin/bash', '-x', '-e', build_file]
+                cmd = ['/bin/sh', '-x', '-e', build_file]
 
                 _check_call(cmd, env=env, cwd=src_dir)
 
--- a/conda_build/post.py
+++ b/conda_build/post.py
@@ -320,7 +320,7 @@ def mk_relative(m, f):
     if not is_obj(path):
         return
 
-    if sys.platform.startswith('linux') or sys.platform.startswith('openbsd'):
+    if sys.platform.startswith('linux') or ('bsd' in sys.platform):
         mk_relative_linux(f, rpaths=m.get_value('build/rpaths', ['lib']))
     elif sys.platform == 'darwin':
         mk_relative_osx(path)
--- a/conda_build/post.py
+++ b/conda_build/post.py
@@ -320,7 +320,7 @@ def mk_relative(m, f):
     if not is_obj(path):
         return
 
-    if sys.platform.startswith('linux'):
+    if sys.platform.startswith('linux') or sys.platform.startswith('openbsd'):
         mk_relative_linux(f, rpaths=m.get_value('build/rpaths', ['lib']))
     elif sys.platform == 'darwin':
         mk_relative_osx(path)
--- a/tests/test-recipes/build_recipes.sh
+++ b/tests/test-recipes/build_recipes.sh
@@ -8,8 +8,10 @@ cd "$(dirname "${BASH_SOURCE[0]}")"
 # Recipes that should fail and give some error
 
 for recipe in metadata/*/; do
-    if [[ $(ls -A "$recipe") ]]; then
-        if [[ $recipe =~ .*osx_is_app.* && $(uname) != "Darwin" ]]; then
+    # Only consider directories
+    if [ -d "$recipe" ]; then
+        # Disable test recipe "osx_is_app" if not on OSX
+        if [ "${recipe#*osx_is_app}" != "$recipe" ] && [ $(uname) != "Darwin" ]; then
             continue
         fi
         conda build --no-anaconda-upload $recipe
--- a/tests/test-recipes/metadata/jinja_vars/build.sh
+++ b/tests/test-recipes/metadata/jinja_vars/build.sh
@@ -1,11 +1,11 @@
 # CONDA_TEST_VAR was inherited via build/script_env
-[ "${CONDA_TEST_VAR}" == "conda_test" ]
+[ "${CONDA_TEST_VAR}" = "conda_test" ]
 
 # CONDA_TEST_VAR_2 was not inherited
-[ "${CONDA_TEST_VAR_2}" == "" ]
+[ "${CONDA_TEST_VAR_2}" = "" ]
 
 # Sanity check: Neither was LD_LIBRARY_PATH
-[ "$LD_LIBRARY_PATH" == "" ]
+[ "$LD_LIBRARY_PATH" = "" ]
 
 # Check the special value we gave the build string, which depends on build number
-[ "${PKG_BUILD_STRING}" == "${CONDA_TEST_VAR}_${PKG_BUILDNUM}" ]
+[ "${PKG_BUILD_STRING}" = "${CONDA_TEST_VAR}_${PKG_BUILDNUM}" ]


conda_build/post.py
```diff
@@ -320,7 +320,7 @@ def mk_relative(m, f):
     if not is_obj(path):
         return
 
-    if sys.platform.startswith('linux'):
+    if sys.platform.startswith('linux') or sys.platform.startswith('openbsd'):
```
ccordoba12: Why not `endswith('bsd')`? i assume this should work the same for the other BSDs (Free and Net, I mean).

conda_build/build.py
```diff
@@ -582,7 +582,7 @@ def test(m, verbose=True, channel_urls=(), override_channels=False):
         else:
             test_file = join(tmp_dir, 'run_test.sh')
             # TODO: Run the test/commands here instead of in run_test.py
-            cmd = ['/bin/bash', '-x', '-e', test_file]
+            cmd = ['/bin/sh', '-x', '-e', test_file]
```
ccordoba12: What if there are bashisms in the `build.sh` script of a recipe? We've never made the promise to work with other shells.

I assume BSDs have a package for `bash`, so why not use the path provided to `bash` for each BSD instead of changing the default shell?

conda_build/post.py
NewbiZ: It _might_ work the same on other BSDs, but I don't have a Net/Free BSD to test. Also, sys.platform form e.g. OpenBSD 5.x is `"openbsd5"`, so matching `endswith("bsd")` wouldn't even work here.

ccordoba12: I understand you don't have other BSDs to test, but given that your changes are so simple, I'd assume they would work on them too.

Well, if `endswith("bsd")` doesn't work, then `bsd in sys.platform` most surely would :-)

conda_build/build.py
NewbiZ: > What if there are bashisms in the build.sh script of a recipe? We've never made the promise to work with other shells.

Then I guess we should make all the scripts work with `sh`, as it's the only reliable shell that one can find at a fixed location on any POSIX compliant OS.
To be honest, let's not over-engineer this change, it's a 0 cost issue. I did not found any script in conda-recipes that rely on bash-specific syntax, it's not like build scripts are very complex, most of them just run some very basic commands, there's no pragmatic need for `bash` here.

> I assume BSDs have a package for bash, so why not use the path provided to bash for each BSD instead of changing the default shell?

Because it's not just a BSD issue here. There is no guarantee even on linux that bash will be installed in `/bin`.

Also note that if a build script _really_ needs bash, then it should call it itself with `/usr/bin/env bash`, but no need to force bash for everyone, when really all these scripts do is run some commands.

conda_build/post.py
NewbiZ: All right, let me install Net/Free BSD and I will change that so it works on them too ;-)

conda_build/build.py
ccordoba12: > sh, as it's the only reliable shell that one can find at a fixed location on any POSIX compliant OS.

We've never promised for conda-build to work on a POSIX OS other than Linux (and OS X :-).

> To be honest, let's not over-engineer this change, it's a 0 cost issue

I really can't say because I don't know the internals of conda-build. It probably relies on running our scripts on bash, but I can't tell.

To minimize even further the impact of this PR, what if you change to

```
if sys.platform.startswith('linux'):
    cmd = ['/bin/bash', '-x', '-e', build_file]
else:
    cmd = ['/bin/sh', '-x', '-e', build_file]
```

?

conda_build/post.py
ccordoba12: Don't take the long path here! :-) Just look for the Python docs on `sys.platform`.

conda_build/build.py
groutr: AFAIK, I don't think conda-build relies on bash.  I think it might just make the assumption that bash the most universally available shell.

ccordoba12: @groutr, so, is it ok for you to change `bash` for `sh` as @NewbiZ proposes?

groutr: I think it should be fine.  Conda build should be shell agnostic.  Enforce bash should, theoretically, be done in the build script (because the build script knows if it is using a specific shell feature or not).

groutr:
Provided that all the tests are passing, I think this is a positive change and should be merged.

conda_build/post.py
ccordoba12: @groutr, I'd like for @NewbiZ to do my suggested change here before merging, so this is not related only to `openbsd` but to all BSDs.

conda_build/build.py
frol: Let me put my two cents here.

I would support using `sh` instead of `bash` if this breaking change will be made with a deprecation step (for at the very least a couple of minor releases) instead of breaking things altogether. There are some quite complex scripts already in conda-recipes repo using features of bash, for example GCC: https://github.com/conda/conda-recipes/blob/master/gcc-4.8/post-link.sh

conda_build/post.py
```diff
@@ -320,7 +320,7 @@ def mk_relative(m, f):
     if not is_obj(path):
         return
 
-    if sys.platform.startswith('linux'):
+    if sys.platform.startswith('linux') or ('bsd' in sys.platform):
```
frol: While it is not really the issue of this exact pull-request, but why don't just use `platform.system()` (`import platform`) instead of this hacking around with `startswith`, `in`?

frol:
@groutr I'm not a supervisor, I'm just a regular opensource contributor, but I would like to ask you: Are you going to fix all broken conda-recipes at once due to `bash` vs `sh` differences? As I have suggested above, please, consider adding a deprecation code and update a documentation.

conda_build/build.py
frol: BTW, `bash` is super handy when arrays are used to craft a command. Using `sh` in such a case is a headache.

Contemplating this `bash` vs `sh`, and remembering `bat` on Windows I thought that it would be great to move from these crazy hacks to a normal language - Python! 

Why not?! Having a simple `run` function (a wrapper around `subprocess.Popen`) to execute external commands would provide a very easy way to write build scripts. There is a project that could be integrated for this very purpose - [PyInvoke](http://www.pyinvoke.org/)

frol: Added a new issue #706 to discuss `bash`/`sh`/`bat` VS Python for build scripts.

ilanschnell:
I'm totally against changing the default behavior of conda-build.  On Unix system we have always been calling out to `/bin/bash`, and changing this will cause many problems for many people.  We cannot do this (not even in a migration process).  Why not simply add something like the following as an option to the `meta.yaml` file:
```
build:
  shell_cmd: /bin/sh
```
This could even maybe be made more flexible to allow calling build scripts with Python (instead of https://github.com/conda/conda-build/issues/706).

NewbiZ:
@frol While I agree with some things you say, let's not transform this PR to a Python vs Bash issue, that would make the whole PR sink in an endless debate.

@ilanschnell If you're going the way to let people choose their interpreter, then just force `+x` on build scripts and let the shebang do its work. This is a totally valid solution, one can use whatever shell or programming language to build their recipes. The only problem with this approach is that you now have a totally free, uncontrollable recipe list which depends on the host system. conda-build did not choose this path, if you want it, create a new issue for this debate please.

On a side note, I do not think we should try to fix the whole `conda-recipes` repository. Only some rare recipes will use bash features, they are trivial to port to sh, and honestly most packages there are broken for a ton of reasons (missing dependencies, incorrect builds, garbage bundled, ...). Fixing this repository should be the subject of a whole new milestone.

conda_build/post.py
NewbiZ: That would pretty much be the same dance with `platform.system()`.

    if platform.system() == 'Linux' or ('BSD' in platform.system()):

pmuller:
Hi there,

On Linux, bash is _usually_ placed in /bin, but that's just a de-facto standard, not a real guarantee.

IF we want to always use bash, we should use a mecanism that locates the real path of bash instead of hard-coding its path. It's the only way to make Conda usable on systems that put bash elsewhere.

The alternative is to let the build scripts chose their interpreter themselves and recommending recipes authors to use /usr/bin/env in the shebang; the only change would be to ensure the build.sh script is executable and calling it directly : 

    #!/usr/bin/env bash
    do something...

This would work everywhere.

frol:
Please, forgive me for my jumping into this discussion, but I don't really like breaking releases without any prior notices.

@NewbiZ (As to the Python vs Bash, that is exactly why I have created another PR. I've just came to this quite obvious thought looking at the problem of downgrading more powerful Bash language to a lightweight and more widely available Ash, yet less feature-rich, language)

/offtop shebang doesn't work on Windows (unless executing in Cygwin) as far as I know, so this will introduce different implementation of finding a proper interpreter, though there is already a difference in Bash vs Bat.

> Only some rare recipes will use bash features, they are trivial to port to sh

Then they should be fixed first and an anouncement should be made for those who maintain their own packages.

@ilanschnell 

> Why not simply add something like the following as an option to the meta.yaml file

It seems, the OpenBSD support will still require to update all the recipes to include these lines. Ash might be a good thing; if my memory serves me correctly, Ubuntu has switched from Bash to Ash. However, I don't know if this change has any benefit at all in this circumstances. Anyway, hardcoded `/bin/bash` is not a good implementation also.

pmuller:
@frol : on windows, the build.bat script is used, so using the shebang in build.sh is not an issue

ilanschnell:
@pmuller `#!/usr/bin/env bash` would work everywhere?
No, it will only work if the `env` program is in `/usr/bin/env`.  I think there a bigger chance of `bash` being in `/bin/bash`.

pmuller:
@ilanschnell : /usr/bin/env is a POSIX standard ; you'll find "env" in /usr/bin on Debian and RHEL based distributions, on BSD and on Mac OS X.

conda_build/post.py
frol: @NewbiZ How about using a set?

```Python
if platform.system() in {'Linux', 'OpenBSD', 'FreeBSD'}:
```

NewbiZ: Please stop flooding this issue. I don't think spending 5 comments on how to rephrase a single line is justified. Both versions are correct, the complexity difference between both is non existent, there really is nothing to discuss here.

frol: @NewbiZ If you wouldn't break things and would write a good code from the beginning, I would keep silence.

P.S. You are welcome.

stuarteberg:
My gut says that the impact of changing from `bash` to `sh` is too costly.  But instead of trusting anyone's gut, the impact could be better understood if somebody were to try running [shellcheck][1] on a large body of recipes, e.g. `conda-recipes`.  (A script that automates that process wouldn't be hard to write... maybe that would be a nice feature to implement if `conda-recipes` ever gets a CI system.)  FWIW, the example @frol cited [above][2] has no errors under `bash`, but has several problems in POSIX `sh`.

[1]: http://www.shellcheck.net
[2]: https://github.com/conda/conda-build/pull/705#discussion_r47051458

ilanschnell:
My guy feeling is also that it is too costly.  Why don't we do what Carlos initially suggested, basically:
```
if sys.platform.startswith('openbsd'):
    cmd = ['/bin/sh', '-x', '-e', build_file]
else:
    cmd = ['/bin/bash', '-x', '-e', build_file]
```

ccordoba12:
I'm +1 with @ilanschnell's suggestion too. It's the safest approach :-)

ilanschnell:
I created https://github.com/conda/conda-build/pull/707
We can continue the discussion there, if needed (but this seems to be the consensus).

PR #707, opened 2015-12-10
use /bin/sh on openbsd

ilanschnell:
This is a followup PR on https://github.com/conda/conda-build/pull/705

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -36,7 +36,7 @@ from conda_build.exceptions import indent
 
 
 on_win = (sys.platform == 'win32')
-if sys.platform.startswith('openbsd'):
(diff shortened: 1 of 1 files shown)


conda_build/build.py
```diff
@@ -36,6 +36,10 @@
 
 
 on_win = (sys.platform == 'win32')
+if sys.platform.startswith('openbsd'):
```
ccordoba12: @ilanschnell, my only suggestion is to change this line to:

```python
if 'bsd' in sys.platform:
```

so that this PR applies to the other BSD's (i.e. FreeBSD and NetBSD).

NewbiZ:
Thanks!

PR #717, opened 2015-12-30
Enhancements for module exec in load_setuptools

richafrank:
Our setup.py relies on `__file__`, so I've added that, as well as `__name__` and `__doc__`, to the namespace used when exec'ing the file.

### DIFF
--- a/conda_build/jinja_context.py
+++ b/conda_build/jinja_context.py
@@ -34,7 +34,8 @@ def load_setuptools(setup_file='setup.py'):
             '__doc__': None,
             '__file__': setup_file,
         }
-        code = compile(open(setup_file).read(), setup_file, 'exec')
+        code = compile(open(setup_file).read(), setup_file, 'exec',
+                       dont_inherit=1)
         exec(code, ns, ns)
         setuptools.setup = setuptools_setup
         del sys.path[-1]
--- a/conda_build/jinja_context.py
+++ b/conda_build/jinja_context.py
@@ -7,13 +7,15 @@ from __future__ import absolute_import, division, print_function
 
 import json
 import os
+from functools import partial
 
 from conda.compat import PY3
 from .environ import get_dict as get_environ
 
 _setuptools_data = None
 
-def load_setuptools(setup_file='setup.py'):
+def load_setuptools(setup_file='setup.py', from_recipe_dir=False,
+                    recipe_dir=None):
     global _setuptools_data
 
     if _setuptools_data is None:
@@ -26,6 +28,9 @@ def load_setuptools(setup_file='setup.py'):
         #Add current directory to path
         import sys
         sys.path.append('.')
+
+        if from_recipe_dir and recipe_dir:
+            setup_file = os.path.abspath(os.path.join(recipe_dir, setup_file))
 
         #Patch setuptools, distutils
         setuptools_setup = setuptools.setup
@@ -50,7 +55,7 @@ def load_npm():
     with open('package.json', **mode_dict) as pkg:
         return json.load(pkg)
 
-def context_processor(initial_metadata=None):
+def context_processor(initial_metadata, recipe_dir):
     """
     Return a dictionary to use as context for jinja templates.
 
@@ -61,7 +66,7 @@ def context_processor(initial_metadata=None):
     environ = dict(os.environ)
     environ.update(get_environ(m=initial_metadata))
 
-    ctx.update(load_setuptools=load_setuptools,
+    ctx.update(load_setuptools=partial(load_setuptools, recipe_dir=recipe_dir),
                load_npm=load_npm,
                environ=environ)
     return ctx
--- a/conda_build/jinja_context.py
+++ b/conda_build/jinja_context.py
@@ -29,7 +29,12 @@ def load_setuptools(setup_file='setup.py'):
         #Patch setuptools
         setuptools_setup = setuptools.setup
         setuptools.setup = setup
-        exec(open(setup_file).read())
+        ns = {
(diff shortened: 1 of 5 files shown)


richafrank:
If there are tests I should add, I'd love some tips on how I should go about that!

richafrank:
The other benefit here is that any functions called by setup.py will be able to access that module's globals, especially imports.  Otherwise, we run into https://bugs.python.org/issue16781.

conda_build/jinja_context.py
```diff
@@ -29,7 +29,13 @@ def setup(**kw):
         #Patch setuptools
         setuptools_setup = setuptools.setup
         setuptools.setup = setup
-        exec(open(setup_file).read())
+        ns = {
+            '__name__': os.path.splitext(os.path.basename(setup_file))[0],
+            '__doc__': None,
+            '__file__': setup_file,
+        }
+        code = compile(open(setup_file).read(), setup_file, 'exec')
```
stuarteberg: I think you should pass `dont_inherit=1` here.  Otherwise, this might choke if `setup.py` is written for python 2 and contains a `print` statement.

stuarteberg:
>If there are tests I should add, I'd love some tips on how I should go about that!

For many test-cases, adding a recipe to `tests/test-recipes/metadata` is the way to go.  But in this case, I'm not sure.  I don't know how to reproduce the problem that spurred you to write this PR... does it only occur when using `conda skeleton`?  If so, perhaps `tests/test-skeleton` is the place to look, but I'm not sure exactly what to do.  Perhaps a conda dev can chime in.

conda_build/jinja_context.py
richafrank: Definitely agreed.

richafrank:
I encountered the problem(s) using `conda build <path to recipe>` with [a recipe that calls load_setuptools] (https://github.com/quantopian/zipline/pull/937/files#diff-67647ffe12f2bec8feb1070bb7ec7176) and a setup.py that tries to access globals from inside a function.  Once I added the namespace dict, I saw that setup.py was raising a NameError on `__file__`.

stuarteberg:
OK, I think we definitely needs a general test for `load_setuptools`.  I tried adding a simple test recipe (see below), but the `load_setuptools()` function seems to assume that the `cwd` is the source directory, which isn't true.  Clearly, I'm either missing something or `load_setuptools()` just doesn't work.

(This isn't due to your PR in particular, but perhaps this is a good opportunity for someone to finally write a proper test.)

```bash
$ cat tests/test-recipes/metadata/jinja_load_setuptools/meta.yaml
```

```yaml
{% set data = load_setuptools() %}

package:
  name: test-jinja-setuptools
  version: 1.0

source:
  path: ../../test-package

requirements:
  build:
    - python
    {% for req in data.get('install_requires', []) -%}
    - {{req}}
    {% endfor %}
```

```bash
$ cd tests
$ PYTHONPATH=. conda build test-recipes/metadata/jinja_load_setuptools

  File "/miniconda/bin/conda-build", line 5, in <module>
    sys.exit(main())
  File "/miniconda/lib/python2.7/site-packages/conda_build/main_build.py", line 191, in main
    args_func(args, p)
  File "/miniconda/lib/python2.7/site-packages/conda_build/main_build.py", line 478, in args_func
    args.func(args, p)
  File "/miniconda/lib/python2.7/site-packages/conda_build/main_build.py", line 357, in execute
    m = MetaData(recipe_dir)
  File "/miniconda/lib/python2.7/site-packages/conda_build/metadata.py", line 328, in __init__
    self.parse_again(permit_undefined_jinja=True)
  File "/miniconda/lib/python2.7/site-packages/conda_build/metadata.py", line 339, in parse_again
    self.meta = parse(self._get_contents(permit_undefined_jinja))
  File "/miniconda/lib/python2.7/site-packages/conda_build/metadata.py", line 608, in _get_contents
    return template.render(environment=env)
  File "/miniconda/lib/python2.7/site-packages/jinja2/environment.py", line 989, in render
    return self.environment.handle_exception(exc_info, True)
  File "/miniconda/lib/python2.7/site-packages/jinja2/environment.py", line 754, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/Users/bergs/Documents/workspace/conda-build/tests/test-recipes/metadata/jinja_load_setuptools/meta.yaml", line 1, in top-level template code
    {% set data = load_setuptools() %}
  File "/miniconda/lib/python2.7/site-packages/conda_build/jinja_context.py", line 32, in load_setuptools
    exec(open(setup_file).read())
IOError: [Errno 2] No such file or directory: 'setup.py'
```

richafrank:
>  I tried adding a simple test recipe (see below), but the load_setuptools() function seems to assume that the cwd is the source directory, which isn't true. Clearly, I'm either missing something or load_setuptools() just doesn't work.

I agree - I think I had found the same issue and have just been running the build from my source dir.

richafrank:
@stuarteberg I added optional support for `load_setuptools` to use a path relative to the recipe dir when looking for setup.py, along with a test case.  It should be backwards compatible with existing recipes.

Does that sound like a reasonable addition, and is the test case valid/comprehensive enough?

stuarteberg:
>Does that sound like a reasonable addition

Yes! I think this is exactly what was needed.  And the test looks reasonable, too.  (It passes on my machine.)

IMHO, this PR is ready for merge.  Ping: @msarahan @kalefranz

richafrank:
@stuarteberg I realized that pip sets `__name__` to `'__main__'` when running setup.py, so I changed this to match that, since that will be common check.

Anything blocking this merge?

msarahan:
Thanks!

richafrank:
Awesome - thanks!

PR #719, opened 2016-01-02
Support for Lua as a built-in language

alexbw:
Includes:
 - Lua is now a supported language, alongside Python, Perl and R.
 - Tags for selecting packages based on the installed Lua version in `meta.yaml`
 - A skeleton for building conda packages from the `luarocks` Lua package manager.

For examples of conda recipes for Lua, see: https://github.com/alexbw/conda-lua-recipes.


stuarteberg:
>For examples of conda recipes for Lua, see: https://github.com/alexbw/conda-lua-recipes.

I know nearly nothing about Lua, but I noticed that all of your example recipes have `post-link.sh`/`pre-unlink.sh` scripts.  Is there any way this PR can be enhanced to eliminate the need for all those post-link scripts?  (They all do the same thing, as far as I can tell.)

(IIUC, the conda devs generally recommend against using post-link scripts unless absolutely necessary.  They can introduce unexpected and subtle problems, they can be tricky to debug, and it can be bothersome to maintain 100% portability across platforms.  Admittedly, all of your post-link scripts are pretty simple, though.)

alexbw:
I'll check, I found I had to use those last April when I started tinkering
with this, and left them in as superstition.

alexbw:
Unfortunately, the Lua package manager, luarocks, requires both of these
scripts to keep its list of installed packages sanitized. Without the
post-link.sh script, luarocks doesn't register a new package has been
installed, and without the pre-unlink.sh script, luarocks doesn't register
that the package has been removed. Although this is less than ideal, I hope
this peculiarity of Lua's packaging system is not a blocker for merging
support.

stuarteberg:
OK, keeping the luarocks package list in sync with the conda package list sounds like a good idea.  (For python, it's certainly convenient that `conda list` and `pip list` are consistent with each other.)

I'm just asking if there's a way to make that synchronization step *built in* to conda itself, rather than relying on recipe maintainers to add `post-link.sh` and `pre-unlink.sh` to every lua recipe they write.  If literally every lua recipe needs this step, shouldn't we just have `conda` do it automatically?

stuarteberg:
On a related note: For python support, the conda devs decided to ensure that `python` and `pip` are always installed together.  They did this by adding a circular dependency between the `python` and `pip` packages (`python` requires `pip`, and `pip` requires `python`).

Since the relationship between `lua` and `luarocks` is analogous to `python`/`pip`, perhaps you should do the same thing: Make `lua` depend on `luarocks` and vice-versa.  I know of at least one "gotcha" to handle, though: `conda.toposort()` should be patched to explicitly handle the circular dependency. (It's a 2-line change; just copy conda/conda#1614 as an example.)  Besides that, I don't know of any other "gotchas", but we should get a second opinion from an actual conda dev to be sure.

alexbw:
- What suggestions do you have for adding facilities in conda for keeping
luarocks and conda in sync? I'm not sure what's the the best way to do
this. It's tricky because the conda name of the package does not
necessarily equal the luarocks name of the package (I enforce lua-* prefix,
that is true only for a small minority of luarocks packages)
 - I'm fine adding in that circular dependency between lua and luarocks,
will submit a PR soon

alexbw:
Can you help me understand this circular dependency? Python recipes do not depend on pip, from what I've seen:

https://github.com/conda/conda-recipes/blob/master/python-2.7.8/meta.yaml
https://github.com/conda/conda-recipes/blob/master/python-3.5/meta.yaml

stuarteberg:
Ah, right.  See conda/conda#1154.

To summarize:

As it turns out, `python` is a special case because `conda` itself relies on python.  Older versions of `conda` couldn't handle circular dependencies, so the `pip` requirement could only be added in a specail (hacky) way to avoid breaking `conda update conda` if you had an old version of `conda`.

But in the case of `lua`/`luarocks`, there's no need for this special hackery.  IIUC, you can just add `luarocks` to the `lua` requirements.  `conda` won't choke on the circular dependency.

alexbw:
Ok, great, so the only change I need to make is in my recipe.

stuarteberg:
>What suggestions do you have for adding facilities in conda for keeping luarocks and conda in sync?

Honestly, this is outside of my expertise.  As it stands now, it seems like you've already found a way to tell `luarocks` about `conda` packages (via your `post-link` scripts).  So I guess the only question is what to do about the reverse case.  I'm not sure exactly what's required to tell `conda` about packages that were installed with `luarocks`.  Perhaps @ilanschnell can chime in here.

Looking at the python case, it appears that [`conda/pip.py`][1] implements the logic for telling `conda` which packages have been installed via `pip`.  I don't know how `pip` magically knows about packages that have been installed with `conda`, though.  Maybe `pip` always searches all packages in `site-packages` every time you type `pip list`?  If that's true, then maybe `conda` isn't doing anything special to inform `pip` about the packages it installed.

[1]: https://github.com/conda/conda/blob/master/conda/pip.py

alexbw:
It's unclear what my task list is at this point to enable a merge — I'd be happy to tick down a concrete todo list of required changes, but the suggestions so far seem to revolve around changes to the stand-alone `conda` package, not `conda-build` (cc @stuarteberg). Am I missing something?

stuarteberg:
>the suggestions so far seem to revolve around changes to the stand-alone `conda` package, not `conda-build`

Right.  Just to summarize the above conversation for other readers, my only suggestions so far have been:

1. See if you can eliminate the need for `post-link.sh` in all lua recipes by having `conda` handle `luarocks` synchronization automatically.
2. Have your `lua` and `luarocks` packages depend on each other, and add a special case to `conda.toposort()` to explicitly account for that.  (If you don't do that, it's not a disaster, but then `post-link.sh` scripts will not always run in the proper order.)

That's all I've got. (Hopefully I added more value than noise :confused:.) But now this PR needs attention from an actual conda dev.  Pinging @ilanschnell @msarahan @groutr:  Should @alexbw continue his work on this PR (and build on it with a separate PR to the `conda` repo), or are you guys (understandably) too swamped to look at it in the near term anyway?

Regarding the actual code of this PR, it all looks reasonable to me, for what that's worth.  I don't know enough about `luarocks` to spot any issues in `conda_build/luarocks.py`, but it seems like you know what you're doing :-)

I guess one item for your task list would be to put a simple lua test recipe in `tests/test-recipes/metadata`.  (Admittedly, there is no existing coverage for `perl` or `R` recipes, so you won't have any examples of non-python tests to mimic.  But I think it will be straightforward nonetheless.)

kalefranz:
@groutr, when you have a moment, could you review this PR and give some input?

conda_build/main_build.py
```diff
@@ -30,6 +30,7 @@
     'numpy': [16, 17, 18, 19, 110],
     'perl': None,
     'R': None,
+    'lua': ["2.0", "5.1", "5.2", "5.3"]
```
kalefranz: I think the `.` could have implications for how conda parses package names.  Probably best to follow the pattern already established for python and numpy.

stuarteberg: I'm not so sure -- `perl` and `R` both allow `.` characters in their versions.  Only python and numpy use the strange `27` and `110` conventions.  Since that doesn't generalize very well (especially when there are more than two version numbers, such as R's `5.18.2`), then IMHO using dots should be the preferred convention.  We should keep the python/numpy version convention around for historical reasons, but just for those two "langs", not for new langs.

stuarteberg: >perl and R both allow . characters in their versions.

Oops, I meant to include a link:
https://github.com/conda/conda-build/blob/master/conda_build/config.py#L21

kalefranz: Ah thanks @stuarteberg.  Still learning the code base :)

kalefranz:
@alexbw Can you update the `CHANGELOG.txt` as part of this PR?

alexbw:
Done

alexbw:
@kalefranz with addition to changelog, what else required for a merge? 

I can't quite figure out why the tests are failing, none of the errors seem to be related to the code changes I've made.

alexbw:
@kalefranz @groutr Any chance for some eyes on this PR?

JanSchulz:
If you are on it: any chance of adding julia as well?

alexbw:
That would be a whole separate PR, but I'd be happy to walk an interested
in person through how to go about doing that.

jakirkham:
Can't tell what failures there are as it looks like there are merge conflicts. Could you try and resolve them so we can see them?

alexbw:
Tests pass. Worth a merge?

kalefranz:
@ilanschnell Can you review the naming conventions used here to make sure they're compatible with conda conventions for build strings and the like?

alexbw:
@kalefranz @ilanschnell Happy to change any naming schemes as needed.

alexbw:
Would it be possible to try a merge soon before the codebase diverges and test fail again?

kalefranz:
Yeah I'm going to try to get to this soon...

alexbw:
Anything I can do to help? Hoping to start spreading conda to the Lua
community.

jakirkham:
Agree that would be nice to see some traction on this (especially after all the hard work you have put into this @alexbw). It looks like there are some pretty cool tools in Lua (like [torch]( http://torch.ch/ )) that would be fun to play with in `conda`.

pelson:
@alexbw - when this does get merged, you might be interested in packaging through conda-forge. There are a few steps we would need to go through (such as we did for Perl and R: https://github.com/SciTools/conda-build-all/pull/27), but it shouldn't be too onerous to get off the ground with community packaged Lua on http://conda-forge.github.io.

alexbw:
I love it! Can you send me an email so we can start discussing this? I have
questions re: how versioning is decided.

kalefranz:
OK @alexbw. Finally able to spend time here again.  Would like to merge it soon as part of a conda-build `1.20.0` release.

I think my only issue with merging at this point is the versioned selectors in `conda_build/metadata.py`.  Right now I'm more comfortable taking the numpy approach, where there's just an `np` selector variable.  So analogously here just have the `lua`, then remove `luajit`, `luajit20`, `lua51`, etc.

If you did that, and then fixed whatever merge conflicts cropped up again, we can merge asap :)

alexbw:
@kalefranz I removed e.g. `lua51`, `luajit21`, but kept `lua` vs `luajit`, which are equivalent to CPython vs PyPy, so it really is required on the Lua end to maintain a distinction.

kalefranz:
Merged!

alexbw:
Hooray!

jakirkham:
:fireworks:

PR #730, opened 2016-01-18
use conda_build.metadata for checking metadata

wojdyr:
A simple generalization of the `__main__` section in metadata.py
to (ab)use it for quick checking of meta.yaml files.

I just came across a syntax error in conda-recipes/test-noarch/meta.yaml
https://github.com/conda/conda-recipes/issues/554
and I was curious if there are more errors. This PR is a quick hack that I used to check all recipes in conda-recipes:

```
$ PYTHONPATH=$PWD/../conda-build ~/anaconda/bin/python -m conda_build.metadata * >/dev/null
***** Error in aubio:
  'numpy x.x' requires external setting
***** Error in binstar_client:
  
***** Error in cyrus-sasl:
  Error: meta.yaml or conda.yaml not found in cyrus-sasl
***** Error in LICENSE.txt:
  
***** Error in matplotlib-tests:
  Error: in section u'requirements': unknown key u'imports'
***** Error in README.md:
  
***** Error in r-not-working:
  Error: meta.yaml or conda.yaml not found in r-not-working
***** Error in r-packages:
  Error: meta.yaml or conda.yaml not found in r-packages
***** Error in test-noarch:
  Error: in section u'build': unknown key u'noarch'
```

It would be more verbose without `>/dev/null`.
 It's not perfect, but it doesn't add much complexity so perhaps can be merged.

### DIFF
--- a/conda_build/metadata.py
+++ b/conda_build/metadata.py
@@ -307,7 +307,7 @@ def handle_config_version(ms, ver):
 class MetaData(object):
 
     def __init__(self, path):
-        assert isdir(path)
+        assert isdir(path), "Not a directory: %s" % path
         self.path = path
         self.meta_path = join(path, 'meta.yaml')
         self.requirements_path = join(path, 'requirements.txt')
@@ -630,13 +630,13 @@ class MetaData(object):
 
 if __name__ == '__main__':
     from pprint import pprint
-    from os.path import expanduser
 
-    recipes = sys.argv[1:] or [expanduser('~/conda-recipes/pycosat')]
-    for recipe in recipes:
+    print('Quick check of meta.yaml files. Args: recipe directories.\n'
+          'Stdout: index info after FIRST-PASS parse. Stderr: parsing errors.')
+    for recipe in sys.argv[1:]:
         try:
-            m = MetaData(recipe)
-            pprint(m.info_index())
-            m.check_fields()
+            met = MetaData(recipe)
+            pprint(met.info_index())
+            met.check_fields()
         except (Exception, SystemExit) as e:
-            sys.stderr.write('***** Error in %s:\n  %s\n' % (recipe, e))
+            print('***** Error in %s:\n  %s' % (recipe, e), file=sys.stderr)
--- a/conda_build/metadata.py
+++ b/conda_build/metadata.py
@@ -632,5 +632,11 @@ if __name__ == '__main__':
     from pprint import pprint
     from os.path import expanduser
 
-    m = MetaData(expanduser('~/conda-recipes/pycosat'))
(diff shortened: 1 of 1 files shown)


JanSchulz:
See also here: https://github.com/conda-forge/conda-smithy/blob/master/conda_smithy/lint_recipe.py

CC: @pelson

wojdyr:
@JanSchulz: yes, I've seen it because I googled for conda lint before :)
I like the idea. This PR is a quick hack not a proper linter.
For myself I also have a small script to check that:
 - source/fn is the same as filename from source/url (unless it's an exception)
 - package/name is the same as the recipe's directory
 - about/license_file is defined (to be sure the license text is distributed with binaries)

and a few other ad-hoc conventions

stuarteberg:
This PR strikes a good balance between finding gross violations of yaml syntax without running all of the preprocessing that would normally be required to fully parse a `meta.yaml` file.  (After all, a `meta.yaml` file can't be guaranteed to be fully parsed without actually downloading the source, and also parsing certain configuration arguments to conda, like `--numpy=1.9`.)

I have two minor comments, neither of which is really a big deal, I think:

- I'm not sure what selectors like `# [np]` will resolve to, since the numpy version won't be defined yet.
- Some otherwise-valid jinja templates may still be undefined at the end of the `MetaData()`constructor.  I think that won't screw anything up, but it does prevent you from expanding this PR into a more general linting tool. For instance, we can call `check_fields()` as this PR does, but we can't really check the *contents* of the fields, because they won't all be known yet.  That's not a criticism of this PR; rather, it's a justification for not asking for more from it. :-)

FWIW, I think maybe a quick 1 or 2 lines of comments would be appropriate in your changes, just to explain what this code does, and point out it's benefits/limitations.

wojdyr:
@stuarteberg thanks for your review. I've added a two-line message to explain what the code is for, stressing that it does only the "first-pass" parsing.

msarahan:
This looks like a helpful tool.  Would you consider making a CLI for it, like the other conda build commands?  You might also want to tie in ```conda render``` - added in https://github.com/conda/conda-build/pull/908 - but this is for a single recipe, whereas you recurse.

We definitely need some kind of linter in conda build - perhaps based on your work here and https://github.com/conda-forge/conda-smithy/blob/master/conda_smithy/lint_recipe.py

wojdyr:
It was only a small enhancement to the test at the end of conda_build/metadata.py,
nothing important. Since that test is gone now, I think it's best to just close this PR.

msarahan:
Thanks.  Regardless, we will be implementing some linter here, and I appreciate you pointing this possibility out.

Issue #362, opened 2015-04-09
Unable to build with tar.Z source archive


vitale232:
I'm attempting to install [HDF-EOS2](http://hdfeos.org/software/library.php) from source using conda-build.
  fn: HDF-EOS2.19v1.00.tar.Z
  url: ftp://edhs1.gsfc.nasa.gov/edhs/hdfeos/latest_release/HDF-EOS2.19v1.00.tar.Z

conda-build does not seem to extract the tar.Z archive properly, thus there is no ./configure file when the build.sh script is invoked.

Is there a workaround to perhaps convert the tar.Z archive to a tar.gz archive in the beginning of the build script?

Here is the conda build output:
----
$ conda build hdf-eos2/
Removing old build directory
Removing old work directory
BUILD START: hdf-eos2-2.19.1-py27_0
Fetching package metadata: ........
Solving package specifications: .
The following NEW packages will be INSTALLED:

    hdf4:       4.2.11-py27_0
    jpeg:       8d-0         
    openssl:    1.0.1k-1     
    pip:        6.1.1-py27_0 
    python:     2.7.9-2      
    readline:   6.2-2        
    setuptools: 15.0-py27_0  
    sqlite:     3.8.4.1-1    
    system:     5.8-2        
    tk:         8.5.18-0     
    zlib:       1.2.8-0      

Linking packages ...
[      COMPLETE      ]|####################################################################################################################################| 100%
This is an install message from the hdf4 package.
Removing old work directory
Source cache directory is: /home/vitale232/miniconda/conda-bld/src_cache
Found source in cache: HDF-EOS2.19v1.00.tar.Z
Extracting download
Warning: Unrecognized source format. Source file will be copied to the SRC_DIR
Package: hdf-eos2-2.19.1-py27_0
source tree in: /home/vitale232/miniconda/conda-bld/work
\+ echo 'This is the prefix: ' /home/vitale232/miniconda/envs/_build
This is the prefix:  /home/vitale232/miniconda/envs/_build
\+ ./configure --with-jpeg=/home/vitale232/miniconda/envs/_build --with-zlib=/home/vitale232/miniconda/envs/_build --with-hdf4=/home/vitale232/miniconda/envs/_build
/home/vitale232/Documents/conda_ledaps-build/hdf-eos2/build.sh: line 5: ./configure: No such file or directory
Command failed: /bin/bash -x -e /home/vitale232/Documents/conda_ledaps-build/hdf-eos2/build.sh

Here is conda info:
[...]

PR #731, opened 2016-01-20
Find uncompress in old Linux

ocefpaf:
This PR changes `uncompress` to `gunzip`.

`uncompress` is an alias for `gunzip` and that alias is not always defined in old Linux systems (like CentOS 6).  (See the failures [here](https://circleci.com/gh/ioos/conda-recipes/929) for an example.)

For more information see the discussion in https://github.com/conda/conda-build/issues/362#issuecomment-101066000.

### DIFF
--- a/conda_build/utils.py
+++ b/conda_build/utils.py
@@ -62,8 +62,10 @@ def tar_xf(tarball, dir_path, mode='r:*'):
     if tarball.lower().endswith('.tar.z'):
         uncompress = external.find_executable('uncompress')
         if not uncompress:
+            uncompress = external.find_executable('gunzip')
+        if not uncompress:
             sys.exit("""\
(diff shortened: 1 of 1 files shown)


msarahan:
+1 - but @ilanschnell should comment, in case there's some weird corner case.

ilanschnell:
How about we look for `uncompress` first, and if that is not found, fall back to `gunzip`.

ocefpaf:
> How about we look for `uncompress` first, and if that is not found, fall back to `gunzip`.

In theory that is not needed, since `uncompress` is `gunzip`.  But better safe than sorry.  I just changed the PR to look for `uncompress` first.

msarahan:
Thanks!

ilanschnell:
Thanks, yes just checking for `gunzip` is good enough.

PR #744, opened 2016-01-28
Bugfix to restore functionality: building for multiple python versions on Windows

Korijn:
There is a bug in conda-build that makes it impossible to build for multiple python versions in one line on Windows (e.g., `--py 35 --py 34 --py 27`), because of this assertion: https://github.com/conda/conda-build/blob/237e8fd2643863b6c1df6a4fd413dba8c24119b5/conda_build/main_build.py#L275 

Obviously that check will fail on any build after the first. The *real* solution for the issue that assert is trying to fix, would be to *not* delete the trash under such conditions, and to warn the user with a message that actually makes sense. This PR does so and restores the functionality.


ilanschnell:
Thanks!

Issue #647, opened 2015-10-21
Git required to build packages without git source


mausch:
anaconda-client: 1.0.2-py34_0 
conda-build:     1.18.1-py34_0

I have a package with a local source path. When I try to build it in an environment without `git` I get the following error:

```
BUILD START: my_package-1.0.22-np19py34_0
Fetching package metadata: ........
Solving package specifications: ................................
The following packages will be downloaded:

    package                    |            build
    ---------------------------|-----------------
    ...
    ------------------------------------------------------------
                                           Total:        47.4 MB

The following NEW packages will be INSTALLED:

    ...

Fetching packages ...
...
An unexpected error has occurred, please consider sending the
following traceback to the conda GitHub issue tracker at:

    https://github.com/conda/conda-build/issues

Include the output of the command 'conda info' in your report.

Traceback (most recent call last):
  File "/opt/conda/bin/conda-build", line 5, in <module>
    sys.exit(main())
  File "/opt/conda/lib/python3.4/site-packages/conda_build/main_build.py", line 190, in main
    args_func(args, p)
  File "/opt/conda/lib/python3.4/site-packages/conda_build/main_build.py", line 468, in args_func
    args.func(args, p)
  File "/opt/conda/lib/python3.4/site-packages/conda_build/main_build.py", line 391, in execute
    override_channels=args.override_channels, include_recipe=args.include_recipe)
  File "/opt/conda/lib/python3.4/site-packages/conda_build/build.py", line 395, in build
    m.parse_again()
  File "/opt/conda/lib/python3.4/site-packages/conda_build/metadata.py", line 357, in parse_again
    self.meta = parse(get_contents(self.meta_path))
  File "/opt/conda/lib/python3.4/site-packages/conda_build/metadata.py", line 300, in get_contents
    env.globals.update(context_processor())
  File "/opt/conda/lib/python3.4/site-packages/conda_build/jinja_context.py", line 44, in context_processor
    ctx = get_environ()
  File "/opt/conda/lib/python3.4/site-packages/conda_build/environ.py", line 106, in get_dict
    d.update(**get_git_build_info(d['SRC_DIR']))
  File "/opt/conda/lib/python3.4/site-packages/conda_build/environ.py", line 47, in get_git_build_info
    env=env)
  File "/opt/conda/lib/python3.4/subprocess.py", line 859, in __init__
    restore_signals, start_new_session)
  File "/opt/conda/lib/python3.4/subprocess.py", line 1457, in _execute_child
[...]

Issue #713, opened 2015-12-23
Versioning conda packages via git no longer works


jhunkeler:
Hi,

It seems `1.18.2` has broken versioning for git builds. The wall of text below shows the build process of a simple package in both the latest release of `conda-build` as well as the previous release. 

The git repository I am building against has an annotated tag "v0.1" and is currently 25 commits ahead.

The expected output file: 
`specview-v0.1.git-py35_25.tar.bz2`

The actual output file: 
`specview-0.0.0.git-py35_0.tar.bz2`

I'm hoping I am just doing something wrong, or maybe the documentation hasn't been updated to reflect a new requirement for building with git. Either way, I've been stuck using `1.18.1` since `1.18.2` was released.

**Quick Notes:** 
* `conda-build`'s git master recipe produces the same result as the `1.18.2` release version. Just in case you wanted me to try it ;)
* In testing I modified the `environ.get()` call to match conda's documentation, but the result was the same. I've been building with these `GIT_` references for several months without encountering any issues.
* This appears to be OS agnostic. The results are the same on OS X and Linux.

## meta.yaml

``` yaml
about:
    home: https://github.com/spacetelescope/specview
    license: BSD
    readme: README.rst
    summary: specview
build:
    number: '{{ environ.get("GIT_DESCRIBE_NUMBER", 0) }}'
package:
    name: specview
    version: '{{ environ.get("GIT_DESCRIBE_TAG", "0.0.0") }}.git'
requirements:
    build:
    - astropy
    - numpy
    - scipy
    - pyqtgraph
    - setuptools
    - python
    run:
    - astropy
    - numpy
    - scipy
    - pyqtgraph
    - python
source:
    git_url: https://github.com/spacetelescope/specview
```

## 1.18.2

```
BUILD START: specview-0.0.0.git-py35_0
Fetching package metadata:
Solving package specifications:
The following packages will be downloaded:

    package                    |            build
    ---------------------------|-----------------
    pyqtgraph-0.9.10           |           py35_0         890 KB

The following NEW packages will be INSTALLED:

    astropy:    1.1-np110py35_0
    freetype:   2.5.5-0
    libpng:     1.6.17-0
    numpy:      1.10.2-py35_0
    openssl:    1.0.2d-0
    pip:        7.1.2-py35_0
    pyopengl:   3.1.1a1-np110py35_0
    pyqt:       4.11.4-py35_1
    pyqtgraph:  0.9.10-py35_0
    python:     3.5.1-0
[...]

PR #745, opened 2016-01-28
fix building when git is not installed

ilanschnell:
This simple PR fixes https://github.com/conda/conda-build/issues/647

### DIFF
--- a/conda_build/environ.py
+++ b/conda_build/environ.py
@@ -117,7 +117,9 @@ def get_dict(m=None, prefix=None):
         except NotImplementedError:
             d['CPU_COUNT'] = "1"
 
-    d.update(**get_git_build_info(d['SRC_DIR']))
+    if m.get_value('build/git_url'):
+        d.update(**get_git_build_info(d['SRC_DIR']))
+
     d['PATH'] = dict(os.environ)['PATH']
     d = prepend_bin_path(d, prefix)
 


ilanschnell:
@kalefranz please review

jasongrout:
As noted in https://github.com/conda/conda-build/issues/713#issuecomment-178235625, this change broke a lot of recipes that depended on git being automatically detected, and probably should be documented in the changelog as a breaking change requiring an update to meta.yaml files (e.g., `path: ... -> git_url: ...`. Also, this change breaks `python -m conda_build.environ` since we don't check to see if `m` is `None`.

ilanschnell:
Sorry for the breakage.

jakirkham:
Just so I understand what broke here, `path` could have been used to specify the a git URL before? Also, I don't think I have ever used `conda_build.environ` before what did this do before it stopped working?

jasongrout:
When debugging, I saw the if statement at the bottom of the file, which
indicates it can be run separately. It looks like it will print out the
environment.

jasongrout:
To be clear, I support the change. It prevents problems I've had in the
past with the auto guessing, like
https://github.com/conda/conda-build/issues/529.  It just caught us by
surprise when it broke our builds.

frol:
<irony>Super change! I love it!</irony>

I have a number of internal recipes, which are cloned by Jenkins, so the ssh credentials for the repos are not available to make another clone on `conda build`, also, it is quite annoying to re-clone a huge project with a number of submodules on every build. `path` options worked just perfect and now it is gone... What is the recommended way of dealing with this kind of workflow?

P.S. The reason I hit by this "feature" only today is that I was using conda-build 1.18.1 because of the broken 1.18.2 release. Today, in a hope of that the bugs of 1.18.2 were solved, I moved on 1.20.0... and here I am.

PR #753, opened 2016-02-02
Add git_recursive option (defaults to true)

gmarkall:
Since the change in #745, the requirement to use `git_url` instead of `path` even for local git repositories forces a clone of any submodules every time a conda build is run (the `--mirror` option used to create the local cache does not mirror submodules). This can be a problem if the submodules are large (e.g. several hundreds of MB is not uncommon) and especially if they are not actually required for the build.

This commit adds a `git_recursive` option, which defaults to `true`. When `true`, the git clone that conda build makes is recursive, and submodules are included; when `false`, the clone is not recursive and submodules are not cloned.

This is one way that I could see around the issue, so I've had a go at the implementation of it here - I thought about adding a test recipe for this too, but I am not really sure how to effectively test this change, though I have tested it locally with a recipe inside a repository that used submodules.

Any feedback is appreciated!

### DIFF
--- a/conda_build/metadata.py
+++ b/conda_build/metadata.py
@@ -181,7 +181,8 @@ def parse(data):
     for field in ('build/osx_is_app', 'build/preserve_egg_dir',
                   'build/binary_relocation',
                   'build/detect_binary_files_with_prefix',
-                  'build/skip', 'app/own_environment'):
+                  'build/skip', 'app/own_environment',
+                  'source/git_recursive'):
         section, key = field.split('/')
         if res.get(section) is None:
             res[section] = {}
@@ -258,6 +259,7 @@ FIELDS = {
     'package': ['name', 'version'],
     'source': ['fn', 'url', 'md5', 'sha1', 'sha256', 'path',
                'git_url', 'git_tag', 'git_branch', 'git_rev', 'git_depth',
+               'git_recursive',
                'hg_url', 'hg_tag',
                'svn_url', 'svn_rev', 'svn_ignore_externals',
                'patches'],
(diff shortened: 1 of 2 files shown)


jakirkham:
Sounds ok to me. Maybe you can copy-paste and modify one of the git test cases to show this in action?

gmarkall:
> Maybe you can copy-paste and modify one of the git test cases to show this in action?

I could do that and it would provide a demonstration of its use, but it would be trickier to actually have the test pass/fail depending on whether the clone was recursive - would the demonstration of its use alone be sufficient though?

stuarteberg:
I wonder how the `source/git_depth` setting in `meta.yaml` relates to this.  I have no idea what happens if you run `git clone --recursive --depth=1 http://foobar.com/yadayada`.  Does the `--depth` setting apply to the submodules, too?  If so, perhaps that at least partially alleviates the concern about huge repos mentioned above.

groutr:
@stuarteberg This might shed some light: http://alblue.bandlem.com/2011/11/git-tip-of-week-git-submodules.html
I don't think the depth setting is propagated to submodules.  You can pass a depth argument to submodules for add and update operations though (see bottom of this page https://git-scm.com/docs/git-submodule/2.6.5)

jakirkham:
While depth is an interesting question, I think it is orthogonal to the OP's proposal. Could we please move the depth discussion into a new issue?

jakirkham:
@gmarkall, fair question, sorry I haven't gotten back to you on this. So, I went ahead and tried to clone a [repo]( https://github.com/jakirkham/docker_449_test ) that has a submodule. If I don't pass the recursive option, it will create an empty directory for the submodule. However, if I do pass the recursive option, the directory will be populated. One particularly good file to check for would be the `.git` file in the submodule. This file is only created if a recursive clone (or later recursive fetch or other operation) is performed to create it.

PR #758, opened 2016-02-03
Don't define GIT_ environment variables if the work directory is out-of-sync

stuarteberg:
Before the package source is downloaded, variables like `GIT_DESCRIBE_TAG` can be given incorrect values if a previous build left a git repo in the source directory.  It's not a huge problem, but it does produce incorrect status messages until the source is finally downloaded.

This PR fixes it by leaving those `GIT_` variables undefined if the `work` directory doesn't correspond to the package being built.

### DIFF
--- a/conda_build/environ.py
+++ b/conda_build/environ.py
@@ -36,14 +36,41 @@ def get_stdlib_dir():
 def get_sp_dir():
     return join(get_stdlib_dir(), 'site-packages')
 
-def get_git_build_info(src_dir):
+def get_git_build_info(src_dir, git_url, expected_rev):
     env = os.environ.copy()
     d = {}
     git_dir = join(src_dir, '.git')
-    if os.path.exists(git_dir):
-        env['GIT_DIR'] = git_dir
-    else:
+    if not os.path.exists(git_dir):
         return d
+
+    env['GIT_DIR'] = git_dir
+    try:
+        # Verify current commit matches expected commit
+        current_commit = subprocess.check_output(["git", "log", "-n1", "--format=%H"], env=env)
+        current_commit = current_commit.decode('utf-8')
+        expected_tag_commit = subprocess.check_output(["git", "log", "-n1", "--format=%H", expected_rev], env=env)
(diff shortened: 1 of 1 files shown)


msarahan:
Sorry, this looks like some Python 3 string issues at https://github.com/conda/conda-build/pull/758/files#diff-8cf4f5bfdd4544775a2f28ef5e24b2c5R59 .  Otherwise OK, thanks for fixing.

stuarteberg:
>looks like some Python 3 string issues

Whoops, thanks.  Fixed now.

conda_build/environ.py
```diff
@@ -118,7 +145,9 @@ def get_dict(m=None, prefix=None):
             d['CPU_COUNT'] = "1"
 
     if m.get_value('source/git_url'):
-        d.update(**get_git_build_info(d['SRC_DIR']))
+        d.update(**get_git_build_info(d['SRC_DIR'],
+                                      m.get_value('source/git_url'),
+                                      m.get_value('source/git_rev', default='master')))
```
msarahan: Is there a way to check for git_tag here, also?

stuarteberg: That was my first try, but it turns out [`git_rev` is a synonym for `git_tag`][1], and it's the canonical spelling.

[1]: https://github.com/stuarteberg/conda-build/blob/7ce7be2095848fc43b61da18492234a1694cef95/conda_build/metadata.py#L219-L228

msarahan: Thanks for pointing that out.  I just want to make sure: this fix works for people who specify git_tag in their recipe, rather than git_rev?

stuarteberg: Right.  `git_tag` and `git_branch` are copied into `git_rev` (and then deleted -- `git_tag` is empty at this point).

msarahan: Thanks for the clarification.  Merging.

jakirkham: Interesting, I saw the introduction of `git_rev` in the docs and was wondering what was going on. Will `git_tag` and `git_branch` then be deprecated at some point.

PR #750, opened 2016-02-01
WIP: Add version control environmental variables for --output flag

ericdill:
**What problem does this solve?**

Gives `conda-build --output` a mechanism to correctly report the version string for a conda package 
that uses environmental variables from version control (e.g., GIT_DESCRIBE_NUMBER).

There is a lot of extra, unnecessary output in this (admittedly) hacky approach.  I copy/pasted two lines from elsewhere in the codebase that give conda-build the ability to use version control env vars to report output. Unfortunately, this resulted in the git clone and git checkout information to be dumped to the screen which breaks the ability to do:
```
$ anaconda upload -u user `conda build /path/to/recipe --output`
```
Is there any way to add this proposed functionality without breaking the aforementioned use case?

**Example of the output before and after this PR**

For example, trying to use `conda-build conda-build/conda-build.recipe --output` currently
returns the following build string:
```
$ conda build conda-build/conda_build.recipe/ --output
/home/edill/miniconda/conda-bld/linux-64/conda-build-1.10alpha.0-py35_GIT_STUB.tar.bz2
```

And with this change --output returns (along with a lot of extra output):
```
$ conda build conda-build/conda_build.recipe/ --output
Removing old work directory
remote: Counting objects: 4, done.
remote: Compressing objects: 100% (4/4), done.
remote: Total 4 (delta 2), reused 0 (delta 0)
Unpacking objects: 100% (4/4), done.
From /home/edill/dev/conda/conda-build/conda_build.recipe/..
   21ea84a..a37551a  master     -> master
   21ea84a..a37551a  origin/HEAD -> origin/HEAD
   21ea84a..a37551a  origin/master -> origin/master
 * [new tag]         1.19.0     -> 1.19.0
checkout: 'c7b0f5b8eb03b99c50198aa7d0e9104d0612061d'
Cloning into '/home/edill/miniconda/conda-bld/work'...
done.
Note: checking out 'c7b0f5b8eb03b99c50198aa7d0e9104d0612061d'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by performing another checkout.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -b with the checkout command again. Example:

  git checkout -b <new-branch-name>

HEAD is now at c7b0f5b... MNT: Add GIT_* env vars for --output
==> git log -n1 <==

commit c7b0f5b8eb03b99c50198aa7d0e9104d0612061d
[...]

PR #759, opened 2016-02-03
Fix --output command by downloading the source first if necessary

stuarteberg:
~~Do not merge yet.  This PR depends on #758, so review that first.~~

This PR builds on #750, but avoids downloading the source when it's not necessary.  Simply put, the `--output` flag doesn't produce the right answer if the `build_id` depends on downloading the source (so the git tag can be checked).  This PR downloads the source first, but only if necessary.

### DIFF
--- a/conda_build/main_build.py
+++ b/conda_build/main_build.py
@@ -385,6 +385,17 @@ def execute(args, parser):
                       "configuration." % m.dist())
                 continue
             if args.output:
+                try:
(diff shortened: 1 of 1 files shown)


msarahan:
would you mind rebasing or merging master on this?  It seems like there's duplicate changes in environ.py

stuarteberg:
>would you mind rebasing or merging master on this?

Done.

msarahan:
Thanks!

Issue #754, opened 2016-02-02
conda build --output doesn't expand GIT_xxx environment variables anymore


pitrou:
This is a regression in 1.19.0. When running `conda build --no-output`, the GIT_DESCRIBE_TAG and GIT_DESCRIBE_NUMBER variables are not expanded anymore in the meta.yaml, therefore the generated filename is wrong.

pitrou:
Ping @jakirkham @meawoppl

jakirkham:
Thanks for doing this. The flag is `--output`, correct?

jakirkham:
So, the package should `<CONDA_ROOT>/conda-bld/<OS>-<ARCH>/<PACKAGE>-<VERSION>-<BUILDSTR>_<FEATURE>_<BUILDNUM>.tar.bz2`. Using this style of path directly should at least provide you a workaround for finding the package.

jakirkham:
@meawoppl, so I don't know that there is a way to specify the path, but all the paths look something like the one above.

ericdill:
@pitrou what version of conda-build did this work in? `conda-build /path/to/recipe --output` does not fill in the GIT_* vars for 1.17, 1.18, 1.18.1 either

pitrou:
On our CI systems this worked this 1.18.x, it seems. Perhaps there are some other factors to the variability?

jakirkham:
Was it 1.18.2, @pitrou? Did you test that as well, @ericdill?

pitrou:
1.18.2 indeed, apparently.

pitrou:
By the way, I think jinja2 should be hard dependency of conda-build. It's too easy to forget installing it and get weird variations.

ericdill:
1.18.2 totally fails with GIT_* vars. 
```
$ conda build --version
conda-build 1.18.2

$ conda-build recipes/databroker/ --output
Error: Failed to parse jinja template in /home/edill/dev/conda/staged-recipes-dev/recipes/databroker/meta.yaml:
'dict object' has no attribute 'GIT_DESCRIBE_TAG'

$ cat recipes/databroker/meta.yaml 
package:
  name: databroker
  version: {{ environ['GIT_DESCRIBE_TAG'] }}.post{{ environ['GIT_DESCRIBE_NUMBER'] }}
<snip>
```

[...]

PR #761, opened 2016-02-03
MNT: Remove git spam from conda build --output

ericdill:
@stuarteberg Could you have a look at this to make sure that I am not introducing any unintended consequences?  

This PR silences the output from the git commands that get called with the `--output` flag and adds a `--verbose` flag to the conda build CLI.  This --verbose flag sets a logger to the debug level so that the output that was being directed to stderr by git and calls to print in source.py and are now sent to the console through the built in logging framework instead.

Deals with some discussion in #754. 
Retains backwards compatibility with certain use cases of `conda build /path --output` that #758 and #759 broke.

### DIFF
--- a/conda_build/main_build.py
+++ b/conda_build/main_build.py
@@ -185,10 +185,28 @@ different sets of packages."""
         metavar="R_VER",
         choices=RVersionsCompleter(),
     )
+    p.add_argument(
+        '-v',
+        '--verbose',
+        action='store_true',
+        help="Enable verbose mode. Turns on extra debug output"
+    )
     add_parser_channels(p)
     p.set_defaults(func=execute)
 
     args = p.parse_args()
+    # enable logging. Default to warning
+    import logging
+    stream = logging.StreamHandler()
+    from .utils import logger
+    if args.verbose:
+        loglevel = logging.DEBUG
+    else:
+        loglevel = logging.WARNING
+    stream.setLevel(loglevel)
+    logger.setLevel(loglevel)
+    logger.addHandler(stream)
+
     args_func(args, p)
 
 
--- a/conda_build/main_build.py
+++ b/conda_build/main_build.py
@@ -197,7 +197,7 @@ different sets of packages."""
     args = p.parse_args()
     # enable logging. Default to warning
     import logging
-    stream = logging.StreamHandler()
+    stream = logging.StreamHandler(sys.stdout)
     from .utils import logger
     if args.verbose:
         loglevel = logging.DEBUG
--- a/conda_build/main_build.py
+++ b/conda_build/main_build.py
@@ -409,7 +409,7 @@ def execute(args, parser):
                     # Something went wrong; possibly due to undefined GIT_ jinja variables.
                     # Maybe we need to actually download the source in order to resolve the build_id.
                     source.provide(m.path, m.get_section('source'))
-                    
+
                     # Parse our metadata again because we did not initialize the source
                     # information before.
                     m.parse_again(permit_undefined_jinja=False)
--- a/conda_build/source.py
+++ b/conda_build/source.py
@@ -135,7 +135,7 @@ def git_source(meta, recipe_dir):
     # assume the user wants the current HEAD
     if not checkout and git_url.startswith('.'):
         checkout = check_output(["git", "rev-parse", "HEAD"], cwd=git_url,
-                                stderr=STDOUT).decode()
+                                stderr=STDOUT).decode().strip()
     if checkout:
         logger.debug('checkout: %r' % checkout)
 
--- a/conda_build/source.py
+++ b/conda_build/source.py
@@ -5,7 +5,7 @@ import sys
 from os.path import join, isdir, isfile, abspath, expanduser
 from shutil import copytree, copy2
 from subprocess import (check_call, CalledProcessError, check_output, STDOUT)
-
+import pdb
 from conda.fetch import download
 from conda.utils import hashsum_file
 
@@ -110,15 +110,7 @@ def git_source(meta, recipe_dir):
 
     # update (or create) the cache repo
     if isdir(cache_repo):
-        try:
-            output = check_output([git, 'fetch'], cwd=cache_repo,
-                                  stderr=STDOUT).decode()
-            logger.debug(output)
-        except CalledProcessError as cpe:
-            # output the actual git error message in case something goes wrong.
-            # then reraise the exception
-            print(cpe.output.decode())
-            raise
+        _git_call([git, 'fetch'], cwd=cache_repo)
     else:
         args = [git, 'clone', '--mirror']
         if git_depth > 0:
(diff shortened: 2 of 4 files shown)


conda_build/source.py
```diff
@@ -155,22 +160,14 @@ def git_info(fo=None):
...
-        stdout = stdout.decode(encoding, 'ignore')
-        stderr = stderr.decode(encoding, 'ignore')
-        if check_error and stderr and stderr.strip():
-            raise Exception("git error: %s" % stderr)
+        stdout = check_output(cmd.split(), cwd=WORK_DIR, env=env,
+                              stderr=STDOUT).decode()
         if fo:
-            fo.write(u'==> %s <==\n' % cmd)
-            fo.write(stdout + u'\n')
+            logger.debugfo.write(u'==> %s <==\n' % cmd)
```
stuarteberg: typo here

conda_build/main_build.py
```diff
@@ -185,10 +185,28 @@ def main():
...
+        action='store_true',
+        help="Enable verbose mode. Turns on extra debug output"
+    )
     add_parser_channels(p)
     p.set_defaults(func=execute)
 
     args = p.parse_args()
+    # enable logging. Default to warning
+    import logging
+    stream = logging.StreamHandler()
```
stuarteberg: If memory serves, `StreamHandler` defaults to `stderr`.  If that's not what you intended, you should write `StreamHandler(sys.stdout)`.

ericdill: Oh wow. That's good to know and you are correct. I definitely just assumed that it went to stdout.

Relevant docstring:
```
Signature: logging.StreamHandler.__init__(self, stream=None)
Docstring:
Initialize the handler.

If stream is not specified, sys.stderr is used.
```

conda_build/source.py
```diff
@@ -112,13 +111,16 @@ def git_source(meta, recipe_dir):
 
     # update (or create) the cache repo
     if isdir(cache_repo):
-        check_call([git, 'fetch'], cwd=cache_repo)
+        output = check_output([git, 'fetch'], cwd=cache_repo, stderr=STDOUT).decode()
+        logger.debug(output)
     else:
```
stuarteberg: By changing `check_call()` into `check_output()`, you are hiding the output (which is the intention here, obviously).  But if something goes wrong in the call, I want to see what the output from git was just before it failed.  I think this would be sufficient:

```python
try:
    output = check_output([git, 'fetch'], cwd=cache_repo, stderr=STDOUT).decode()
    logger.debug(output)
except subprocess.CalledProcessError as e:
    print(ex.output)
    raise # Or maybe sys.exit()?
```

ericdill:
@stuarteberg Thanks for the review.  I would imagine that all the git calls should be wrapped in a try/except block that will print the output on a failed call?

stuarteberg:
>I would imagine that all the git calls should be wrapped in a try/except block that will print the output on a failed call?

I guess so, except for the ones that are expected to possibly fail, such as the ones in `environ.get_git_build_info()`

ericdill:
Oh, yeah, I forgot that there are git calls in other modules.  I was really only thinking about the git calls in source.py

conda_build/utils.py
```diff
@@ -15,6 +15,8 @@
 
 from conda_build import external
 
+import logging
+logger = logging.getLogger('conda-build')
```
stuarteberg: I don't know what conventions other projects use, but in my projects I follow the convention that every source file uses it's own logger (`logger = logging.getLogger(__name__)`), as suggested in the python logging module documentation.  In this PR, you are only declaring a single logger, which is shared across all modules.  Maybe a brief discussion of the pros/cons of each convention is warranted.

ericdill: I actually use `logging.getLogger(__name__)` too.  I thought that this was a convention that was used in `conda`, but I was mixing that up with some other project and overall just not following the correct convention.  I'll rework this to use __name__ in modules that now use the logging framework

stuarteberg:
I think this PR is a step in the right direction, but it feels like a half-measure in it's current form.  If we start using `logger.[debug/info/warn]` instead of `print` in some places, why not start using them basically everywhere?  (I think that would be my preference.)

Also, the boilerplate around the subprocess calls might be getting excessive.  I know there was a previous attempt at eliminating that boilerplate (#621), but it broke a lot of stuff and had to be reverted.  Is it maybe worth someone's time to revisit the idea?

Also, it's kinda nice to see the actual git output as it's downloading the source code, so you can monitor progress.  For instance, if I'm having a network issue (a common problem when I'm using a VM on my laptop and moving from one network to another), sometimes the first symptom is that git merely hangs while fetching a repo.  But with this PR, the entire output from git is only shown after the operation completes.  I don't think it will be obvious what is hanging.

PS -- Gotta step away for a while. Will continue discussion later today.

ericdill:
> feels like a half-measure in it's current form. 

Yes, I agree.

> If we start using logger.[debug/info/warn] instead of print in some places, why not start using them basically everywhere? (I think that would be my preference.)

I agree.

> Also, it's kinda nice to see the actual git output as it's downloading the source code, so you can monitor progress. For instance, if I'm having a network issue (a common problem when I'm using a VM on my laptop and moving from one network to another), sometimes the first symptom is that git merely hangs while fetching a repo. But with this PR, the entire output from git is only shown after the operation completes. I don't think it will be obvious what is hanging.

My primary goal for this PR is to retain this workflow:

```
conda build some-recipe
anaconda -t <TOKEN> upload -u some-other-user `conda build some-recipe --output` 
```

because I am depending on that to build the master branches of our software stack right now.  However, as you point out, silencing the git output has the very real downside of masking what is hanging, when conda-build hangs.  One alternative that actually feels cleaner and would enable the same functionality is to do what @pitrou suggested in #754/#755 and allow conda-build to pass a user and token for uploading.

Perhaps this PR should die and be replaced with one that adds `-u/--user` and `-t/--token` flags to conda-build. Then the git output doesn't need to be silenced and switching to the logging framework can be punted on (though I agree it is a good idea).

Thoughts @stuarteberg @pitrou @jakirkham?

ericdill:
@pitrou @jakirkham @stuarteberg Any thoughts on killing this in favor of `-u/--user` and `-t/--token` flags to conda-build?

pitrou:
That's entirely fine with me.

stuarteberg:
Fine by me, too.  But for your use case, isn't there a straightforward way to find the new package without needing to ask conda what the name is?

Instead of:

```bash
conda build some-recipe
anaconda -t <TOKEN> upload -u some-other-user `conda build some-recipe --output` 
```
If you just assume that the most recently created tarball is the one to upload, then there's no need to ask conda for the name, right?

```bash
conda build some-recipe
if [ $? -eq 0 ]; then
    NEWEST_TARBALL=$(ls -t $(conda info --root)/conda-bld/*-64/*.tar.bz2 | head -n1)
    anaconda -t <TOKEN> upload -u some-other-user $NEWEST_TARBALL
fi
```

jakirkham:
If you wish to drop this it is obviously your call @ericdill, however I still think this has value. Personally, as conda does provide this flag, I do think it should work, which IMHO it does not.

stuarteberg:
Wait a minute:

>My primary goal for this PR is to retain this workflow:
>
>      conda build some-recipe
>      anaconda -t <TOKEN> upload -u some-other-user `conda build some-recipe --output` 

That example already works.  Since you're building the recipe immediately before you call `conda build --output`, the source will already be downloaded and cached, and no git output will be produced.

Now, if your use case is more like this, then yes, there will be extra noise you'll have to ignore:

```bash
conda build some-recipe-A
conda build some-recipe-B
anaconda -t <TOKEN> upload -u some-other-user `conda build some-recipe-A --output` 
anaconda -t <TOKEN> upload -u some-other-user `conda build some-recipe-B --output` 
```

>as conda-build does provide this flag, I do think it should work, which IMHO it does not

It still "works" in the sense that the last line of output is exactly what you asked for.  It's just that there *might* be extra noise that gets printed first, iff your source is a `git_url` and it is out-of-date with the `conda-bld/work` directory.

ericdill:
> That example already works. Since you're building the recipe immediately before you call conda build --output, the source will already be downloaded and cached, and no git output will be produced.

Yeah, you're right.  I'm not sure how I concluded that this example no longer worked. I see nothing of value that this PR adds and am sorry for wasting everyone's time.

Issue #612, opened 2015-09-30
Change behavior of undefined variables to simply not define them


jakirkham:
I ran into a problem because of this change recently. ( https://github.com/conda/conda-build/blob/46a605cb5326eb92583db30869ea4ebd2f746a67/CHANGELOG.txt#L177-L180 ) It seems surprising to me that an undefined variable should be defined. Further, it seems strange that they should be defined to be anything other than an empty string as one might expect from bash. Is it possible this behavior could be changed either to not define these variables or make them an empty string?

jakirkham:
Any thoughts on this? I think this is actually breaking things now that there is this two pass jinja templating scheme. The easy solution would be to quit assign variable value if we don't want them assigned and simply warn us that they are unassigned instead of making them "<UNDEFINED>". I think I'll write up a PR.

PR #763, opened 2016-02-05
Stop setting values for undefined environment variables

jakirkham:
Fixes https://github.com/conda/conda-build/issues/612

In a recipe, a user can specify variables that can be defined externally to influence the recipe. The previous practice was to leave these undefined. However, a commit was added that began setting them to `<UNDEFINED>`. If a user intentionally left these variables undefined, they must now go back and unset them afterwards. The intent of this was likely to smoke out errors. This could initially be easily worked around in bash or batch. However, 2 pass jinja templating has forced us to handle this inside the `meta.yaml`, which is more challenging and unnecessarily ugly. Instead of continuing this trend of setting intentionally undefined variables, this breaks from that change and simply warns the user when they have an undefined environment variable.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,3 +1,10 @@
+2016-01-29   1.19.1:
+--------------------
+  * Environment variables defined in the 'script_env' build section of
+    the meta.yaml file were previously assigned the value '<UNDEFINED>'
+    if not found in the environment. Now they are left unset and a
+    warning is raised instead, #763.
+
 2016-01-29   1.19.0:
 --------------------
   * normalize unicode in conda skeleton cran, #681
--- a/conda_build/environ.py
+++ b/conda_build/environ.py
@@ -5,6 +5,7 @@ import sys
 from os.path import join, normpath, isabs
 import subprocess
 import multiprocessing
+import warnings
 
 import conda.config as cc
 
@@ -134,8 +135,12 @@ def get_dict(m=None, prefix=None):
         for var_name in m.get_value('build/script_env', []):
             value = os.getenv(var_name)
             if value is None:
-                value = '<UNDEFINED>'
-            d[var_name] = value
+                warnings.warn(
+                    "The environment variable '%s' is undefined." % var_name,
+                    UserWarning
+                )
+            else:
+                d[var_name] = value
 
     if sys.platform == "darwin":
         # multiprocessing.cpu_count() is not reliable on OSX
--- a/tests/test-recipes/metadata/build_env/meta.yaml
+++ b/tests/test-recipes/metadata/build_env/meta.yaml
@@ -1,0 +1,7 @@
+package:
+  name: conda-build-test-env
+  version: 1.0
+
+build:
+  script_env:
+    - UNDEF_VAR
--- a/tests/test-recipes/metadata/build_env/run_test.py
+++ b/tests/test-recipes/metadata/build_env/run_test.py
@@ -1,0 +1,9 @@
+import os
+
+def main():
+    undef_var = os.environ.get("UNDEF_VAR")
+
+    assert undef_var is None
+
+if __name__ == '__main__':
+    main()


msarahan:
+1

Ping @kalefranz

jakirkham:
Given I believe the change was added by @ilanschnell, he might have some thoughts about whether this works for the intended use case or not.

jakirkham:
To see an example of this behavior checkout this package ( https://github.com/nanshe-org/rank_filter ). If I build this with `conda-build<1.19`, then it works as intended. `USE_CYTHON` is set to "<UNDEFINED>" and then is unset in `build.sh` using a simple check. However, as I need `USE_CYTHON` to determine what dependencies I need install I use it in the `meta.yaml` file. This causes no problems for `conda-build<1.19`.

Once we switch to `conda-build>=1.19`, we see that there is two pass jinja templating in play. This results in `USE_CYTHON` getting set to "<UNDEFINED>" in the `meta.yaml` before reaching `build.sh`. This also ends up being impossible to debug in any clear way for a less knowledgeable user because they can't see that the variable was set in the `meta.yaml`. I am able to tell here as I can see the build string getting set to have a `_vg` and see that it now depends on VIGRA in some way.

FTR I think having to pass jinja templating is a neat idea and can be very useful for simplifying the jinja templates that are required and would much prefer it to stay around. However, we either to accept that changes similar to the one seen in this issue are required to keep the system easy to follow or we need to switch to a simpler model of parsing. IMHO the change I proposed here is the right one.

jakirkham:
@stuarteberg, this is what I was mentioning to you earlier today. Maybe this will give you a better idea of what is I am seeing here.

jakirkham:
I suppose another solution that would allow us to recover the old behavior, if needed, would be allow the environment variables to have specified default values in the `meta.yaml`. For instance, taking my example from before, something like the following could be done.

```
build:
  script_env:
    - CC
    - CXX
    - CXX_FLAGS
    - CXX_LDFLAGS
    - USE_CYTHON: true
```

This would allow one to explicitly set `USE_CYTHON` to `true`. Not specifying a value results in it being undefined unless it is defined externally. This makes it very clear how environment variables and avoids some internal specification by `conda build`.

Additionally, in this framework, I think it should be explicitly disallowed to set variables using other variables. If that is the intended behavior, it should be done external to the `meta.yaml`.

jakirkham:
Also added a test.

stuarteberg:
To start with a minor nit-pick, the I think following two sentences are not *quite* accurate:

>The previous practice was to leave these undefined. However, a commit was added that began setting them to `<UNDEFINED>`.

In `build.sh`, `<UNDEFINED>` has been the default value for all `script_env` variables ever since the `script_env` feature was first [introduced](https://github.com/conda/conda-build/commit/1cc940bc14485c2e8d2f2f703f717ef96ddedb20).

However, there was an inconsistency between `meta.yaml` and `build.sh`. In `meta.yaml`, missing `script_env` variables were left undefined, but in `build.sh`, they were defined (with contents `<UNDEFINED>`).  That inconsistency was remedied in #667.  After that, such variables always have the same contents in both `meta.yaml` and `build.sh`.

That said, I do agree that giving undefined variables a value at all seems strange. (If Bertrand Russel knew bash, he'd probably have something to say about statements like `export MY_VAR="This variable is undefined!"` :wink:)

Personally, I think I prefer this PR over your other one (#767).  This one is simpler, and seems to align more closely with the behavior I would expect if I hadn't read the docs.

jakirkham:
> In `build.sh`, `<UNDEFINED>` has been the default value for all `script_env` variables ever since the `script_env` feature was first introduced.

Previously *all* environment variables where just inherited from the user's environment so this changed it so only some explicitly stated environment variables were inherited. This sounds like a good idea. However, it did break the old behavior as now undefined ones that were inherited now were given a value instead of being left undefined.

> After that, such variables always have the same contents in both `meta.yaml` and `build.sh`.

Having them defined in `meta.yaml` and `build.sh`/`bld.bat` seems like the right way to go.

> That said, I do agree that giving undefined variables any value at all seems strange....

Exactly.

> Personally, I think I prefer this PR over your other one (#767).

Fair enough. That one was only designed to provide some sort of continuation of the old behavior if needed, but it still breaks it I guess. If it seems less intuitive than this, I am happy to stick with the simple warning.

stuarteberg:
In #767, @jakirkham wrote:

>it was my understanding that @ilanschnell made this change for some sort of client ... Though I don't know if they cared about the `<UNDEFINED>` setting in particular

The decision to implement the `script_env` feature comes from this old thread:
https://groups.google.com/a/continuum.io/d/msg/conda/5rgQ6H28UDc/2S77yhRTaRgJ

In that thread, there is no discussion of the behavior for undefined variables, so I wouldn't worry about it too much.

>That one was only designed to provide some sort of continuation of the old behavior if needed, but it still breaks it I guess.

Right.  Either way, recipes that depend on the `<UNDEFINED>` convention will need to be updated  after this PR is merged. (I have a few such recipes myself.)  But I don't think that's a big deal.  There seem to be very few public recipes that depend on `script_env`, much less the behavior of undefined `script_env` variables.  I took a look at a few large recipe repos (conda-recipes, conda-forge), and could not find a single use of `script_env`.  Maybe you and I are the only people using this feature right now. :-)

OK, that's probably enough bike-shedding for this one.  My vote is for this (#763) to be merged, and call it a day.  (It might be worth a quick note in `CHANGELOG.txt`, though.)

jakirkham:
Added a note in the changelog. @msarahan, thoughts?

ilanschnell:
I like it.  Not sure why I added `<UNDEFINED>` originally, but I think it was a way to communicate that a variable requested is not defined.

jakirkham:
Ok, just wanted to make sure there wasn't something else I was missing here.

jakirkham:
I think the docs need to be updated. See they still refer to [`<UNDEFINED>`]( http://conda.pydata.org/docs/building/environment-vars.html#inherited-environment-variables ). Not sure where there source lives.

stuarteberg:
>Not sure where there source lives.

The doc to change is [`conda-docs/source/building/environment-vars.rst`](
https://github.com/conda/conda-docs/blob/master/docs/source/building/environment-vars.rst#inherited-environment-variables)

jakirkham:
See this PR ( https://github.com/conda/conda-docs/pull/271 ) for doc change.

PR #765, opened 2016-02-05
Fix GIT_ variables when the repo is local

stuarteberg:
Fixes an issue introduced via #758.  If `git_url` points to a local directory instead of a remote url, the `GIT_` variables remained undefined.


msarahan:
Thanks for getting to this so quickly.

PR #766, opened 2016-02-08
Fix Python 3 incompatibility in cran.py

johanneskoester:
This PR fixes cran.py to use compat.iteritems instead of the old iteritems method of Python 2 dictionaries, which is no longer present in Python 3.

### DIFF
--- a/conda_build/cran.py
+++ b/conda_build/cran.py
@@ -571,7 +571,7 @@ def main(args, parser):
         name = d['packagename']
 
         #Normalize the metadata values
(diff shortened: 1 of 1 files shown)


johanneskoester:
The third commit fixes the normalization procedure, which converted metadata items to ascii bytes, but failed to convert them back to string objects. I hope that this was the original intention of the author. @groutr is that correct?

groutr:
Yes, I believe these were oversights on my part.  Thanks for fixing them.

jakirkham:
Needs a rebase.

groutr:
Closing in favor of #817

PR #767, opened 2016-02-08
Provide optional default environment variables

jakirkham:
Related: https://github.com/conda/conda-build/pull/763

Includes the change from this PR ( https://github.com/conda/conda-build/pull/763 ). Instead of simply leaving undefined environment values as such, this provide a mechanism for giving them a default value in `meta.yaml`. Users requiring or desiring to set the environment variables to some default value when not picked up externally can do so. However, this is better than the original strategy as this does it in a manner that is explicit.

Here is an example of the new syntax. All variables without a default value will be left undefined (if not defined in the external environment) during the build. All variables with a default value will be defined (if not defined in the external environment) using the default value.

```
build:
  script_env:
    - CC
    - CXX
    - CXX_FLAGS
    - CXX_LDFLAGS
    - USE_CYTHON: true
```

A test for this behavior is included in this PR.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -3,7 +3,8 @@
   * Environment variables defined in the 'script_env' build section of
     the meta.yaml file were previously assigned the value '<UNDEFINED>'
     if not found in the environment. Now they are left unset and a
-    warning is raised instead, #763.
+    warning is raised instead. However, it is possible to override
+    their default values by using a single element dictionary, #767.
 
 
 2016-01-29   1.19.0:
--- a/conda_build/environ.py
+++ b/conda_build/environ.py
@@ -1,5 +1,6 @@
 from __future__ import absolute_import, division, print_function
 
+import collections
 import os
 import sys
 from os.path import join, normpath, isabs
@@ -13,6 +14,12 @@ from conda_build.config import config
 
 from conda_build import source
 from conda_build.scripts import prepend_bin_path
+
+
+try:
+    basestring = basestring
+except NameError:
+    basestring = (str, bytes)
 
 
 def get_perl_ver():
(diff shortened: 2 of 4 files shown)


jakirkham:
@msarahan, this proposal might be a little better than the other as it provides some recourse if people want to define default values for environment variables.

jakirkham:
@stuarteberg, maybe you will like this proposal better than PR ( https://github.com/conda/conda-build/pull/763 ).

jakirkham:
@ilanschnell, is this similar enough to your original intent. People trying to smoke out illegal usage can do this explicitly and keeps the magic out of `conda-build`.

msarahan:
I like it.  If @stuarteberg thinks it is OK, I'd like to merge.

jakirkham:
Cool, thanks @msarahan.

FWIW, @stuarteberg, I think the [warnings]( https://travis-ci.org/conda/conda-build/jobs/107827728#L705-L708 ) are pretty clear anyways, but this will keep the option open to set some intended defaults in a clear manner or smoke out problematic cases.

stuarteberg:
IMHO, #763 is better.

- It's simpler
- I think it feels more intuitive for those that may not have memorized the docs (if a variable isn't set ... then it isn't set. Period.)
- The syntax proposed here makes it look like you might be setting the final values of these variables (possibly overriding the user's shell environment), not merely providing default values in the case of an undefined variable.

This gives the user extra flexibility, but it's not clear why that flexibility would be useful.  To me, it feels like this violates the "There should be one ... obvious way to do it" mantra.

jakirkham:
Sure, I only added this based on our conversation last week because it sounded like you were against ( https://github.com/conda/conda-build/pull/763 ) as the warnings might not be obvious. So, this provided a way of setting these variables to get back the old behavior and maybe make things more obvious.

Also, it was my understanding that @ilanschnell made this change for some sort of client so I wanted to make sure they still could get something out of it that they may still want.

Honestly, I am happy with either one. Just as long as environment variables aren't getting set in some way different from the recipe author's explicit intent.

ilanschnell:
Now that #763 is merged, I think we should close this one.  I agree with @stuarteberg

jakirkham:
Yes.

jakirkham:
A use case for something like this can be seen here ( https://github.com/conda-forge/staged-recipes/pull/188#discussion_r57181155 ).

PR #775, opened 2016-02-10
Add -m32 to CFLAGS and CXXFLAGS for multilib gcc

msarahan:
Most distributions provide a multilib gcc option.  These provide the capability to compile 32-bit code with a 64-bit GCC.  Our experimental compiler docker image ( https://github.com/ContinuumIO/docker-images/pull/20 & https://hub.docker.com/r/msarahan/conda_builder_linux/ ) uses this.  Moreover, there is no 32-bit only CentOS docker image, so this is our only real option with Docker/CentOS.

Ping @groutr @ilanschnell @kalefranz 

@stuarteberg @jakirkham @ukoethe - if any of you are familiar with this stuff, I'd appreciate your input, too.

### DIFF
--- a/conda_build/environ.py
+++ b/conda_build/environ.py
@@ -191,6 +191,9 @@ def get_dict(m=None, prefix=None):
 
     elif sys.platform.startswith('linux'):      # -------- Linux
         d['LD_RUN_PATH'] = prefix + '/lib'
+        if cc.bits == 32:
+            d['CFLAGS'] = '-m32'
+            d['CXXFLAGS'] = d['CFLAGS']
 
     if m:
         d['PKG_NAME'] = m.name()
--- a/conda_build/environ.py
+++ b/conda_build/environ.py
@@ -192,8 +192,8 @@ def get_dict(m=None, prefix=None):
     elif sys.platform.startswith('linux'):      # -------- Linux
         d['LD_RUN_PATH'] = prefix + '/lib'
         if cc.bits == 32:
-            d['CFLAGS'] = '-m32'
-            d['CXXFLAGS'] = d['CFLAGS']
+            d['CFLAGS'] = ' '.join([os.getenv('CFLAGS', ''), d.get('CFLAGS', ''), '-m32'])
+            d['CXXFLAGS'] = ' '.join([os.getenv('CXXFLAGS', ''), d.get('CXXFLAGS', ''), '-m32'])
 
     if m:
         d['PKG_NAME'] = m.name()
--- a/conda_build/environ.py
+++ b/conda_build/environ.py
@@ -182,18 +182,22 @@ def get_dict(m=None, prefix=None):
         d['PKG_CONFIG_PATH'] = join(prefix, 'lib', 'pkgconfig')
         d['R'] = join(prefix, 'bin', 'R')
 
+    cflags = d.get('CFLAGS', '') # in case CFLAGS was added in the `script_env` section above
+    cxxflags = d.get('CXXFLAGS', '')
+    ldflags = d.get('LDFLAGS', '')
+
     if sys.platform == 'darwin':         # -------- OSX
         d['OSX_ARCH'] = 'i386' if cc.bits == 32 else 'x86_64'
-        d['CFLAGS'] = '-arch %(OSX_ARCH)s' % d
-        d['CXXFLAGS'] = d['CFLAGS']
-        d['LDFLAGS'] = d['CFLAGS']
+        d['CFLAGS'] = cflags + ' -arch %(OSX_ARCH)s' % d
+        d['CXXFLAGS'] = cxxflags + ' -arch %(OSX_ARCH)s' % d
+        d['LDFLAGS'] = ldflags + ' -arch %(OSX_ARCH)s' % d
         d['MACOSX_DEPLOYMENT_TARGET'] = '10.6'
 
     elif sys.platform.startswith('linux'):      # -------- Linux
         d['LD_RUN_PATH'] = prefix + '/lib'
         if cc.bits == 32:
-            d['CFLAGS'] = ' '.join([os.getenv('CFLAGS', ''), d.get('CFLAGS', ''), '-m32'])
-            d['CXXFLAGS'] = ' '.join([os.getenv('CXXFLAGS', ''), d.get('CXXFLAGS', ''), '-m32'])
+            d['CFLAGS'] = cflags + ' -m32'
+            d['CXXFLAGS'] = cxxflags + ' -m32'
 
     if m:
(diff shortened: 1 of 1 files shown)


ilanschnell:
This might break existing recipes.  Wouldn't it be better to address this problem by adding those environment variables in `build.sh` of the recipe itself?
```
if [ $ARCH == 32 ]; then
    export CFLAGS='-m32'
    export CXXFLAGS='-m32'
fi
```

msarahan:
I will test this with a 32-bit CentOS vm.  I'm pretty sure it's harmless.  The idea of modifying every recipe we have with that block makes me cringe more than a little bit.

jakirkham:
I think this works with Mac, as well. Though do we want to add the flag there?

msarahan:
On my 32-bit CentOS VM:

```
[msarahan@localhost ~]$ gcc -m32 hello.c 
[msarahan@localhost ~]$ ./a.out 
Hello World
```
Checking what happens with duplicate flags (in case recipes already have the flag):
```
[msarahan@localhost ~]$ gcc -m32 -m32 hello.c 
[msarahan@localhost ~]$ ./a.out 
Hello World
```

If people do not have multilib gcc, and they try to compile 32-bit code, it will of course fail.  That won't be this flag's fault, though.

msarahan:
@jakirkham , I am inclined not to do that.  Thankfully, we really only support 64-bit on that platform.  Apple's zeal in platform standardization is good in that regard.

jakirkham:
That's pretty much what I thought.

jakirkham:
@ilanschnell, don't think it matters if the flag gets repeated. Most compilers accept duplication of many flags as long as they don't conflict adding yet another `-m32` shouldn't matter.

stuarteberg:
LGTM

jakirkham:
Is this possibly overwriting any values of `CFLAGS` or `CXXFLAGS`?

msarahan:
Yes, possibly.  I'll fix that.  I also worry a bit about other stuff in build.sh scripts clobbering this, but I think that's a job for documentation.

jakirkham:
When it comes to recipes, I think we should just recommend a practice of doing things like this.

```
CFLAGS="${CFLAGS} ..."
CXXFLAGS="${CXXFLAGS} ..."
```

There still might be some weird cases, but that should be case by case and relatively uncommon.

stuarteberg:
>Is this possibly overwriting any values of CFLAGS or CXXFLAGS?

Sigh.  There's a corner case, yes.  Normally, the user's environment variables are not inherited by `build.sh` at all.  But if the user's recipe includes this:

```
build:
  script_env:
    - CFLAGS
```

...then he would expect his own `CFLAGS` to be available in `build.sh`.  On Mac, this was already broken.  With this PR, it will be broken on 32-bit linux, too.

But honestly, I don't know if this is a corner case worth worrying about.  If a recipe relies on the user's environment having a particular value for `CFLAGS`, then that recipe isn't very good in the first place.

But if you want to fix it, I think this would do the trick:

```python
if cc.bits == 32:
    cflags = d.get('CFLAGS', '') # in case CFLAGS was added in the `script_env` section above
    cxxflags = d.get('CXXFLAGS', '')
    d['CFLAGS'] = cflags + ' -m32'
    d['CXXFLAGS'] = cxxflags + ' -m32'
```

...and do something similar for the osx case above that.

And if you really want to go the extra mile, I suppose you could check a list of important environment variables (e.g. `LD_LIBRARY_PATH`, `CFLAGS`, etc.) and warn the user that they will be ignored if they have them defined, but not mentioned in `script_env`.

jakirkham:
I think @stuarteberg's proposal sounds good. We have to keep in mind that we may also want to manipulate these again somewhere else and then we wouldn't want to lose our earlier changes.

Sadly, I think I have some ancient recipes like this. I might be able to remove these environment variables. Something I have been thinking about doing anyways, but it will require some investigation.

msarahan:
Thanks @stuarteberg and @jakirkham .  I think my last commit here addresses the issue of recipe CFLAGS and environment CFLAGS.  I would really hate to break things that work.  I'll be testing this PR with our docker build image, mentioned above.  I'll also search through conda-recipes and try to correct any recipes that might be clobbering this setting.

conda_build/environ.py
```diff
@@ -191,6 +191,9 @@ def get_dict(m=None, prefix=None):
 
     elif sys.platform.startswith('linux'):      # -------- Linux
         d['LD_RUN_PATH'] = prefix + '/lib'
+        if cc.bits == 32:
+            d['CFLAGS'] = ' '.join([os.getenv('CFLAGS', ''), d.get('CFLAGS', ''), '-m32'])
+            d['CXXFLAGS'] = ' '.join([os.getenv('CXXFLAGS', ''), d.get('CXXFLAGS', ''), '-m32'])
```
stuarteberg: I think this isn't right.  You're copying`CFLAGS` from the user's environment in all cases, whereas I think it should only be copied iff they listed `CFLAGS` in their recipe's `script_env` section.  If you just get the `CFLAGS` via `d.get()` and not `os.getenv()`, then you'll get the right behavior.

stuarteberg: Also, if you're going to fix the linux-32 case, why not go ahead and fix it in the darwin case, just above.

msarahan: I see, thanks.  Sorry I misunderstood your earlier suggestion.

jakirkham: It was decided earlier the flag should not be applied to osx 32 as it isn't a support platform any more.

jakirkham:
Is there a similar flag for MSVC?

msarahan:
The 32-bit one?  I don't think so.  I think they say "activate a different toolchain"

msarahan:
@stuarteberg I think this fixes the mac case in a similar way to how you fixed the linux case.  Please confirm.

stuarteberg:
Looks good to me!

jakirkham:
I think you can use the same flag in both cases, no?

msarahan:
@jakirkham , regarding CFLAGS/CXXFLAGS?  I would rather keep these distinct.  I think most of the time, they are the same.  I just don't want to prevent anyone from setting them differently.

jakirkham:
This is on Mac 10.9. The file is basically inconsequential (does nothing really).

```
$ clang -m32 test.c
$ ./a.out
$ file a.out 
a.out: Mach-O executable i386
```

stuarteberg:
I say leave it alone.  If it ain't broke, why fix it?  This PR is good to go.

msarahan:
Thanks all for the reviews and improvements.  Merging.

PR #785, opened 2016-02-17
Py2/3 compatible iteration through dict

zachcp:
Conda skeleton in python3  tripping on this line.  Not sure is this is the best way to handle this but it will work.


msarahan:
Good catch, thanks!

PR #788, opened 2016-02-18
Fix CYGWIN_PREFIX for other drive letters other than C:

tadeu:
If the drive is on D: (for example), the CYGWIN_PREFIX environment variable will be wrong, which may cause it to miss paths to be replaced in text files.


msarahan:
Good generalization. Thanks.

PR #799, opened 2016-02-28
Building noarch packages on Windows

### LINKED ISSUE #787: Setting `noarch_python: False` seems to have no effect
Setting `noarch_python` to `False` in the recipe still builds a noarch package.

### LINKED ISSUE #796: Building noarch packages on Windows
Noarch packages are awesome, thanks for that!

However, when I try to build a noarch package on Windows, the build process is aborted near the end with a message: `[noarch_python] Error: Python noarch packages can currently not be created on Windows systems.` The message is produced [here](https://github.com/conda/conda-build/blob/master/conda_build/noarch_python.py#L89).

Is there intent to make building noarch package available on Windows? If so, how hard would it be to implement it? I would not mind doing some work towards this.

### LINKED ISSUE #797: When building noarch package, the conda-bld/linux-64 dir is not initialized
I get this when I build a noarch package in a clean root conda env (e.g. when building Bokeh's conda package on Travis):

```
almar@almar-ubuntu64:~/py/bokeh$ conda build conda.recipe
Removing old work directory
BUILD START: bokeh-nightly-py_1
Using Anaconda Cloud api site https://api.anaconda.org
Fetching package metadata: ...........Error: Could not find URL: file:///home/almar/anaconda34/conda-bld/linux-64/
```

Manually creating the `linux-64` dir is not enough, the `repodata.json` and `repodata.json.bz2` need to be there too.

almarklein:
* Fixes that the `build/script` entry was ignored on Windows
* Fixes that setting `noarch_python` to `False` still builds it as a noarch package (#787)
* Enables making noarch packages on Windows (fixes #796)

### DIFF
--- a/conda_build/metadata.py
+++ b/conda_build/metadata.py
@@ -179,7 +179,7 @@ def parse(data):
     trues = {'y', 'on', 'true', 'yes'}
     falses = {'n', 'no', 'false', 'off'}
     for field in ('build/osx_is_app', 'build/preserve_egg_dir',
-                  'build/binary_relocation',
+                  'build/binary_relocation', 'build/noarch_python',
                   'build/detect_binary_files_with_prefix',
                   'build/skip', 'app/own_environment'):
         section, key = field.split('/')
--- a/conda_build/noarch_python.py
+++ b/conda_build/noarch_python.py
@@ -106,6 +106,7 @@ def transform(m, files):
     _force_dir(bin_dir)
 
     # Create *nix prelink script
+    # Note: it's important to set newline to LF or it wont work if we build on Win
     with open(join(bin_dir, '.%s-pre-link.sh' % name), 'w', newline='\n') as fo:
         fo.write('''\
 #!/bin/bash
--- a/conda_build/noarch_python.py
+++ b/conda_build/noarch_python.py
@@ -118,7 +118,7 @@ $PREFIX/bin/python $SOURCE_DIR/link.py
     _force_dir(scripts_dir)
 
     # Create windows prelink script (be nice and use Windows newlines)
-    with open(join(scripts_dir, '.%s-pre-link.bat' % name), 'w') as fo:
+    with open(join(scripts_dir, '.%s-pre-link.bat' % name), 'wb') as fo:
         fo.write('''\
 @echo off
 "%PREFIX%\\python.exe" "%SOURCE_DIR%\\link.py"
--- a/conda_build/noarch_python.py
+++ b/conda_build/noarch_python.py
@@ -107,22 +107,22 @@ def transform(m, files):
     _force_dir(bin_dir)
 
     # Create *nix prelink script
-    # Note: it's important to set newline to LF or it wont work if we build on Win
-    with open(join(bin_dir, '.%s-pre-link.sh' % name), 'w', newline='\n') as fo:
+    # Note: it's important to use LF newlines or it wont work if we build on Win
+    with open(join(bin_dir, '.%s-pre-link.sh' % name), 'wb') as fo:
         fo.write('''\
 #!/bin/bash
 $PREFIX/bin/python $SOURCE_DIR/link.py
-''')
+'''.encode('utf-8'))
 
     scripts_dir = join(prefix, 'Scripts')
     _force_dir(scripts_dir)
 
     # Create windows prelink script (be nice and use Windows newlines)
-    with open(join(scripts_dir, '.%s-pre-link.bat' % name), 'w', newline='\r\n') as fo:
+    with open(join(scripts_dir, '.%s-pre-link.bat' % name), 'w') as fo:
         fo.write('''\
 @echo off
 "%PREFIX%\\python.exe" "%SOURCE_DIR%\\link.py"
-''')
+'''.replace('\n', '\r\n').encode('utf-8'))
 
     d = {'dist': m.dist(),
          'site-packages': [],
--- a/conda_build/noarch_python.py
+++ b/conda_build/noarch_python.py
@@ -66,8 +66,9 @@ def handle_file(f, d):
 
     # The presence of .so indicated this is not a noarch package
     elif f.endswith(('.so', '.dll', '.pyd', '.exe', '.dylib')):
-        if f.endswith('.exe') and f.startswith('Scripts'):
-            os.unlink(path)  # we use the xx-script.py
+        if f.endswith('.exe') and (isfile(f[:-4] + '-script.py') or
+                                   basename(f[:-4]) in d['python-scripts']):
+            os.unlink(path)  # this is an entry point with a matching xx-script.py
             return
         _error_exit("Error: Binary library or executable found: %s" % f)
 
--- a/conda_build/noarch_python.py
+++ b/conda_build/noarch_python.py
@@ -4,7 +4,7 @@ import sys
 import json
 import shutil
 import locale
-from os.path import basename, dirname, isdir, join
+from os.path import basename, dirname, isdir, join, isfile
 
 from conda_build.config import config
 from conda_build.post import SHEBANG_PAT
@@ -50,7 +50,7 @@ def rewrite_script(fn):
     # Rewrite the file to the python-scripts directory
     dst_dir = join(config.build_prefix, 'python-scripts')
     _force_dir(dst_dir)
-    with open(join(dst_dir, fn), 'w', newline='\n') as fo:
+    with open(join(dst_dir, fn), 'w') as fo:
         fo.write(new_data)
     return fn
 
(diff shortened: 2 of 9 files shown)


almarklein:
With t his I was able to build both Bokeh and Flexx on windows, and install + run at Linux.

conda_build/noarch_python.py
```diff
@@ -85,15 +98,15 @@ def handle_file(f, d):
...
 
     prefix = config.build_prefix
     name = m.name()
 
+    bin_dir = join(prefix, 'bin')
+    _force_dir(bin_dir)
+
     # Create *nix prelink script
-    with open(join(prefix, 'bin/.%s-pre-link.sh' % name), 'w') as fo:
+    with open(join(bin_dir, '.%s-pre-link.sh' % name), 'w', newline='\n') as fo:
```
almarklein: This newline setting is important

conda_build/noarch_python.py
```diff
@@ -103,7 +117,7 @@ def transform(m, files):
     _force_dir(scripts_dir)
 
     # Create windows prelink script
-    with open(join(scripts_dir, '.%s-pre-link.bat' % name), 'w') as fo:
+    with open(join(scripts_dir, '.%s-pre-link.bat' % name), 'w', newline='\r\n') as fo:
```
almarklein: This one is not necessary per-see, but just cleaner.

almarklein:
Ready from my end.

conda_build/noarch_python.py
```diff
@@ -24,7 +26,7 @@ def rewrite_script(fn):
     directory after it passes some sanity checks for noarch pacakges"""
 
     # Load and check the source file for not being a binary
-    src = join(config.build_prefix, 'bin', fn)
+    src = join(config.build_prefix, 'Scripts' if ISWIN else 'bin', fn)
```
msarahan: May also need to check in Library\bin for exes or DLLs.

conda_build/metadata.py
```diff
@@ -179,7 +179,7 @@ def parse(data):
     trues = {'y', 'on', 'true', 'yes'}
     falses = {'n', 'no', 'false', 'off'}
     for field in ('build/osx_is_app', 'build/preserve_egg_dir',
-                  'build/binary_relocation',
+                  'build/binary_relocation', 'build/noarch_python',
```
jakirkham: Nice.

conda_build/noarch_python.py
```diff
@@ -55,7 +65,7 @@ def handle_file(f, d):
         os.unlink(path)
 
     # The presence of .so indicated this is not a noarch package
-    elif f.endswith('.so'):
+    elif f.endswith(('.so', '.dll', '.pyd', '.dylib')):
```
msarahan: add .exe here?

conda_build/noarch_python.py
jakirkham: Wouldn't it do this by default? If so, I would just assume drop it.

conda_build/noarch_python.py
```diff
@@ -131,7 +151,7 @@ def transform(m, files):
         link_code = fi.read()
 
     # Write the package metadata, and bumper with code for linking
-    with open(join(prefix, 'link.py'), 'w') as fo:
+    with open(join(prefix, 'link.py'), 'w', newline='\n') as fo:
```
jakirkham: Same story as above.

conda_build/noarch_python.py
```diff
@@ -70,13 +80,16 @@ def handle_file(f, d):
...
+            os.unlink(path)  # we use the xx-script.py
+            return
         fn = basename(path)
-        rewrite_script(fn)
+        fn = rewrite_script(fn)
         d['python-scripts'].append(fn)
 
     # Include examples in the metadata doc
-    elif f.startswith('Examples/'):
+    elif os.path.split(f)[0] == 'Examples':
```
msarahan: A slightly more pythonic line would be:

    os.path.dirname(f)

conda_build/windows.py
```diff
@@ -121,7 +121,6 @@ def build(m):
             os.makedirs(path)
 
     src_dir = source.get_dir()
-    bld_bat = join(m.path, 'bld.bat')
```
msarahan: Is this code right?  You require bld_bat as an argument, but then you overwrite it here?

almarklein: That line is deleted :)

conda_build/noarch_python.py
almarklein: no, because every entry point comes with an `.exe`, we ignore exes down below.

conda_build/windows.py
msarahan: Sorry, the monitors they have us using here are truly miserable contrast.  Thanks for pointing that out. =)

conda_build/noarch_python.py
almarklein: mmm, but in contrast to entry points, these exe's could not be translated to native linux/osx executables, so it could not be a noarch package. I think.

conda_build/noarch_python.py
almarklein: In case the f is `Examples/foo/bar.py`, dirname would give `Examples/foo`, so we need split in this case.

conda_build/noarch_python.py
almarklein: On Windows, it would be `\r\n` by default. But when building on Linux, it would be cleaner if it still produced a .bat with Windows newlines, IMO.

conda_build/noarch_python.py
almarklein: This case is not strictly necessary, but I reckoned it would be cleaner to be consistent. I'd be happy to get rid of all the newline stuff, though (except for the .bat and .sh).

conda_build/noarch_python.py
msarahan: My thought was to prevent things that are executables from being made into noarch packages.  So, packages like 7zip or Qt or curl or ... - just those kinds of things.  File extensions don't make sense for those on *nix, but they do on Windows.

I doubt that people will try to package windows exe's for mac/linux, but they are likely to try to make something noarch for the sake of using only one msvc runtime, possibly.  That won't work, and we should try to help them understand why, ideally with an error message here?

There are ways to make EXE-only packages not depend on a python version (hard-code the compiler, and give it the correct runtime dependency) - and this is sort of what people would want as "no-arch" - but it's different from noarch stuff here.

conda_build/noarch_python.py
msarahan: Cool, thanks for correcting me. =)

conda_build/noarch_python.py
almarklein: Ah, right. Good point. Will have a look at that.

almarklein: Or maybe we should just check for .exe as you suggest below. And raise an error there, unless its an exe `foo.exe` that has a matching `foo-script.py`.

msarahan: This sounds like a sane approach.

almarklein: Done

conda_build/noarch_python.py
```diff
@@ -55,8 +65,11 @@ def handle_file(f, d):
         os.unlink(path)
 
     # The presence of .so indicated this is not a noarch package
-    elif f.endswith('.so'):
-        _error_exit("Error: Shared object file found: %s" % f)
+    elif f.endswith(('.so', '.dll', '.pyd', '.exe', '.dylib')):
+        if f.endswith('.exe') and f.startswith('Scripts'):
```
msarahan: I think this is still not quite right.  People should install executables into Library/bin, but not everyone knows that.  If someone installs a non-entry-point exe in Scripts, we should complain.

how about something like:

    if not os.path.exists(os.path.splitext(f)[0] + "-script.py"):
        # barf because this is an exe in the wrong place

almarklein: Yeah ... this is about what I did at first, but the `xx-script.py` could already have been processed and moved. Maybe keep track of these then?

almarklein: Oh wait, we already keep track of them. Fix on the way.

conda_build/noarch_python.py
almarklein: better now?

msarahan:
Lgtm. Are there any tests that can be added to make sure that this doesn't regress?

almarklein:
Adding tests would be great. A stub package that has an entry point and examples should touch quite a bit of this code. 

Do we have any CI on this project, or elsewhere, where a regression would be detected if we write the correct test?

msarahan:
Yes, Travis CI is enabled on this repo. Please see the tests folder.

msarahan:
Sorry, just realized Travis won't run the intent here. Please come up with tests, even if only pseudocode fit us to implement. We need to get appveyor or equivalent up ASAP.

almarklein:
Added tests.

msarahan:
Tests look great.  Thanks for your extra effort.  Merging.

PR #800, opened 2016-03-01
read into memory by chunks in `have_prefix_files` to avoid

brentp:
huge memory-use on large files. 

See: https://github.com/Anaconda-Server/anaconda-client/pull/293
for motivation and similar changes to the client.

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -81,7 +81,7 @@ def create_post_scripts(m):
         os.chmod(dst, int('755', 8))
 
 
-def have_prefix_files(files):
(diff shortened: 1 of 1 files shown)


msarahan:
This looks like a very good change.  Can you implement a test to make sure it works the way you intend and we don't break it in the future?

brentp:
I realized it was incomplete. But the solution involves creating *another* copy of a large data file. Is there a way to tell conda not to perform this check at all? --I mean is there a way to tell conda that it's just a static data file and doesn't need to be modified?

PR #801, opened 2016-03-02
Fixing source/path and GIT_* issues

groutr:
The overarching goal of this PR is to fix issues with source/path (https://github.com/conda/conda-build/issues/783).

I uncovered a subtle issue with the Metadata.get_value() method not behaving as expected (see https://github.com/conda/conda-build/commit/3bd490d6a280be6444781e885e07faf3edadba7a)

PR #803, opened 2016-03-03
Fix git with local branches

stuarteberg:
This PR addresses an issue that can occur when building a conda recipe from a local git repository. For example:

```yaml
source:
  git_url: ../
```

If the user *switches branches* and attempts to build a recipe, conda-build's `git_cache` will still point to the old `HEAD`, meaning that the user is not building the branch she thinks she's building.

For example:

```bash
cd my-git-repo
git checkout master

# This clones my-git-repo to the cache, and sets HEAD to 'master'
conda build my-conda-recipe

# Switch branches
git checkout develop

# This updates my-git-repo via 'git fetch', but HEAD is still 'master'.
# Oops! We're just re-building the 'master' recipe, not 'develop'.
conda build my-conda-recipe 
```

It's surprising that this hasn't been noticed before, but maybe most people use `source/path` instead of `source/git_url` for local repositories, so this problem doesn't affect them.

Aside from that issue, I'm fixing a few other git-related problems.  See the commit history for details.

### DIFF
--- a/conda_build/environ.py
+++ b/conda_build/environ.py
@@ -38,7 +38,7 @@ def get_sp_dir():
     return join(get_stdlib_dir(), 'site-packages')
 
 def get_git_build_info(src_dir, git_url, expected_rev):
-    expected_rev = expected_rev or 'master'
+    expected_rev = expected_rev or 'HEAD'
     env = os.environ.copy()
     d = {}
     git_dir = join(src_dir, '.git')
--- a/conda_build/source.py
+++ b/conda_build/source.py
@@ -116,7 +116,7 @@ def git_source(meta, recipe_dir):
     else:
         args = [git, 'clone', '--mirror']
         if git_depth > 0:
-            args += ['--depth', git_depth]
+            args += ['--depth', str(git_depth)]
 
         check_call(args + [git_url, cache_repo_arg],  cwd=recipe_dir)
         assert isdir(cache_repo)
--- a/conda_build/source.py
+++ b/conda_build/source.py
@@ -112,7 +112,19 @@ def git_source(meta, recipe_dir):
 
     # update (or create) the cache repo
     if isdir(cache_repo):
-        check_call([git, 'fetch'], cwd=cache_repo)
+        if meta.get('git_rev', 'HEAD') != 'HEAD':
+            check_call([git, 'fetch'], cwd=cache_repo)
+        else:
+            # Unlike 'git clone', fetch doesn't automatically update the cache's HEAD,
+            # So here we explicitly store the remote HEAD in the cache's local refs/heads,
+            # and then explicitly set the cache's HEAD.
+            # This is important when the git repo is a local path like "git_url: ../",
+            # but the user is working with a branch other than 'master' without
+            # explicitly providing git_rev.
+            check_call([git, 'fetch', 'origin', '+HEAD:_conda_cache_origin_head'],
+                       cwd=cache_repo)
+            check_call([git, 'symbolic-ref', 'HEAD', 'refs/heads/_conda_cache_origin_head'],
+                       cwd=cache_repo)
     else:
         args = [git, 'clone', '--mirror']
         if git_depth > 0:


stuarteberg:
Update: As an added side-benefit, this implementation only fetches the specific `git_rev` that the recipe asks for (and all of its upstream commits, of course).  I think this will result in time-savings for large repos.

groutr:
Is this ready for review/merging?

stuarteberg:
Yeah, it's good to go.

stuarteberg:
OK, I had 4 commits here until just a few minutes ago.  The last one was unrelated to the rest of this PR: it tried to improve an error message under a very specific (but common) scenario.  But I realized that my "improvement" didn't actually work, so I removed it.  **Now** this PR is good to go. :-)

stuarteberg:
Sigh.  Not sure what's going on with the tests.  They were working before, and they still pass on my machine...

jakirkham:
That build failure does not seem to be unique to your PR. So, I think there is a build failure in master. It would be nice to know what exactly is going wrong and fix it first.

stuarteberg:
My guess is that somehow the new conda 4.0 release is incompatible with the conda-build test suite.  But I don't have time to investigate right now...

jakirkham:
Yeah, there are two problems.

1. The error message changed in conda 4.0. I have a fix for that here ( https://github.com/conda/conda-build/pull/809 ).

2. Python 3.3 support was dropped with conda 4.0, but conda-build still has support. So, we need to either use an older `conda` (seems like a bad idea) or drop Python 3.3 support. ( https://github.com/conda/conda-build/issues/810 )

stuarteberg:
Nice detective work.

jakirkham:
Added this proposal ( https://github.com/conda/conda-build/pull/812 ), which should fix all of the testing issues.

stuarteberg:
I rebased this branch from the master branch to get @jakirkham's testing fixes.  Now Travis is happy again.  This PR is back in "good to go" state.

groutr:
Regarding the source/path vs source/git_url (in your OP).  I don't believe git_url was originally designed to take a local path (which was the express purpose of source/path).

jakirkham:
Hmm...maybe I am misunderstanding what you are saying, but my understanding was that `git_url` can take any remote that `git` can which includes local paths. That is what the [recipe for conda build]( https://github.com/conda/conda-build/blob/master/conda_build.recipe/meta.yaml#L6 ) does after all.

jakirkham:
Also, should add that source/path now don't take URLs as this caused a lot of weird ambiguous cases.

groutr:
The ability `source/git_url` to accept local paths was introduced in this commit https://github.com/conda/conda-build/commit/90de286720dd04433622890b15b61453ab3ec87b.

The reason the conda recipe uses git_url is because that is the only way to get GIT_* variables (which #801 is attempting to rectify).  Maybe there should be a discussion to sort out when to use `source/git_url` or `source/path`, since their key use cases overlap.

stuarteberg:
Git supports both local and remote repos, so I think most users would just assume that both are supported in conda, too.  Discovering that they don't work would be likely to cause unnecessary surprise and/or puzzlement.

>Maybe there should be a discussion to sort out when to use source/git_url or source/path, since their key use cases overlap.

Perhaps, but I don't think that should delay acceptance of this 7-line change.  After all, the current implementation is broken, and this PR fixes it.  We can always decide to scrap support for local git repos in the future.  (I'm not sure I would be in favor of it, though.  If you open a separate thread on that subject, I can elaborate on my reasoning.)

groutr:
I'm very much in favor of local git support.  I'm just not to excited about having two very similar ways to get that (with some subtle differences).

stuarteberg:
Thanks for the merge.  I agree there are some subtle differences between the behavior of `path` and `git_url` with a local directory.  The definition of `GIT_` variables is one.  Another is the fact that `git_url` will clone the current `HEAD` and discard any un-committed changes in your working directory when the package is built.  If you can enable `GIT_` variables for the `path` case, then that's a win.

jakirkham:
I see what you are saying no @groutr. My preference would be that `git_url` be the only thing working with `git`. Though changes like this certainly should be discussed and probably go through a deprecation cycle or two to make sure people are aware of the impending change.

PR #808, opened 2016-03-07
Work-around System Integrity Protection on El Capitan.

mingwandroid:
The first commit is about making otool more useful / informative, then removing a hack and a bit of refactoring.

The 2nd commit builds on it and is important. Any package that depends on DYLD_FALLBACK_LIBRARY_PATH on OS X can't be built currently on 10.11 and this change addresses that. My approach is to add an -rpath to LDFLAGS during the build and then remove it from the executables at the end.

I went down a different path for fixing this initially (modifying the LC_ID_DYLIB load command names in the dylib files in the build environment) but backed it out when I realized it would've required copying instead of hard-linking when installing packages into the build environment. Also doing it as-per this PR involves less code (but it does rely on people not overriding LDFLAGS in build.sh)

We could mark packages that need this work-around explicitly, but I feel that it needs to work in every case anyway.

Finally, some packages that have dylibs with incorrect LC_ID_DYLIB load command names (they are missing '@rpath/') and they need to be rebuilt. I'm going to get a list together.

### DIFF
--- a/conda_build/environ.py
+++ b/conda_build/environ.py
@@ -191,7 +191,8 @@ def get_dict(m=None, prefix=None):
         d['OSX_ARCH'] = 'i386' if cc.bits == 32 else 'x86_64'
         d['CFLAGS'] = cflags + ' -arch %(OSX_ARCH)s' % d
         d['CXXFLAGS'] = cxxflags + ' -arch %(OSX_ARCH)s' % d
-        d['LDFLAGS'] = ldflags + ' -arch %(OSX_ARCH)s' % d
+        rpath = ' -rpath %(PREFIX)s/lib' % d # SIP workaround, DYLD_* no longer works.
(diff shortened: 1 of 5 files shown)


mingwandroid:
@groutr, @ilanschnell I'd appreciate a review when you guys have a chance, I'm not sure why travis is failing, the same test succeeded locally.

groutr:
I'm not sure exactly why the test is failing either.  The code looks good to me, though.

jakirkham:
Tests are failing because Python 3.3 is not supported by conda 4. So, must install conda 3. Also, the error message has changed for one test failure on conda 4. For the fix to these issues, see this PR ( https://github.com/conda/conda-build/pull/812 ).

groutr:
@mingwandroid I think this needs to be rebased off master again to pick up the changes from #812

mingwandroid:
@groutr thanks, done.

groutr:
Thanks

PR #809, opened 2016-03-08
tests: Update failure message for recursive build

jakirkham:
The error message for a missing package has [changed]( https://github.com/conda/conda/blob/6a6409a6810cc1689de10e76a5d516814e197c47/conda/resolve.py#L100 ) for conda 4.0. Here we grep the two last lines to see if they match. We do so in a manner that works on OS X and hopefully will work on Linux too.


jakirkham:
Works everywhere except Python 3.3. Where conda 4.0 is not supported. ( https://github.com/conda/conda-build/issues/810 )

jakirkham:
Changes included in PR ( https://github.com/conda/conda-build/pull/812 ), which has been merged.

Issue #810, opened 2016-03-08
CI: Python 3.3 has unsatisfiable dependencies


jakirkham:
Running into failures on Python 3.3. Should we be dropping support for it as it is no longer supported for `conda`?

jakirkham:
See ( https://github.com/conda/conda-build/pull/812 ) for the fix for CI.

groutr:
What is the output from conda?  I'm not getting unsatisfied dependencies?

jakirkham:
The issue occurs when trying to install the latest version of `conda` into the root environment without enforce dependency constraints as is done with Travis CI. Here is an example of the error as seen [on Travis]( https://travis-ci.org/conda/conda-build/jobs/114597276#L172-L174 ). This has been fixed by ( https://github.com/conda/conda-build/pull/812 ), which has already been merged so it is a non-issue now.

PR #811, opened 2016-03-08
travis: Fix testing on Python 3.3.

jakirkham:
Related: https://github.com/conda/conda-build/issues/810

This will bypass an issue testing on Python 3.3.

PR #812, opened 2016-03-08
Fix Travis CI for Python 3.3 and conda 4.x

jakirkham:
Fixes https://github.com/conda/conda-build/issues/810
Includes https://github.com/conda/conda-build/pull/811
Includes https://github.com/conda/conda-build/pull/809

As of conda 4.0, Python 3.3 is not supported. However, testing continues to happen here for Python 3.3. We have added a workaround to handle installing a version of conda 3.x here for Python 3.3. Also, an error message changes in conda 4.0 so we have included a different check for that warning message. However, if conda 3.x is installed for the Python 3.3 case the old error message check is used instead.

### DIFF
--- a/tests/test-recipes/build_recipes.sh
+++ b/tests/test-recipes/build_recipes.sh
@@ -31,9 +31,19 @@ echo "$OUTPUT" | grep "Error" | wc -l | grep 6
 ! OUTPUT=$(conda build --no-anaconda-upload conda-meta/ 2>&1)
 echo "$OUTPUT" | grep 'Error: Untracked file(s) ('\''conda-meta/nope'\'',)'
 
+# Get the version of conda as the error message changed in conda 4.
+set +e
+echo $(conda -V 2>&1) | grep "conda 3\..*"
+export IS_CONDA_3=$?
+set -e
+
 ! OUTPUT=$(conda build --no-anaconda-upload recursive-build/ 2>&1)
-echo "$OUTPUT" |  tail -n2 | head -n1 | grep 'Error:  Package missing in current .* channels: '
-echo "$OUTPUT" |  tail -n1 | grep '  - recursive-build2 2.0'
+if [[ $IS_CONDA_3 -eq 0 ]]; then
+    echo "$OUTPUT" | grep 'No packages found in current .* channels matching: recursive-build2 2\.0'
+else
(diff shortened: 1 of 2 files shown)


jakirkham:
@groutr, would you mind having a look at this and if there are no issues merge it. This will fix Travis CI for conda build to work on Python 3.3 (using conda 3.x) and conda 4.x on all other platforms.

jakirkham:
Passes all tests as expected.

stuarteberg:
:+1:

groutr:
Inspecting it now.

groutr:
So, there might be a subtle difference.  Python 3.3 is should still be supported as a target (ie conda 4.x should be able to still build packages for python 3.3).  However, it looks like conda 4.x itself has dropped support for Python 3.3 (or at least it doesn't test against 3.3).

jakirkham:
Not sure what you are proposing. The changes here do support Python 3.3 still. We just must install conda 3.x on Python 3.3 for the reasons you have mentioned. The test suite has been updated to support conda 3.x and conda 4.x.

tests/test-recipes/build_recipes.sh
```diff
@@ -31,8 +31,19 @@ echo "$OUTPUT" | grep "Error" | wc -l | grep 6
 ! OUTPUT=$(conda build --no-anaconda-upload conda-meta/ 2>&1)
 echo "$OUTPUT" | grep 'Error: Untracked file(s) ('\''conda-meta/nope'\'',)'
 
+# Get the version of conda as the error message changed in conda 4.
+set +e
+echo $(conda -V 2>&1) | grep "conda 3\..*"
+export IS_CONDA_3=$?
```
jakirkham: Switched to treat conda 3 as the special case instead of conda 4. The effect is the same.

groutr:
The code looks good to me.  Pending the CI tests, I think this is good to merge.  Thanks for the investigative work.

jakirkham:
Of course. Looks like it just passed. Please let me know if there is anything else.

jakirkham:
There was a [build failure]( https://travis-ci.org/conda/conda-build/jobs/114630730#L2566 ) on master, but it seems to be caused by too many svn requests. Should probably just rebuild it.

    svn: Server sent unexpected return value (429 Too Many Requests) in response to OPTIONS request for 'https://github.com/conda/conda-build'

PR #814, opened 2016-03-08
skip source_svn test because github doesn't like it

kalefranz:
@groutr Merge on approval


jakirkham:
I think this test still works just GitHub got overwhelmed by the number of requests.

kalefranz:
Happy to revert the commit later if we figure out why github is throttling this all the sudden...

PR #821, opened 2016-03-10
macho.py: Change the right dylib name

mingwandroid:
Fixes a bug I introduced in 38f333 where
install_name_tool -change was being called
on the last LC_LOAD_DYLIB instead of the
one that was meant to be changed.


mingwandroid:
Merging as obvious.

PR #827, opened 2016-03-10
Revert "travis: Fix testing on Python 3.3."

jakirkham:
This reverts commit ( https://github.com/conda/conda-build/commit/f85dcdb1b511ffd059992133d1ad2adb35a9375a ).

As it appears Python 3.3 CI has been dropped, there is no longer a need for this check.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,11 +19,7 @@ install:
   - export PATH="$HOME/miniconda/bin:$PATH"
   - hash -r
   - conda config --set always_yes yes
-  - if [[ "$TRAVIS_PYTHON_VERSION" == "3.3" ]]; then
-      conda install --force --no-deps 'conda=3.*' requests;
-    else
-      conda install --force --no-deps conda requests;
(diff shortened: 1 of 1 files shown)


kalefranz:
@jakirkham Do you know of anyone who still needs conda-build itself to run under python 3.3?  I think it really *is* fine to drop the 3.3 runtime testing. Do you agree?

jakirkham:
Not personally no. There are very few projects that still support 3.3.

Given that `conda` has dropped 3.3 support, I can't really see a reason `conda-build` should keep it.

The only reason one would want to keep an older version of Python 3 with `conda` and `conda-build` installed in it is because people are building packages with `bdist_conda`. That requires the `root` environment have the same version of Python as the one you intend to build for. Now that constraint may not make sense, but fixing it has proved tricky.

kalefranz:
Thanks for the input.

jakirkham:
Sure.

PR #833, opened 2016-03-15
Add new option --keep-old-work

mingwandroid:
The point is to allow engineers to debug issues with callstacks spanning
multiple conda packages without having to hack conda-build or mess around
with multiple miniconda environments.

The contents of the old work directory are moved into a temporary dir
before the build commences and moved back after it has completed. It was
done this way because the source unpacking logic is fairly hostile to
this kind of change.

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -13,6 +13,7 @@ import sys
 import time
 import tarfile
 import fnmatch
+import tempfile
 from os.path import exists, isdir, isfile, islink, join
 import mmap
 
@@ -361,7 +362,7 @@ def bldpkg_path(m):
     '''
     return join(config.bldpkgs_dir, '%s.tar.bz2' % m.dist())
 
-def build(m, get_src=True, post=None, include_recipe=True):
+def build(m, get_src=True, post=None, include_recipe=True, keep_old_work=False):
     '''
     Build the package with the specified metadata.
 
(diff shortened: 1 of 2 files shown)


mingwandroid:
@msarahan here's my first pass at this. I couldn't get it to work in the system so had to do it as a pre- and post- step, moving then moving things back.

See what you think anyway, it was useful for me immediately.

msarahan:
I think this is a pretty good addition.  We should definitely do this or something like this.  @kalefranz - please weigh in.  

The tempfile is not a pretty way to do it, though.  Is this a valuable enough feature to spend more time on?  I mean should we make the unpacking logic less aggressive?

mingwandroid:
I agree, but if you dig into source.py, particularly get_dir() and unpack() there wasn't much I could do, without significant time reworking things.

get_dir() looks for the single child of the work dir, and if it doesn't find it, it returns the work dir itself and everything goes wrong after that .. so .. I can see why it's been done that way; getting sources is a heterogeneous thing here, even the unpacking part is scary in and of itself.

I'd elect to spend time in the background bringing libarchive up to scratch and replacing all the tar, unzip, uncompress, subprocess stuff with a uniform call to that library, then determining the top level folder name for the version control schemes is pretty simple after that.

mingwandroid:
I wonder if mtime could be used to find the most recently modified directory in work instead of moving then unmoving, but it's probably even more of a hack.

My initial attempt involved attempting to determine the name of the directory created by the unpack() or source control functions and then passing that around but it was ugly to say the least.

We can leave this open as a useful feature to aim for anyway.

msarahan:
Merging this.  @mingwandroid would you please document this feature at https://github.com/conda/conda-docs/?  Please ask @electronwill about which file exactly.

PR #838, opened 2016-03-21
fix Examples directory in noarch_python

kalefranz:


almarklein:
Thanks Kale! Looks like I was the one who introduced the bug. I owe you a beer :)

PR #843, opened 2016-03-23
Execute source fetchers (e.g., git, hg, svn) in the _build environment

mjuric:
This patch makes `conda-build` execute any source fetching commands (i.e., `git`, `hg`, `svn`, etc.) with the `_build` environment on path. This makes it possible to provide the source fetchers as dependencies in a package's `meta.yaml` file, rather than require the developer to install them by hand, externally.

I.e., something like:
```
source:
  svn_url: https://github.com/ilanschnell/bsdiff
  svn_rev: 1.1.4
  svn_ignore_externals: True # (defaults to False)

requirements:
  build:
    - svn
```
should be able to fetch the source even if `svn` has not been installed on the developer's system.

n.b.: The motivation for this patch came from the need to build some packages that are backed by `git-lfs` (github's large file storage service). This requires `git-lfs`, which is somewhat a pain to install and configure manually (even for seasoned developers). We realized that by providing `git-lfs` as a package (plus another config package), the build system could transparently take care of this by pulling it in as a build dependency. Alas, to make this work the executables pulled in as build dependencies need to be visible to `git` when it's fetching the source; this is how this patch came about originally.


mjuric:
Hi everyone,
I wonder if there's anything I could do to help with getting this patch reviewed and merged?

msarahan:
This is great!  Sorry it went unnoticed.  Is there a test you could write?  Something pretty much exactly like the example you have in your description here would be fine.  I'd recommend a new test recipe in ```tests/test-recipes/metadata```

mjuric:
@msarahan, I added a recipe to test the feature (see https://github.com/conda/conda-build/pull/843/commits/0fb9fa0541ad516a65ad1baf453ffe041b4bb8b5). It uses `svn` to check out sources to `conda-build` (just as a test).

That said, the because TravisCI images come with `svn` preinstalled, this recipe won't fail there even if `svn` isn't picked up from the `_build` environment. I have an alternative (if slightly more complicated) test [here](https://github.com/mjuric/conda-build/commit/a73cc87fea450424fa95038db31ad675519265a1) that's immune to that. Should I push it?

msarahan:
@mjuric yes, please on your push.  That looks like what I would suggest: hide the system svn as necessary (add .bak to filename?) for the test, and restore it afterwards.

mjuric:
@msarahan, I pushed what I think is a better solution:
* I discovered that there was already a `source_svn` recipe in `tests/skip` subdirectory, re-enabled it, and added `svn` as a build dependency.
* To hide the system `svn`, I introduced `_setup.sh`/`_teardown.sh` hooks into `build_recipes.sh` that allow one to override the environment on a per-recipe bases. This is a generic facility, and can be used by any recipe. In the case of `source_svn`, these two create/remove a dummy `svn` script that shadows the system `svn` (by having its directory prepended to the `PATH`). See the diff for details, it should be fairly self-explanatory.

Hope this looks OK!

mjuric:
Fixed the exit code of the dummy `svn`; should be ready to merge.

msarahan:
@mjuric, two things (and I'm really sorry I didn't explain more earlier):

- The SVN test was disabled intentionally because it sporadically fails because github says we're making too many requests.  I'm not sure that reenabling it is a good idea.  Maybe a test that uses git instead would be more robust?
- I have reworked the test suites to no longer use shell scripts, but rather py.test: https://github.com/conda/conda-build/pull/864

I haven't merged the latter PR there, but since it would create a merge conflict (one of us would have to adapt your setup and teardown hook equivalents there), I wanted to check to see what you want to do.  Should I merge that other PR first, and you can adapt yours here, or should I merge yours first and I'll fix your recipe on my side?

mjuric:
@msarahan Good idea re `git` -- I'll see if I can cook that up. Go ahead and merge #864; I should be able to rewrite this in py.test framework (may take a day or two, the day-job is beginning to interfere... :) ).

msarahan:
Excellent.  I have merged # 864, and I'll look forward to merging this when you have time.  Let me know if you have questions about the new py.test replacements for the shell scripts.

mjuric:
@msarahan OK, here's the reworked PR.

It wasn't possible to use `git`, as `git` is very intertwined with `conda-build` in many places. It's very difficult to set up a test where we pretend git doesn't exist on the system.

I used `svn` again instead. To work around the issue of hammering github, I made a tiny svn repository and included it in the repository. There's some additional trickery that I had to add there (i.e., to create a fully-qualified `file://` URL, as svn doesn't do relative ones), but nothing too difficult to understand.

I added a mechanism (a tiny 3-line change to `test_build_recipes.py`) that checks for the existence of `_driver.sh` in the recipe directory, and instead of directly calling `conda build <RECIPE>`, it calls `_driver.sh conda build <RECIPE>`. It's like the previously proposed `_startup/_teardown` facility, but cleaner. I use this to expand the included `svn` repository, and generate the `meta.yaml` with the fully-qualified `file://` URL.

Take a look, hope it's OK to merge.

msarahan:
Thanks for your continuing effort.  Your approach looks fine to me.  What happens on Windows, though?  Is there a way to translate _driver.sh to python, so that it is platform independent?  Or does that even matter?

mjuric:
Good point. One could teach `test_build_recipes.py` to look for `_driver.bat` on Windows (and write the analogous .bat file), but I don't have the capability to test on Windows, unfortunately (and my `.bat` writing skills are some 20-years out of date...).

At the very least, I could teach `test_build_recipes` to ignore `_driver.sh` on Windows. Would that be acceptable? Also, is there an existing mechanism to mark a recipe to be skipped on windows?

msarahan:
I'll try to translate it tonight and send you a PR against this branch.  I don't think we should skip it.  I have one other more pressing PR (https://github.com/conda/conda-build/pull/861), but will work on yours next.  The way to skip on a per-platform basis is like this: https://github.com/conda/conda-build/blob/master/tests/test_build_recipes.py#L40-L42

mjuric:
Thanks!

mjuric:
Looks great, @msarahan! I just added a small commit to make the nomenclature consistent (changed 'fetcher' to 'checkout tool' everywhere).

msarahan:
Thanks for working with me on this. Nice addition.

msarahan:
@mjuric we have an Appveyor build failure: https://ci.appveyor.com/project/ironmancio54716/conda-build/build/1.0.2/job/3ck2r3awh3eh5405

namely:
    svn: E235000: In file '..\..\..\subversion\libsvn_client\checkout.c' line 87: assertion failed (svn_uri_is_canonical(url, scratch_pool))

Any ideas what that means?  I am not hugely familiar with SVN.  If you have ideas, I'd appreciate it.  Sorry we didn't catch this earlier, I didn't have Appveyor's webhook set up correctly, so it wasn't getting build notifications.

mingwandroid:
Hi guys,

I really need this PR for MSYS2 which I'm adding to Conda at present. I'm going to give it a whirl and see how it works for me as I'm actually right at the point of requiring it. I will be testing it with git.

PR #847, opened 2016-03-24
fixed race condition between removal and creation of tmp_dir on win

ukoethe:
I already proposed this PR last fall, but Aaron wanted to implement a better fix. Now, as Aaron has left and the problem still exists, I'm trying again. 

The problem is that Windows removes the test dir asynchronously and can't recreate it before removal is finished, resulting in occasional test failures. I just avoid the problem pragmatically by waiting for a second between the two operations.

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -10,6 +10,7 @@ import shutil
 import stat
 import subprocess
 import sys
+import time
 import tarfile
 import fnmatch
 from os.path import exists, isdir, isfile, islink, join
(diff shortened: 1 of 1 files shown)


msarahan:
Does it make sense to have a while loop with the sleep in it, checking if the folder still exists?

ukoethe:
This doesn't seem to be necessary, it just works. But then, I have an SSD, maybe it's different on hard drives. I also don't know if it is possible to check if the file still exists when deletion is already in progress. I'm sure this can be solved in a better way, but for now it seems to be good enough and does no harm.

msarahan:
OK, works for me.  Is there a test that can be written for this (break it in a known way, then apply this fix and make sure that it is fixed?  Or is this behavior not deterministic enough for a test?

ukoethe:
For me, it fails about once in every five builds, but it is totally unpredictable.

msarahan:
I really don't like arbitrary sleeps (who does?) but I don't see a better way here right now.  Merging.

PR #855, opened 2016-03-30
environ: Removed invalid assertion.

stuarteberg:
This assertion in `environ.py` isn't correct because the conda-build root directory might not be `conda-bld` (it might be customized in `.condarc`).  The assertion didn't provide much value anyway, so I'm just removing it.

Fixes the error in conda-forge/nn-feedstock#3.


msarahan:
Thanks @stuarteberg

PR #859, opened 2016-03-31
Merge folders if they already exist when packaging a program

183amir:
This patch prevents conda from complaining that a folder 
already exists when trying to package a program.
Fixes conda/conda-build#858

### DIFF
--- a/conda_build/utils.py
+++ b/conda_build/utils.py
@@ -10,6 +10,7 @@ import operator
 import fnmatch
 from os.path import dirname, getmtime, getsize, isdir, join
 from collections import defaultdict
+import distutils.dir_util
 
 from conda.utils import md5_file
 from conda.compat import PY3, iteritems
@@ -32,7 +33,7 @@ def copy_into(src, dst):
         dstname = os.path.join(dst, afile)
 
         if os.path.isdir(srcname):
-            shutil.copytree(srcname, dstname)
+            distutils.dir_util.copy_tree(srcname, dstname)
         else:
             shutil.copy2(srcname, dstname)
 
--- a/conda_build/utils.py
+++ b/conda_build/utils.py
@@ -8,7 +8,7 @@ import zipfile
 import subprocess
 import operator
 import fnmatch
-from os.path import dirname, getmtime, getsize, isdir, join, is_file
+from os.path import dirname, getmtime, getsize, isdir, join, isfile
 from collections import defaultdict
 import distutils.dir_util
 
@@ -35,7 +35,7 @@ def copy_into(src, dst):
         if os.path.isdir(srcname):
             # make sure we will not replace files while we merge folders
             files_to_be_copied = distutils.dir_util.copy_tree(srcname, dstname, dry_run=1)
-            existing_files = [path for path in files_to_be_copied if is_file(path)]
+            existing_files = [path for path in files_to_be_copied if isfile(path)]
             if existing_files:
                 raise Exception('The following files already exist:\n' +
                                 '\n'.join(existing_files))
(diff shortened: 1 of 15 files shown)


msarahan:
Thanks for submitting this.  Couple of comments:

Please always submit PRs with titles of what they change, not what issues they fix.  It is good to put something like "closes #858" in the description, though, and github will automatically close that issue when this PR is merged.

What does this change do if files already exist?  I see folder merging add a positive change, but if files overwrite other files, that leads to undefined behavior and would be very bad.

Can you write a simple test for these two behaviors as part of this PR?
  - does merging two folders succeed?
  - does it throw an error if a file being copied already exists?

183amir:
from documentation: *If the file exists, it will be ruthlessly clobbered.*

msarahan:
In that case, I'm afraid we need a different fix.  Clobbering files (especially ruthlessly) is very dangerous.

183amir:
Yes, I also think it is very bad. I will look into another solution maybe a custom function.

stuarteberg:
Maybe this new folder-merging behavior should only be enabled conditionally, based on a new field in `meta.yaml`.

For most recipes, it would be an error to touch directories that already exist in `site-packages`, and I would want to see an error to know that I did something wrong.  But if your recipe *expects* the folder to exist, `meta.yaml` should say so.

Maybe the risk of accidentally clobbering files will be acceptably small if you're limiting it to special cases like yours.

183amir:
Most package managers do not complain if the folder already exists but do complain when trying to overwrite a file.
In some package managers you have to specify which folders are yours and which ones are assumed to be owned by another package (.deb packages).

183amir:
How about now?

msarahan:
Nice.  Good way to do it.  I think you might want ```is_file``` rather than ```exists```, though?  I think ```exists``` would raise errors for directories.

Is there a (pair of) test recipes that you could add to tests/test-recipes/metadata?

183amir:
The return of `distutils.dir_util.copy_tree` is [only files](https://docs.python.org/2/distutils/apiref.html#distutils.dir_util.copy_tree) that are going to be copied. So there will be no directory name in that list. Also, even if you have empty directories in source, they will be created but they will not be in the output of this function.

183amir:
Sorry I am not familiar with your test setup. How do I add test for it?
I have recipes for `bob.extension` and `bob.blitz` which should make test this case very well.
As `bob.blitz` ships a `__init__.py` file which is also in `bob.extension`. But `bob.blitz` depends also on `libblitz` which is not packaged in `conda`.

msarahan:
Thanks for the docs on distutils.  Please use ```is_file``` instead anyway.  I think it makes your intent more explicit, and requires less thought for anyone else trying to figure this out later.

To add tests, you just create recipe(s) that require your new functionality.  I think your functionality requires at least 2 new recipes.  For the recipe that is only a dependency, I recommend starting the folder name with _, so that it is not considered its own test.  For examples, see the tests/test-recipes/metadata/recursive-build-two-layers folder, which depends on the _recursive-build-a folder.  Make sure to create whatever folder structure you want to test in those recipes.

If things still are not clear, please let us know.

183amir:
@msarahan Looks like I am on the right track with these tests but my tests raises an `Exception` which is supposed to happen but it is seen as a failed test.

msarahan:
It looks like you might need to rename one or more of these folders (the ones throwing this exception), and put an underscore first so that they don't get picked up.  Next, manually run the test in an entry in test_build_recipes, and wrap the subprocess call in a try/catch block.

test_checkout_tool_as_dependency is an example of what I'm suggesting here.  Rather than re-raising as that example does, I'd instead verify that the error contains your desired message.

msarahan:
@183amir any time to revisit this?

msarahan:
Superseded by #1090

PR #866, opened 2016-04-06
use memory map rather than read() to reduce memory usage.

brentp:
currently conda-build uses `data = fi.read()` on an entire file. It then uses `data = data.replace(a, b)`
even if `a` is never found in the file. The replace means that two copies of the entire data file are in memory at that point.

This uses a memory-map instead of fi.read() and only incurs the memory cost of the `.replace()` if the `a` is found in the file.

### DIFF
--- a/conda_build/post.py
+++ b/conda_build/post.py
@@ -31,8 +31,7 @@ if sys.platform.startswith('linux'):
 elif sys.platform == 'darwin':
     from conda_build import macho
 
-SHEBANG_PAT = re.compile(r'^#!.+$', re.M)
-
+SHEBANG_PAT = re.compile(br'^#!.+$', re.M)
 
 def is_obj(path):
     assert sys.platform != 'win32'
@@ -58,7 +57,7 @@ def fix_shebang(f, osx_is_app=False):
         mm = mmap.mmap(fi.fileno(), 0)
         m = SHEBANG_PAT.match(mm)
 
-        if not (m and 'python' in m.group()):
+        if not (m and b'python' in m.group()):
             return
 
         data = mm[:]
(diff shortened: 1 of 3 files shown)


conda_build/post.py
```diff
@@ -45,14 +46,22 @@ def fix_shebang(f, osx_is_app=False):
         return
     elif os.path.islink(path):
         return
-    with io.open(path, encoding=locale.getpreferredencoding()) as fi:
+
+    with io.open(path, encoding=locale.getpreferredencoding(), mode='r+') as fi:
         try:
-            data = fi.read()
-        except UnicodeDecodeError: # file is binary
+            data = fi.read(100)
```
msarahan: Can you explain the arbitrary 100 here?  Is it just to trigger the exception?

brentp: yes, I did a few checks and with the file open invocation, even fh.read(1) gives something like

```
UnicodeDecodeError: 'ascii' codec can't decode byte 0xf3 in position 1: ordinal not in range(128)
```

But, I'm not positive how fragile this is without reading the entire file.

msarahan:
Looks like this works fine for Python2, but barfs on Python3: https://travis-ci.org/conda/conda-build/jobs/121206085

If you fix that, I think this is ready.  Thanks for your work here.

brentp:
got it working on python3. I did a couple of force-pushes so I wouldn't have 10 commits fixing the issues. Looks like appveyor is failing for other reasons.

msarahan:
Thanks.  I'll follow up on the appveyor issues.  Merging yours, since it works fine on Travis.

brentp:
cheers!

jakirkham:
So I understand this PR was fixing memory issues, but am unclear on the severity of the issues. Was memory usage ever profiled before or after this change?

brentp:
I made this PR because I couldn't build a (data) package before on my laptop with 8GB of memory. After this, I could build much larger recipes. 
Granted, it's a rare use-case since most software packages (and files) are tiny. So the severity is exactly dependent on the size of the package.

I guess the solution would be to try mmap, catch the OSError and just read the file into memory. And then have a check to not call close() on the string.

msarahan:
We're in the midst of anacondacon right now. I'm open to what @brentp suggested. If you submit a PR, it'll get done much sooner.

PR #870, opened 2016-04-07
add files in order of increasing size to improve access times to tar

brentp:
downstream utilities like the binstar/anaconda client access the info/ stuff in a tar.bz2. If those occur after a large binary or data file then extra work is done to decompress. This mitigates that by sorting files before they so that smaller ones are added to the tar ball first.

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -524,7 +524,15 @@ can lead to packages that include their dependencies.""" %
 
         path = bldpkg_path(m)
         t = tarfile.open(path, 'w:bz2')
-        for f in sorted(files3 - files1):
+
+        def size(f):
+            # we don't care about empty files so send them back via 100000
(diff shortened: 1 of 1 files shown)


brentp:
this is the related change in binstar_client that takes advantage of the change here: https://github.com/Anaconda-Server/anaconda-client/pull/311

dsludwig:
Hahaha nice, I was just looking into this when I didn't see an improvement in performance with your branch. I was thinking put everything in `info` into the tarfile first.

You beat me to it!

brentp:
I see a huge difference on e.g. https://anaconda.org/bspuu/reference/files with both changes. I will update the PR sort info/ to be first (and then by size) tomorrow morn.

msarahan:
Nice. Will wait for your update, then merge.

brentp:
this should be good to go.

msarahan:
Good idea. Thanks!

PR #872, opened 2016-04-08
VS_YEAR, VS_VERSION, VS_MAJOR and CMAKE_GENERATOR Env Variables

patricksnape:
Following up from my work in #861, this does two things:

  1. Refactor the path logic so that the paths are defined once at the top of the `windows.py` file. This should reduce the risk of the tests going out of sync with the file.
  2. Add a new ENV variable `VS_VERSION` that supplies the visual studio version (9.0, 10.0, 12.0, 14.0 etc). This should remove the need for the 'black magic' from the Wiki as the user is supplied with whatever VS version command was run.

If everyone is happy with this then I will also add it to the documentation.

### DIFF
--- a/conda_build/environ.py
+++ b/conda_build/environ.py
@@ -287,6 +287,7 @@ def unix_vars(prefix):
     return {
         'HOME': os.getenv('HOME', 'UNKNOWN'),
         'PKG_CONFIG_PATH': join(prefix, 'lib', 'pkgconfig'),
+        'CMAKE_GENERATOR': 'Unix Makefiles',
         'R': join(prefix, 'bin', 'R'),
     }
 
--- a/conda_build/windows.py
+++ b/conda_build/windows.py
@@ -14,6 +14,39 @@ from conda_build.utils import _check_call
 
 assert sys.platform == 'win32'
 
+# Set up a load of paths that can be imported from the tests
+if 'ProgramFiles(x86)' in os.environ:
+    PROGRAM_FILES_PATH = os.environ['ProgramFiles(x86)']
+else:
+    PROGRAM_FILES_PATH = os.environ['ProgramFiles']
+
+# Note that we explicitly want "Program Files" and not "Program Files (x86)"
+WIN_SDK_BAT_PATH = os.path.join(os.path.abspath(os.sep),
+                                'Program Files', 'Microsoft SDKs',
+                                'Windows', 'v7.1', 'Bin', 'SetEnv.cmd')
+VS_TOOLS_PY_LOCAL_PATH = os.path.join(
+    os.getenv('localappdata', os.path.abspath(os.sep)),
+    'Programs', 'Common', 'Microsoft', 'Visual C++ for Python', '9.0',
+    'vcvarsall.bat')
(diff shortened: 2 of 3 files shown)


jakirkham:
@msarahan, this sounds incredibly handy. Do you have thoughts on it?

msarahan:
This is a very nice addition.  The black magic may still be necessary where the VS version differs, though.  For example, CMake generator expects the form:

Visual Studio 9 2008

or

Visual Studio 9 2008 Win64

with the black magic, I generally output the "9 2008" part.  Thus, I think we need several forms: 9, 9.0, 2008 - and possibly the architecture, if we want to keep that logic out of recipes.  We have architecture, but we have it as either 32 or 64, and we always have to change it into some other string (x64, win64, whatever)

so:

VS_MAJOR  (9)
VS_VERSION (9.0)
VS_YEAR (2008)

Does that seem reasonable?  You could omit VS_VERSION and just have VS_MAJOR (call it VS_VERSION if you want), because it's probably less voodoo to add on the ".0" than it is to strip it off in batch files.

%VS_VERSION%.0
vs
%VS_VERSION:.0=%

Either in this PR, or in a follow-on, if we also have the CMAKE_GENERATOR defined, then I think we'll cover the vast majority of bases.  We used to have that, and I ripped it out because NMakefiles made things simpler.  However, that doesn't work in all cases, and even where it does, CMake solutions built with MSBuild will use multiple CPUs and build much faster.  I think ripping out the CMake solution generator stuff from conda build was a mistake.

ccordoba12:
> VS_MAJOR (9)
VS_VERSION (9.0)
VS_YEAR (2008)

I agree with this proposal, it seems very handy :-)

patricksnape:
As discussed, I have added `VS_YEAR`, `VS_VERSION`, `VS_MAJOR` and `CMAKE_GENERATOR`. I also added `CMAKE_GENERATOR` to Unix for symmetry.

Where do I update the docs for this?

msarahan:
Thanks @patricksnape.  Docs are at https://github.com/conda/conda-docs/blob/master/docs/source/building/environment-vars.rst

Merging.  Ping @electronwill @kerrywatson1 - docs changes incoming.

anntzer:
`CMAKE_GENERATOR` should probably be quoted on Windows, so that a `bld.bat` that contains
```
cmake .. -C%CMAKE_GENERATOR%
```
does not fail on a quoting error.
Note that scripts that correctly wrote
```
cmake .. -C"%CMAKE_GENERATOR%"
```
would still work.

msarahan:
Either way, someone's going to mess up.  I think it's fine the way it is.  If we quote it when setting it here, then the double quotes break everyone who already has quotes in place.

jakirkham:
Agreed. Let's leave it alone.

anntzer:
No, the whole point is that even if quotes are already there, *both* are stripped off by `cmd.exe`.

msarahan:
Please write a test that proves it works.  Given that test, I'll make the change you suggest.

anntzer:
Hum, sorry, I made a typo in my tests, leading to the wrong conclusions.  Please ignore the noise, indeed the suggested patch would mess up those who do include quotes.

PR #878, opened 2016-04-11
make the prefix longer

brentp:
currently conda install will fail when the user installs into a deeply nested directory structure.

this happens because the build_prefix is limited to 80 characters. See: https://github.com/conda/conda/issues/1087#issuecomment-207559278

This extends the 80 character limit to 1000 as suggested by @msarahan


msarahan:
Thanks @brentp - duplicate at https://github.com/conda/conda-build/pull/877.  Closing this one to consolidate discussion.

Issue #889, opened 2016-04-22
BUG: Trouble fixing shebangs


jakirkham:
Running into some sort of [ASCII/unicode bug]( https://travis-ci.org/conda-forge/vigra-feedstock/jobs/124919857#L953-L977 ) in `conda-build` with on Python 3.5 with `conda-build` version [1.20.1]( https://travis-ci.org/conda-forge/vigra-feedstock/jobs/124919857#L218 ) on Mac.

```
updating shebang: bin/vigra-config
An unexpected error has occurred, please consider sending the
following traceback to the conda GitHub issue tracker at:

    https://github.com/conda/conda-build/issues

Include the output of the command 'conda info' in your report.
Traceback (most recent call last):
  File "/Users/travis/miniconda3/bin/conda-build", line 5, in <module>
    sys.exit(main())
  File "/Users/travis/miniconda3/lib/python3.5/site-packages/conda_build/main_build.py", line 208, in main
    args_func(args, p)
  File "/Users/travis/miniconda3/lib/python3.5/site-packages/conda_build/main_build.py", line 493, in args_func
    args.func(args, p)
  File "/Users/travis/miniconda3/lib/python3.5/site-packages/conda_build/main_build.py", line 438, in execute
    include_recipe=args.include_recipe)
  File "/Users/travis/miniconda3/lib/python3.5/site-packages/conda_build/build.py", line 515, in build
    post_build(m, sorted(files2 - files1))
  File "/Users/travis/miniconda3/lib/python3.5/site-packages/conda_build/post.py", line 332, in post_build
    fix_shebang(f, osx_is_app=osx_is_app)
  File "/Users/travis/miniconda3/lib/python3.5/site-packages/conda_build/post.py", line 76, in fix_shebang
    fo.write(new_data)
TypeError: write() argument must be str, not bytes
```

stuarteberg:
Just ran into this on Linux, too.

```
$ conda --version
conda 4.0.5
$ conda build --version
conda-build 1.20.1
```

stuarteberg:
Downgrading to conda-build 1.20.0 fixes the problem.  Presumably, d398611fde9395b8f921d4ada6ae3e245f82c71f is to blame.  Do you have any thoughts, @brentp?

brentp:
yes, that's me. looks like we can just add .decode('ascii')  to the write. trying to figure out how to trigger this error in the existing tests...

stuarteberg:
[...]

PR #892, opened 2016-04-22
fix for #889

brentp:
any advice on how to trigger this with the tests?


brentp:
I tried to get the tests to fail in #893, but no luck. This should fix #889, but any insight on making a test break would be great.

stuarteberg:
I can confirm that the problem I reported in #889 is not evident any more when I use this patch.

stuarteberg:
Thanks for the quick fix, Brent!

PR #897, opened 2016-04-26
Fix skip-existing with custom build string.

johanneskoester:
This PR fixes a bug in the skip-existing code: if a custom build string was used, it could happen that e.g. `PKG_BUILDNUM` was not initialized. Hence, skip-existing was assuming a wrong package name and errorneously assumed the package was either already present or vice versa. 

Thanks!
Johannes Köster

EDIT: The PR also includes a small improvement to an error message and a fix for a regression introduced in 1.20.1 (see the comments in the diff).

### DIFF
--- a/conda_build/metadata.py
+++ b/conda_build/metadata.py
@@ -139,7 +139,7 @@ def ensure_valid_fields(meta):
     if pin_depends not in ('', 'record', 'strict'):
         raise RuntimeError("build/pin_depends cannot be '%s'" % pin_depends)
 
-def parse(data):
+def parse(data, path=None):
     data = select_lines(data, ns_cfg())
     res = yamlize(data)
     # ensure the result is a dict
@@ -149,8 +149,8 @@ def parse(data):
         if field not in res:
             continue
         if not isinstance(res[field], dict):
-            raise RuntimeError("The %s field should be a dict, not %s" %
-                               (field, res[field].__class__.__name__))
+            raise RuntimeError("The %s field should be a dict, not %s in file %s." %
+                               (field, res[field].__class__.__name__, path))
 
 
 
@@ -325,7 +325,7 @@ class MetaData(object):
         # Start with bare-minimum contents so we can call environ.get_dict() with impunity
         # We'll immediately replace these contents in parse_again()
         self.meta = parse("package:\n"
-                          "  name: uninitialized")
+                          "  name: uninitialized", path=self.meta_path)
 
         # This is the 'first pass' parse of meta.yaml, so not all variables are defined yet
         # (e.g. GIT_FULL_HASH, etc. are undefined)
@@ -342,7 +342,7 @@ class MetaData(object):
         """
         if not self.meta_path:
             return
-        self.meta = parse(self._get_contents(permit_undefined_jinja))
+        self.meta = parse(self._get_contents(permit_undefined_jinja), path=self.meta_path)
 
         if (isfile(self.requirements_path) and
                    not self.meta['requirements']['run']):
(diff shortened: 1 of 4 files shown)


johanneskoester:
Aparently the tests fail because there are errors in the master branch (independent of the changes in this PR). Against which branch should I provide the PR?

mingwandroid:
`master` is the correct branch. Your change is still merge-able despite the test failures. It seems we are in a PR-race against the same bit of code too, mine is https://github.com/conda/conda-build/pull/898

conda_build/metadata.py
```diff
@@ -148,9 +148,12 @@ def parse(data):
     for field in FIELDS:
         if field not in res:
             continue
+        # ensure that empty fields are dicts (otherwise selectors can cause invalid fields)
+        if not res[field]:
+            res[field] = {}
```
johanneskoester: I have added above three lines from 1.20.0 to fix a regression introduced in 1.20.1. Without these, the following code in a meta.yaml would lead to an error:
```
build:
    skip: True # [osx]
```

Because build is turned into a string.

conda_build/metadata.py
```diff
@@ -148,9 +148,12 @@ def parse(data):
...
         if not isinstance(res[field], dict):
-            raise RuntimeError("The %s field should be a dict, not %s" %
-                               (field, res[field].__class__.__name__))
+            raise RuntimeError("The %s field should be a dict, not %s in file %s." %
+                               (field, res[field].__class__.__name__, path))
```
johanneskoester: Here, I have extended the error message to also show the path of the meta.yaml. This is important in cases where conda build is applied to more than one recipe.

johanneskoester:
Any updates or opinions on this? Thanks!

msarahan:
Since @mingwandroid has been working on something similar, I'll defer to him.  He should be back on Monday if not sooner.  I appreciate your effort in fixing this, though.

mingwandroid:
I will look at this on Monday. I'm away from computers this weekend. Conda and conda-build issues will be firmly on my radar on Monday.

mingwandroid:
Sorry for the delay, LGTM.

msarahan:
Suggested merge: https://github.com/bioconda/conda-build/pull/1

Please either merge that, or resolve the merge conflicts yourself, @johanneskoester - let me know if you have any questions.

johanneskoester:
Thanks, I have merged the PR.

msarahan:
We need to talk about @chapmanb's change to allow multiple meta.yaml files.  It appears to be non-intrusive, in that it maintains behavior for projects that don't follow your structure.  My question is more one of "can we settle on a standard way of having multiple versions of a recipe?"  This is something we're debating at conda-forge, and there the direction seems to be towards using git branches rather than folders (https://github.com/conda-forge/staged-recipes/pull/370#issuecomment-214490289)  My own opinion is that git branches are a better way to go, but harder to work with, especially for newcomers.

This is a minor point to split a whole community over, so I'm not going to get dogmatic about this.  However, I'd like for us all to do the same thing where it is reasonably achievable.

chapmanb:
Mike;
Thanks for the comments. My changes were made more in the spirit of "I need to get builds working again with new code" rather than a permanent fix so happy to go with whatever works best with you. Practically, I don't think branches are a great choice for bioconda since we keep all the recipes together -- you'd quickly be overloaded with branches for everything. Independent of that it would be nice to have some mechanism to resolve having more than one meta.yaml definition with a directory -- so whatever way you think best to resolve that is great with me. Thanks much.

msarahan:
OK.  For now, please add a test case for your recipe directory structure, so that we make sure to not break your workflow (again).

msarahan:
Merging this, but please submit a PR with a test later - or else I'm afraid later breakage of your workflow is likely.

PR #898, opened 2016-04-26
Fix --skip-existing since channel prefixes

mingwandroid:
@mcg1969, @kalefranz, please review.

Now, local packages have a 'local::' prefix so we must add
that when checking for a package in the index.


msarahan:
Closing in favor of #897

Issue #895, opened 2016-04-26
Python 3.4 build failures


jakirkham:
Running into a strange issue in a number of cases where conda-build 1.20.1 cannot build Python 3.4 64-bit packages. This problem does not occur in conda-build 1.20.0. See related issue ( https://github.com/conda-forge/staged-recipes/issues/448 ) and use cases documented within.

cc @msarahan @patricksnape

msarahan:
These seem to be caused by MSYS2 path conversion, plus extra CL environment variable settings set by the win 7 SDK's setenv.cmd:

    CL=/AI C:\Program Files\...

becomes
    
    CL=C:/root/of/msys2/AI C:\Program Files\...

@mingwandroid has been very helpful in debugging this.  We investigating a fix, which might include changing / to - in the CL variable, since MS tools recognize arguments like that also.

Alternatives:
  - set MSYS2_ARG_CONV_EXCL=“/AL;/OUT;%MSYS2_ARG_CONV_EXCL%" - MSYS2 should ignore paths starting with patterns here.  Semicolon delimited.
  - set MSYS2_ENV_CONV_EXCL=“CL" - MSYS2 should ignore path conversion in the CL environment variable

msarahan:
This will require more test coverage - ideally, building some small project that uses a configure script.  If anyone has suggestions, I'd like to hear them.

mingwandroid:
Path conversion is a fact of MSYS2 life, arguments that 'look like' POSIX paths get turned into their Windows equivalent when calling Windows programs from POSIX programs.

The same happens for env. vars and can be excluded with `MSYS2_ENV_CONV_EXCL`.

It's better than Windows programs getting POSIX paths and not being able to do anything with them.

mingwandroid:
I should clarify, in this case because configure interpolates the CL variable into the command line args, `MSYS2_ENV_CONV_EXCL` won't be useful for this one.

msarahan:
I think this is fixed between #900 and #933 - I'm going to close it for now, but please comment here or reopen it if any further issues like this crop up.  I don't think this will be truly, reliably fixed until we have conda-build activating environments rather than changing PATH directly.

PR #900, opened 2016-04-27
Add MSYS2 path translation exclusion for CL with MSVC builds

msarahan:
This might work - needs testing.  If it doesn't, I think we might need to remove the /Release flag (but this really should work: https://msdn.microsoft.com/en-us/library/ff660764(v=vs.100).aspx)

ping @patricksnape @jakirkham - is there a good way for us to test this with conda-forge?

### DIFF
--- a/conda_build/windows.py
+++ b/conda_build/windows.py
@@ -111,17 +111,15 @@ def msvc_env_cmd(bits, override=None):
     def build_vcvarsall_cmd(cmd, arch=arch_selector):
         return 'call "{cmd}" {arch}'.format(cmd=cmd, arch=arch)
 
-    msvc_env_lines.append('set VS_VERSION="{}"'.format(version))
-    msvc_env_lines.append('set VS_MAJOR="{}"'.format(version.split('.')[0]))
-    msvc_env_lines.append('set VS_YEAR="{}"'.format(VS_VERSION_STRING[version][-4:]))
-    msvc_env_lines.append('set CMAKE_GENERATOR="{}"'.format(VS_VERSION_STRING[version] +
+    msvc_env_lines.append('set "VS_VERSION={}"'.format(version))
+    msvc_env_lines.append('set "VS_MAJOR={}"'.format(version.split('.')[0]))
+    msvc_env_lines.append('set "VS_YEAR={}"'.format(VS_VERSION_STRING[version][-4:]))
+    msvc_env_lines.append('set "CMAKE_GENERATOR={}"'.format(VS_VERSION_STRING[version] +
                                                             {64: ' Win64', 32: ''}[bits]))
     if version == '10.0':
-        # Unfortunately, the Windows SDK takes a different command format for
-        # the arch selector - debug is default so explicitly set 'Release'
-        win_sdk_arch = '/x86 /Release' if bits == 32 else '/x64 /Release'
+        win_sdk_arch = '/Release /x86' if bits == 32 else '/Release /x64'
         win_sdk_cmd = build_vcvarsall_cmd(WIN_SDK_BAT_PATH, arch=win_sdk_arch)
-        
+
         # Always call the Windows SDK first - if VS 2010 exists but was
         # installed using the broken installer then it will try and call the 
         # vcvars script, which will fail but NOT EXIT 1. To work around this,
@@ -134,7 +132,6 @@ def msvc_env_cmd(bits, override=None):
 
         # First, check for Microsoft Visual C++ Compiler for Python 2.7
         msvc_env_lines.append(build_vcvarsall_cmd(VS_TOOLS_PY_LOCAL_PATH))
-        
         msvc_env_lines.append(error1.format(
             build_vcvarsall_cmd(VS_TOOLS_PY_COMMON_PATH)))
         # The Visual Studio 2008 Express edition does not properly contain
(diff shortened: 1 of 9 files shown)


jakirkham:
Is this meant to fix ( https://github.com/conda/conda-build/issues/895 )?

patricksnape:
Sorry I'm away this week with work and they are fairly intense, long days. I did have a bit of a look at this and I think it might be to do with delayed expansion within the SDK script. We may need to flip it on before calling it - as is done [in conda-forge right now](https://github.com/conda-forge/staged-recipes/blob/master/appveyor.yml#L6).

conda_build/windows.py
```diff
@@ -111,17 +110,15 @@ def msvc_env_cmd(bits, override=None):
...
+    msvc_env_lines.append('set "VS_VERSION={}"'.format(version))
+    msvc_env_lines.append('set "VS_MAJOR={}"'.format(version.split('.')[0]))
+    msvc_env_lines.append('set "VS_YEAR={}"'.format(VS_VERSION_STRING[version][-4:]))
+    msvc_env_lines.append('set "CMAKE_GENERATOR={}"'.format(VS_VERSION_STRING[version] +
                                                             {64: ' Win64', 32: ''}[bits]))
     if version == '10.0':
-        # Unfortunately, the Windows SDK takes a different command format for
-        # the arch selector - debug is default so explicitly set 'Release'
-        win_sdk_arch = '/x86 /Release' if bits == 32 else '/x64 /Release'
+        win_sdk_arch = '/Release /x86' if bits == 32 else '/Release /x64'
```
patricksnape: Line 97 above. That is documented as being required if the Windows SDK is used. Currently that only gets set if the user overrides the default version - when in fact it may need to be set for VS 2010 only (which is the only version that uses the Win SDK.

msarahan: I understand those flags above as telling distutils "stand back, I'm going to try science!"

In other words, it overrides distutils' search for compiler.  Maybe I'm wrong - haven't looked deeply at this.

Are there changes you'd suggest here, or just commenting?

msarahan:
@jakirkham - hopefully.  Testing now with ICU: https://ci.appveyor.com/project/conda-forge/icu-feedstock/build/1.0.17

@patricksnape good idea on delayed expansion.  I'll try that next if this doesn't work.

tests/test-skeleton/sympy-0.7.5/meta.yaml
```diff
@@ -4,7 +4,7 @@ package:
 
 source:
   fn: sympy-0.7.5.tar.gz
-  url: https://pypi.python.org/packages/source/s/sympy/sympy-0.7.5.tar.gz
+  url: https://pypi.python.org/packages/8c/a5/5fa8adee81837687f7315122769fc0b0e8b042c69e2fe5809c41191c7183/sympy-0.7.5.tar.gz
```
jakirkham: Unrelated, but it is here. Why do we now have to do this with the URLs?

I have seen this is needed on conda-forge too. Would be nice if we could handle a redirect or whatever so that we can avoid this ugly URL somehow.

cc @ocefpaf

msarahan: I don't know when this changed, but pypi changed their web api.  They don't seem to offer redirects that I know of.  My search terms have been "pypi url change" and stuff like that.  Haven't been able to find anything.

jakirkham: Yeah, I have tried to search for this change and found nothing either, but thought maybe you might know more than I. If you figure out anything, please let me know.

grlee77: @msarahan, @jakirkham 
Seems to be quite a bit of complaints/discussion related to the pypi URL changes here:
https://bitbucket.org/pypa/pypi/issues/438/backwards-compatible-un-hashed-package

TLDR: doesn't look like it will be fixed on the current pypi, but backwards compatible URLs are featured in "Warehouse" which is PyPI 2.0.  

I was able to verify the expected behaviour for the package pyface:
The links listed at https://pypi.io/project/pyface/#files have the hashes in them, but the following old-style URL also works: https://pypi.io/packages/source/p/pyface/pyface-5.1.0.tar.gz

jakirkham: Awesome, works for sympy too.

`https://pypi.io/packages/source/s/sympy/sympy-0.7.5.tar.gz`

msarahan: I'm going to split that commit off of this PR and make its own.  That really needs a fix.

jakirkham: Was debating asking about just that exactly.

msarahan: https://github.com/conda/conda-build/pull/905

msarahan:
Ultimately, the cause of this issue is that MSYS2 automatically converts things that look like

    /AI

into paths, like C:\root\of\msys2\ai

This confuses MSVC a lot.  The main additions here are some environment variables that prevent MSYS2's path conversion from acting on a few things.  Otherwise, there are some tests that are probably good to have, anyway.

msarahan:
Merging.  Passing appveyor build in https://ci.appveyor.com/project/msarahan/conda-build/build/1.0.54

PR #908, opened 2016-04-29
split off recipe render from build; recurse to find meta.yaml

### LINKED ISSUE #887: Option to generate static `meta.yaml` file(s)
We have some options to do this with tools in [conda-forge]( https://conda-forge.github.io/ ), but I am wondering if it wouldn't be valuable to have this in conda-build proper. The idea here is we complete the required jinja passes and fill in all selectors, but never actually build anything (though we may need to download stuff temporarily (`GIT_*` vars). All we do is generate a static `meta.yaml` (maybe called `meta_static.yaml`). This allows for the final product to be observed and used in other process where reasoning about this result is necessary to proceed.

I think this would be incredibly powerful in debugging recipes. Particular ones using these dynamic features like the jinja passes. This might also prove very useful in combination with highly dynamic recipes like those proposed in this PR ( https://github.com/conda/conda-build/pull/848 ). Though other strategies to handle this ( https://github.com/conda/conda-build/issues/857 ) could also benefit from a clean static output. Other strategies of pinning while maybe less dynamic ( https://github.com/conda/conda-build/pull/728 ) ( https://github.com/conda/conda-build/pull/747 ) could at least be turned into their components. Also, this could play a role in generating final product recipes of a variety of different potential combinations that are used in matrix builds (potentially with exclusion) for example.

While I am proposing this, I don't want it to distract from efforts of fixing API/ABI in packages. Though I do want us to keep this in mind as a potential part of a more general solution. Also, this may tease out one of the important issues with regards to how this pinning happens and is used in later steps.

Then again maybe this totally off base. Please let me know your thoughts.

cc @pelson @ukoethe @stuarteberg @msarahan @johanneskoester @kalefranz

msarahan:
Closes #887 

This is the start of splitting off recipe rendering from building.  I am trying to avoid duplicating code, so I'm splitting the existing code off into other modules.

Right now, it only prints the processed metadata, not a yaml file.  That's easy enough to write as an addon option later - I'm expecting that the raw dictionary will be the most useful thing to people, rather than going back to some file on disk.  My idea is that the build module will just call functions from this file to render each recipe's meta.yaml.

This works right now, with a trivial test of a recipe from conda-forge (zlib).  Original recipe at https://github.com/conda-forge/zlib-feedstock, which I have cloned as a submodule into the packages/zlib folder:

```
$ conda render packages/zlib
{u'about': {u'home': u'http://zlib.net/',
            u'license': u'zlib (http://zlib.net/zlib_license.html)',
            u'license_family': u'Other',
            u'license_file': u'license.txt',
            u'summary': u'Massively spiffy yet delicately unobtrusive compression library'},
 u'build': {u'features': u''},
 u'extra': {u'recipe-maintainers': [u'groutr', u'msarahan', u'ocefpaf']},
 u'package': {u'name': u'zlib', u'version': u'1.2.8'},
 u'requirements': {u'build': [u'cmake']},
 u'source': {u'fn': u'zlib-1.2.8.tar.gz',
             u'sha256': u'36658cb768a54c1d4dec43c3116c27ed893e88b02ecfcb44f2166f9c0b7f2a0d',
             u'url': u'http://zlib.net/zlib-1.2.8.tar.gz'},
 u'test': {u'commands': [u'test -f ${PREFIX}/lib/libz.a']}}
```

This change also makes conda-build recursively search for meta.yaml in any folder you provide, so it should make conda build more directly compatible with the conda-forge layout of having the recipe one level deeper.

I'll be cleaning this up and coming up with test cases - please kick the tires and point out any details I should pay special attention to.

CC @jakirkham @stuarteberg @kalefranz @pelson @groutr

### DIFF
--- a/conda_build/main_render.py
+++ b/conda_build/main_render.py
@@ -13,7 +13,8 @@ from locale import getpreferredencoding
 import os
 from os.path import exists, isdir, isfile, join, abspath
 
-# import conda.config as config
+import yaml
+
 from conda_build.config import config
 from conda.compat import PY3
 from conda.cli.common import add_parser_channels
@@ -203,15 +204,52 @@ def render_recipe(recipe_path, download_source=False):
     return m
 
 
+# Next bit of stuff is to support YAML output in the order we expect.
+# http://stackoverflow.com/a/17310199/1170370
+class MetaYaml(dict):
+    fields = ["package", "source", "build", "requirements", "test", "extra"]
+    def to_omap(self):
+        return [(field, self[field]) for field in MetaYaml.fields]
+
+
+def represent_omap(dumper, data):
+   return dumper.represent_mapping(u'tag:yaml.org,2002:map', data.to_omap())
+
+def unicode_representer(dumper, uni):
+    node = yaml.ScalarNode(tag=u'tag:yaml.org,2002:str', value=uni)
+    return node
+
+
+class IndentDumper(yaml.Dumper):
+    def increase_indent(self, flow=False, indentless=False):
+        return super(IndentDumper, self).increase_indent(flow, False)
+
+
+yaml.add_representer(MetaYaml, represent_omap)
+if PY3:
+    yaml.add_representer(str, unicode_representer)
+else:
+    yaml.add_representer(unicode, unicode_representer)
+
+
 def main():
     import pprint
     p = get_render_parser()
+    p.add_argument(
+        '-y', '--yaml',
+        action="store_true",
+        help="print YAML, as opposed to printing the metadata as a dictionary"
+    )
 
     args = p.parse_args()
     set_language_env_vars(args)
 
     metadata = render_recipe(find_recipe(args.recipe), download_source=args.source)
-    pprint.pprint(metadata.meta)
+    if args.yaml:
+        print(yaml.dump(MetaYaml(metadata.meta), Dumper=IndentDumper,
+                        default_flow_style=False, indent=4))
+    else:
+        pprint.pprint(metadata.meta)
 
 
 if __name__ == '__main__':
--- a/conda_build/main_render.py
+++ b/conda_build/main_render.py
@@ -52,13 +52,6 @@ platform specifics, making it simple to create working environments from
         help="Obtain the source and fill in related template variables.",
     )
     p.add_argument(
-        'recipe',
-        action="store",
-        metavar='RECIPE_PATH',
-        choices=RecipeCompleter(),
-        help="Path to recipe directory.",
-    )
-    p.add_argument(
         '--python',
         action="append",
         help="""Set the Python version used by conda build. Can be passed
@@ -239,6 +232,14 @@ def main():
         '-y', '--yaml',
         action="store_true",
         help="print YAML, as opposed to printing the metadata as a dictionary"
+    )
+    # we do this one separately because we only allow one entry to conda render
+    p.add_argument(
+        'recipe',
+        action="store",
+        metavar='RECIPE_PATH',
+        choices=RecipeCompleter(),
+        help="Path to recipe directory.",
     )
 
     args = p.parse_args()
--- a/conda_build/main_render.py
+++ b/conda_build/main_render.py
@@ -7,11 +7,9 @@
 from __future__ import absolute_import, division, print_function
 
 import sys
-from collections import deque
-from glob import glob
 from locale import getpreferredencoding
 import os
-from os.path import exists, isdir, isfile, join, abspath
+from os.path import isdir, isfile, abspath
 
 import yaml
 
@@ -104,7 +102,7 @@ platform specifics, making it simple to create working environments from
     return p
 
 
-def set_language_env_vars(args):
+def set_language_env_vars(args, parser, execute=None):
     """Given args passed into conda command, set language env vars"""
     for lang in all_versions:
         versions = getattr(args, lang)
@@ -118,7 +116,8 @@ def set_language_env_vars(args):
         if len(versions) > 1:
             for ver in versions[:]:
                 setattr(args, lang, [str(ver)])
-                execute(args, parser)
+                if execute:
+                    execute(args, parser)
                 # This is necessary to make all combinations build.
                 setattr(args, lang, versions)
             return
@@ -152,7 +151,6 @@ def render_recipe(recipe_path, download_source=False):
 
     with Locked(config.croot):
         arg = recipe_path
-        try_again = False
         # Don't use byte literals for paths in Python 2
         if not PY3:
             arg = arg.decode(getpreferredencoding() or 'utf-8')
@@ -183,19 +181,7 @@ def render_recipe(recipe_path, download_source=False):
             source.provide(m.path, m.get_section('source'), patch=False)
             print('Source tree in:', source.get_dir())
 
-        try:
-            m.parse_again(permit_undefined_jinja=False)
-        except SystemExit:
-            # Something went wrong; possibly due to undefined GIT_ jinja variables.
-            # Maybe we need to actually download the source in order to resolve the build_id.
-            source.provide(m.path, m.get_section('source'))
-
-            # Parse our metadata again because we did not initialize the source
-            # information before.
-            m.parse_again(permit_undefined_jinja=False)
-
-            print(build.bldpkg_path(m))
-            raise
+        m.parse_again(permit_undefined_jinja=False)
 
         if need_cleanup:
             shutil.rmtree(recipe_dir)
@@ -264,7 +250,7 @@ def main():
     )
 
     args = p.parse_args()
-    set_language_env_vars(args)
+    set_language_env_vars(args, p)
 
     metadata = render_recipe(find_recipe(args.recipe), download_source=args.source)
     if args.output:
--- a/conda_build/main_render.py
+++ b/conda_build/main_render.py
@@ -45,9 +45,10 @@ platform specifics, making it simple to create working environments from
         version = 'conda-build %s' % __version__,
     )
     p.add_argument(
-        '-s', "--source",
+        '-n', "--no_source",
         action="store_true",
-        help="Obtain the source and fill in related template variables.",
+        help="When templating can't be completed, do not obtain the \
+source to try fill in related template variables.",
     )
     p.add_argument(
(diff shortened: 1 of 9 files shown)


msarahan:
This now also renders YAML, keeping the keys in the order we expect:

```
$ conda render ~/Documents/code/anaconda/packages/zlib --yaml
package:
    name: zlib
    version: 1.2.8
source:
    fn: zlib-1.2.8.tar.gz
    sha256: 36658cb768a54c1d4dec43c3116c27ed893e88b02ecfcb44f2166f9c0b7f2a0d
    url: http://zlib.net/zlib-1.2.8.tar.gz
build:
    features: ''
requirements:
    build:
        - cmake
test:
    commands:
        - test -f ${PREFIX}/lib/libz.a
extra:
    recipe-maintainers:
        - groutr
        - msarahan
        - ocefpaf
```

jakirkham:
Could we tweak it to space out the main sections?

jakirkham:
Also, just to clarify, will this proceed through the download (but not build phase)? What is the option to enable/disable this in case people don't want to proceed that far?

msarahan:
By default, it will not download the source.  You can pass the -s/--source option to do so.  If it can't fill in fields without the source, it will throw errors (Jinja strict undefined stuff that @stuarteberg came up with).

You can also pass the -o/--output flag to print the output filename (as conda build did/does), rather than the meta.yaml data.

Sorry, but adding spaces is not something I want to fight PyYAML to do.  It was enough fun getting it to where it is now.  If you want to tackle that, I think the right place would be https://github.com/conda/conda-build/pull/908/files#diff-154513cee3e5765ecebd315978958f40R212

stuarteberg:
Looks good to me, but I have a couple of questions:

`conda build --output` first attempts parse `meta.yaml` without downloading the source, but then it automatically downloads the source if necessary.  But as you noted above, the new `render` command doesn't do that:

>By default, it will not download the source. You can pass the -s/--source option to do so. If it can't fill in fields without the source, it will throw errors

So in other words, the user is expected to interpret the error and add `--source` when necessary?  I can see why we wouldn't want to preemptively download the source (it won't be necessary for all recipes), but in the case of an error, isn't it nicer to just go ahead and give it a try on the user's behalf, as we do for `conda build --output`?

Also, I have an unrelated question about the output format. (If this PR is still WIP, then this question may be premature...)

The current version of your PR writes the fully-rendered YAML to `stdout`.  Are you planning to change that before this PR is merged?  It seems like a file output would be more desirable because misc. functions throughout the conda code base write various status messages to `stdout`, which will pollute the output with non-YAML junk.

msarahan:
@stuarteberg all good ideas.  This is why I ping you.  I also had the thought that there need not be a --yaml option - who wants output in any other format?  If they want the metadata itself as a python object, they can call the ```render_recipe``` function themselves.  I will replace it with a -f/--file option that writes to a file (the argument to -f/--file), but keep the default as writing to stdout.  Sound good?

As for downloading the source: you're right.  I like the transparent failover better than assuming interpretation of error messages.

stuarteberg:
Sounds good!

jakirkham:
I think I missed the recurse portion when skimming this quickly before. Would this be useful for [feedstocks]( http://github.com/conda-forge/feedstocks )?

msarahan:
@jakirkham Yes, that was my motivation.  I can now replace things in the main repo with submodules to feedstocks.  Closing the loop, little by little.

I still need to tie in rendering the recipe for tools that are expecting static content, but it's progress.

stuarteberg:
>This change also makes conda-build recursively search for meta.yaml in any folder you provide

I've been contemplating a kinda similar feature, but haven't taken the time for a PR yet:  What do you guys think about a `RECIPE_PATH` environment variable that conda-build uses to search for recipes it can't find?

```
RECIPE_PATH=/path/to/some-dir:/path/to/some-other-dir conda build mypackage
```

... if that's off-topic I can raise it in a different PR.

jakirkham:
I love it. Sorry, I got distracted by the static rendering. Lots of great stuff in this PR. 👍

> I can now replace things in the main repo with submodules to feedstocks. Closing the loop, little by little.

👍  Do you want help doing this after the release? It sounds like we are entering [phase 3]( https://github.com/conda-forge/conda-forge.github.io/issues/22#issuecomment-201619228 ).

cc @pelson

> I still need to tie in rendering the recipe for tools that are expecting static content, but it's progress.

Are you meaning like a public API or are there other tools that you guys have that will need this?

msarahan:
@stuarteberg That's not a bad idea.  I don't think I would use it, but I'm sure others would.  I have many folders of recipes, with very high degrees of overlap.  Having this feature would confuse me a lot on which collection I was getting.  I do think it deserves its own PR.

msarahan:
@jakirkham 

> Do you want help doing this after the release? It sounds like we are entering phase 3.

Help removing recipes from conda-recipes would be great.  I will ask again about the CLA - if it would be necessary for this.  Internal people have to do the internal repo, obviously.

> Are you meaning like a public API or are there other tools that you guys have that will need this?

Most just the legacy internal build system.  It does a lot of moving packages around and examining metadata, so I need to make sure that wherever it looks, it finds fully rendered recipes.

jakirkham:
> Help removing recipes from conda-recipes would be great. I will ask again about the CLA - if it would be necessary for this. Internal people have to do the internal repo, obviously.

I mean if you guys would rather merge PRs that is fine to. I'm not picky here. Whatever works really. I'm just asking as to how we and others at conda-forge should start to act in this regard.

> Most just the legacy internal build system. It does a lot of moving packages around and examining metadata, so I need to make sure that wherever it looks, it finds fully rendered recipes.

I see. Thanks for clarifying.

jakirkham:
So, I'm thinking this might be part of our solution to automating recipe generation to cut down time spent on minutiae. See this [comment]( https://github.com/conda-forge/staged-recipes/issues/535#issuecomment-216650342 ) for some thoughts.

jakirkham:
So, I tried to build this [example package]( https://github.com/conda/conda-build/tree/master/example_packages/setuptools ), but it doesn't quite work as it doesn't like `conda.yaml`. So, I renamed the file to `meta.yaml` and was able to build it with `conda build`. However, I can't get `conda render` to work.

msarahan:
@jakirkham - I'll have it look for conda.yaml too.  I'll test with conda render, too.  Thanks for testing.

jakirkham:
Sure, I don't know if I really understand the code path that the example package linked above is following, but maybe if you could point out what is going on with this I could try to help. There seems to be a lot of flexibility to that approach and I think this could resolve some of our issues with generating recipes quickly in a standard way.

msarahan:
so the issue with conda render is that it expects the few required fields too early - before the first template pass.  I'm going to see about changing that.

msarahan:
@jakirkham should work better now.

jakirkham:
Thanks. I'll give it another try.

jakirkham:
Works. Thanks @msarahan.

```
package:
    name: test_package
    version: 1.1.0
build:
    entry_points: ''
    number: '1'
    script:
        - cd $RECIPE_DIR
        - $PYTHON setup.py install --single-version-externally-managed --record=record.txt
requirements:
    build:
        - python
        - setuptools
    run:
        - python
        - Flask
        - werkzeug
test:
    imports: ''
    requires: ''
```

jakirkham:
So, I tried basically using this on the `setup.py` from `pandas`, but it got a little picky about having things like `pytz >=2011k` instead of `pytz >= 2011k`. Now we can and probably should clean these sorts of whitespace things in our rendering process (I added my own hack for now), but maybe it could be a little more relaxed about this sort of stuff for rendering as it can be tricky to catch all of that stuff on the first pass and it can be cleaned up right after.

jakirkham:
Kind of weird that the build number keeps getting turned into a string. Not sure where that is coming from.

jakirkham:
When you are doing this rendering @msarahan, are you including the `about` section? It seems to be missing in many cases.

conda_build/main_render.py
```diff
@@ -0,0 +1,275 @@
...
+def get_package_build_string(metadata):
+    import conda_build.build as build
+    metadata = parse_or_try_download(metadata)
+    return build.bldpkg_path(metadata)
+
+
+# Next bit of stuff is to support YAML output in the order we expect.
+# http://stackoverflow.com/a/17310199/1170370
+class MetaYaml(dict):
+    fields = ["package", "source", "build", "requirements", "test", "extra"]
```
jakirkham: Could we please add `about` here?

jakirkham:
> When you are doing this rendering @msarahan, are you including the `about` section? It seems to be missing in many cases.

Nevermind. Answered my own question. It would be great if we could add `about` here.

conda_build/main_render.py
jakirkham: Should come after `test`, but before `extra`.

msarahan: Good catch. Oversight. Will add.

jakirkham: No worries. With this change things appear to work for me.

jakirkham:
So, I made this variant based on the included template, which looks like [this]( https://gist.github.com/jakirkham/726ed57f1fef0c94fce22b7429124730 ) still kind of WIP, but it is coming along nicely. From that, I can generate the `pandas` recipe as long as it is included in the source.

Would be good to figure out how we can work this into more of a `conda skeleton pypi` operation. That way we can get checksums, fill out the download url, and such.

Also, we could probably get the `recipe-maintainers` filled out with a user's handle given a little more detective work.

The license unfortunately is always a mess as no one seems to think it worthwhile to specify the license version in the `license` field. We might be able to get around this by using `classifiers` if they added one specifying the license version. Though in the case of `pandas` they don't. Maybe this is the tedious step we do manually. I guess it is not so bad as we really don't want to get this wrong.

jakirkham:
Added this issue ( https://github.com/conda/conda-build/issues/914 ) to try and refactor `conda skeleton` and expose that functionality to `meta.yaml` in a straightforward manner. We could push the existing approach the rest of the way with this change I think.

msarahan:
Discussed with @kalefranz.  Long-term, need to move logic out of CLI files (main_*).  OK for now, though.  Merging.

PR #912, opened 2016-05-04
Fix issue with cpan skeleton generating invalid recipes for core modules.

dan-blanchard:
Since the conda skeleton was written `source` and `test` are now required to be dicts if present instead of allowing empty strings.  This fix comments `source` and `test` out if they would have been empty.


msarahan:
Thanks for this.  Merging.

PR #919, opened 2016-05-05
Restore old conda-build behavior for empty sections

groutr:
This is something that I inadvertently broke in #801.  This restores the old behavior until #874 is finished.


msarahan:
LGTM.  Thanks for catching the problem.

Do you want to add a test for this case, so it doesn't regress?

msarahan:
Thanks, @groutr

PR #920, opened 2016-05-06
remove upload always true, fix some args in download functions

msarahan:
This makes the --output argument of conda render respect the --no-source argument.  It also tries to fix some logic regarding anaconda.org uploading that looked broken.


msarahan:
lint: binstar_args created but not used.  Merging.  Will fix in next PR.

Issue #755, opened 2016-02-02
conda build doesn't allow passing anaconda token for upload


pitrou:
The `anaconda` client has a `--token` option to pass an authentication token. However, `conda build` doesn't reflect that option, making it more tedious to upload files in a CI system.

pitrou:
@jakirkham

jakirkham:
Definitely a good idea.

ivoflipse:
You also can't upload to any other account than your own, which is a pain if you want to add packages to an Organization account

pitrou:
> You also can't upload to any other account than your own, which is a pain if you want to add packages to an Organization account 

Excellent point too.

jakirkham:
Isn't there a user flag that solves this when uploading normally?

jakirkham:
Token should resolve the organization account, as well.

kalefranz:
I'm not sure we should address this with conda-build. It seems like this is the job of the anaconda client. If it's not already, the anaconda client should be looking for an `ANACONDA_TOKEN` environment variable or something similar. At most, conda-build should be simply shelling out to the anaconda client here. 

If I remember right, there's a ticket somewhere to make conda-build more robust in how it gives file paths for the anaconda client to upload.  I think that would be more in scope here.

pitrou:
Perhaps @stephenakearns wants to chime in here.

PR #921, opened 2016-05-06
WIP: add command line options to pass tokens and user through to anaconda upload

msarahan:
Closes https://github.com/conda/conda-build/issues/755

Note: anaconda upload has no --token argument.  Might need a separate call to anaconda, then anaconda upload?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ install:
   - hash -r
   - conda config --set always_yes yes
   - conda update -q --all
-  - conda install --force --no-deps conda 
+  - conda install --force --no-deps conda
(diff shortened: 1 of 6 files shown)


msarahan:
@pitrou - if I could please place the testing burden on you - I have never used this feature, and I'd like to avoid the learning time overhead right now.  Please propose a set of test steps, so we can add a test for this feature to our test suite.

pitrou:
I checked that `anaconda -t TOKEN upload -u USER ...` works on the command-line, so testing that such a command is issued would be enough, IMHO.

msarahan:
Passing on Travis - Will resolve Windows issues separately.  Merging.

PR #924, opened 2016-05-06
Set up anaconda.org builds for conda-build

msarahan:
Also, simplify in-source conda recipe for conda-build, using jinja more simply (depends on git tags, intended only for development releases right now)


msarahan:
Travis failures are lint that was removed in master.  Merging.

PR #926, opened 2016-05-07
add test for environ.get access to jinja vars

msarahan:
@stuarteberg - Not sure if you caused this, but you probably know more than anyone about when this might have changed.

The issue is that right now, the officially documented way of using jinja:

    {{ environ.get("GIT_DESCRIBE_TAG") }}

is not working.

    {{ GIT_DESCRIBE_TAG }}

is working.

This PR copies the jinja test you wrote, and keeps the environ.get syntax.  It fails for me, but I don't understand why - from following the code, it seems that those variables all are provided to the environment with the environ.get_dict(metadata) function in build.py.

Any insight you can offer would be much appreciated.  Did jinja change?  Should we solve this by updating the docs?  What about people with broken recipes?


msarahan:
I think I understand this.  The problem is that the environ.get() syntax *can* be evaluated in the first pass, and so it is evaluated, but it evaluates to None.

I'll try to think up a way to work around this.

stuarteberg:
I'm not able to reproduce the problem you're seeing here, @msarahan.  Should I expect to see a crash?  Or the wrong `git_tag`?  (Anyway, both seem correct when I test...)

>I think I understand this. The problem is that the environ.get() syntax can be evaluated in the first pass, and so it is evaluated, but it evaluates to None.

That doesn't make sense to me.  No matter what happened in the first pass, the entire file is parsed again for the second pass.  (The values we found the first pass are discarded.)

msarahan:
These guys: https://github.com/numba/numba/pull/1882/files were not seeing the GIT_* variables at all.  They tried both source/path and source/git_url, but neither was showing correct info.  Looking back, I may have been confused, and forgotten to check on whether their complaint was just about the initial build string, which is a well known issue (but I think fixed by #908)

Our test suite here seems happy enough with the new test.  I guess we should still update the docs, and consider the old environ.get as legacy, or preferred for situations where a fallback default is wanted?

stuarteberg:
>Our test suite here seems happy enough with the new test.

BTW, Is this PR missing a commit?  AFAICT, the "new" test is identical to the old one.

stuarteberg:
>AFAICT, the "new" test is identical to the old one.

(But, I should add, that I made sure to use the `environ.get()` syntax when I tested on my machine...)

msarahan:
@stuarteberg the "old" test used environ.get.  I added a test for using variable names directly ("source_git_jinja2_raw_vars")  You're right that my logic is flawed.  I'm at a loss as to why the numba recipe was failing.  You test the "raw_vars" in metadata/jinja_vars - so this PR does not add anything important.

Closing.  May revisit if I can nail down the issue later.

PR #927, opened 2016-05-08
Reactivate skip-existing and two minor fixes.

johanneskoester:
@msarahan, your PR bioconda/conda-build#1 accidentally removed the entire skip-exisiting check. Since our fork was merged into upstream master afterwards, the master branch currently lacks skip-existing functionality. This PR reactivates it. 

Further, I have renamed `get_package_build_string` to `get_package_build_path`, because it returns a path and build string would be ambiguous with the build string from the meta.yaml.

Finally, I have changed `already_built` into a set, which is the better DS for adding and lookups.

### DIFF
--- a/conda_build/main_build.py
+++ b/conda_build/main_build.py
@@ -225,7 +225,7 @@ def execute(args, parser):
             update_index(d)
         index = build.get_build_index(clear_cache=True)
 
-    already_built = {}
+    already_built = set()
     to_build_recursive = []
     with Locked(config.croot):
         recipes = deque(args.recipe)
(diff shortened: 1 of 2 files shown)


msarahan:
Sorry about the bad merge.  Thanks for catching it.  Again, please add a test for your use case!

msarahan:
Thanks @johanneskoester - again, please add a test.  It makes everyone's lives better to know that your use case will be protected by a loudly complaining test case.  If you need help knowing how/where to add this test, let me know.

conda_build/main_build.py
```diff
@@ -223,8 +223,9 @@ def execute(args, parser):
             if not isdir(d):
                 makedirs(d)
             update_index(d)
+        index = build.get_build_index(clear_cache=True)
```
msarahan: I think I removed this because pyflakes was complaining about index being set but not used.  I see you use it below.  Is there a reason to have it up here, and not down, immediately under

    if args.skip_existing:

?

msarahan:
Merging in #935 - @johanneskoester please look over that PR and ensure that it does what you intended.

PR #931, opened 2016-05-09
add library path to shebang patches

kalefranz:
Proposed addition to shebang rewrites.

CC @msarahan @ilanschnell

### DIFF
--- a/conda_build/post.py
+++ b/conda_build/post.py
@@ -70,10 +70,13 @@ def fix_shebang(f, osx_is_app=False):
     py_exec = ('/bin/bash ' + config.build_prefix + '/bin/python.app'
                if sys.platform == 'darwin' and osx_is_app else
                config.build_prefix + '/bin/' + basename(config.build_python))
-    new_data = SHEBANG_PAT.sub(b'#!' + py_exec.encode(encoding), data, count=1)
+    lib = "{0}_LIBRARY_PATH={1}/lib".format('DYLD' if sys.platform == 'darwin' else 'LD',
+                                            config.build_prefix)
+    new_shebang = '#!/usr/bin/env {0} {1}'.format(lib, py_exec).encode(encoding)
+    new_data = SHEBANG_PAT.sub(new_shebang, data, count=1)
     if new_data == data:
         return
-    print("updating shebang:", f)
+    print("updating shebang of {0} with {1}".format(f, new_data.split('\n')[0]))
     with io.open(path, 'w', encoding=locale.getpreferredencoding()) as fo:
         fo.write(new_data.decode(encoding))
     os.chmod(path, int('755', 8))


msarahan:
Can you point to an issue that this fixes?  I'm not intimately familiar with shebangs, outside of beginner linux knowledge.  There, I'm used to seeing them point at executables, but here you seem to point at libraries or folders of libraries?

conda_build/post.py
```diff
@@ -70,10 +70,13 @@ def fix_shebang(f, osx_is_app=False):
     py_exec = ('/bin/bash ' + config.build_prefix + '/bin/python.app'
                if sys.platform == 'darwin' and osx_is_app else
                config.build_prefix + '/bin/' + basename(config.build_python))
-    new_data = SHEBANG_PAT.sub(b'#!' + py_exec.encode(encoding), data, count=1)
+    lib = "{0}_LIBRARY_PATH={1}/lib".format('DYLD' if sys.platform == 'darwin' else 'LD',
+                                            config.build_prefix)
+    new_data = SHEBANG_PAT.sub('#!/usr/bin/env {0} '.format(lib) + py_exec.encode(encoding),
+                               data, count=1)
```
stuarteberg: On Mac, it's usually best to use `DYLD_FALLBACK_LIBRARY_PATH` instead of `DYLD_LIBRARY_PATH`.  (I don't know if this use-case is an exception to that rule, though.)

kalefranz:
> Can you point to an issue that this fixes?

On my mac, I have homebrew python installed at `/usr/local/bin/`, and also that location first on `$PATH`.  I have conda installed at `/conda` and `/conda/bin/` on path, but not first.

With the patch, `/conda/bin/virtualenv` works as expected.  Without the patch, I get a  symbol error

    kfranz@0283:~/forge/staged-recipes *supervisor ❯ /conda/bin/virtualenv ve
    New python executable in ve/bin/python
    Installing setuptools, pip, wheel...
      Complete output from command /Users/kfranz/forge/...ecipes/ve/bin/python -c "import sys, pip; sys...d\"] + sys.argv[1:]))" setuptools pip wheel:
      File "<string>", line 1, in <module>
      File "/conda/lib/python2.7/site-packages/virtualenv_support/pip-7.0.1-py2.py3-none-any.whl/pip/__init__.py", line 13, in <module>
      File "/conda/lib/python2.7/site-packages/virtualenv_support/pip-7.0.1-py2.py3-none-any.whl/pip/utils/__init__.py", line 15, in <module>
      File "/conda/lib/python2.7/zipfile.py", line 6, in <module>
        import io
      File "/conda/lib/python2.7/io.py", line 51, in <module>
        import _io
    ImportError: dlopen(/Users/kfranz/forge/staged-recipes/ve/lib/python2.7/lib-dynload/_io.so, 2): Symbol not found: __PyCodecInfo_GetIncrementalDecoder
      Referenced from: /Users/kfranz/forge/staged-recipes/ve/lib/python2.7/lib-dynload/_io.so
      Expected in: dynamic lookup

    ----------------------------------------
    ...Installing setuptools, pip, wheel...done.
      File "/conda/bin/virtualenv", line 6, in <module>
        sys.exit(main())
      File "/conda/lib/python2.7/site-packages/virtualenv.py", line 832, in main
        symlink=options.symlink)
      File "/conda/lib/python2.7/site-packages/virtualenv.py", line 1004, in create_environment
        install_wheel(to_install, py_executable, search_dirs)
      File "/conda/lib/python2.7/site-packages/virtualenv.py", line 969, in install_wheel
        'PIP_NO_INDEX': '1'
      File "/conda/lib/python2.7/site-packages/virtualenv.py", line 910, in call_subprocess
        % (cmd_desc, proc.returncode))
    OSError: Command /Users/kfranz/forge/...ecipes/ve/bin/python -c "import sys, pip; sys...d\"] + sys.argv[1:]))" setuptools pip wheel failed with error code 1

Starts to--but does not completely--address issues like https://github.com/conda/conda/issues/1367.


> but here you seem to point at libraries or folders of libraries?

Not quite.  For virtualenv for example, the shebang without the patch was

    #!$PREFIX/bin/python

and with the patch the shebang becomes

    #!/usr/bin/env [DY]LD_LIBRARY_PATH=$PREFIX/lib $PREFIX/bin/python

So it sets/overrides the library path for the current executable only.

conda_build/post.py
kalefranz: @stuarteberg can you expand on "usually best"?  I think `DYLD_LIBRARY_PATH` is the appropriate environment variable in this particular case.

msarahan:
Thanks for the explanation.  Pending @stuarteberg's comment on the variable to use, I am +1 on this.

stuarteberg:
>@stuarteberg can you expand on "usually best"? I think `DYLD_LIBRARY_PATH` is the appropriate environment variable in this particular case.

Yeah, you're right to call me out for being vague here.  I'm afraid all I've got to offer is "In my experience, `DYLD_LIBRARY_PATH` sometimes causes weird problems, whereas `DYLD_FALLBACK_LIBRARY_PATH` doesn't."

I know, that's a pretty lame answer. :-P  If it helps, here's an example of someone attempting to explain: http://stackoverflow.com/a/3172515/162094

Anyway, I'm not an expert on this, so if you've got deeper knowledge than I, don't mind my comments.

Moving on:

>and with the patch the shebang becomes
>
>```bash
>#!/usr/bin/env [DY]LD_LIBRARY_PATH=$PREFIX/lib $PREFIX/bin/python
>```

Wouldn't it be better not to throw away the existing value (if any)?

```
#!/usr/bin/env LD_LIBRARY_PATH=$PREFIX/lib:$LD_LIBRARY_PATH $PREFIX/bin/python
```

I sometimes rely on those variables when I'm working in development mode, before I've created a proper conda-package.

stuarteberg:
As a last side note (sorry to be a thorn in your side), I generally agree with @ilanschnell's [opinion regarding these variables](https://github.com/conda/conda/issues/1367#issuecomment-178813795): we should avoid using them.  I'd prefer an alternative solution if possible.

Regarding the virtualenv issue specifically, I just [posted an alternative workaround](https://github.com/conda/conda/issues/1367#issuecomment-217958169), inspired by your [explanation of the problem](https://github.com/conda/conda/issues/1367#issuecomment-195596461).  Maybe it can be adapted into an upstream patch for virtualenv?

kalefranz:
@stuarteberg 

From the stackoverflow comment:

> This will look for symbols in libraries provided in the default paths and if the symbol is not found there, fall back to the specified path.

This is actually exactly the opposite behavior we want here.  We need to prepend to the library search path, not append.

> Wouldn't it be better not to throw away the existing value (if any)?

Yes, I wish that were possible.  Shebang lines are incredibly limited in their ability though.  With the runtime variable expansion you propose in 

    #!/usr/bin/env LD_LIBRARY_PATH=$PREFIX/lib:$LD_LIBRARY_PATH $PREFIX/bin/python

You'd literally end up with your python executable seeing that environment variable as something like "/conda/lib:$LD_LIBRARY_PATH".

> As a last side note (sorry to be a thorn in your side), I generally agree with @ilanschnell's opinion regarding these variables: we should avoid using them.

Ilan's opinion there *I think* refers to the general case of setting library path in your bash config or at least for your whole terminal session.  This is a very specific case that only explicitly sets the library path for the specific subprocess being executed.

kalefranz:
Also CC @mingwandroid.  Any thoughts?

kalefranz:
As a side note, dropping `$PREFIX/bin/` from the python part of the shebang could be a solution to shebangs being too in certain environments.  That can be done independently of if we implement the library path part.  So, if a python shebang ends up being too long, we can just make it `#!/usr/bin/env python`.

stuarteberg:
>You'd literally end up with your python executable seeing that environment variable as something like `/conda/lib:$LD_LIBRARY_PATH` without any additional expansion.

Yikes, good point.  That's a bummer, though...

>I don't think this problem is limited to just virtualenv. But if it actually is, we can always just patch the conda version of virtualenv.

As you explained in conda/conda#1367, the problem occurs because virtualenv copies the `python` binary to a new location without also copying `libpython2.7.dylib`.  Besides virtualenv, what other packages actually copy the python interpreter itself?

ilanschnell:
Needless to say that I am very strongly opposed to this change.  The `virtualenv` problem would be fixed on their end, not here.  Adding `/usr/bin/env` as a requirement for very Python script introduces another subprocess, and more complexity.

kalefranz:
I guess I would have assumed `/usr/bin/env` would just `exec` and pass the subprocess off.  But now I'm going to check.

ilanschnell:
Maybe, but my main point is that it adds more overhead for no good reason.  Moreover, it will "hide" link errors, and by setting LD_LIBRARY_PATH, it will also also likely introduce new problems.

mingwandroid:
Regarding DYLD_*, none of it works since SIP anyway, so anything you hope to fix by using this isn't a fix for the latest Apple OS, or any future ones unfortunately.

kalefranz:
@mingwandroid 

> Regarding DYLD_*, none of it works since SIP anyway, so anything you hope to fix by using this isn't a fix for the latest Apple OS, or any future ones unfortunately.

Pretty sure SIP only affects executables at /[s]bin, /usr/[s]bin, and maybe /usr/local/[s]bin.  Anything outside of those paths is untouched by SIP--at least for now.

kalefranz:
I'm going to go ahead and decline this PR for now.  We can patch the conda virtualenv package directly.  I think there are other examples that this would help, but I can't think of them right now.  If we come across the issue again, I guess we can revisit it then.

mingwandroid:
@kalefranz Unfortunately SIP disables DYLIB_* in a global manner since `/bin/sh` is a SIP protected executable and at the end of the day, that's what launches our software nearly all the time. To prevent this we could build our own `bash` and given that Apple have stuck to 3.2, there's a good amount of merit to that suggestion, but users would have to hook it into Terminal.app (Preferences->General->Shells open with) .. or whatever terminal they are using. For reference:

http://apple.stackexchange.com/questions/212945/unable-to-set-dyld-fallback-library-path-in-shell-on-osx-10-11-1

PR #932, opened 2016-05-09
limit scope of locks in main_build

msarahan:
Fixes an issue where outer lock of build would conflict with inner lock of metadata rendering.  The scope of that lock was far too large, anyway.  Now the lock is applied for each build process, and for each test process, but not in between.

cc @jakirkham

### DIFF
--- a/conda_build/main_build.py
+++ b/conda_build/main_build.py
@@ -226,119 +226,120 @@ def execute(args, parser):
 
     already_built = []
     to_build_recursive = []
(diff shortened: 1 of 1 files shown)


jakirkham:
Awesome. Thanks for looking into this, @msarahan.

If you need a place to test this, feel free to build off this PR ( https://github.com/conda-forge/apptools-feedstock/pull/2 ), which I'm guessing you have already seen.

PR #933, opened 2016-05-10
write env vars to bld.bat

msarahan:
This also rolls back to an earlier behavior of writing the build env vars to the bld.bat script, rather than passing the env to the subprocess call.  This is simpler for debugging, but also (partially) fixes an issue observed in https://github.com/conda-forge/apptools-feedstock/pull/3 - there, with a Python 3.5 root install, the VS 2010 py 3.4 configuration was seeing the root environment rather than the build environment.  This PR changes that so at least the correct Python is on PATH - but the package is still broken somehow.

It should be said that this PR might actually somewhat mask some other issue - the issue being that the _build environment is not always being used, as it should be.

### DIFF
--- a/conda_build/windows.py
+++ b/conda_build/windows.py
@@ -3,7 +3,7 @@ from __future__ import absolute_import, division, print_function
 import os
 import sys
 import shutil
-from os.path import dirname, isdir, isfile, join, exists
+from os.path import dirname, isdir, isfile, join
 
 import conda.config as cc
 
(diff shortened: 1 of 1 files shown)


msarahan:
Very interesting data point here from apptools-feedstock test pr mentioned above:

root env 2.7: builds 2.7, 3.4, 3.5 OK.
root env 3.5: builds 2.7, 3.5 OK. Fails to import apptools.
Not sure yet where that points. I have fixed part of the issue in this PR, but the issue is not done.

msarahan:
Merging this for its value in debugging.  Will explore activation of _build and_test in another PR.

PR #938, opened 2016-05-11
enable py3.5 win; quiet updates, use flake8; fix pep8

msarahan:


msarahan:
flake failures appear to be related to https://bugs.launchpad.net/pyflakes/+bug/1560134 - which seems to be an upstream bug in pyflakes/flake8.  Going to try pinning to an older version.

msarahan:
@kalefranz @groutr @mingwandroid @ilanschnell 

I think these are uncontroversial changes - but please make sure you are OK with this before I go forward.  Changes here are only whitespace, plus CI changes to use flake8 instead of pyflakes, and enabling py3.5 on AppVeyor.

mingwandroid:
LGTM.

msarahan:
Thanks @mingwandroid - merging.

PR #940, opened 2016-05-11
fix win activate VS order; allow VS{ver}0COMNTOOLS env var for vs path

msarahan:
This is one that grew too much.  It began as just moving the VS activation down, so that other env var settings did not overwrite it.  

@ikalev suggested a good improvement to how we look for VS (using environment variables that VS sets.  We try those first, but fall back to the old way.  

There is a little bit in here for building on Anaconda.org - the test suite now runs there.

Finally, I fought a lot with Appveyor to test this PR.  There are lot of good lessons here - perhaps most notably knowledge from @patricksnape in getting CMake to find VS 2008 adequately on appveyor.  Also note that Appveyor now uses their existing miniconda installations, rather than downloading miniconda each build.


ikalev:
Note that ``VSx0COMNTOOLS`` points to the common tools directory, not vsvarsall:

```
VS140COMNTOOLS=C:\Program Files (x86)\Microsoft Visual Studio 14.0\Common7\Tools\
```

This implementation should work fine:

```Python
vstools = "VS{0}0COMNTOOLS".format(version)

if vstools in os.environ: 
    return os.path.join(os.environ[vstools], '..\\..\\VC\\vcvarsall.bat')
else:
    # prefer looking at env var; fall back to program files defaults
    return os.path.join(PROGRAM_FILES_PATH,
                        'Microsoft Visual Studio {}'.format(version), 'VC',
                        'vcvarsall.bat')
```

msarahan:
Thanks @ikalev - I integrated your changes.

msarahan:
@patricksnape I am having a great deal of trouble with CMake and VS 2008: https://ci.appveyor.com/project/msarahan/conda-build/build/1.0.122/job/lgpe1afpib3qnpc5

I think you fixed this with your comment at http://help.appveyor.com/discussions/kb/38-visual-studio-2008-64-bit-builds

Is there a good way to incorporate that stuff into conda-build, or are we stuck pointing people who use Appveyor to that information (and possibly also incorporating it into the conda-forge appveyor master template)?

msarahan:
Note that the path-finding stuff works great for python building C/C++ extensions, thanks to you, but the trouble is that CMake doesn't know where to look, as you point out.

patricksnape:
Ok so I totally agree with using that env variable - but I've got something in the pipeline that will go one step further and check the registry - since this should be a pretty bullet proof way of picking up the VS path no matter where it was installed.

Unfortunately Appveyor were not keen to edit the environment despite my explanation so I maintain this shim at the moment for Menpo. Does CMake not pick it up if you call the correct `vcvarsall.bat` beforehand?

msarahan:
Agreed that registry would be better. Looking forward to your contribution.

PR #942, opened 2016-05-12
cran: Use MSYS2

mingwandroid:
Unify Windows and Unix R dependencies:
 * Dependencies on packages in r-recommended are now explicitly listed
   on Windows just the same as for the other systems.
 * Add m2-base and m2w64-toolchain dependencies.
 * Use r-base as the build-string identifying dependency, as well as r
   (r is kept for legacy installations)

### DIFF
--- a/conda_build/cran.py
+++ b/conda_build/cran.py
@@ -28,6 +28,9 @@ from conda import compat
 from conda_build import source, metadata
 
 CRAN_META = """\
+{{% set posix = 'm2-' if win else '' %}}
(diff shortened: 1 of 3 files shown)


msarahan:
Breakage on appveyor is my fault - python 3 does not like some of the work I did with PEP8 fixes.  I'll ping you here when I have a fix for you to merge/rebase on.

mingwandroid:
Thanks @msarahan.

PR #947, opened 2016-05-12
inspect: use conda.install for dist names and metadata

mcg1969:
Fixes crashes in `conda inspect`, in part by avoiding any reliance on the filenames or dist names for package info. Instead uses the `conda-meta` content for that info.

### DIFF
--- a/conda_build/ldd.py
+++ b/conda_build/ldd.py
@@ -3,11 +3,11 @@ from __future__ import absolute_import, division, print_function
 import sys
 import re
 import subprocess
-import json
 from os.path import join, basename
 
 from conda.utils import memoized
 from conda.misc import untracked
+from conda.install import linked_data
 
 from conda_build import post
 from conda_build.macho import otool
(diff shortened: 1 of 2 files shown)


mingwandroid:
LGTM.

mcg1969:
@mingwandroid @msarahan, some of the techniques here might need to be incorporated more widely in the code. For example, using `linked_data` instead of `linked` is actually no more costly, because `linked` calls `linked_data`---and you get the full set of metadata instead of a list of names.

conda_build/ldd.py
```diff
@@ -50,23 +50,21 @@ def get_linkages(obj_files, prefix):
             res[f] = ldd(path)
         elif sys.platform.startswith('darwin'):
             links = otool(path)
-            res[f] = [(basename(l), l) for l in links]
+            res[f] = [(basename(l['name']), l['name']) for l in links]
```
mcg1969: Ran this fix by @mingwandroid already.

conda_build/ldd.py
```diff
@@ -50,23 +50,21 @@ def get_linkages(obj_files, prefix):
...
 
     return res
 
 
 @memoized
 def get_package_obj_files(dist, prefix):
-    with open(join(prefix, 'conda-meta', dist +
-            '.json')) as f:
-        data = json.load(f)
+    data = linked_data(prefix).get(dist)
```
mcg1969: Note that we've already read the entire contents of the `conda-meta` directory at this point. Using this command avoids a re-reading.

conda_build/main_inspect.py
```diff
@@ -229,26 +230,27 @@ def test_installable(channel='defaults', verbose=True):
         print("######## Testing platform %s ########" % platform)
         channels = [channel] + get_default_urls()
         index = get_index(channel_urls=channels, prepend=False, platform=platform)
-        for package in sorted(index):
-            if channel != 'defaults':
-                # If we give channels at the command line, only look at
-                # packages from those channels (not defaults).
-                if index[package]['channel'] not in normalize_urls([channel], platform=platform):
-                    continue
-            name, version, build = package.rsplit('.tar.bz2', 1)[0].rsplit('-', 2)
```
mcg1969: This is critical: we never want to depend on the package filename for this info. We need to extract it from the index instead.

mcg1969:
Note also that this fix is agnostic about the channel prefix stuff (e.g., `local::`). All we're doing is removing the assumption that the package/dist name has any informational content.

msarahan:
Thanks @mcg1969 - this is nice work.

PR #950, opened 2016-05-13
fix extraneous output when outputting build path

### LINKED ISSUE #949: Erroneous extra output in `conda build --output conda.recipe`
Before release 1.20.2, the command `conda build --output conda.recipe` only printed the path of the output file. Now, on all operating systems, it prints something like:

```
Removing old work directory
C:\Anaconda3\conda-bld\win-64\libdynd-0.7.2-0.tar.bz2
```

This is a pain when trying to redirect its output to another command (like anaconda upload).
For example, `anaconda --token $ANACONDA_TOKEN upload $(conda build --output conda.recipe) --user dynd --channel dev` (or the equivalent powershell version of that statement) is run in the libdynd CI tests.

If that was a deliberate change, what's the recommended workaround?

CC: @izaid

msarahan:
conda build --output was outputting other stuff, when external tools required strictly the package output path.  This fixes it, and adds tests for the behavior.

Closes #949 (sorry everyone)

### DIFF
--- a/conda_build/source.py
+++ b/conda_build/source.py
@@ -7,6 +7,7 @@ from shutil import copytree, copy2
 from subprocess import check_call, Popen, PIPE, CalledProcessError
 import locale
 
+from conda.compat import StringIO
 from conda.fetch import download
 from conda.install import move_to_trash
 from conda.utils import hashsum_file
@@ -69,12 +70,13 @@ def download_to_cache(meta):
     return path
 
 
-def unpack(meta):
+def unpack(meta, hide_output=False):
     ''' Uncompress a downloaded source. '''
     src_path = download_to_cache(meta)
 
     os.makedirs(WORK_DIR)
-    print("Extracting download")
+    if not hide_output:
+        print("Extracting download")
     if src_path.lower().endswith(('.tar.gz', '.tar.bz2', '.tgz', '.tar.xz',
             '.tar', 'tar.z')):
         tar_xf(src_path, WORK_DIR)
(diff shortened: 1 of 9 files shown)


stefanseefeld:
Now I see the "old" (and wrong) behavior again, i.e. it reports a label that doesn't correspond to what's in my local tree.

msarahan:
Yeah, I was trying to be clever about avoiding downloading the source, but that just doesn't work.  It will successfully render without it in your case, but it will not be accurate.

stefanseefeld:
Now `conda build ... --output` again generates more output than is desirable:
```
(py34)[stefan@quasimodo numba]$ conda build --numpy=1.10 --python=2.7 buildscripts/condarecipe.local --output
Copying /home/stefan/projects/numba to /usr/local/Continuum/anaconda/conda-bld/work
/usr/local/Continuum/anaconda/conda-bld/linux-64/numba-0.26.0.dev-np110py27_199.tar.bz2
```

johanneskoester:
Can you quickly clarify when the source is downloaded now? To me, it looks like it is downloaded always, regardless of if it is needed or not. This would be a huge problem for bioconda, because we use --skip-existing and --output to determine if recipes need a rebuild. However, we have thousands of recipes in our repository and downloading the source for all of them would need ages.

msarahan:
This isn't finished yet.  The idea is that it will only download source when it needs to.  For example, if your recipe uses git variables in jinja templates, you can't know the build output path without downloading the source.  However, if that's not necessary, then we should not download the source.

johanneskoester:
That's good, thanks!

conda_build/source.py
```diff
@@ -69,25 +72,34 @@ def download_to_cache(meta):
...
     if src_path.lower().endswith(('.tar.gz', '.tar.bz2', '.tgz', '.tar.xz',
             '.tar', 'tar.z')):
         tar_xf(src_path, WORK_DIR)
     elif src_path.lower().endswith('.zip'):
         unzip(src_path, WORK_DIR)
     else:
         # In this case, the build script will need to deal with unpacking the source
-        print("Warning: Unrecognized source format. Source file will be copied to the SRC_DIR")
+        if verbose:
+            print("Warning: Unrecognized source format. Source file will be copied to the SRC_DIR")
```
mwcraig: Should this warning be raised even if `verbose` is `False`? I normally think of verbose as "no chatter, but if something goes wrong please do tell me."

mwcraig:
Just one fairly minor comment. The tests do seem to be checking for #949 properly. Haven't tried running this locally, but can if you want.

conda_build/source.py
msarahan: Thanks, I'll fix that.

PR #953, opened 2016-05-14
Create python api

msarahan:
This is the start of providing an external-facing API for conda-build.  Functions in the api module should be considered both safe to outside people, and changing them implies bumping the major version number.

As an added bonus, this will finally start giving us relevant code coverage information for our test recipes.

CC @kalefranz @ilanschnell @jakirkham @ocefpaf @pelson @stuarteberg @mingwandroid 

- [ ] main_* scripts broken up to isolate CLI:
    - [x] main_build
    - [x] main_convert
    - [x] main_develop
    - [x] main_index
    - [x] main_inspect
    - [x] main_metapackage
    - [x] main_render
    - [x] main_sign
    - [x] main_skeleton


jakirkham:
I think it is worth keeping issue ( https://github.com/conda/conda-build/issues/1113 ) in mind while we try to sort this out. This failure mode has caused us a bit of grief at conda-forge.

msarahan:
Yes, good call. I would call that needing better tests of metadata parsing.

jakirkham:
Should we have a corresponding Python API for conda too? Or am I missing that it is already a thing somehow.

msarahan:
That's a good idea, but you need to raise it on conda's issue tracker.  Libconda is sort of an effort in that direction, but it is more of a copy of particular common functionality, rather than an API layer on top of currrent code.

jakirkham:
That's reasonable. Added this issue ( https://github.com/conda/conda/issues/3138 ) and copied you on it.

jakirkham:
I'd like to stay subscribed to this thread. However, the Codecov comments are really noisy. Any chance I could convince you guys to turn them off? It seems that info is readily available from the links provided in the status. Would this be a question for, @kalefranz? If not, I may just have to unsubscribe. Just feel free to ping me when I should be looking at something.

msarahan:
It's a question for me.  Sorry you find them noisy.  I will look into if there's a way to keep the status updates in the checks, but avoid the message posts.

jakirkham:
Ok, wasn't sure if it was an org standard or something. Thanks for looking into it.

msarahan:
Merging this to kill it and branch off smaller PRs for further work.  Present appveyor failure is due to a race condition in an anaconda.org test that needs improvement.

PR #956, opened 2016-05-15
cran: License guess fix and use 'posix' package

mingwandroid:
Please review @msarahan, nothing major here.


msarahan:
Thanks @mingwandroid

PR #957, opened 2016-05-16
metadata: Fixed selector regex to avoid confusion with jinja templates

### LINKED ISSUE #955: Regression regarding `[]` in jinja
Running into a regression when using the `[]` operator in jinja. This functionality was added by @stuarteberg in this PR ( https://github.com/conda/conda-build/pull/771 ). However, this appears to now be broken in `conda-build` 1.20.2. The current workaround would be to downgrade to `conda-build` 1.20.1, which does not have this problem. Here are some examples.

```
Error: Invalid selector in meta.yaml line 40:
    - python {{ environ['PY_VER'] + '*' }}  # [win]
```

See this [comment]( https://github.com/conda-forge/hdf5-feedstock/pull/13#issuecomment-219302083 ) and PR for more details.

```
Error: Invalid selector in meta.yaml line 12:
  fn:  graphviz-{{ version[:-2] }}.zip  # [win]
```

Also see this [comment]( https://github.com/conda-forge/staged-recipes/pull/568#issuecomment-219250321 ) for context.

cc @ccordoba12 @msarahan

stuarteberg:
The fix here is to replace `(.+)` with `([^\[\]]+)`:

```diff
-sel_pat = re.compile(r'(.+?)\s*(#.*)?\[(.+)\](?(2).*)$')
+sel_pat = re.compile(r'(.+?)\s*(#.*)?\[([^\[\]]+)\](?(2).*)$')
```

That is, selectors don't contain "any character", they contain "any non-bracket character".

Fixes #955

### DIFF
--- a/conda_build/metadata.py
+++ b/conda_build/metadata.py
@@ -69,7 +69,15 @@ def ns_cfg():
     return d
 
 
-sel_pat = re.compile(r'(.+?)\s*(#.*)?\[(.+)\](?(2).*)$')
+# Selectors must be either:
+# - at end of the line
+# - embedded (anywhere) within a comment
+#
+# Notes:
+# - [([^\[\]]+)\] means "find a pair of brackets containing any
+#                 NON-bracket chars, and capture the contents"
+# - (?(2).*)$ means "allow trailing characters iff group 2 (#.*) was found."
+sel_pat = re.compile(r'(.+?)\s*(#.*)?\[([^\[\]]+)\](?(2).*)$')
 
 
 def select_lines(data, namespace):
--- a/tests/test-recipes/metadata/jinja_vars/meta.yaml
+++ b/tests/test-recipes/metadata/jinja_vars/meta.yaml
@@ -4,7 +4,7 @@ package:
 
 build:
   number: 123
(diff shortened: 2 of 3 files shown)


msarahan:
Thanks @stuarteberg - I have been fighting regexes on this one for a while this morning.  Thanks for putting and end to that.

I think a better place to test (or perhaps in addition to this) would be test_metadata.py, in test_select_lines.  I had added:

    {{ environ["test"] }}  # [abc]

there.  The tests you added are fine - but adding this bit in test_metadata is a more direct test of this regex.  Otherwise, the only failure on travis was flake8.

stuarteberg:
>I have been fighting regexes on this one for a while this morning.

Yeah, the current PR was not my first attempt.  I had a different PR written up that changed the regex entirely, but then I had a head-smacking "ah-ha" moment before I hit the "Create Pull Request" button.

>in test_select_lines. I had added:
>
>     {{ environ["test"] }}  # [abc]

The current PR already adds a few equivalent cases to `test_select_lines()`, but I'll add this one as well.

>Otherwise, the only failure on travis was flake8.

Oops, I'll fix that, too.

msarahan:
Very nice, thanks @stuarteberg

stuarteberg:
regex

PR #958, opened 2016-05-16
WIP: Improve Windows MSVC logic

patricksnape:
Use setuptools/distutils to do the heavy lifting. Instead of needing all
the logic around the Microsoft C++ Compiler for Python 2.7, we just use
setuptools which already supports this. We do however, maintain the shim
for x64 builds if VS2008 is required. This should also fix up the Python
3.3/3.4 builds since there is better logic for detecting if the Windows
SDK failed (checking an environment variable).

@msarahan This is better than using the registry (it actually does use the registry) because setuptools and distutils are doing all the hard work for us.


patricksnape:
As a note there is one possible scenario not currently covered: Falling back to the Microsoft SDK v7.0 for Python 2.7. In this case we would need to set `DISTUTILS_USE_SDK` and `MSSdk` before calling the `SetEnv.cmd` in a manner similar to VS2010 (but with v7.0).

patricksnape:
The last commit seems to make my simple test case [pass](https://ci.appveyor.com/project/patricksnape/test-new-conda-build/build/1.0.21). This builds using both CMAKE/nmake and setuptools.

msarahan:
Looks like a nice simplification.  I'd like to hold off on merging until we have tests and documentation in place for the override.  If you want to tackle that here, go for it.  Otherwise I'll try to get to it soon.

patricksnape:
For which override sorry? Don't mind adding tests just not sure what you would like testing.

msarahan:
The msvc_compiler specification, where you can override the default hard
coded python - msvc version association.

patricksnape:
Ok as an update - I'm just having some troubles with the VS2010 tests I have failing despite everything looking correct... Need to investigate.

msarahan:
Which tests?  You are aware that we needed to tweak MSYS2 so that it would not mangle paths, right?  Have you pulled in a recent copy of origin?  That was an evil bug to find.

patricksnape:
New tests that I wrote to make sure everything actually works now! These tests check that the corrected Visual Studio versions actually get activated so hopefully we don't have any more bugs. Appveyor is a bit backed up right now so I have run the tests on [my account](https://ci.appveyor.com/project/patricksnape/conda-build).

There are a few issues:

  1) `test_build_msvc_compiler[10.0]` fails because for some bizarre reason calling `call "C:\Program Files\Microsoft SDKs\Windows\v7.1\Bin\SetEnv.cmd" /Release /x64` returns a non-zero exit code despite no errors seeming to occur.
  2) `test_cmake_generator[2.7-32/64]` fails because this logic prefers the Python tools and doesn't allow for a way to override this logic. This is something we should discuss. How do we want to mitigate this? Patch CMake? Prefer VS 2008 over the Python tools? Differentiate between vs9 and the Python tools?
  3) `test_cached_source_not_interfere_with_versioning ` fails due to an access denied error? This is new to me?

msarahan:
> 1) test_build_msvc_compiler[10.0] fails because for some bizarre reason calling call "C:\Program Files\Microsoft SDKs\Windows\v7.1\Bin\SetEnv.cmd" /Release /x64 returns a non-zero exit code despite no errors seeming to occur.

We need to figure out an alternative check for this.  What exactly is the return code?  Can we add a check for that return code, and change it to 0 if matches the "correct" value?

> 2) test_cmake_generator[2.7-32/64] fails because this logic prefers the Python tools and doesn't allow for a way to override this logic. This is something we should discuss. How do we want to mitigate this? Patch CMake? Prefer VS 2008 over the Python tools? Differentiate between vs9 and the Python tools?

Are you sure this is due to your new logic?  I think this is just CMake refusing to see the Python compiler tools - which your other patch addresses.  We need to ensure that your vs2008 patch/registry stuff is being done.

> 3) test_cached_source_not_interfere_with_versioning fails due to an access denied error? This is new to me?

New to me, too.  It looks like those files may still be in use somehow when the cleanup happens.  I doubt this is anything you've caused with your changes, but I'm not sure why it's happening.  Is it always failing here, or is it sporadic?

patricksnape:
For 1) I'm not entirely sure - I need to spend some more time debugging it.
For 2) yes this is the CMake issue I believe. I don't think you need the registry entries, you just need to copy the file. Our builds are working without the registry entries now.
For 3) I'm not sure if this is one off or not, would need to re-run the build.

patricksnape:
Just kicking the builds off again.

patricksnape:
@msarahan Appveyor doesn't appear to be starting?

msarahan:
Sorry, I changed over to a different AppVeyor account, and had not updated the webhook.  This is now on a paid plan, so things should go faster.

junit.xml
```diff
@@ -0,0 +1,538 @@
...
+writing C:\Program Files (x86)\Microsoft Visual Studio 9.0\VC\bin\vcvars64.bat (exit /b 1)
+

+C#x08L#x08I#x08N#x08K#x08 #x08#x1B[1;32;40mC:\Users\pts08\Documents\GitHub\conda-build {git}{hg} 

+#x1B[1;30;40m{lamb} #x1B[0mset DISTUTILS_USE_SDK=1 

+

+C#x08L#x08I#x08N#x08K#x08 #x08#x1B[1;32;40mC:\Users\pts08\Documents\GitHub\conda-build {git}{hg}
```
janschulz: Should this file be in the commit? The output is from a cmder session (`{git}{hg}` and `{lamb}`) and won't be the same in a plain cmd

patricksnape: Thanks @janschulz - this was indeed an accidental commit!

patricksnape:
I see that we are nearing a 2.0.0 conda-build release. It would be really fantastic if we could try and sort this stuff out some time soon. The biggest blocker at the moment is how to handle the case where invoking the windows compiler scripts fails. Should we be hard resetting the `ERRORLEVEL` to 0? How would we test for this as in tests we don't want that to happen?

Do we still want all this logic within conda-build? Or maybe it would be better as a pre-link script within a package? The take home from this PR (once it actually 100% works) is that setuptools actually does 95% of the hard work and we should use that rather than reinventing the wheel.

msarahan:
I will merge this before the 2.0 release.  Too busy with a conda release to do that right now, though.  More soon.

msarahan:
I'm actually not sure that this is the right way to go.  Though I wholeheartedly agree with using setuptools existing logic, I think the VS activate scripts themselves are a problem.  I have spend most of today fighting them.  When PATH is already too long, then running the script (for VS 2008, at least) clobbers PATH completely, leaving an utterly incapable prompt.  This has been a really difficult thing to debug about conda-build.

@mingwandroid has experimented with how long variables can be in shell, and got up to ~8100 characters.  I'm going to try to manually add the contents that the vsvars scripts take care of.

patricksnape:
OK. We would need to review how tools like setuptools and CMake determine the path to the correct compiler and ensure that conda build enables those tools to find the correct compiler. In actual fact, it seemed to me that both setuptools and CMake prefer the registry to environment variables and that is why I had preferred to use setuptools in the first place (to avoid having to parse the registry myself).

msarahan:
AFAICT, setuptools just has logic for finding the vcvarsall.bat files.  CMake seems to do more, since it looks for actual executables.

My latest strategy (suggested by @mingwandroid) is to have a clean shell (nothing but core OS paths on PATH), and to get the difference of environment variables when we run vcvarsall.  We can then apply that difference to our variables.

patricksnape:
Sounds reasonable. Though I do wonder in that case if we should drop all this logic from conda build and have it be a build-time package?

msarahan:
I have done what you suggest, and I'm very excited by it.  PR coming up soon.

msarahan:
Superseded by #1036 - closing here.  Thanks for your contributions @patricksnape - I have included them in that new PR.

PR #959, opened 2016-05-17
MatchSpec.strictness is no more

mingwandroid:
It's been replaced with MatchSpec.is_simple() and
MatchSpec.is_exact():

https://github.com/conda/conda/commit/6ddc6d9845a49f25fa67ed6702c409168b2be1e8

### DIFF
--- a/conda_build/metadata.py
+++ b/conda_build/metadata.py
@@ -304,10 +304,10 @@ def handle_config_version(ms, ver):
     configuration, e.g. for ms.name == 'python', ver = 26 or None,
     return a (sometimes new) MatchSpec object
     """
-    if ms.strictness == 3:
+    if ms.is_exact():
         return ms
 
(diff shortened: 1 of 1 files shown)


mingwandroid:
@mcg1969, @msarahan please review.

mcg1969:
Well, crap. Sorry Ray. I clearly need to do a better job of surveying how my changes to `conda` are affecting these other projects. We can put `strictness` back if necessary. But this LGTM.

mcg1969:
In fact we probably should put `strictness` back so that there isn't quite the tight dependency between `conda` versions and `conda-build` versions. What do you think @msarahan?

mingwandroid:
No problem @mcg1969, I can use the change locally to ensure that my linked_data_ change is working as expected for now.

mingwandroid:
Strictness is back, closing.

PR #967, opened 2016-05-23
Two tiny fixes

stuarteberg:
These commits were meant to be tacked on to #964, but I didn't push them in time before it was merged.  Sorry about that.

The `jinja_context` commit is pretty minor, but will hopefully it will help avoid potential surprises in the future.

The `environ` change here is important if you build a recipe with a `git_url` and then accidentally change the capitalization in the URL.

For example, if your recipe says:

```yaml
source:
  git_url: https://github.com/Foo/bar
```

and later you accidentally change it to:

```yaml
source:
  git_url: https://github.com/Foo/Bar
```

... then you'll encounter a weird error. `verify_git_repo()` will think that the git cache is out-of-sync, so `GIT_*` variables won't work.  Oops.  This PR just makes the comparison case-insensitive.

cc @msarahan -- Sorry again for the double-PR.

### DIFF
--- a/conda_build/environ.py
+++ b/conda_build/environ.py
@@ -93,7 +93,7 @@ def verify_git_repo(git_dir, git_url, expected_rev='HEAD'):
 
         # If the current source directory in conda-bld/work doesn't match the user's
         # metadata git_url or git_rev, then we aren't looking at the right source.
-        if remote_url != git_url:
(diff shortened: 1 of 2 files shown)


stuarteberg:
Travis is happy now (first build had a random network issue, apparently).

msarahan:
Thanks @stuarteberg

Issue #968, opened 2016-05-23
Handle `a/` and `b/` in patches better


jakirkham:
When `git` makes a patch, it looks like this with `a/` and `b/` at the beginnings of each filename. However, `conda-build` doesn't seem to handle this correctly and gets confused as to where to apply the patch. It is a bit of an annoyance to go back and fix these patches afterwards. So, it would be nice if `conda-build` just let these patches work.

```
From 2bea755a2139907d331bfa0229154b6dab7a61c1 Mon Sep 17 00:00:00 2001
From: Denis Barbier <bouzim@gmail.com>
Date: Tue, 10 May 2016 12:47:02 +0200
Subject: [PATCH] Instantiate abstract template classes *Function in
 assembly.cpp

This fixes issue #41.
---
 src/assembly.cpp | 15 +++++++++++++++
 1 file changed, 15 insertions(+)

diff --git a/src/assembly.cpp b/src/assembly.cpp
index 471ed6b..3ff8e54 100644
--- a/src/assembly.cpp
+++ b/src/assembly.cpp
@@ -214,6 +214,21 @@ void Function<T>::prepareBlock(const ClusterData*, const ClusterData*,
 
 
 // Template declaration
+template class Function<S_t>;
+template class Function<D_t>;
+template class Function<C_t>;
+template class Function<Z_t>;
+
+template class SimpleFunction<S_t>;
+template class SimpleFunction<D_t>;
+template class SimpleFunction<C_t>;
+template class SimpleFunction<Z_t>;
+
+template class AssemblyFunction<S_t>;
+template class AssemblyFunction<D_t>;
+template class AssemblyFunction<C_t>;
+template class AssemblyFunction<Z_t>;
+
 template class BlockFunction<S_t>;
 template class BlockFunction<D_t>;
 template class BlockFunction<C_t>;
```

PR #969, opened 2016-05-23
Handle -p1 patch level

jschueller:
fixes https://github.com/conda/conda-build/issues/968

### DIFF
--- a/conda_build/source.py
+++ b/conda_build/source.py
@@ -336,7 +336,12 @@ Error:
     try:
         check_call([patch] + patch_args, cwd=src_dir)
     except CalledProcessError:
-        sys.exit(1)
+        # fallback to -p1, the git default
+        patch_args[0] = '-p1'
+        try:
+            check_call([patch] + patch_args, cwd=src_dir)
+        except CalledProcessError:
+            sys.exit(1)
     if sys.platform == 'win32' and os.path.exists(patch_args[-1]):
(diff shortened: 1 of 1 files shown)


msarahan:
Thanks @jschueller

conda_build/source.py
```diff
@@ -336,7 +336,12 @@ def apply_patch(src_dir, path):
     try:
         check_call([patch] + patch_args, cwd=src_dir)
     except CalledProcessError:
-        sys.exit(1)
+        # fallback to -p1, the git default
+        patch_args[0] = '-p1'
```
jakirkham: This feels a little fragile (not too bad) because if the arg order is changed somehow. This could start breaking. Could we find another way to change this.

Also, I don't think we are going to hit this `except` because patch will wait for user input current. Can we add an argument to make sure patch errors out.

jakirkham:
I guess we have already merged it, but I have some concerns I raised above.

mingwandroid:
Yeah, I tried to fix this before and gave up. I concluded that we should either:
* Add metadata for patch-level to the patches in meta.yaml.
* Use https://github.com/conda/conda-recipes/blob/master/common-scripts/patch.py for all of our patching needs. It does some patch level normalization, I'm not sure it'd work immediately though, but we could probably add better patch level detection.

jakirkham:
Both great ideas. I think I prefer the second one.

PR #973, opened 2016-05-24
add --dirty flag to skip download/extraction step

msarahan:
I've been working on Qt5 builds.  Extracting the source takes 5-10 *minutes*.  This PR avoids it.  This is a gun pointed at your foot - use it to iterate quickly on recipes, but keep in mind that it may bite you if you forget to capture any source code changes as patches or changes to your recipe.

### DIFF
--- a/conda_build/source.py
+++ b/conda_build/source.py
@@ -346,7 +346,7 @@ Error:
         os.remove(patch_args[-1])  # clean up .patch_unix file
 
 
-def provide(recipe_dir, meta, verbose=False, patch=True):
+def provide(recipe_dir, meta, verbose=False, patch=True, dirty=False):
     """
     given a recipe_dir:
       - download (if necessary)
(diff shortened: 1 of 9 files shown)


patricksnape:
This would be extremely useful.

msarahan:
Any other thoughts on improving this, @patricksnape?  If not, I'll merge this soon.

conda_build/source.py
```diff
@@ -346,43 +346,45 @@ def apply_patch(src_dir, path):
...
+                move_to_trash(WORK_DIR, '')
+        else:
+            rm_rf(WORK_DIR)
+
+    if not os.path.exists(WORK_DIR):
+        if any(k in meta for k in ('fn', 'url')):
+            unpack(meta, verbose=verbose)
+        elif 'git_url' in meta:
+            git_source(meta, recipe_dir, verbose=verbose)
+        # build to make sure we have a work directory with source in it.  We want to make sure that
```
patricksnape: The comments here are repeated

patricksnape:
Do we really want to keep the `_build` environment around? I can see maintaining the `work` directory but generally creating the `_build` environment is usually fairly quick? I can foresee the `_build` environment potentially becoming quite 'dirty' if say files are being installed into the wrong places.

conda_build/source.py
msarahan: hah!  That's a weird copy/paste.  Thanks for pointing it out.

msarahan:
Thanks @patricksnape.  I agree, and have integrated your suggestions.

PR #978, opened 2016-05-25
FEATURE_ environment variables

ilanschnell:
In this PR, I want to make it easier to create recipes with support features.  By setting `FEATURE_*` environment variables, the user can select whatever feature they need for building a recipe with and without that feature enabled.
For example, having `FEATURE_DEBUG=0` will add the `debug` variable (set to `False`) to the selector namespace, and set `DEBUG=0` in the environment of the build script.

It is also important to see that during building and testing, the feature is enabled (when set to `1`).

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -20,7 +20,7 @@ import mmap
 import conda.config as cc
 import conda.plan as plan
 from conda.api import get_index
-from conda.compat import PY3
+from conda.compat import PY3, iteritems
 from conda.fetch import fetch_index
 from conda.install import prefix_placeholder, linked, move_to_trash
 from conda.lock import Locked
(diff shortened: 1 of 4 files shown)


msarahan:
This is a good idea, much in line with #848.  I don't think environment variables are the best way to do this, though.  I'm going to ponder this, review the code, and try to incorporate it with my efforts to incorporate #848 ( #966 )

msarahan:
Thanks @ilanschnell - merging.

For those observing, we had a meeting where I asked Ilan not to generalize this PR, so that it did not conflict in spirit with the discussions from #848 et al.  This PR is necessary to push forward with our internal goal of making our legacy build recipes into conda recipes.  We anticipate that the syntax that this PR supports will eventually be supplanted by the final implementation of proposals from earlier discussion about customizing build pinning, features/variants, compilers, and the rest.

ilanschnell:
Thanks for merging!

PR #979, opened 2016-05-26
MSYS2: Add build paths

mingwandroid:
Please review @masarahan.

This mirrors what we do in conda when we activate enivornments.


msarahan:
LGTM.  Thanks @mingwandroid

PR #981, opened 2016-05-27
Only set os.environ for non-None config vars

mingwandroid:
Otherwise we end up stringifying None as 'None' and the
version parsing code falls over later (if you call conda
from a build script for example).

Specifically, for me, the rendered batch file contained:
"set CONDA_NPY=None"


msarahan:
LGTM.  Travis errors seem to be issues with our package repository being over capacity.  Merging.

PR #984, opened 2016-05-27
refine tests for default values for environ.get in jinja2

msarahan:
cc @stuarteberg

### DIFF
--- a/tests/test-recipes/metadata/_source_git_jinja2_relative_git_url/meta.yaml
+++ b/tests/test-recipes/metadata/_source_git_jinja2_relative_git_url/meta.yaml
@@ -1,0 +1,10 @@
+package:
+  name: flex
+  version: {{ environ.get('GIT_DESCRIBE_TAG', '0.0') }}
+
+build:
+  number: {{ environ.get('GIT_DESCRIBE_NUMBER', 0) }}
+  string: {{ environ.get('GIT_DESCRIBE_NUMBER', '0') + '_JPMC' }}
+
+source:
+  git_url: ../../../../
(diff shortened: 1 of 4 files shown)


stuarteberg:
Not sure if this matters per se, but it looks like this branch is based off of an out-dated commit of the master branch.  (The condition in `parse_or_try_download()` function is slightly different now, for instance.)  Might be worth rebasing just to avoid confusion, unless this is what you intended.

stuarteberg:
OK, I see the problem.  It's in your test.  You wrote:

```
version: {{ environ.get(GIT_DESCRIBE_TAG.replace('v', ''), "WRONG_ANSWER") }}
```

During the second pass, that will evaluate to something like this:

```
version: {{ environ.get('1.2.3'.replace('v', ''), "WRONG_ANSWER") }}
```

... and since '1.2.3' (or whatever) doesn't exist in `environ`, you'll always get `WRONG_ANSWER` as a result.


I think you meant to write this:

```
version: {{ environ.get("GIT_DESCRIBE_TAG", "WRONG_ANSWER").replace('v', '') }}
```

which will evaluate to this (assuming `"GIT_DESCRIBE_TAG"` exists in the environment.):

```
version: {{ '1.2.3'.replace('v', '') }}
```

msarahan:
Ah!  good catch on both counts.  Rebasing and changing the test.

msarahan:
@joelhullcio, I believe these tests exercise your recipe and do not show the error.  Can we please investigate which version of conda build you have installed?  I don't see any version output in the logs you sent me, so I think you're current, but I want to make certain.

msarahan:
Merging these tests - they're good to have, anyway.

PR #986, opened 2016-05-28
add  selectors for Intel platforms

ilanschnell:
In this PR, we add new selectors for `x86` and `x86_64` for Intel platforms.  Note that `x86` is `True` on either 32 or 64-bit.  This is consistent with the selectors currently in use in constructor.


ilanschnell:
@msarahan This is required for conversion of legacy recipes.

msarahan:
Thanks

PR #989, opened 2016-06-01
Always use the RequestsTransport to access pypi

### LINKED ISSUE #550: conda skeleton ignores the proxy environment variables
When an Enterprise client is operating from behind a proxy server, the conda skeleton command appears to ignore the availability of proxy settings in the $HTTP_PROXY, $HTTPS_PROXY and $NO_PROXY environment variables and fails connecting to pypi unless proxy settings are specified in the conda configuration file (observed with conda 3.16.0 on linux rhel6).

It would be really great if this behavior could be fixed and conda skeleton made use of the environment settings (as conda already does in most cases).

filmor:
This ensures that proxies from the environment are used even if they are not defined in the `.condarc`.

Fixes #550.

### DIFF
--- a/conda_build/pypi.py
+++ b/conda_build/pypi.py
@@ -22,7 +22,6 @@ import yaml
 
 from conda.cli.common import spec_from_line
 from conda.compat import input, configparser, StringIO, string_types, PY3
-from conda.config import get_proxy_servers
 from conda.connection import CondaSession
 from conda.fetch import (download, handle_proxy_407)
 from conda.install import rm_rf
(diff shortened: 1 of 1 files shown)


msarahan:
I'm not familiar with this code, nor really familiar with proxies.  Can you describe what problem led you to this change?  How does this affect proxy settings in condarc?  What would it take to add a test for the functionality you desire?

filmor:
Check issue #550. `conda skeleton pypi` (implemented in `pypi.py`) has code to use proxy servers (the mentioned `RequestsTransport`) correctly through the `CondaSession` object. However, this code is only invoked if the `.condarc` file contains the `proxies` entry, in particular, it skips the (otherwise working) `$HTTP_PROXY` etc. environment variables.

Regarding tests, I guess for those you would need a proxied environment on the CI servers, *not* configure the proxy server in `.condarc` but instead only in the environment and run `conda skeleton pypi some-package` there.

msarahan:
@filmor thanks for this.  I don't fully grok all of this, but it looks like a good improvement to me.  I'm going to merge it.  Is it OK to ping you if I run across issues/concerns?

PR #990, opened 2016-06-01
fix missing dirty args.  Cleanup.

msarahan:
This adds a test that currently fails on Windows, because Conda's move_to_trash isn't working correctly.


msarahan:
Remaining test failures should be resolved by next conda release.  They are related to trash not being cleaned up.  @mingwandroid fixed in https://github.com/conda/conda/pull/2587

Issue #910, opened 2016-05-02
Conda needs to activate _build and _test environments - not monkey with PATH


jakirkham:
When using the following recipe.

```
package:
  name: blas_openblas
  version: 1.0.0

build:
  number: 0
  track_features:
    - blas_openblas

test:
  commands:
    - conda list blas_openblas

about:
  home: NA
  license: NA
  summary: Metapackage to enable the use of OpenBLAS in an environment.

extra:
  recipe-maintainers:
    - jakirkham
```

I run the build, but it appears not to activate the `_test` environment during the testing phase, but remains in the `_build` environment. Here is a snippet that I see in the log.

```
TEST START: blas_openblas-1.0.0-0
Using Anaconda Cloud api site https://api.anaconda.org
Fetching package metadata: ..............
.Solving package specifications: .........

The following packages will be downloaded:

    package                    |            build
    ---------------------------|-----------------
    blas_openblas-1.0.0        |                0          657 B  file:///zopt/conda2/conda-bld/osx-64/

The following NEW packages will be INSTALLED:

    blas_openblas: 1.0.0-0 file:///zopt/conda2/conda-bld/osx-64/

Fetching packages ...
blas_openblas- 100% |#############################################################################################################| Time: 0:00:00 947.94 kB/s
Extracting packages ...
[      COMPLETE      ]|################################################################################################################################| 100%
Linking packages ...
[      COMPLETE      ]|################################################################################################################################| 100%
+ conda list blas_openblas
Error: Error: environment does not exist: /zopt/conda2/envs/_build
#
# Use 'conda create' to create an environment before listing its packages.
TESTS FAILED: blas_openblas-1.0.0-0
```

msarahan:
This is very related to your compiler configuration PR on conda-forge.  IMHO, we need conda-build actually activating environments.  That will both clean up code in conda-build, and make it more predictable.

jakirkham:
Also, this issue ( https://github.com/conda/conda-build/issues/355 ) seems to be related.

jakirkham:
Also, maybe this issue ( https://github.com/conda/conda-build/issues/353 ) too.

jakirkham:
Here is a use case for configuring compiler tools ( https://github.com/conda-forge/thrift-cpp-feedstock/pull/4 ).

jakirkham:
Just so I know (as there is a 2.0.0beta out), is this still planned for 2.0.0?

msarahan:
[...]

Issue #355, opened 2015-04-02
have the ENVs set by etc/conde/activate.d scripts available in build enviroment


tacaswell:
Some packages either need to have ENVs set to work properly or use ENVs to find their dependencies.

jakirkham:
Believe this should be addressed by PR ( https://github.com/conda/conda-build/pull/1002 ).

PR #1002, opened 2016-06-06
Activate envs

### LINKED ISSUE #364: Fully activate test environment prior to executing tests
For packages that rely on custom environment variables that are set as part of the activate / deactivate scripts you can't run tests and get reliable results without fully activating the environment.

msarahan:
Closes #364 #910 #355 

CC @jakirkham 

@kalefranz - this is very poor use of conda.  I'd really love to have an API to be doing this sort of thing.  Let me if/how I can help.


jakirkham:
Also, think this closes issue ( https://github.com/conda/conda-build/issues/355 ).

jakirkham:
cc @tacaswell

jakirkham:
Would this rely on PR ( https://github.com/conda/conda/pull/2365 ) to work correctly on Windows?

cc @janschulz

msarahan:
Yes, I think so.  I think it is working on my computer because I have the development versions of the activate scripts in place.  Those are one of the hardest things to be "clean" about.

jakirkham:
So true. That's one thing I like about docker containers. Easy to create a test and easy to toss afterwards without messing up your system. Still that doesn't really work if you want to test non-Linux (excepting Windows docker, but I have no experience with that).

jakirkham:
Though in the end I think this will be ok. It is just we need to ensure we have a corresponding `conda` release to make sure this all works when it gets out.

msarahan:
yeah... the situation with this is pretty dire on Windows until we get 4.1 out.  I will heckle @kalefranz and see if I can help him get it out.

jakirkham:
Well maybe not "heckle". 😆 Though yes let's see if we can get that change into a release.

msarahan:
Tagging 1.21.0 without this.  Will add this in when Conda 4.1 is released, and potentially try to roll it in, or do another release.

jakirkham:
Ok, that makes sense.

janssen:
Does this have anything to do with #1034?

msarahan:
Yes, something is going wrong with this for you. Does deactivate work by itself?

conda_build/environ.py
```diff
@@ -366,6 +376,70 @@ def system_vars(env_dict, prefix):
...
+def activate_or_deactivate_env(action, env_name_or_path=""):
+    """
+    Strategy is to open a subprocess, run activate, record the variables,
+    then apply them in our process
+
+    action should be "activate" or "deactivate"
+    env_name_or_path should only be provided for activation.
+    """
+    if sys.platform == "win32":
+        cmd = '"{0}\\Scripts\\{1}.bat" "{2}" && set'.format(sys.prefix,
```
janschulz: This might be a little bit more robust to wrap the  `set` into an `echo unique_string`, so that output from the conda command isn't added to the variables (this currently seems to work because the activate call only prints to stderr)

E.g.:

` ... && echo __CONDA_BUILD_SEPARATOR___ && set && echo __CONDA_BUILD_SEPARATOR___`

and then only take the stuff in between the  `__CONDA_BUILD_SEPARATOR___`.

PR #1006, opened 2016-06-07
patch: Switch to using Python code

mingwandroid:
@msarahan, @jakirkham what do you think?

`patch.exe` from gnuwin32 has fallen over on several of my bigger
patches now and it's becoming a serious impediment.

I don't want to add MSYS2's `patch.exe` since that requires a lot
of infrastructure to work correctly (though I could build a native
Windows `patch.exe` if necessary).

However rather than relying on some arbitrary `patch` executable,
it's better to use Python to do the work and avoid the subprocess
overhead (though there will be a native-vs-python difference).

Also add automatic patch skip level detection logic, which allows
us to directly use patches generated using `git format-patch` and
also those authored by external entities without needing to add
metadata (although this may not be a bad idea). This part could be
moved into patch.py at a later point perhaps?

I've not added tests for this since upstream has a comprehensive
testsuite (and I did add tests to check my changes to my PR).

Upstream: https://github.com/techtonik/python-patch
Pull Req: https://github.com/techtonik/python-patch/pull/41

### DIFF
--- a/conda_build/patch.py
+++ b/conda_build/patch.py
@@ -1,0 +1,1242 @@
+#!/usr/bin/env python
+"""
+    Patch utility to apply unified diffs
+
+    Brute-force line-by-line non-recursive parsing 
+
+    Copyright (c) 2008-2016 anatoly techtonik
+    Available under the terms of MIT license
+
+    https://github.com/techtonik/python-patch/
+
+"""
+from __future__ import print_function
+
+__author__ = "anatoly techtonik <techtonik@gmail.com>"
+__version__ = "1.16"
+
+import copy
+import logging
+import re
+import tempfile
+
+# cStringIO doesn't support unicode in 2.5
+try:
+  from StringIO import StringIO
+except ImportError:
+  from io import BytesIO as StringIO # python 3
+try:
+  import urllib2 as urllib_request
+except ImportError:
+  import urllib.request as urllib_request
+
+from os.path import exists, isfile, abspath
+import os
+import posixpath
+import shutil
+import sys
+
+
+PY3K = sys.version_info >= (3, 0)
+
+# PEP 3114
+if not PY3K:
+  compat_next = lambda gen: gen.next()
+else:
+  compat_next = lambda gen: gen.__next__()
+
+def tostr(b):
+  """ Python 3 bytes encoder. Used to print filename in
+      diffstat output. Assumes that filenames are in utf-8.
+  """
+  if not PY3K:
+    return b
+
+  # [ ] figure out how to print non-utf-8 filenames without
+  #     information loss
+  return b.decode('utf-8')    
+
+
+#------------------------------------------------
+# Logging is controlled by logger named after the
+# module name (e.g. 'patch' for patch.py module)
+
+logger = logging.getLogger(__name__)
+
+debug = logger.debug
+info = logger.info
+warning = logger.warning
+
+class NullHandler(logging.Handler):
+  """ Copied from Python 2.7 to avoid getting
+      `No handlers could be found for logger "patch"`
+      http://bugs.python.org/issue16539
+  """
+  def handle(self, record):
+    pass
+  def emit(self, record):
+    pass
+  def createLock(self):
+    self.lock = None
+
+streamhandler = logging.StreamHandler()
+
+# initialize logger itself
+logger.addHandler(NullHandler())
+
+debugmode = False
+
+def setdebug():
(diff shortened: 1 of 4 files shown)


jakirkham:
Moving to Python sounds great to me. There were lots of annoyance to using `patch` on Unix platforms. We just need to make sure that this code is really well tested. I was looking earlier to see if there were any decent libraries for patching. Unfortunately there are only refs to `difflib`, which doesn't cut it. Other than that there is [python-patch]( https://github.com/techtonik/python-patch ), which is actually Python and C++ so that's not great for adding to `conda-build`, but it could be useful for inspiration and tests. There's also [whatthepatch]( https://github.com/cscorley/whatthepatch ), but it is still very young. Though it does have a decent number of features, quite a few test cases, and is pure Python. In any event, both of those are MIT so that should aid in reusability.

mingwandroid:
@jakirkham https://github.com/techtonik/python-patch doesn't use any C++ and (as the PR states) is the upstream for this :-)

It also has quite a comprehensive (if somewhat fiddly to add to) testsuite. But I agree that it needs testing and more testing, so please, go to town.

ccordoba12:
@mingwandroid, we were `python-patch` before moving to `patch.exe`. If I recall correctly, @pelson (among others) asked for this change. So maybe he has something to say about it :-)

jakirkham:
> doesn't use any C++ and (as the PR states) is the upstream for this :-)

I see. I didn't look too closely. Maybe GitHub was just pointing out the sample code for patch tests.

msarahan:
I don't remember anyone asking to use patch.exe.  IIRC, I did so because the old one was failing miserably in my efforts to add python 3.5 and all its packages to Anaconda.

msarahan:
I'm OK with this addition, but I would like some tests to accompany it.  What I have in mind is a project with some dummy source code, and a patch that applies to that source code.  We need to parse or examine the patched output to ensure that the patch applied cleanly.

I guess we need patches both with p0 and p1 (without and with prefix, respectively), just for completeness?

mingwandroid:
For the first one, can I copy a test from upstream? They have extensive tests for this sort of thing.

The patch strip level detection code needs a new test for sure (actually it should be able to detect -p<arbitrary-value> too).

.. but first, I need to figure out what's up with the CI on this.

msarahan:
Copying is A-OK, but any testing here should cover our use of patch.py, not
just the functionality of patch.py.  Integration tests, of sorts, more than
unit-tests of the patch functionality.

ilanschnell:
Is the Python patch only for Windows?  Also, I would argue that if we move to a different patch program, we should only change the dependency, and package `patch` separately.  I think `conda-build` should not be tied to a specific patch implementation.

mingwandroid:
@ilanschnell python-patch is for all systems, it is pure python and it works on 2.7.x and 3.x.

I've split this PR two bits, one that I hope will be merged soon (automatic detection of patch strip level): https://github.com/conda/conda-build/pull/1011, the other bit I don't plan to push for inclusion at present (see later).

IMHO the patch executables in the wild are too variable. For example, gnuwin32's patch.exe crashes on completely valid patches. There are pros and cons to using Python implementations of the tools we depend on: not needing to use subprocess, consistency of operation, python's speed versus some-other-language's speed, maintainability, amount of code, and other things I've missed.

Therefore, the decision of whether to import a python implementation of a given tool needs to be done on a case-by-case basis and I think a (working) python-patch would be a good thing (provided the code isn't too large, which in this case, I don't think it is). Using different patch implementations *could* generate different source code and that could result in horrible to track down bugs. The other major issue with gnu patch executables is that they are always interactive on patch failure which breaks CI systems. It's interesting you mention `install_name_tool` since broken or incompatible versions of that tool appear to be hurting us on OS X.

That being said, python-patch can't handle offsets or fuzzing, both of which are features we need, so for that reason I am closing this request, but I'd be interested in carrying on the conversation about whether continuing on this path (by implementing offsets and fuzzing) in future is worthwhile or not.

ilanschnell:
Thanks for your response, Ray.  I've never had any problems with `patch` on Unix.  It sounds like all your concerns could be equally addressed by creating a `python-patch` package, and have conda depend on that.  I wold be in favor of this approach.

mingwandroid:
That approach would also work for me, thanks Ilan.

msarahan:
@mingwandroid, ok to close this?

jezdez:
Closing since it's been a while.

PR #1016, opened 2016-06-10
chdir to recipe dir before using git

fschlimb:
It checks for .git in the recipe dir, not in cwd, so we have to chdir there first.
This might need fixing for hg, svn etc. E.g. by chdir first and then check for the .git, .hg in '.'.
@ilanschnell @msarahan

### DIFF
--- a/conda_build/source.py
+++ b/conda_build/source.py
@@ -286,8 +286,11 @@ def get_repository_info(recipe_path):
     """This tries to get information about where a recipe came from.  This is different
     from the source - you can have a recipe in svn that gets source via git."""
     if isdir(join(recipe_path, ".git")):
+        owd = os.getcwd()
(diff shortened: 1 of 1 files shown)


msarahan:
Thanks @fschlimb.  I think this is unnecessary, though.  The cwd argument to check_output dictates what directory that subprocess call occurs in.  We should already be executing the git commands in ```recipe_path```: https://github.com/conda/conda-build/pull/1016/files#diff-9fdc13ead8018ee54de2ac73289ec2e1R291

Is something about that not working?  Can I help you debug it?

fschlimb:
Ooops, I did not look closely enough. The version I use (conda-build 1.20.3) does not have the cwd arg. I didn't see the version here has it....
Sorry. I close this then.

PR #1020, opened 2016-06-13
patch: strip-level guess includes target files

mingwandroid:
As often you'll get a source file with .old appended
and the target file is the only one that exists.

Without this, you get a warning and a patch strip level
of 0 will be used in some common cases.

Also, I overthought the common prefix stuff and it
is just wrong.


msarahan:
Thanks, @mingwandroid

PR #1030, opened 2016-06-15
add force option when installing packages in test env

msarahan:
to force overwrite in pkgs

CC @ilanschnell


msarahan:
Status: this wreaks havoc on Windows, causing all kinds of permissions errors when we try to replace/reinstall things like the VS runtime.  Closing this PR for now.

PR #1036, opened 2016-06-17
add --no-activate option; simplify msvc activation

msarahan:
Without this, conda build was dropping the active env that the user may have had.  This still activates and deactivates _build and _test, but ultimately in between those steps, the originally active env is maintained.

CC @ilanschnell


patricksnape:
Is this coupled with packages that activate the correct compiler?

msarahan:
So locally, distutils worked wonderfully.  I didn't test well enough on appveyor.  I'm going back to what you had, which I see as a partial step towards what I did here.  

for distutils (and probably setuptools - not sure if there's additional code there, or if it just uses distutils), query_vcvarsall returns a dictionary with the relevant environment variables.  No activation is necessary - you just have to merge that dictionary in with the rest of our environment.

I'm going to look into patching distutils or setuptools to add support for the win 7 sdk compiler once we have this figured out.

msarahan:
This was merged into 1.21.x manually, and has now found its way to master via #1057.  Closing.

PR #1045, opened 2016-06-23
revert MSVC disutils changes

msarahan:


msarahan:
rolled into #1046 with fd6b11a.  Too hard to fix these bugs independently.

PR #1046, opened 2016-06-24
Revert MSVC to use vcvarsall; improve detection of needed VCS install

msarahan:
This addresses several small testing tweaks, especially on Windows.  It adds better support for knowing when conda needs to install some VCS in order to build a recipe (since activation may take other VCS off PATH).

This reverts MSVC activation to running the vcvarsall.bat scripts, and incorporates @patricksnape's updates to use distutils/setuptools to locate the scripts (originally #958)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ install:
 
 script:
   - flake8 .
-  - py.test --cov conda_build --cov-report xml tests
+  - py.test -v --cov conda_build --cov-report xml tests
   - conda build --help
 
 notifications:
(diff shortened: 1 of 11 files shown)


msarahan:
Windows test for environment activation is failing due to a bug that https://github.com/conda/conda/pull/2856 addresses.

Linux is failing due to an as-yet unidentified bug with cleaning up environments.  The failing tests do not intentionally create a __conda_version__.txt file, but they somehow have one when the test suite is run all at once.  Each of these tests pass when run individually locally.  Will investigate this, but it should not hold up release.

PR #1067, opened 2016-06-29
centralize folders to config.  Make new build folder per build.

### LINKED ISSUE #1062: Need ability to specify the build output directory as a command line parameter or env var
On shared systems where multiple users may be performing builds from the same root Anaconda installation it is desirable to be able to specify the build output directory on a per build basis.

Add a command line parameter and/or and env var where the output directory can be specified to override the default output location.

This is an immediate need at the JPM client site.

msarahan:
Closes #1062 

Allows, but does not implement parallelism in building.  You can, for example, run the test suite in parallel with pytest-xdist.

It needs to:

- [x] provide per-build environments and work/test folders
- [x] Rework the --dirty option to allow continuation of old builds
- [x] Add a cleanup method, as the new individual folders should not be deleted by the next build (because builds may happen concurrently)


conda_build/source.py
```diff
@@ -376,20 +371,14 @@ def apply_patch(src_dir, path):
         os.remove(patch_args[-1])  # clean up .patch_unix file
 
 
-def provide(recipe_dir, meta, verbose=False, patch=True):
+def provide(recipe_dir, meta, verbose=False, patch=True, build_id=None):
```
msarahan: build_id should go

msarahan:
Superseded by #1103 - closing.

Issue #1074, opened 2016-06-30
conda-build crashes when building PEP420 namespace packages


ikalev:
There is a very trivial bug in ``conda_build.utils.copy_into()``, causing it to crash when building PEP420 namespace packages. To reproduce the issue, all of the following conditions must be met:

 - your package must be a native PEP420 namespace package (``python >=3.3``)
 - your package must depend on (and therefore install) other packages mapping to the same namespace
 - you must build all of these packages with ``recipe['build']['preserve_egg_dir'] = False``
 - your namespace must be a multi-level directory 

This is easier to explain by example:

```
found egg dir: C:\ANACONDA\envs\_build\Lib\site-packages\example-4.3.0-py3.5-win-amd64.egg
...
Traceback (most recent call last):
  File "C:\ANACONDA\Scripts\conda-build-script.py", line 4, in <module>
    sys.exit(main())
  File "C:\ANACONDA\lib\site-packages\conda_build\main_build.py", line 144, in main
    args_func(args, p)
  File "C:\ANACONDA\lib\site-packages\conda_build\main_build.py", line 389, in args_func
    args.func(args, p)
  File "C:\ANACONDA\lib\site-packages\conda_build\main_build.py", line 332, in execute
    dirty=args.dirty, activate=args.activate)
  File "C:\ANACONDA\lib\site-packages\conda_build\build.py", line 606, in build
    preserve_egg_dir=bool(m.get_value('build/preserve_egg_dir')))
  File "C:\ANACONDA\lib\site-packages\conda_build\post.py", line 162, in post_process
    remove_easy_install_pth(files, preserve_egg_dir=preserve_egg_dir)
  File "C:\ANACONDA\lib\site-packages\conda_build\post.py", line 119, in remove_easy_install_pth
    utils.copy_into(join(egg_path, fn), join(sp_dir, fn))
  File "C:\ANACONDA\lib\site-packages\conda_build\utils.py", line 54, in copy_into
    shutil.copytree(srcname, dstname)
  File "C:\ANACONDA\lib\shutil.py", line 177, in copytree
    os.makedirs(dst)
  File "C:\ANACONDA\lib\os.py", line 157, in makedirs
    mkdir(name, mode)
WindowsError: [Error 183] Cannot create a file when that file already exists: 'C:\ANACONDA\envs\_build\Lib\site-packages\ns\sub'
```

Here ``ns`` is the top level namespace directory and ``sub`` is a sub-package, but also a namespace package:

```
+ ns
[...]

Issue #649, opened 2015-10-26
Build fails if using namespace packages


remram44:
I have two libs that share a namespace package, and one of them depends on the other.

When building the one that has the dependency, the build fails with "`OSError: File exists: <path to the namespace directory>`". It looks like you are not handling the fact that the directory will already have been created while installing the first package.

```
Traceback (most recent call last):
  File "/Users/remirampin/anaconda/bin/conda-build", line 5, in <module>
    sys.exit(main())
  File "/Users/remirampin/anaconda/lib/python2.7/site-packages/conda_build/main_build.py", line 190, in main
    args_func(args, p)
  File "/Users/remirampin/anaconda/lib/python2.7/site-packages/conda_build/main_build.py", line 468, in args_func
    args.func(args, p)
  File "/Users/remirampin/anaconda/lib/python2.7/site-packages/conda_build/main_build.py", line 391, in execute
    override_channels=args.override_channels, include_recipe=args.include_recipe)
  File "/Users/remirampin/anaconda/lib/python2.7/site-packages/conda_build/build.py", line 455, in build
    post_process(sorted(files2 - files1), preserve_egg_dir=bool(m.get_value('build/preserve_egg_dir')))
  File "/Users/remirampin/anaconda/lib/python2.7/site-packages/conda_build/post.py", line 148, in post_process
    remove_easy_install_pth(files, preserve_egg_dir=preserve_egg_dir)
  File "/Users/remirampin/anaconda/lib/python2.7/site-packages/conda_build/post.py", line 105, in remove_easy_install_pth
    utils.copy_into(join(egg_path, fn), join(sp_dir, fn))
  File "/Users/remirampin/anaconda/lib/python2.7/site-packages/conda_build/utils.py", line 34, in copy_into
    shutil.copytree(srcname, dstname)
  File "/Users/remirampin/anaconda/lib/python2.7/shutil.py", line 177, in copytree
    os.makedirs(dst)
  File "/Users/remirampin/anaconda/lib/python2.7/os.py", line 157, in makedirs
    mkdir(name, mode)
OSError: [Errno 17] File exists: '/Users/remirampin/anaconda/envs/_build/lib/python2.7/site-packages/reprounzip/unpackers'
```

remram44:
Bump. Please look into this, still happening, super annoying.

kalefranz:
The python community has abandoned namespace packages. Because getting them to work is *super annoying*.

[...]

PR #1090, opened 2016-07-04
fix PEP420 namespace packages 

ikalev:
See #1074  for details.

### DIFF
--- a/conda_build/utils.py
+++ b/conda_build/utils.py
@@ -10,6 +10,7 @@ import subprocess
 import operator
 from os.path import dirname, getmtime, getsize, isdir, join
 from collections import defaultdict
+from distutils.dir_util import copy_tree
 
 from conda.utils import md5_file, unix_path_to_win
 from conda.compat import PY3, iteritems
@@ -51,7 +52,7 @@ def copy_into(src, dst):
         dstname = os.path.join(dst, afile)
 
         if os.path.isdir(srcname):
-            shutil.copytree(srcname, dstname)
+            copy_tree(srcname, dstname)
         else:
             shutil.copy2(srcname, dstname)
 
--- a/conda_build/utils.py
+++ b/conda_build/utils.py
@@ -8,7 +8,7 @@ import tarfile
 import zipfile
 import subprocess
 import operator
-from os.path import dirname, getmtime, getsize, isdir, join
+from os.path import dirname, getmtime, getsize, isdir, isfile, join
 from collections import defaultdict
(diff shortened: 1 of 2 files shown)


msarahan:
@ikalev sorry I forgot earlier, this has been proposed once before: https://github.com/conda/conda-build/pull/859

There, I suggested one change: throw an exception if copying would overwrite any file(s).  I think that this is very important for sanity - otherwise it is ill-defined what the contents of your files are.  Any ideas on that point?

ikalev:
@msarahan It's a bit of a grey zone: If we are merging namespace packages some overwriting might happen and that might be okay in edge cases... But otherwise it would be an error indeed.

I think it's reasonable to ignore said edge cases and assume that this is an error. I'll try to extend the implementation with an explicit check. Should be easy.

ikalev:
@msarahan should be good to merge now.

msarahan:
Looks great.  Minor PEP8 errors and existing errors on Travis.  Will wait for AppVeyor, then merge.

msarahan:
(I am happy to fix PEP8 - you don't need another push.)

msarahan:
Thanks again, @ikalev

ikalev:
Great, thanks!

183amir:
Thank you this is awesome. Sorry I was busy and could not find a time to write proper test cases for my pull request.

ikalev:
Note that this does not solve all cases of #649 . Python 2 namespace packages (declared with setuptools or distutils) contain ``__init__.py`` files, which ``copy_into()`` might attempt to overwrite. This will obviously not pass our test for merge conflicts. The workaround on Python 2 is to use ``pkg_resources. declare_namespace()`` and enable ``preserve_egg_dir``, which hides shared namespace files in private egg folders.

ikalev:
@msarahan I don't see these changes in the source tree at the moment. I deleted my fork after the merge, but this should be fine, right? I'm confused.

msarahan:
@ikalev could you please resubmit this?  I don't know how, but I don't see it in the git history.  Nor do I see any force pushes that I would have overwritten it with...

ikalev:
@msarahan Unfortunately I have deleted my fork of the repo, but I can try to extract a patch from this diff and prepare a new PR: https://github.com/conda/conda-build/pull/1090/files

msarahan:
Sorry for the mess.  I can submit this PR for you, if you'd prefer.

ikalev:
No problem. Here is the new PR: #1105.

PR #1103, opened 2016-07-07
centralize folders to config. Make new build folder per build.

### LINKED ISSUE #1062: Need ability to specify the build output directory as a command line parameter or env var

msarahan:
Closes #1062 
Supersedes #1067 - to apply to master.

Allows, but does not implement parallelism in building.  You can, for example, run the test suite in parallel with pytest-xdist.

It needs to:

- [x] provide per-build environments and work/test folders
- [x] Rework the --dirty option to allow continuation of old builds
- [x] Add a cleanup method, as the new individual folders should not be deleted by the next build (because builds may happen concurrently)


kalefranz:
Any reason this set of appveyor tests are taking 45+ minutes per?

kalefranz:
Cancelled the appveyor test run for 4f263d7. The first test was at 49 minutes and the second at 45 minutes when I cancelled it.  The remaining in the grid hadn't started yet.

msarahan:
I don't know why, but I'll look into it.  A lot of it has to do with https://github.com/conda/conda/issues/2672 - but still, these things have been taking ~15 min per configuration.

kalefranz:
ahhhh 

Definitely understand the importance of conda/conda#2672.  But am also extra sensitive to putting anything but *strict* bug fixes in both 4.0 and 4.1 at this point.  4.2 canary out Thursday.

msarahan:
Oh, I'm not proposing to change anything.  Just commenting on why things might be slow.

msarahan:
The more I think about it, I'm guessing this is due to lock problems with conda.  I'm going to pause work on this PR until you have a canary 4.2 candidate.

msarahan:
Closing here.  This has been merged into #953, and will be broken out before merge.

Issue #629, opened 2015-10-08
meta.yaml selectors are not processed correctly


ikalev:
Selectors in ``meta.yaml`` work fine as long as the yaml lines containing them are not enclosed in single or double quotes:

```yaml
 - pandas >=1.0   # [win]
```

However, "#" is a special character in yaml, which means that the entire value should be enclosed in quotes. As a result, yaml recipes generated with ``yaml.dump()`` look like this:

```yaml
 - 'pandas >=1.0   # [win]'
 - "pandas >=1.0   # [win]"
```

When processing the selector, ``conda build`` strips out the selector itself (`# [win]`), but leaves the opening single/double quote:

```yaml
 - 'pandas >=1.0
```

which is no longer a valid yaml file and ``conda build`` fails.

stuarteberg:
I think this is just a reflection of the fact that conda's defined syntax for `meta.yaml` is not, in fact, valid YAML. Therefore, we can't expect the raw output of `yaml.dump()` to necessarily valid input for the `meta.yaml` parser.

Perhaps, if they were to do it all over again, the conda designers would have named the file `meta.yaml.in` to reflect the fact that some preprocessing will be performed (selectors, jinja).

Anyway, if you're programmatically producing recipes, I think you'll need to come up with a workaround for this issue.  Something like:

```python
yaml_text = yaml.dump({"requirements" : {"run" : ["pandas <HASH> [win]"]}}, default_flow_style=False)
print yaml_text.replace("<HASH>", "#")
```
```yaml
requirements:
  run:
  - pandas # [win]
```

janssen:
One simple solution would be to change the file extension to "notyaml".

ikalev:
We need to fix the parser. It's a trivial but. I'll do some experiments and submit a PR.

ikalev:
Fixed in #1104

PR #1104, opened 2016-07-07
Handle quotes when matching selectors 

ikalev:
This fixes an edge case in preprocessing selectors (#629), where quoted lines contain imbalanced quotes after removing the selector part. Generated yaml files will typically enclose in single quotes all entries containing ``#`` characters, so this PR adds support for generated recipes.

### DIFF
--- a/conda_build/metadata.py
+++ b/conda_build/metadata.py
@@ -89,8 +89,14 @@ sel_pat = re.compile(r'(.+?)\s*(#.*)?\[([^\[\]]+)\](?(2).*)$')
 
 def select_lines(data, namespace):
     lines = []
+
     for i, line in enumerate(data.splitlines()):
-        line = line.strip(' \'"')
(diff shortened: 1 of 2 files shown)


msarahan:
Looks great.  I'm not sure there is going to be another 1.21 release though.  I'm hoping for a 2.0.0 preview release next week - do you mind resubmitting this PR against master?

ikalev:
@msarahan No problem, will do. 

But I noticed that #1090 didn't make it into the 1.21.5 release. If you look at the source code of the release, my patch is just not there. Can't see it in any repo branch either. Should I redubmit this PR too? Against master?

msarahan:
That was a git SNAFU on my part.  Sorry.  I'm going to just merge this here and do another conda-build release with #1090.

ikalev:
Great, let me know if I need to do anything.

PR #1105, opened 2016-07-08
fix PEP420 namespace packages

ikalev:
(resubmitting #1090)


msarahan:
Thanks.  To be clear to the record, this is being resubmitted because I lost the earlier commits with some git mistake.  Merging this here.  Thanks @ikalev.

PR #1107, opened 2016-07-08
make activate script file extensions explicit

msarahan:
Seems to be an issue where in some cases the bash scripts are being called in windows:

https://github.com/conda-forge/ca-certificates-feedstock/pull/4#issuecomment-231399300

This makes the file extension explicit, and should clear up confusion.

CC @jakirkham

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -746,8 +746,9 @@ def test(m, move_broken=True, activate=True):
 
         with open(test_script, 'w') as tf:
             if activate:
-                tf.write("{source}activate _test\n".format(source="call " if on_win
-                                                           else "source "))
+                source = "call " if on_win else "source "
+                ext = ".bat" if on_win else ""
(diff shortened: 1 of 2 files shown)


jakirkham:
I can't see how this would hurt.

Though I really still don't know how the bash scripts got in there in the first place. I still feel like I must be doing something wrong, but can't figure out what it is.

msarahan:
The bash scripts are supposed to be on Windows, from conda.  Bash (msys2 or otherwise) is supported with the 4.1 series onward.

jakirkham:
I see. So there is some activation script floating around on the `conda-build` end that is getting run. Sorry, I guess I'm being slow today.

jakirkham:
Is this planned for a release?

msarahan:
Yes, probably tonight or tomorrow

jakirkham:
Thanks for the info. Though no need to rush. Just was curious where we stood on it.

PR #1119, opened 2016-07-15
expose CFLAGS and friends if defined in os.environ

msarahan:
This allows easier pass-through of these, until the build customization provides a better way.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,3 +1,22 @@
+2016-07-14 1.21.6:
+------------------
+
+Bug fixes:
+----------
+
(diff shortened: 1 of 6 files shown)


fschlimb:
:+1:

msarahan:
Windows is pretty terrible here.  It always includes the parent shell's environment, no matter how we try to limit its scope.  If we try to restrict it to only conda-builds known subset, things go really strange, with the activate scripts failing to run.  If we don't do that, then we capture all variables, including any potentially sensitive token data that the user may have set.

I'm leaving this PR out of the next release, but intend to address this issue soon.

msarahan:
There are much better ways to do this with #1585.  Closing.

PR #1128, opened 2016-07-20
Don't assume `token` or `user` args on the command line

necaris:
If `token` or `user` args aren't passed to `conda build`, then the `handle_binstar_upload` function throws an AttributeError trying to find them. This ensures that they are treated as `None` if not found.

_Note_: It's not clear to me where the `binstar_upload` argument the function is also expecting comes from -- does this need to be renamed to `anaconda_upload`?

### DIFF
--- a/conda_build/main_build.py
+++ b/conda_build/main_build.py
@@ -149,16 +149,16 @@ def handle_binstar_upload(path, args):
     from conda_build.external import find_executable
 
     upload = False
-    if args.token or args.user:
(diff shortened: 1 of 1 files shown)


msarahan:
As discussed on Flowdock, this issue is already fixed in #953.  I'm closing this issue, but will work to get #953 out as soon as I can.

Thanks @necaris for your contribution to resolving this issue.

Issue #865, opened 2016-04-04
Mac relative paths to dynamic libraries not converted to rpaths 


jakirkham:
If a library links to another library, but does not contain the full prefix and instead includes some relative path, then this path will not be corrected to an rpath. Instead it remains a relative path. This causes issues when trying to load dynamic libraries. Here is an [example](https://travis-ci.org/conda-forge/staged-recipes/jobs/120711298#L2731-L2735) of this behavior.

patricksnape:
One adhoc way to fix this would be to change the logic that if `install_name_tool` finds either `@loader_path` or a relative path then it tries two things:

  1) Resolve the path and see if the library exists - if so then change it
  2) Just try and see if the library exists in the `$PREFIX/lib` directory and if it does then change it

jakirkham:
It seems like 1 is a good choice. With 2, there are some situations where that might break.

PR #1132, opened 2016-07-21
post: Use os.path.normpath() in find_lib()

mingwandroid:
dylibs can end up with path names such as:
'/Users/ray/qt5-x64-2.7/envs/_build/gcc/../lib/libstdc++.6.dylib'
.. and, after removing the build prefix we are left with:
'gcc/../lib/libstdc++.6.dylib'
find_lib will fail to find that in a list that contains:
'lib/libstdc++.6.dylib' because it compares strings, so we
must normalize the path first.

Rebuilding GCC 4.8.5 on OS X ran afoul of this.

### DIFF
--- a/conda_build/post.py
+++ b/conda_build/post.py
@@ -7,7 +7,7 @@ import sys
 import stat
 from glob import glob
 from os.path import (basename, dirname, join, splitext, isdir, isfile, exists,
-                     islink, realpath, relpath)
+                     islink, realpath, relpath, normpath)
 try:
     from os import readlink
 except ImportError:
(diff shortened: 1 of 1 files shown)


msarahan:
LGTM.  Is this worth adding a test for?  Perhaps something is os.path.isabs?

mingwandroid:
I guess I would need to link something to a shared library which introduces some toolchain dependencies and I didn't really want to go there, and I figured it was minor enough to not bother adding a test for.

I did check that normpath when presented with '../../blah' doesn't except, and just returns the input value instead.

Famous last words :-)

msarahan:
OK.  Merging.

jakirkham:
Would this have any relation to and/or solve issue ( https://github.com/conda/conda-build/issues/865 )?

mingwandroid:
Well, my code is in the OS X post code path, and won't make any difference on any other platforms, but perhaps they share a common cause and could share a similar solution?

Issue #789, opened 2016-02-21
Compiling pyc files without python dependency fails with stacktrace


danielfrg:
This is a really annoying bug and the exit from `conda-build` is far from ideal.

If you are building a python package and don't have `python` on the dependencies in the `meta.yaml` you are going to get a stacktrace error because `conda-build` will try to compile the `.py` files to `.pyc` files: https://github.com/conda/conda-build/blob/master/conda_build/post.py#L130-L144

This will trigger a subprocess at https://github.com/conda/conda-build/blob/master/conda_build/utils.py#L54-L58 and it will just fail because `python` is not on the `_build` environment.

I am ok with all this issues, I am ok with `conda-build` trying to execute a python command when python is not a dependency and I am even ok with the stacktrace but please just print the command that `conda-build` is executing in [_check_call](https://github.com/conda/conda-build/blob/master/conda_build/utils.py#L54) so its easier to reproduce and see that `python` was a missing dependency on the `meta.yaml`.

kalefranz:
Would be happy to accept a PR with your proposal :-p

PR #1146, opened 2016-07-26
Don't exit on compileall failure

minrk:
Packages may include sources that are only valid and used on certain Python versions, in which case compileall will fail, but installation should still succeed.

### DIFF
--- a/conda_build/post.py
+++ b/conda_build/post.py
@@ -153,7 +153,7 @@ def compile_missing_pyc():
                 break
     if need_compile:
         print('compiling .pyc files...')
-        utils._check_call([config.build_python, '-Wi',
+        call([config.build_python, '-Wi',
                            join(stdlib_dir, 'compileall.py'),
                            '-q', '-x', 'port_v3', sp_dir])
 


msarahan:
This is a good way to do this, IMHO.  Other people have discussed this at #789 #686 #715 

My concerns here are:

 * does this approach simply stop compiling at the first failure?
 * what are the implications for usability on systems where the person using the package is not the person installing the package (and may not have write authority to write the .pyc)?

For the latter, it is my understanding that there will be a speed hit in byte compiling the module each time, but it will not fundamentally break anything.

If we want to keep .pyc files as part of conda packages (and @ilanschnell and @kalefranz have argued for this strongly internally), then I think a better approach to files that are known to not be compilable is to list them in meta.yaml, where selectors can be used, and then hide them somehow from the compileall call.

minrk:
> does this approach simply stop compiling at the first failure?

Compilation still hits every file.

> what are the implications for usability on systems where the person using the package is not the person installing the package (and may not have write authority to write the .pyc)?

None, that I'm aware of. In the cases I've seen, these are files that won't import on the Python versions where compilation fails (e.g. safely ignored Python 3-specific files installed on Python 2), so these files will never be imported.

ocefpaf:
Thanks for this @minrk! I maintain a certain number of packages where I need to patch them (removing the py3k parts) to be able to build them. My approach is not sustainable and does not scale. I would love to start using this instead!

msarahan:
OK, sounds good. Do you mind implementing a test to ensure that all files are compiled? Something like 2-3 files named such that the non compileable one is in the middle? 

Alternatively, if there are docs that specify the behavior, please add a comment to this pr pointed at those docs.

msarahan:
@minrk, I'm hoping to tag a bugfix release later today.  Do you want to get this in, or wait for 2.0 (in a couple of weeks)?

msarahan:
Merging. This is important enough and you are important enough that I'll finish this up with a separate pr. Thanks @minrk

PR #1147, opened 2016-07-27
Add filename argument with default to load_npm.py

srossross:


msarahan:
LGTM.  Could you please clarify what folder this looks in, by default - whether recipe dir, source dir, or build prefix dir?

srossross:
I just checked and I think this just runs from the curdir .... :( 

This works for me, but is not portable. what do you recommend? I would want my recipe relative to 
the `recipe_dir`

srossross:
>  I would want my recipe relative to 
the recipe_dir

* I think the default should be `source_dir` , maybe like his?:

```
def load_npm(package_fn=None):
    if package_fn is None:
        package_fn = os.path.join(get_environ()['SRC_DIR'], 'package.json')
   ...
```

this would need tests

msarahan:
SRC_DIR seems pretty reasonable.  Are you going to add tests, or you want me to?  I have not done anything with NPM, and would appreciate some spec/guidance on what info you want to ensure is available from this process.

srossross:
@msarahan can you point me to where I can add a test?

msarahan:
I think copying https://github.com/conda/conda-build/tree/master/tests/test-recipes/metadata/jinja_load_setuptools is a good start

msarahan:
Note that folders not starting with _ are auto-discovered and built as test cases - so all you need to do to add a test is create a recipe that should work under your desired conditions.

msarahan:
@srossross, I'm hoping to tag a bugfix release later today.  Do you want to try to get this in, or wait for 2.0 (in a couple of weeks)?

srossross:
Thanks for the reminder. I'll try and get the tests in today, but don't wait for me.

msarahan:
Looks good. Does windows matter? 

One small improvement would be checking the actual version in the built package as part of a run_test.sh script. 

Flake8 is complaining. That's not a strict barrier to merge, but would be nice to fix.

PR #1148, opened 2016-07-27
ensure need_source_download exists

daler:
This is a quick fix to an issue where 

`conda build <recipe> --dirty`

would raise:

```
File "/conda-build/conda_build/render.py", line 113, in parse_or_try_download
    return metadata, need_source_download, need_reparse_in_env
UnboundLocalError: local variable 'need_source_download' referenced before assignment
```

### DIFF
--- a/conda_build/render.py
+++ b/conda_build/render.py
@@ -81,6 +81,9 @@ def parse_or_try_download(metadata, no_download_source, verbose,
     need_reparse_in_env = False
     if (force_download or (not no_download_source and (metadata.uses_vcs_in_meta() or
                                                        metadata.uses_setuptools_in_meta()))):
+        if dirty:
+            need_source_download = False
+
         # this try/catch is for when the tool to download source is actually in
         #    meta.yaml, and not previously installed in builder env.
         try:


msarahan:
This is interesting.  I think that this function should actually not be accessed at all when dirty is True.  Maybe I'm thinking about some of the refactor work in #953.  I'll look into it, and either merge this, or point out where it's fixed in #953.  Either way, thanks for raising this PR.

daler:
Great, thanks.

msarahan:
Here's the relevant part in that PR: https://github.com/conda/conda-build/pull/953/files#diff-a273b977a04f704063ec213a6230e203R81

I think that covers it, but would appreciate your eyes on it as well.  If that code is OK in your eyes, can we close this PR?

daler:
Not sure I have time at the moment to read up on which exceptions are raised where under which conditions if `config.dirty` is True, but that PR certainly fixes the issue I was having. 

thanks!

msarahan:
I'm going to reopen and merge this, to put it into a bugfix release tonight.

Issue #1117, opened 2016-07-15
Allow CPU_COUNT to be overridden


jakirkham:
It would be nice if users could set `CPU_COUNT` externally via an environment variable that gets automatically passed through by `conda-build` to the recipe(s). In other words, doing `CPU_COUNT=2 conda build recipe` would result in `CPU_COUNT` being `2` within that recipe.

This is very useful when working with CIs or other shared infrastructure where this value is often misreported. As a result, a user could override this to respect the CI's constraints, but use the same recipe they would locally.

cc @pelson @ocefpaf @msarahan @jschueller

jakirkham:
Please see PR ( https://github.com/conda/conda-build/pull/1149 ) for a proposal to do exactly this.

PR #917, opened 2016-05-05
Whitelist MAKEFLAGS env var

jschueller:
Allows to override MAKEFLAGS from shell for ci builds.
See https://github.com/conda-forge/staged-recipes/pull/531#discussion_r62131583

Issue #784, opened 2016-02-17
CPU_COUNT can be too high on Travis CI


izaid:
In DyND, we've found that the CPU_COUNT environment variable can sometimes be reported very high (.e.g CPU_COUNT=32) on Travis CI.

msarahan:
That's probably the native CPU's on the machine.  Docker just passes it through.  Is this causing failed builds?  I feel like this is a special case for Travis, not really a special case for conda-build to handle.  We should definitely provide a way for people to override any auto-detected CPU_COUNT, though, so that we don't break Travis.

izaid:
Yeah, it's causing failed builds for us. Would love a fix for it. Ping @insertinterestingnamehere

msarahan:
For now, I'd recommend changing your recipe to not use -j${CPU_COUNT} -  at least for your checked-in copy of the recipe.

izaid:
Already done, thanks. :)

groutr:
I wonder if this might be of any help
http://serverfault.com/questions/691659/count-number-of-allowed-cpus-in-a-docker-container

`nproc` is a newish command present in coreutils.  It seems possible to parse `/proc/self/status`.  The `Cpus_allowed_list` gives the correct info inside a docker container.  Multiprocessing simply parses `/proc/cpuinfo`.
(http://manpages.ubuntu.com/manpages/natty/man7/cpuset.7.html)

jakirkham:
I opened this issue ( https://github.com/conda/conda-build/issues/1117 ) to try and allow for `CPU_COUNT` to be overridden externally via an environment variable.

PR #1149, opened 2016-07-28
Automatically pass through `CPU_COUNT`

jakirkham:
Fixes https://github.com/conda/conda-build/issues/1117

In some situations, users find the processor information to be reported incorrectly by the infrastructure they may be running a build on. For instance, a user may be running a build on a node where the node supports more cores than the user is actually provided. As a result, the detection of the number of cores will be too high and cause issues completing the build.

Previously it was thought `MAKEFLAGS` would be an acceptable way to accomplish this. Please see PR ( https://github.com/conda/conda-build/pull/917 ). In fact, that strategy does work. Most of the cases where a parallel build is desired are using `make` and thus can easily leverage this. Though it doesn't really catch all cases and so something else still needs to be done for them. Also, using it in this way can result in an implicit setting of parallel builds. Thus there were some cases where a parallel build was not reliable. As a result, users had to unset `MAKEFLAGS` entirely. This is problematic as other flags not related to parallelism may be lost in the process. Please see PR ( https://github.com/conda-forge/conda-forge-build-setup-feedstock/pull/18 ) for references to this problem.

The strategy proposed here is to allow the environment variable `CPU_COUNT` to be automatically passed through to the recipe. This allows users to opt-in to parallelism if it helps their build. However, if their build does not work well in parallel it is not forced upon them. As it simply overrides an existing environment variable used to specify the number of cores available, recipes that use this feature can be safely built in different environments by setting this environment variable to the appropriate value in that setting. Further strategies build strategies that don't use `make`, but do have the ability to build in parallel can be set appropriately using the `CPU_COUNT` environment variable. Finally, by allowing `CPU_COUNT` to be passed through we provide a useful tool in debugging recipes. Instead of having to tinker with hard-coded values in recipes, a user can set this environment variable to different values externally and see how it affects there build.

In short, this proposal should hopefully provide a useful feature for those eager to support parallel builds by putting that control more firmly in the hand of the users. Further, this should help separate what is appropriate for build infrastructure from the recipes themselves resulting in greater flexibility.

cc @ocefpaf @pelson @msarahan @jschueller @frol

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,3 +1,19 @@
+XXXX-YY-ZZ 1.21.x:
+------------------
+
+Enhancements:
+-------------
+
+* Whitelist the CPU_COUNT environment variable. #1149
+
+Bug fixes:
+----------
+
+Contributors:
+-------------
+
+* @jakirkham
+
 2016-08-06 1.21.11:
 -------------------
 
--- a/conda_build/environ.py
+++ b/conda_build/environ.py
@@ -345,7 +345,11 @@ def system_vars(env_dict, prefix):
     if 'MAKEFLAGS' in os.environ:
         d['MAKEFLAGS'] = os.environ['MAKEFLAGS']
 
-    d['CPU_COUNT'] = get_cpu_count()
+    if 'CPU_COUNT' in os.environ:
+        d['CPU_COUNT'] = os.environ['CPU_COUNT']
+    else:
+        d['CPU_COUNT'] = get_cpu_count()
+
     if "LANG" in os.environ:
         d['LANG'] = os.environ['LANG']
     d['PATH'] = os.environ['PATH']


CHANGELOG.txt
```diff
@@ -1,3 +1,19 @@
+XXXX-YY-ZZ 1.21.x:
+------------------
+
+Enhancements:
+-------------
+
+* Whitelist the CPU_COUNT environment variable. Setting this outside conda
```
msarahan: Please shorten this and just link to the PR.  People can read more if they're interested.

jakirkham: Yeah, it got a bit out of hand. Will update it later.

conda_build/environ.py
```diff
@@ -345,7 +345,11 @@ def system_vars(env_dict, prefix):
     if 'MAKEFLAGS' in os.environ:
         d['MAKEFLAGS'] = os.environ['MAKEFLAGS']
 
-    d['CPU_COUNT'] = get_cpu_count()
+    if 'CPU_COUNT' in os.environ:
+        d['CPU_COUNT'] = os.environ['CPU_COUNT']
```
msarahan: This is a good way to accomplish this.  Correct me if I'm wrong, but it means that recipes need to add stuff like:

make -j${CPU_COUNT}

right?  So we are now saying that people make the choice to be parallel explicitly?  I agree with that, but I just want to make sure that I understand it clearly.

ocefpaf: @msarahan that is correct and it is the same strategy as many Linux distro use. To make the flag available, so people who use wish to use it can use.

jakirkham: > make -j${CPU_COUNT}

This would indeed be an adjustment for conda-forge. Though this is exactly what many of us had done with our recipes previously.

jakirkham: > Although I am not sure this below in conda-build.

Could you please clarify this, @ocefpaf? I believe "below" was suppose to be "belong". Assuming that is the case, I have some questions.

Why shouldn't this change be here? `CPU_COUNT` is already provided for by `conda-build`. We are merely trying to get some control over that value in case it was set improperly. It is also something that other people struggle with too ( https://github.com/conda/conda-build/issues/784 ). By making this change, this will be one less thing we have to watch out for in recipe review.

In the long run, I don't think any pass through environment variables should be hard-coded in `conda-build` source. Instead we should be able to specify some sort of configuration file that allows one to set what environment variables qualify for automatic pass through. This should cutdown on maintenance burden and increase flexibility. Though this last point is interesting, it is out of scope for this proposal and I do not have the bandwidth to pursue it ATM.

msarahan:
@jakirkham, I'm hoping to tag a bugfix release later today.  Do you want to try to get this in, or wait for 2.0?

jakirkham:
Looks like the bug fix has already happened. In any event, please don't hold bug fixes on my account. I'm out on vacation this week.

Hoping to look at this again next week around the middle of the week.

CHANGELOG.txt
jakirkham: Done.

jakirkham:
Running into some failures, but I don't think they are caused by this change. 😕

msarahan:
@jakirkham please sign the CLA, then I'm happy to merge.

jakirkham:
Thanks @msarahan. Have gotten approval to sign the CLA and have signed it. Info in the signed CLA.

msarahan:
Thanks @jakirkham - merging.  I hope the legal permissions weren't too arduous for you.  If you have feedback on that, please email me and I'll pass it along.

jakirkham:
> I hope the legal permissions weren't too arduous for you. If you have feedback on that, please email me and I'll pass it along.

Nothing really on your end. More a matter of finding the right people to talk to and getting their ok. Just takes time is all.

jakirkham:
Is it alright if we make sure this lands in some `1.21.x` release?

Periodically discussions come up where people still want to use `$CPU_COUNT`. It would be nice if we can ensure a smooth transition for them to conda-forge without laborious conversations about how we do things and how CIs act. Not to mention, it would resolve some issues that setting `$MAKEFLAGS` has caused.

FWIW I don't think it is urgent for the `1.21.12` release to occur soon. Just it would be nice to address this cause of stress before looking to `2.0.0`. 😄

msarahan:
Yes, I will put one more 1.21 release out soon.

jakirkham:
Put in play at conda-forge. ( https://github.com/conda-forge/conda-forge-build-setup-feedstock/pull/24 ) 😄

Issue #1023, opened 2016-06-14
__conda_buildnum__.txt feature is broken (and its deprecation is not documented)


mwiebe:
Got this in a build:

```
2016-06-13 18:50:45:  0: STDOUT: Deprecation warning: support for __conda_buildnum__ will be removed in Conda build 2.0.Try Jinja templates instead: http://conda.pydata.org/docs/building/environment-vars.html#git-environment-variables
2016-06-13 18:50:45:  0: STDOUT: Setting build number from __conda_buildnum__.txt: 2
2016-06-13 18:50:45:  0: STDOUT: Traceback (most recent call last):
2016-06-13 18:50:45:  0: STDOUT:   File "C:\Anaconda3\Scripts\conda-build-script.py", line 4, in <module>
2016-06-13 18:50:45:  0: STDOUT:     sys.exit(main())
2016-06-13 18:50:45:  0: STDOUT:   File "C:\Anaconda3\lib\site-packages\conda_build\main_build.py", line 134, in main
2016-06-13 18:50:45:  0: STDOUT:     args_func(args, p)
2016-06-13 18:50:45:  0: STDOUT:   File "C:\Anaconda3\lib\site-packages\conda_build\main_build.py", line 373, in args_func
2016-06-13 18:50:45:  0: STDOUT:     args.func(args, p)
2016-06-13 18:50:45:  0: STDOUT:   File "C:\Anaconda3\lib\site-packages\conda_build\main_build.py", line 320, in execute
2016-06-13 18:50:45:  0: STDOUT:     need_source_download=need_source_download)
2016-06-13 18:50:45:  0: STDOUT:   File "C:\Anaconda3\lib\site-packages\conda_build\build.py", line 523, in build
2016-06-13 18:50:45:  0: STDOUT:     get_build_metadata(m)
2016-06-13 18:50:45:  0: STDOUT:   File "C:\Anaconda3\lib\site-packages\conda_build\post.py", line 394, in get_build_metadata
2016-06-13 18:50:45:  0: STDOUT:     m.meta['build']['number'] = build_number
2016-06-13 18:50:45:  0: STDOUT: KeyError: 'build'
```

[...]

PR #1154, opened 2016-07-31
add CONDA_BUILD_RENDERING env var that is set while rendering

msarahan:
Ping @janssen for comment.  This is meant to be my implementation of our discussion in https://github.com/conda/conda-build/issues/1023#issuecomment-236064273

### DIFF
--- a/conda_build/metadata.py
+++ b/conda_build/metadata.py
@@ -382,7 +382,13 @@ class MetaData(object):
         if not self.meta_path:
             return
 
-        self.meta = parse(self._get_contents(permit_undefined_jinja), path=self.meta_path)
+        try:
+            os.environ["CONDA_BUILD_RENDERING"] = "True"
+            self.meta = parse(self._get_contents(permit_undefined_jinja), path=self.meta_path)
+        except:
+            raise
+        finally:
+            del os.environ["CONDA_BUILD_RENDERING"]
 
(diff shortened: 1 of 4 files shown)


msarahan:
Travis failure is the race condition on anaconda.org.  This is ready to go.

janssen:
Looks OK.  Is there a corresponding CONDA_BUILD_BUILDING that is set while actually building?

msarahan:
Not presently.  Can you enlighten me on a use case?  I think CONDA_BUILD_BUILDING is sort of the default state (formerly the only state), but I am also wary of the dangers that a default state entail.  If you have even a half-compelling use case, I'll put it in.  Might as well go ahead and introduce the whole notion of state, and add in a state for testing, too.  Perhaps CONDA_BUILD_RENDERING would be better represented with a CONDA_BUILD_STATE variable that can be RENDERING or BUILDING or TESTING or whatever else is down the road.

janssen:
I don't have a use case, I was just thinking that if there are going to be these explicit modes of operation, it would be useful to be able to find out which mode the build process was in.  CONDA_BUILD_STATE would work.

msarahan:
OK.  I will work this into the 2.0 release.

msarahan:
Implemented in #1232

PR #1155, opened 2016-07-31
test compileall;  only compile files in packages (not deps)

msarahan:
Corrollary to #1146, plus a bit of improved consistency in terms of compiling on all Pythons.  This uses py_compile in favor of compileall, and only compiles files that are fed in from ```post_process```


msarahan:
Ping @ilanschnell - this implements your idea that we should not package .cpython-35.pyc (for example) files.  It renames files and moves them one level higher, adjacent with the origin .pyc files.

I welcome any feedback you might have on the implementation.

jreback:
@msarahan no dice with 1.21.11: https://ci.appveyor.com/project/jreback/pandas-465/build/job/yy70e82akg14my9j (also the conda-build version tag looks like a dev tag!)

msarahan:
Don't worry too much about the dev tag.  @ilanschnell had to patch the setup.py to use distutils in order to build it with his build system.  We're working on figuring out some misbehavior with argparse and our entry points.  I'll look more at your build log on the .pyc stuff, though - that really should be fixed.

msarahan:
Yuck, so the error is FileNotFoundError: [WinError 206] The filename or extension is too long
 - not that compiling is failing, per se, but that the compile command can't even run because we're falling afoul of Window's path length limits somehow.

I'm not sure how to fix this right now.  Please try setting CONDA_BLD_PATH as an environment variable to some very short temporary folder (perhaps ```C:\bld```?) - this might be an adequate workaround for the time being.

What this will do is change conda build's root folder from ```C:\Python34_64\conda-bld``` to ```c:\bld```, which might help here.

jreback:
so I fixed conda-build to 1.21.7 see [here](https://github.com/pydata/pandas/commit/cfbb24f2c0e33ee221c54fd4944458cb0c6f4e84) and this works successfully (well we are failing but on a test now): https://ci.appveyor.com/project/jreback/pandas/build/job/q22klk7yjb0c4osl

msarahan:
Ok. If you don't need any of the fixes between 1.21.7 and now, that sounds fine. I'll see about adding a test and fixing this in the meantime.

PR #1163, opened 2016-08-02
Compile files ending in `.py` not those ending in `py`

mingwandroid:
Otherwise files like h5copy get found and the build will fail


msarahan:
Yikes, good catch.  That was silly of me.  Does it actually kill the build, though?  I thought that with #1146 it would barf but not die?

I'm asking because this is the difference between merging and doing another 1.21.x release, or just merging and proceeding to 2.0.

PR #1165, opened 2016-08-02
Use setuptools entry points

msarahan:
Something was going wrong with the conda-based entry points in the conda-build conda-forge feedstock: https://github.com/conda-forge/pyferret-feedstock/pull/8

As near as I can tell, these entry points were manually created, and then the legacy build system did the right thing with the contents of the scripts entry.  This apparently is not working with conda-build.

This PR uses setuptools entry points instead.  It introduces the constraint that pip be used to install during the build process - otherwise, egg-based entry points are created.  With pip, we get entry points similar to the ones that were manually created.


msarahan:
@ilanschnell - I don't think that this necessarily requires adjustment to your recipe on the anaconda repository, but it might.  Please take a look and make sure.

PR #1169, opened 2016-08-03
Add build/skip_compile_pyc option

### LINKED ISSUE #686: turn off pyc compilation?

mingwandroid:
@msarahan: It's strangely coincedental that I'm making changes in the same area of compiling `.pyc` files two days in a row, as other than the location, this is not related to yesterday's fix.

Some packages ship .py files that should not be compiled because either they
cannot be compiled (for example they contain templates) or should not be
compiled yet because the Python interpreter that will be used cannot be known
at build-time.

An example of such a package is `qtcreator` which uses GDB's embedded Python
interpreter to give a richer debugging experience. This package falls foul of
both scenarios given above.

Closes #686

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -766,7 +766,8 @@ def build(m, config, post=None, need_source_download=True, need_reparse_in_env=F
                      prefix=config.build_prefix,
                      config=config,
                      preserve_egg_dir=bool(m.get_value('build/preserve_egg_dir')),
-                     noarch=m.get_value('build/noarch'))
+                     noarch=m.get_value('build/noarch'),
+                     skip_compile_pyc=m.get_value('build/skip_compile_pyc'))
 
         # The post processing may have deleted some files (like easy-install.pth)
         files2 = prefix_files(prefix=config.build_prefix)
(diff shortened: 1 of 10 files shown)


msarahan:
I would prefer to add explicit lists of files/folders to skip.  Is that sane/feasible?

Realistically, I think the answer to all of this is to push the pyc compilation over to the conda side, so that it happens at install time rather than compile time.  There, I think that the current approach of trying to compile .py files (but not exiting on failure) should work ok, and explicit lists of files to skip would smooth over any confusing messages.

msarahan:
Tagging 1.21.11 without this, but I'm interested in getting this in for the next release.  Let me know what you think about file/folder lists.

msarahan:
As discussed in person, @mingwandroid would prefer to close this for now.  If he or anyone else would like to re-raise it, I'm happy to consider it.

mingwandroid:
I'm going to reopen this now @msarahan and make it a list of files. I need it to build menuinst.

msarahan:
Okie dokie.  Happy to look at this whenever you're ready.

msarahan:
LGTM.  Test failures are my fault, monkeying with the test package.  By checking out master, we are getting some other tests that should only affect one test case, but here is affecting many tests.

Merging.

PR #1171, opened 2016-08-03
allow case insensitive matches of prefix path in binaries

msarahan:
I've seen prefixes that are all lowercase embedded in pip-created entry point exe's.  This PR should catch those as well as ordinary ones.

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -109,7 +109,7 @@ def have_prefix_files(files):
         double_backslash_prefix = prefix.replace('\\', '\\\\')
         double_backslash_prefix_bytes = double_backslash_prefix.encode('utf-8')
         # moar escapes for regex
-        prefix_bytes = prefix_bytes.replace('\\', '\\\\')
+        prefix_bytes = prefix_bytes.replace(b'\\', b'\\\\')
 
     for f in files:
         if f.endswith(('.pyc', '.pyo', '.a')):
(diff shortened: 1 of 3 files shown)


msarahan:
Cycling to run this through CI's again.

minrk:
@msarahan since conda-build 1.21.11 produces packages that cannot be installed by conda < 4.2.3, what's the best way to make sure that the packages require conda >= 4.2.3?

msarahan:
Copying comment from https://github.com/conda-forge/nbformat-feedstock/pull/3#issuecomment-239433797 for easier reading:

Sorry, but I don't think that reverting conda build will work. It's not that conda build changed how pip behaves, but rather just that it now properly detects binary prefixes embedded in the entry point executables. Because pip is embedding these prefixes into the executables, you instead need to figure out when pip started doing that, and pin pip. I am pretty sure that older versions of conda build will still have broken entry points, and moreover they simply won't be fixable by any version of conda.

I have also thought about using pip on Mac and Linux, but the old way on Windows to avoid these embedded prefix exes, until the conda support for them is more generally available. I am not sure if that is palatable to you.

kalefranz:
from https://github.com/conda/conda/pull/3262#issuecomment-239410077

> @minrk For conda-forge users, since conda-forge is now packaging conda itself, the answer is whenever the feedstock is updated. https://github.com/conda-forge/conda-feedstock

> I'd have no objections to the conda-forge community updating before canary is released into defaults.  For 4.2 (currently at 4.2.3), that date is Aug 28.

PR #1182, opened 2016-08-06
Merge 1.21.11

msarahan:


msarahan:
Travis failures are due to insufficiently long prefixes in existing conda-forge packages that are installed into a longer build prefix.  No way around that right now - planning on a fallback, where if build environment creation fails with that message, we set build prefix to 80 characters and try again, but with a huge warning that anything built with this fallback will also be limited to 80 character or less prefixes.

PR #1189, opened 2016-08-08
Re-add logic for activating C++ Compiler for Python 2.7

patricksnape:
Prefer the 'full' VS and fallback to Python tools after. At the moment, conda-build will never seek the Python tools which is a regression in behaviour. I understand why VS9 'full' is preferred (CMake) but we should still fallback to looking for the tools if available. @msarahan

### DIFF
--- a/conda_build/windows.py
+++ b/conda_build/windows.py
@@ -174,8 +174,7 @@ def msvc_env_cmd(bits, override=None):
 
         error1 = 'if errorlevel 1 {}'
 
-        # Setuptools captures the logic of preferring the Microsoft Visual C++
-        # Compiler for Python 2.7 - falls back to VS2008 if necessary
+        # Prefer VS9 proper over Microsoft Visual C++ Compiler for Python 2.7
         msvc_env_lines.append(build_vcvarsall_cmd(vcvarsall_vs_path))
         # The Visual Studio 2008 Express edition does not properly contain
         # the amd64 build files, so we call the vcvars64.bat manually,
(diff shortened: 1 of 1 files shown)


msarahan:
Sorry, I don't understand this one.  AFAICT, we do find the c++ compiler for python at https://github.com/conda/conda-build/pull/1189/files#diff-4fa0e94bf0114b071e08a8bfc07b41a2R185 - but only for 64-bit.  Is this a problem only with 32-bit, or am I misunderstanding.  Happy to add this either way, just want to understand where the issue is.

patricksnape:
If the intended functionality of this line:
```
VCVARS64_VS9_BAT_PATH = os.path.join(os.path.dirname(distutils_find_vcvarsall(9)),
                                                 'bin', 'vcvars64.bat')
```
is to find the C++ compiler then that will not working as that is not the directory structure of the C++ compiler. ``distutils_find_vcvarsall`` will return something like ``%LOCALAPPDATA%\Programs\Common\Microsoft\Visual C++ for Python\9.0\`` which contains the folders `VC`, `WinSDK, `License` and the `vcvarsall.bat`. There is no `bin` folder - though there is within the `VC` folder. However, it the C++ compiler is not found,  then distutils falls back to VS2008 which presumably will work with that path structure.

Also, since the C++ compiler is a lot simpler to install I've just been installing it on our build cluster and then finding that conda-build no longer works because it forces the use of VS2008 at the moment. I am not finding, on my machines, that even 64-bit builds work with the C++ compiler.

msarahan:
Yuck. I thought we had this stuff figured out. 

Merging this pr. Some day I need to return to this and write it all out.

msarahan:
Edits welcome at https://github.com/conda/conda-build/wiki/Windows-Compilers - also https://wiki.python.org/moin/WindowsCompilers is a pretty great resource.

PR #1190, opened 2016-08-08
source: Use `git am` to apply patches when possible.

mingwandroid:
If the source has been cloned using `git` and the patch is a
git format patch, then use `git am` to apply it. This allows
the recipe's author to work directly in the `work` directory
iterating on the patch-set using all of the power of git and
to finally move the patches back into the recipe folder with
`git format-patch` and `mv *.patch ...`

### DIFF
--- a/conda_build/source.py
+++ b/conda_build/source.py
@@ -167,7 +167,7 @@ def git_source(meta, recipe_dir, verbose=False):
     if not verbose:
         FNULL.close()
 
-    return WORK_DIR
+    return git
 
 
 def git_info(fo=None, verbose=False):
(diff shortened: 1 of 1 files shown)


mingwandroid:
Hi @msarahan, what do you think about this?

msarahan:
Lgtm. Any tests you can add for it?

mingwandroid:
Yeah, I'll push a change to test the detection of a git patch vs a non-git patch, would that be enough or do you want tests of running git and checking that the commits get made? I guess that would be better?

Also, I wasn't sure, but I assume that our patches are applied in the order they are listed in `meta.yaml`? This is pretty important for this (and generally!)

msarahan:
Superseded by #1222.  Closing.

PR #1192, opened 2016-08-08
fix python 3 exception string handling with prefix too long errors

msarahan:
Follow-up to #1183, which had issues with Python 3 and the exception message string.

CC @pelson @jakirkham


jakirkham:
LGTM

PR #1195, opened 2016-08-09
add simple function for finding packages with short binary prefixes

msarahan:
With the extended length prefixes coming soon, it is helpful to know which packages in an ecosystem need to be rebuilt.  This PR introduces a function for identifying packages with short prefixes.  It can be used as a Python function (see tarcheck.py, check_prefix_lengths(folder, min_prefix_length=255)), returning a dictionary of filenames and prefix lengths, or as a command-line tool:

```
conda inspect prefix-lengths
```

This command takes one mandatory argument: the folder to inspect; and one optional argument - a prefix length threshold, by default set to the value of the default prefix length in our config.  Only packages with prefix lengths shorter than this value will be show.

This functionality is perhaps best employed on a package host, because otherwise we have to download packages to inspect them.  Perhaps an SSH filesystem would allow remote inspection without downloading complete packages?

CC @jakirkham @bollwyvl @ilanschnell @csoja

### DIFF
--- a/conda_build/main_inspect.py
+++ b/conda_build/main_inspect.py
@@ -29,6 +29,8 @@ from conda_build.main_build import args_func
 from conda_build.ldd import get_linkages, get_package_obj_files, get_untracked_obj_files
 from conda_build.macho import get_rpaths, human_filetype
 from conda_build.utils import groupby, getter, comma_join
+from conda_build.tarcheck import check_prefix_lengths
+from conda_build.config import config
 
 logging.basicConfig(level=logging.INFO)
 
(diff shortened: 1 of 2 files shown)


asmeurer:
I'm glad you're finally making this change. Should have happened a long time ago. 

Do you plan to scan conda-forge and push build number bumps for rebuilds?

msarahan:
Yeah - haven't quite figured out how to scan conda-forge as a channel, though.  Will work the the anaconda.org team to figure out the best way to do that.

Rebuilds on conda-forge should happen once conda-build 2.0 is released and proven to be safe (hopefully soon.)

PR #1196, opened 2016-08-10
Re-evaluate prefix when need to use 80 char version

mingwandroid:
Hi @msarahan, I ran into this just now.


msarahan:
doh.  That's a really dumb one.  Thanks for the fix.  I think it might be cleaner to pass config.build_prefix in the function call directly, though.  Want to change to that, or want me to do it?

mingwandroid:
Done.

mingwandroid:
.. and actually done :-)

msarahan:
Thanks @mingwandroid - good catch here.

PR #1198, opened 2016-08-10
fix: detect VS installation directory 

### LINKED ISSUE #560: Idea: More robust detection of Visual Studio
I'm looking into the source code of ``conda_build.windows.msvc_env_cmd()``:

  https://github.com/conda/conda-build/blob/master/conda_build/windows.py#L56

I've just implemented something similar, not realizing that such functionality already exists in ``conda-build`` (this is something worth mentioning in the documentation by the way). The only thing that concerns me a bit is ``msvc_env_cmd()`` assumes that Visual Studio is installed in Program Files rather than getting it from the registry. Perhaps you can refactor that by relying on ``setuptools``:

```python
from distutils.msvc9compiler import VERSION, PLAT_TO_VCVARS, get_platform
from setuptools.msvc9_support import find_vcvarsall

vc = find_vcvarsall(VERSION)
plat = PLAT_TO_VCVARS[get_platform()]

header = 'call "{vc}" {plat}".format(vc=vc, plat=plat)
```

This code gives you the full path to ``vcvarsall.bat`` matching the version of the current Python interpreter and its ``plat`` argument. The generated ``header`` can be inserted in ``bld.bat``. It needs to be executed using ``%PYTHON%`` to produce correct results as it decides what version of Visual Studio to look for based on the version of the current interpreter.

ikalev:
Really fix #560. 

Allow ``build_vcvarsall_vs_path(v)`` to be invoked with VS version strings in ``Major.Minor`` format. Currently this function assumes that its argument is in ``MajorMinor`` format, which is not always the case. This causes VS installation path detection to fail.

### DIFF
--- a/conda_build/windows.py
+++ b/conda_build/windows.py
@@ -66,7 +66,7 @@ def build_vcvarsall_vs_path(version):
     """
     Given the Visual Studio version, returns the default path to the
     Microsoft Visual Studio vcvarsall.bat file.
-    Expected versions are of the form {9, 10, 12, 14}
+    Expected versions are of the form {9.0, 10.0, 12.0, 14.0}
     """
     # Set up a load of paths that can be imported from the tests
     if 'ProgramFiles(x86)' in os.environ:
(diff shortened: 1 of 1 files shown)


msarahan:
Thanks.  This is a good consistency improvement.  As far as I can tell, the environment variable stuff was working, but the manual fallback would have been broken.  This makes everything work.  Is that right?

ikalev:
It's actually the other way around -- ``version`` is passed in ``x.0`` format (contrary to what the docstring used to say), which resulted in building an incorrect environment variable name: ``VSx.00COMNTOOLS`` instead of ``VSx0COMNTOOLS``. So the environment variable stuff was broken.

msarahan:
Ah, yes.  Thanks for clarifying.  Merging.

PR #1200, opened 2016-08-10
conda interface

kalefranz:
This PR creates an interface file that isolates all of conda-build's imports from conda.

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -18,15 +18,15 @@ import tempfile
 from os.path import exists, isdir, isfile, islink, join
 import mmap
 
-import conda.config as cc
-import conda.plan as plan
-from conda.api import get_index
-from conda.compat import PY3
-from conda.fetch import fetch_index
-from conda.install import prefix_placeholder, linked, move_to_trash, symlink_conda
-from conda.lock import Locked
-from conda.utils import url_path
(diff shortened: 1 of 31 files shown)


msarahan:
Thanks.  This is definitely a step in the right direction.  It's still strange to have this pointed at several conda internals.

I think this PR reflects an idea of "I'm going to keep track of where everyone is using conda internals" - which you certainly simplify with this PR, and presumably with PRs like it to other projects that depend on conda.

I think a more scalable idea that respects your time more is to define an API on the conda side, and make absolutely no promise of support to any external client going around, not through that API.  People should not have interfaces for conda on their side (like this PR) unless they just want to consolidate the imports for their own cleanliness.  This would absolve you of responsibility to track or test these projects.  Instead, you should test your API directly, and discuss additions and changes to that API as external clients request them.

kalefranz:
This is only meant to be a step in the right direction--not at all a final solution. 

I always prefer that dependencies be isolated in my own projects. Conda does it now with ruamel_yaml and requests. 

Maintaining this interface will do at least two things for us. 

1) Help me identify and sketch out what a good API for conda would look like and contain
2) In the meantime before that actually ships, help provide more stability for both of us

msarahan:
Thanks @kalefranz

PR #1208, opened 2016-08-12
patch debug test for conda 4.2 loggers

kalefranz:


msarahan:
Thanks for this fix, but please do not self merge here.

PR #1212, opened 2016-08-15
do not force download if vcs is used as source

### LINKED ISSUE #1211: conda build --output should not check out sources
It used not to, but at some point that changed, and now trying to figure where the built package is wastes a lot of time.

msarahan:
Fixes #1211

### DIFF
--- a/conda_build/metadata.py
+++ b/conda_build/metadata.py
@@ -728,19 +728,10 @@ class MetaData(object):
         return self.__str__()
 
     def uses_vcs_in_meta(self):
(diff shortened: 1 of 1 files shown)


msarahan:
@whitequark - I think this should fix your issue.  I'm looking into how easily I can mock our tests, so that I can test this particular case better.

msarahan:
Merging this here.  A proper mock test requires the testing through the API in #953.  I'll add it when I can.

whitequark:
Thanks!

PR #1219, opened 2016-08-17
fix breakage related to Windows env vars being strings

msarahan:
Windows variables are touchy - you should never change the type of the key, or else it is no longer the same key.  This steps back a bit on #1216.  It also makes the trash function local, for wrapping permission setting (https://github.com/conda/conda/pull/3280), though in practice, I could not get permission setting to work here.

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -24,7 +24,7 @@ from .conda_interface import plan
 from .conda_interface import get_index
 from .conda_interface import PY3
 from .conda_interface import fetch_index
-from .conda_interface import prefix_placeholder, linked, move_to_trash, symlink_conda
+from .conda_interface import prefix_placeholder, linked, symlink_conda
 from .conda_interface import Locked
 from .conda_interface import url_path
 from .conda_interface import Resolve, MatchSpec, NoPackagesFound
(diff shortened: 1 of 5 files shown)


conda_build/build.py
```diff
@@ -394,11 +394,11 @@ def create_env(prefix, specs, clear_cache=True, debug=False):
...
         plan.display_actions(actions, index)
-        plan.execute_actions(actions, index, verbose=debug)
+
         if on_win:
-            os.environ = {k.encode(codec) if hasattr(k, 'encode') else k:
-                            v.encode(codec) if hasattr(v, 'encode') else v
-                            for k, v in os.environ.items()}
+            for k, v in os.environ.items():
+                os.environ[k] = str(v)
+        plan.execute_actions(actions, index, verbose=debug)
```
mingwandroid: Just checking, is the change of order between `plan.execute_actions()` and changing `os.environ` deliberate?

msarahan: Yes.  I'm pretty sure that any error originating from this issue happens in the execute_actions call.  Thus, it is essential that our stringification happens before then.

PR #1221, opened 2016-08-18
Set prefix to config.build_prefix

mingwandroid:
After we've discovered we need to use a smaller prefix, that variable
is re-used, but it still contains the bigger value.


msarahan:
Did I lose this in merging things yesterday? 

Same story as last time:please put config.build_prefix directly in the create env call.

msarahan:
@mingwandroid I'm going to absorb this into another PR and fix things.  The string coercion is all wrong here, too.

mingwandroid:
It wasn't lost no.

I was right first time around .. `prefix` itself is used later in the same function so we need to set that too.

msarahan:
ugh.  ok.  good catch, then.  Merging.

PR #1225, opened 2016-08-18
remove egg removal.  Simplifies and fixes issue with pyc compilation

msarahan:
Discussed with @ilanschnell.  Should fix any issues with compilation, which was finding python files to compile, but then this removal was moving them out from under it before compilation.

CC @quasiben

### DIFF
--- a/conda_build/post.py
+++ b/conda_build/post.py
@@ -191,9 +191,9 @@ def compile_missing_pyc(files, cwd=config.build_prefix, python_exe=config.build_
 
 
 def post_process(files, preserve_egg_dir=False):
+    compile_missing_pyc(files)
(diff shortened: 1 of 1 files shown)


quasiben:
👍

mingwandroid:
What was up with .pyc compilation? I didn't spot that from the diff ..

msarahan:
Just a matter of ordering: https://github.com/conda/conda-build/blob/master/conda_build/post.py#L192-L195

```
def post_process(files, preserve_egg_dir=False):
    remove_easy_install_pth(files, preserve_egg_dir=preserve_egg_dir)
    rm_py_along_so()
    compile_missing_pyc(files)
```

We could just reorder this, so that compilation happens before remove_easy_install_pth.  Perhaps that is better as a less disruptive change.

PR #1237, opened 2016-08-22
add source_files test arg that copies source files into test folder

msarahan:
A frequent source of confusion is that test files are copied from the recipe directory, and there is no clear way to copy files from the source directory into the test folder.  This PR adds a new entry to the test section of meta.yaml, ```source_files``` that copies files from the source folder into the test folder.  It should also download the source if necessary (for example, if a package is being tested separately from a build).

### DIFF
--- a/conda_build/create_test.py
+++ b/conda_build/create_test.py
@@ -4,11 +4,12 @@ Module to handle generating test files.
 
 from __future__ import absolute_import, division, print_function
 
+import glob
+from os.path import join, exists, isdir
 import sys
 
-from os.path import join, exists
-
 from conda_build.utils import copy_into
+from conda_build import source
 
 
 header = '''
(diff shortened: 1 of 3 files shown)


msarahan:
Appveyor failures are due to conda downloads being slow.  Merging.

jakirkham:
Does this work with directories?

msarahan:
Yep.

jakirkham:
So if one did something like...

```yaml
test:
  source_files:
    - {{ SRC_DIR }}
```

Would that actually copy all of the source directory? Thinking of the cases where someone wants to run `make check` at this stage.

msarahan:
I think you'd actually want

```
test:
  source_files:
    - {{ SRC_DIR }}/*
```

because you'll want the files to sit at CWD.  If you do what you proposed, you'd have a folder for the sources, probably something like ```mypackage-myversion```

Glob patterns, directories, and filenames should all work.

msarahan:
There is a "bug" here - it should be using ```source.get_dir(config)```, but it is using config.work_dir.  This means it most often drops you one folder higher than you expect.  I'll fix that.

msarahan:
https://github.com/conda/conda-build/pull/1288

PR #1256, opened 2016-08-24
Build environment used during tooling and testing.

mingwandroid:
Hi @masarhan, please review this.

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -403,42 +403,43 @@ def create_env(prefix, specs, config, clear_cache=True):
             lock = filelock.SoftFileLock(lock_file)
             lock.acquire(timeout=config.timeout)
 
-            try:
-                actions = plan.install_actions(prefix, index, specs)
-                plan.display_actions(actions, index)
-                if on_win:
-                    for k, v in os.environ.items():
-                        os.environ[k] = str(v)
-                plan.execute_actions(actions, index, verbose=config.debug)
-            except SystemExit as exc:
-                if "too short in" in str(exc) and config.prefix_length > 80:
-                    log.warn("Build prefix failed with prefix length {0}."
-                            .format(config.prefix_length))
-                    log.warn("Error was: ")
-                    log.warn(str(exc))
-                    log.warn("One or more of your package dependencies needs to be rebuilt "
-                            "with a longer prefix length.")
-                    log.warn("Falling back to legacy prefix length of 80 characters.")
-                    log.warn("Your package will not install into prefixes > 80 characters.")
-                    config.prefix_length = 80
+            with path_prepended(prefix):
+                try:
+                    actions = plan.install_actions(prefix, index, specs)
+                    plan.display_actions(actions, index)
+                    if on_win:
+                        for k, v in os.environ.items():
+                            os.environ[k] = str(v)
+                    plan.execute_actions(actions, index, verbose=config.debug)
+                except SystemExit as exc:
+                    if "too short in" in str(exc) and config.prefix_length > 80:
+                        log.warn("Build prefix failed with prefix length {0}."
+                                .format(config.prefix_length))
+                        log.warn("Error was: ")
+                        log.warn(str(exc))
+                        log.warn("One or more of your package dependencies needs to be rebuilt "
+                                "with a longer prefix length.")
+                        log.warn("Falling back to legacy prefix length of 80 characters.")
+                        log.warn("Your package will not install into prefixes > 80 characters.")
+                        config.prefix_length = 80
 
-                    # Set this here and use to create environ
-                    #   Setting this here is important because we use it below (symlink)
-                    prefix = config.build_prefix
+                        # Set this here and use to create environ
+                        #   Setting this here is important because we use it below (symlink)
+                        prefix = config.build_prefix
+                        lock.release()
+                        if os.path.isfile(lock_file):
+                            os.remove(lock_file)
+                        create_env(prefix, specs, config=config,
+                                    clear_cache=clear_cache)
+                    else:
+                        lock.release()
+                        if os.path.isfile(lock_file):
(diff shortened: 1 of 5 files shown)


msarahan:
LGTM.  Do you want to add any regression tests based on your reasons for adding this?

mingwandroid:
Yeah, let's see how the existing tests fare on CI then I'll think about a test tomorrow.

conda_build/build.py
```diff
@@ -615,8 +616,9 @@ def build(m, config, post=None, need_source_download=True, need_reparse_in_env=F
 
                     # There is no sense in trying to run an empty build script.
                     if isfile(build_file) or script:
-                        env = environ.get_dict(config=config, m=m, dirty=config.dirty)
-                        env["CONDA_BUILD_STATE"] = "BUILD"
+                        with path_prepended(config.build_prefix):
+                            env = environ.get_dict(config=config, m=m, dirty=config.dirty)
+                            env["CONDA_BUILD_STATE"] = "BUILD"
```
msarahan: not terribly important, but I think the scope of env["CONDA_BUILD_STATE"] is one level out, since it is agnostic of the PATH.

conda_build/build.py
```diff
@@ -769,9 +771,10 @@ def test(m, config, move_broken=True):
 
         create_env(config.test_prefix, specs, config=config)
 
-        env = dict(os.environ.copy())
-        env.update(environ.get_dict(config=config, m=m, prefix=config.test_prefix))
-        env["CONDA_BUILD_STATE"] = "TEST"
+        with path_prepended(config.test_prefix):
+            env = dict(os.environ.copy())
+            env.update(environ.get_dict(config=config, m=m, prefix=config.test_prefix))
+            env["CONDA_BUILD_STATE"] = "TEST"
```
msarahan: same here.  I think the only place the path prepending matters is the env.update() call.

conda_build/utils.py
```diff
@@ -430,3 +430,19 @@ def guess_license_family(license, allowed_license_families):
...
+def _func_defaulting_env_to_os_environ(func, *popenargs, **kwargs):
+    if not 'env' in kwargs:
+        kwargs = kwargs.copy()
+        env_copy = os.environ.copy()
+        kwargs.update({'env': env_copy})
+    return func(*popenargs, **kwargs)
+
+
+def check_call_env(*popenargs, **kwargs):
+    return _func_defaulting_env_to_os_environ(subprocess.check_call, *popenargs, **kwargs)
```
msarahan: nice wrapper here

tests/test_api_build.py
```diff
@@ -429,58 +442,77 @@ def test_numpy_setup_py_data(test_config):
...
+    sys_git_env['GIT_COMMITTER_NAME'] = 'conda-build'
+    sys_git_env['GIT_COMMITTER_EMAIL'] = 'conda@conda-build.org'
+
+    # Put the broken git on os.environ["PATH"]
+    dummy_executable(testing_workdir, 'git')
+    old_path = os.environ["PATH"]
+    os.environ["PATH"] = os.pathsep.join([testing_workdir, os.environ["PATH"]])
+    # .. and ensure it gets run (and fails).
+    dummy_git_ran = False
+    try:
```
msarahan: There's a nice pytest construct for this:

```
with pytest.raises(exception type you expect):
    do stuff
```

conda_build/build.py
mingwandroid: If move this down one line, then os.environ won't contain our modified path, then env.update won't include it either.

conda_build/build.py
mingwandroid: Sure.

conda_build/build.py
msarahan: yes, right, because path_prepended is modifying os.environ, not env.  Got it.

msarahan:
Not bad. Only 3 fails on win, all due to command length limits. Let's talk more tomorrow about ways to shorten the path.

mingwandroid:
Since this is nearly ready to merge (I hope!) I've tacked on a new feature. I wasn't sure about how I transfer `noarch` from the `metadata` to the `config`. If there's a better way or more appropriate place then please advise.

I added the test to the same place as it was a very minor change to the test.

Regarding the AppVeyor failures,

`test_relative_git_url_submodule_clone` is failing because of the usual Windows permissions problems:

```
$ echo C:\\Users\\appveyor\\AppData\\Local\\Temp\\1\\tmpe_6ci39s\\.git\\objects\\56\\a6051ca2b02b04ef92d5150c9ef600403cb1de | wc -c
     106
```

I'm not sure what to recommend here, using `try:`/`finally:`/`with:` to handle automatic teardown is of course nice, but I'd like to be able to do some `exp_backoff` retries in some teardowns too, thoughts? I would hate to (re-)exclude any Git tests from running on Windows as they are more useful on Windows than other platforms, esp. given the uncertainty about which way to go with Git on/for Windows in conda.

.. the other two I'm not too sure about, seems like some massive env. var (PATH) is getting compared to a small value in a batch file overflowing the batch file command line processor for at least one of them.

msarahan:
I prefer that things don't get lumped together. It makes release notes easier to write. Unless later work depends on the original pr, please split it.

mingwandroid:
Do you want me to split the test into 2 test or to split this PR into two?

I bundled the test to save on test-suite execution time and code redundancy.

msarahan:
More just don't try to lump noarch stuff into this PR.  I'm happy to quickly merge this without it, if it helps you keep building noarch on top of this stuff.

mingwandroid:
Good to go now.

msarahan:
Python 3.5 errors on appveyor are due to PATH being too long.  There's a fix for conda at https://github.com/conda/conda/pull/3349

Merging.

PR #1274, opened 2016-08-25
find vs2010 before looking for win7sdk

msarahan:
Shuffles logic to try VS 2010 first.  Falls back to Win7SDK if cl.exe is not on PATH after trying to activate VS 2010.

Why? - I'm trying to build SVN, and the Win7SDK isn't handled correctly there.  It gets activated properly, but the build barfs anyway.  It's an error with their build, but I don't want to fix it.  Plus, it's really annoying that SetEnv.cmd turns my console green and clears it.

CC @patricksnape


msarahan:
FWIW, this does not fix my issue with SVN, I think.  It's also still activating the Win7 SDK on my system.  I'm just going to close this PR for now.

PR #1280, opened 2016-08-27
disallow backslashes in always_include_files pattern matching

### LINKED ISSUE #1194: Handling double slash
Would be nice if we could handle a double backslash [here]( https://github.com/conda/conda-build/blob/a8dbb9fa1919e91dbc28a2166b7e06054a414484/conda_build/build.py#L262 ). This came up in this [context]( https://github.com/conda-forge/conda-feedstock/pull/3#issuecomment-238559684 ).

cc @janschulz

msarahan:
fixes #1194 

CC @janschulz @jakirkham

### DIFF
--- a/tests/test_api_build.py
+++ b/tests/test_api_build.py
@@ -423,6 +423,8 @@ def test_debug_build_option(testing_workdir, test_config, caplog, capfd):
 @pytest.mark.skipif(not on_win, reason="only Windows is insane enough to have backslashes in paths")
 def test_backslash_in_always_include_files_path(test_config):
     api.build(os.path.join(metadata_dir, '_backslash_in_include_files'))
+    with pytest.raises(RuntimeError):
+        api.build(os.path.join(fail_dir, 'backslash_in_include_files'))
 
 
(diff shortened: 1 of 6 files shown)


mingwandroid:
From the subject it sounds like you are allowing them, which would be horrible, but you're actually explicitly disallowing them, which is good.

tests/test_api_build.py
```diff
@@ -420,6 +420,11 @@ def test_debug_build_option(testing_workdir, test_config, caplog, capfd):
     assert debug_message in caplog.text()
 
 
+@pytest.mark.skipif(not on_win, reason="only Windows is insane enough to have backslashes in paths")
+def test_backslash_in_always_include_files_path(test_config):
+    api.build(os.path.join(metadata_dir, '_backslash_in_include_files'))
+
+
```
janschulz: Shouldn't there also be a negative test (recipe with backslashes raise error)?

msarahan: Yes, that's a good improvement.  I will add one.

PR #1282, opened 2016-08-28
Prevent GIT_* env. vars containing newlines

mingwandroid:
Here you go @msarahan.


msarahan:
Good catch.  Thanks @mingwandroid

PR #1285, opened 2016-08-28
Add `build/noarch` to recipe metadata.

mingwandroid:
This is needed for many nodejs projects that generate OS independent
.js and .css files.


msarahan:
Python 3.5 errors on appveyor are due to PATH being too long. There's a fix for conda at conda/conda#3349

Merging.

PR #1290, opened 2016-08-30
add setuptools import to monkeypatch distutils

msarahan:
In playing with Vagrant, I've discovered a flaw with our handling of VC9: it only works if setuptools is imported before distutils, because setuptools monkeypatches distutils.  See https://bugs.python.org/issue23246

CC @patricksnape


patricksnape:
👍

PR #1304, opened 2016-09-01
remove pycache removal

### LINKED ISSUE #1301: Python 3 packages missing __pycache__ have .pyc
Seeing a number of occurrences of packages built on Python 3 with `conda-build` that have `.pyc` files instead of `__pycache__` directories. This is actually causing a world of hurt in some cases as it results in other packages potentially containing `__pycache__` directories for their dependencies due to our compilation of all Python files in the build directory. In turn this results in challenging to debug problems like this one ( https://github.com/conda-forge/conda-smithy/issues/289 ).

xref: https://github.com/conda-forge/setuptools-feedstock/issues/18

msarahan:
Fixes #1301

### DIFF
--- a/conda_build/post.py
+++ b/conda_build/post.py
@@ -22,6 +22,7 @@ from conda_build.os_utils import external
 from .conda_interface import lchmod
 from .conda_interface import walk_prefix
 from .conda_interface import md5_file
+from .conda_interface import PY3
 
 from conda_build import environ
 from conda_build import utils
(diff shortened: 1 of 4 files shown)


jakirkham:
What about this [line]( https://github.com/conda/conda-build/pull/1304/files#diff-60400bbbdb51124a0a8a06ca17ecec5cR114 )?

jakirkham:
Tried building `setuptools` with this patch and got the following.

```
import: 'easy_install'
  File "/opt/conda/conda-bld/test-tmp_dir/run_test.py", line 29, in <module>
    import easy_install
ImportError: No module named 'easy_install'
TESTS FAILED: setuptools-26.1.0-py35_0
```

That being said, I think that import test was wrong to begin with. That file doesn't get add to `site-packages` at that level. We were merely getting away with the fact a `pyc` file was kept for it. Though now that `pyc` file is in `__pycache__`. Arguably it shouldn't be there at all.

msarahan:
@jakirkham the line that you point out has to do with conda trying to take egg folders, rip all the stuff out, and make them "flat" packages.  It's unrelated to this PR.  It is certainly up for debate whether we should also remove that behavior, but that behavior has been around much longer.

Issue #1305, opened 2016-09-01
conda-build 2.0.0 bdist_conda error with metadata.name


jakirkham:
Running into an error building a package with `bdist_conda` on `conda-build` version `2.0.0`. Traceback below. More details in this [log]( https://travis-ci.org/jakirkham/splauncher/jobs/156896002#L355 ).

```
$ python setup.py bdist_conda
/home/travis/miniconda/lib/python2.7/distutils/dist.py:267: UserWarning: Unknown distribution option: 'build_requires'
  warnings.warn(msg)
running bdist_conda
Traceback (most recent call last):
  File "setup.py", line 105, in <module>
    zip_safe=True
  File "/home/travis/miniconda/lib/python2.7/distutils/core.py", line 151, in setup
    dist.run_commands()
  File "/home/travis/miniconda/lib/python2.7/distutils/dist.py", line 953, in run_commands
    self.run_command(cmd)
  File "/home/travis/miniconda/lib/python2.7/distutils/dist.py", line 972, in run_command
    cmd_obj.run()
  File "/home/travis/miniconda/lib/python2.7/distutils/command/bdist_conda.py", line 161, in run
    d['package']['name'] = metadata.name().lower()
TypeError: 'str' object is not callable
```

jakirkham:
Please see PR ( https://github.com/conda/conda-build/pull/1308 ).

PR #1308, opened 2016-09-01
Fix metadata retrieval in bdist_conda

jakirkham:
Fixes https://github.com/conda/conda-build/issues/1305

This seems to work locally for me based on testing with a simple package using `bdist_conda`.

However, it does raise some questions as this appears to be partially reverting PR ( https://github.com/conda/conda-build/pull/1281 ). Maybe I'm missing something. 😕


msarahan:
I'm going to make all of those things be properties long-term, but did not want to change too much in 2.0.  Thanks for the fix.

PR #1309, opened 2016-09-01
remove existing package with rm_rf instead of conda.

msarahan:
These files sometimes stick around, leading frustratingly to cached packages showing up in tests.  Very confusing.  Hopefully this helps.

CC @mingwandroid

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,7 +51,8 @@ script:
       python setup.py bdist_conda;
       popd;
     else
-      py.test -v -n 2 --cov conda_build --cov-report xml tests;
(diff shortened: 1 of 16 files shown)


mingwandroid:
This will help me out as I spent a very frustrating hour debugging a non-problem just now, but I wonder if it indicates something that needs to be fixed in conda? I know that `conda-forge` were complaining that openssl seemed to be leaving artifacts of the previous version [1] and it is possible that these share the same root cause.

 cc'ing @kalefranz

https://github.com/conda/conda/issues/3319

PR #1311, opened 2016-09-02
add disable_pip build option to disable conda's automatic add of pip

msarahan:
Conda makes it very hard to build pip and setuptools.  It has a config option, add_pip_as_python_dependency, but I could not seem to get that config option to change from conda-build.  This is a hacky workaround, removing pip and setuptools from the link instructions.

This adds a key to the build section of meta.yaml: ```disable_pip```

### DIFF
--- a/tests/test_api_build.py
+++ b/tests/test_api_build.py
@@ -603,3 +603,16 @@ def test_noarch(testing_workdir):
         output = api.get_output_file_path(testing_workdir)
         assert ("noarch" in output or not noarch)
         assert ("noarch" not in output or noarch)
+
+
+def test_disable_pip(test_config):
+    recipe_path = os.path.join(metadata_dir, '_disable_pip')
+    metadata, _, _ = api.render(recipe_path, config=test_config)
+
+    metadata.meta['build']['script'] = 'python -c "import pip"'
+    with pytest.raises(SystemExit):
+        api.build(metadata)
+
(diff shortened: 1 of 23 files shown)


msarahan:
CC @jakirkham - unless you are setting add_pip_as_python_dependency in condarc somehow, then setuptools and pip probably have some issues.  This should help sort them out.

jakirkham:
SGTM. Though I have set the aforementioned global flag as it has generally helped with building.

tests/test_api_build.py
```diff
@@ -603,3 +603,16 @@ def test_noarch(testing_workdir):
...
+    recipe_path = os.path.join(metadata_dir, '_disable_pip')
+    metadata, _, _ = api.render(recipe_path, config=test_config)
+
+    metadata.meta['build']['script'] = 'python -c "import pip"'
+    with pytest.raises(SystemExit):
+        api.build(metadata)
+
+    metadata.meta['build']['script'] = 'python -c "import setuptools"'
+    with pytest.raises(SystemExit):
+        api.build(metadata)
```
jakirkham: We should add a `wheel` import test too just for good measure.

jakirkham:
Referenced this to some of you. As for the rest, this provides a recipe level setting to disable `pip` and `setuptools`. We can discuss this more later like what role it may or may not play at conda-forge.

cc @pelson @ocefpaf @jjhelmus @minrk

PR #1313, opened 2016-09-02
Do not pass a Package object to VersionOrder

mingwandroid:
I was getting a problem with trying to apply strip to a Package() without this @msarahan.


mingwandroid:
Actually hold fire, there seems to be two instances of this ..

msarahan:
Dang. I SAS testing with strings. Need to be more careful. Thanks for looking into it.

PR #1326, opened 2016-09-06
provide means of obtaining rendered recipe with selectors still in

msarahan:


msarahan:
This was intended primarily as a tool for converting recipes from conda-forge, which follow the commented selector convention, to internal anaconda recipes, which do not comment selectors.  I don't think it's worth the trouble.  Closing.

PR #1328, opened 2016-09-06
Delete .pyo files before compiling .pyc files

mingwandroid:
Seems they are considered harmful:

https://www.python.org/dev/peps/pep-0488/

(anaconda-verify complains about them too)


msarahan:
Thanks @mingwandroid

PR #1330, opened 2016-09-07
Build wheels with conda-build

### LINKED ISSUE #1294: Use conda build to create Python wheels
Would there be support or interest for having the ability to use conda build to create Python [wheels](https://www.python.org/dev/peps/pep-0427/)?  

My though would be to use the conda build framework to create a conda environment where the command/script (typically `python setup.py bdist_wheel`) to build the wheel is executed.  Then copy this whl file to the conda build output directory in place of a post step.  The test phase would install this whl file  into the test environment before running the tests.  

Is this feature of interest to others?  I have a proof of concept implementation that I am willing to continue to develop if there is interest.

jjhelmus:
Add a --wheel argument to the 'conda build' command that when included uses
conda build to create and test a Python wheel file.

closes #1294

### DIFF
--- a/conda_build/api.py
+++ b/conda_build/api.py
@@ -63,7 +63,7 @@ def check(recipe_path, no_download_source=False, config=None, **kwargs):
 
 
 def build(recipe_paths_or_metadata, post=None, need_source_download=True,
-          build_only=False, notest=False, whl_build=False, config=None, **kwargs):
+          build_only=False, notest=False, config=None, **kwargs):
     import os
     from conda_build.build import build_tree
 
@@ -80,8 +80,7 @@ def build(recipe_paths_or_metadata, post=None, need_source_download=True,
             absolute_recipes.append(os.path.join(os.getcwd(), recipe))
 
     return build_tree(absolute_recipes, build_only=build_only, post=post, notest=notest,
-                      need_source_download=need_source_download,
-                      whl_build=whl_build, config=config)
+                      need_source_download=need_source_download, config=config)
 
 
 def test(recipedir_or_package_or_metadata, move_broken=True, config=None, **kwargs):
--- a/conda_build/api.py
+++ b/conda_build/api.py
@@ -63,7 +63,7 @@ def check(recipe_path, no_download_source=False, config=None, **kwargs):
 
 
 def build(recipe_paths_or_metadata, post=None, need_source_download=True,
-          build_only=False, notest=False, config=None, **kwargs):
+          build_only=False, notest=False, whl_build=False, config=None, **kwargs):
     import os
     from conda_build.build import build_tree
 
@@ -80,7 +80,8 @@ def build(recipe_paths_or_metadata, post=None, need_source_download=True,
             absolute_recipes.append(os.path.join(os.getcwd(), recipe))
 
     return build_tree(absolute_recipes, build_only=build_only, post=post, notest=notest,
-                      need_source_download=need_source_download, config=config)
+                      need_source_download=need_source_download,
+                      whl_build=whl_build, config=config)
 
 
 def test(recipedir_or_package_or_metadata, move_broken=True, config=None, **kwargs):
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -525,8 +525,7 @@ to get the latest version.
 """ % (installed_version, available_packages[-1]), file=sys.stderr)
 
 
-def build(m, config, post=None, need_source_download=True,
-          need_reparse_in_env=False, whl_build=False):
+def build(m, config, post=None, need_source_download=True, need_reparse_in_env=False):
     '''
     Build the package with the specified metadata.
 
@@ -748,7 +747,7 @@ can lead to packages that include their dependencies.""" % meta_files))
             copy_into(tmp_path, path, config.timeout)
         update_index(config.bldpkgs_dir, config)
 
-    elif whl_build:
+    elif config.wheel:
         # copy whl file to the build output directory
         src_dir = source.get_dir(config)
         whl_files = glob(os.path.join(src_dir, 'dist', '*.whl'))
@@ -819,7 +818,7 @@ def clean_pkg_cache(dist, timeout):
                 os.remove(lock._lock_file)
 
 
-def test(m, config, move_broken=True, whl_build=False):
+def test(m, config, move_broken=True):
     '''
     Execute any test scripts for the given package.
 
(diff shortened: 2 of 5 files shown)


jjhelmus:
The following meta.yaml file (based on the conda-forge [jinja2-feedstock](https://github.com/conda-forge/jinja2-feedstock)) can be used to create a Python wheel for Jinja2:

``` yaml
{% set version="2.8" %}

package:
  name: jinja2
  version: {{ version }}

source:
  fn: Jinja2-{{ version }}.tar.gz
  url: https://pypi.io/packages/source/J/Jinja2/Jinja2-{{ version }}.tar.gz
  md5: edb51693fe22c53cee5403775c71a99e

build:
  script: python setup.py bdist_wheel

requirements:
  build:
    - python
    - setuptools
    - wheel
    - markupsafe
  run:
    - python
    - setuptools
    - wheel
    - markupsafe

test:
  imports:
    - jinja2

about:
  home: http://jinja.pocoo.org
  license: 3-Clause BSD
```

The build is run as follows:

```
$ conda build --wheel .
BUILD START: jinja2-2.8-py35_0
    (actual version deferred until further download or env creation)
...
===== testing package: jinja2-2.8-py35_0 =====
import: 'jinja2'
===== jinja2-2.8-py35_0 OK =====
TEST END: jinja2-2.8-py35_0
INFO:/Users/jhelmus/dev/conda-forge/conda-build/conda_build/config.py:--keep-old-work flag not specified.  Removing source and build files.

# If you want to upload this package to anaconda.org later, type:
#
# $ anaconda upload /Users/jhelmus/anaconda/conda-bld/osx-64/Jinja2-2.8-py2.py3-none-any.whl
#
# To have conda build upload to anaconda.org automatically, use
# $ conda config --set anaconda_upload yes
```

jjhelmus:
A preprocessing selector for "wheel" builds would be a good addition if this is accepted as is so wheel only requirements or build scripts can be specified in the recipe metadata.

msarahan:
Very interesting.  I think we have quite a bit of work before this is final, but it's a great start.  In particular:

- I'd like it if the setup command did not have to change, somehow.  Rather, that conda-build provided the bdist_wheel bit when building for a wheel
- I think it would be better to add the whl_build argument as a new part of the config, rather than an argument.  You can pass keyword arguments to the Config constructor, or to ```get_or_merge_config```, and they'll be stored as attributes.  In particular, by adding a new argument to the api build function, I feel like I've promised to bump the major version.
- I'm itching to break up output types into modular options, but I might be able to restrain myself for a little while.

Thanks for putting this forward!  Do you want to iterate on it here more, or would you prefer that I merge and take over?

jjhelmus:
I'm happy to iterate on this PR. I agree that there are many things than can (and should) be corrected before this is finalized.

> I'd like it if the setup command did not have to change, somehow. Rather, that conda-build provided the bdist_wheel bit when building for a wheel

I would like this too, but I think there are cases where just running `python setup.py bdist_wheel` is not sufficient and being able to run a build script with all the capabilities conda build provides will be a great help.  Some examples I can think of are:

* Setting pre-build environment variables or flags.
* Running [`auditwheel repair`](https://github.com/pypa/auditwheel) or [`delocate`](https://github.com/matthew-brett/delocate) to include additional compiled libraries in the wheel file.

Some of utilities could be included in conda build itself but I think a more flexible (and likely maintainable) options is to allow a build script to call out to other tools that might be needed.  

Perhaps adding a `wheel_script` entry to the `build` section of the yaml would be logical.  If this was missing then `python setup.py bdist_wheel` would be used.  As I mentioned in an earlier comment a `whl` selector would be another option.

> I think it would be better to add the whl_build argument as a new part of the config, rather than an argument. You can pass keyword arguments to the Config constructor, or to get_or_merge_config, and they'll be stored as attributes. In particular, by adding a new argument to the api build function, I feel like I've promised to bump the major version.

Agreed, I'll work on changing this.  Do you have an issue or alternative to storing the path to the produced whl file in the config?

msarahan:
> Some of utilities could be included in conda build itself but I think a more flexible (and likely maintainable) options is to allow a build script to call out to other tools that might be needed.

Yes, definitely.  The tricky part is *when* the external tool gets called.  There can and probably should be several different hooks.  For python, I think we can get away with jinja stuff pretty well:

in meta.yaml:
```
build:
  script:
    - python setup.py {{ SETUP_PY_COMMAND }}
```

and we can have conda-build set that to install, or (install + setuptools junk) or bdist wheel, as necessary.

Now, the interesting question is whether the collection of files is something that should be packaged up (this is our current assumption), or something that is a package itself, and should be passed through.

> Do you have an issue or alternative to storing the path to the produced whl file in the config?

Not especially, though I think answering the question of whether the list of changed files is a collection of things to be packaged or actual packages to be collected sort of obviates this.

jjhelmus:
I adjusted the wheel building logic to store the `--wheel` flag setting in the config object which preserves the currently functions in conda_build.api.  

Added a _wheel_ selector which can be used in the `meta.yaml` file to specify lines which should be included only for wheel builds (or non-wheel builds with the [not wheel] selector).  This allows a single `meta.yaml` to be used to build both wheels and conda packages.  For example:

``` yaml
{% set version="2.8" %}
{% set setup_py_command="bdist_wheel" %}  # [wheel]
{% set setup_py_command="install --single-version-externally-managed --record=record.txt" %}  # [not wheel]

package:
  name: jinja2
  version: {{ version }}

source:
  fn: Jinja2-{{ version }}.tar.gz
  url: https://pypi.io/packages/source/J/Jinja2/Jinja2-{{ version }}.tar.gz
  md5: edb51693fe22c53cee5403775c71a99e

build:
  script: python setup.py {{ setup_py_command }}

requirements:
  build:
    - python
    - setuptools
    - wheel        # [wheel]
    - markupsafe
  run:
    - python
    - setuptools
    - wheel        # [wheel]
    - markupsafe

test:
  imports:
    - jinja2

about:
  home: http://jinja.pocoo.org
  license: 3-Clause BSD
```

jjhelmus:
> Now, the interesting question is whether the collection of files is something that should be packaged up (this is our current assumption), or something that is a package itself, and should be passed through.

A number of output formats (wheels, rpms, deb) produce a single file archive which is a collection of files.  I'm guessing conda build may in the future want to to produce and handle these formats so being able to pass these file through seems to make sense to me.  Copying them somewhere under the conda build output directory also makes sense to me, although the same directory as as the conda packages, as this PR is currently doing, might not be the best solutions.  Would a CONDA_BLD_PATH/whl directory be more reasonable?

msarahan:
Superseded by #1576 - @jjhelmus please take a look there and let me know what you think.  I'm writing up a docs PR now.

msarahan:
Docs PR at https://github.com/conda/conda-docs/pull/388

PR #1332, opened 2016-09-07
Consider all given recipes when printing --output.

johanneskoester:
Previously, only the first recipe was considered.
In addition, if build/skip is true for a recipe and configuration, "skipped" will be printed instead of the output file.

@msarahan this basically restores the old behavior of --output, which is needed for the build system of Bioconda. Please let me know what you think.


conda_build/cli/main_build.py
```diff
@@ -158,8 +158,10 @@ def parse_args(args):
 
 def output_action(metadata, config):
     silence_loggers(show_warnings_and_errors=False)
-    print(bldpkg_path(metadata, config))
-    sys.exit(0)
+    if metadata.skip():
+        print("skipped")
```
msarahan: This is confusing to me.  What is the point of printing just "skipped"?  You have no information about what was skipped.

msarahan:
Thanks @johanneskoester.  I don't see where this is looping over all input recipes, though.  Can you add a test case that makes sure that this is doing what you expect (and to make sure that if anyone breaks your desired behavior in the future, that it is a conscious choice, not an accident)?  See the tests/test_cli.py file for example of how to do it - you just create a list of arguments, then pass them to ```main_build.execute```

Side note: please avoid PRs from master in general.  This one is fine, but please make a branch next time.

conda_build/cli/main_build.py
johanneskoester: In bioconda, we zip the output of this with the recipe names and filter accordingly. I can add more information though.

johanneskoester:
Sorry for the master issue. I'll do that next time. The loop is in your code [here](https://github.com/conda/conda-build/pull/1332/files#diff-e392cf87d7fdbe7c8fbbb88d1c6f7935L217). Previously, the output stopped after the first item because of the `sys.exit`. I'll see if I can provide a test case today.

tests/test_cli.py
```diff
@@ -85,6 +85,24 @@ def test_render_output_build_path_set_python(testing_workdir, capfd):
...
+    test_path = lambda pkg: os.path.join(sys.prefix, "conda-bld", test_config.subdir, pkg)
+    test_paths = [test_path(
+        "conda-build-test-python-run-1.0-py{}{}_0.tar.bz2".format(
+        sys.version_info.major, sys.version_info.minor)),
+        "Skipped: {} defines build/skip for this "
+        "configuration.".format(os.path.abspath(skip_recipe))]
+
+    output, error = capfd.readouterr()
+    assert error == ""
+    assert output.rstrip().splitlines() == test_path, error
```
msarahan: I think you mean

```
== test_paths
```

here you are instead comparing to your lambda.

johanneskoester: Exactly :-)

johanneskoester:
@msarahan I believe this is ready for merging now. We have a proper skip message, which is shared shared between normal build and --output. There is a test case. Please note that I had to move the test case above test_render_output_build_path_set_python, because that test apparently causes a side-effect, leading to all following tests assuming either python 3 or python 2, depending on whether I was on Python 2 or 3 before.

msarahan:
Thanks for pointing out the state modification - that's a big mistake waiting to happen.  Sorry for any trouble it caused you.  Appveyor is in crazy shape right now.  I'll wait for it, but will still merge here as long as your tests are passing.

johanneskoester:
No problem. Thank you! Can you give me a rough estimate when the next release will come (containing these changes)?

msarahan:
probably early-mid next week.  Are you OK running your patched master until then?

johanneskoester:
Yes, that should be fine. Thanks!

msarahan:
Appveyor looks clear on your test.  Merging.

PR #1333, opened 2016-09-07
More pyo pyc

mingwandroid:
Placates `anaconda-verify` and saves some junk from getting into our packages.

I slipped the `LGPL` thing in too as it's so minor. If that's a problem I can revisit.


msarahan:
LGTM.  Will wait on CI in case I've missed anything dumb.

PR #1334, opened 2016-09-07
Insert value of noarch into index file

soapy1:
Raise error for invalid noarch values, in this case 'none'


soapy1:
@msarahan when you have a second please review 🌵 

cc/ @kalefranz

msarahan:
Looks great to me.  Can you please also add a test?  The simplest way would be to just copy this folder: https://github.com/conda/conda-build/tree/master/tests/test-recipes/metadata/noarch

and then change the meta.yaml inside.  That's all you need - the test suite finds recipe folders in the metadata folder and builds each one as a test case.

msarahan:
Looks done to me.  I'll let the CIs run, and merge if/when they look happy.

tests/test-recipes/metadata/noarch_foo/meta.yaml
```diff
@@ -0,0 +1,6 @@
+package:
+  name: noarch_foo_test_package
+  version: "1.0"
+
+build:
+  noarch: foo
```
msarahan: I think it would be good to add the test for this, so that you verify that the metadata accurately reflects your setting here.  I think you might need to break this test out (add an _ to the folder name) and manually run the test, like you do for the noarch_none_value.

The test probably looks something like:

```
def sometest(test_config):
    fn = api.get_output_file_path(recipe_path, config=test_config)
    api.build(recipe_path, config=test_config)
    metadata = json.loads(package_has_file(fn, 'index.json'))
    assert 'noarch' in metadata
    assert metadata['noarch'] == foo
```

I think index.json is the right file, but it might be info.json.

kalefranz: > I think index.json is the right file, but it might be info.json.

I think index.json is the safest file.  I don't yet perfectly clear visibility on all the ways we might use this, and it's possible that the solver might need to have knowledge of noarch type at some point.  I think putting it in index.json will be a hedge so we don't limit ourselves going forward.

soapy1: Added this test in commit  d86dc72

tests/test_api_build.py
```diff
@@ -62,21 +62,6 @@ def describe_root(cwd=None):
     return tag
 
 
-def package_has_file(package_path, file_path):
```
msarahan: For the record, I asked @soapy1 to remove this.  It is a duplicate definition, also defined in conda_build/utils.py - and the definition there is better.

tests/test_api_build.py
```diff
@@ -600,3 +585,18 @@ def test_disable_pip(test_config):
...
+
+
+def test_noarch_foo_value():
+    recipe = os.path.join(metadata_dir, "noarch_foo")
+    fn = api.get_output_file_path(recipe)
+    api.build(recipe)
+    metadata = json.loads(package_has_file(fn, 'info/index.json').decode())
+    assert 'noarch' in metadata
+    assert metadata['noarch'] == "foo"
+
```
msarahan: Extra newline here.  Some people care.  I don't.  Just pointing it out.

msarahan:
Appveyor is a mess of unrelated failures.  Merging.  Thanks @soapy1

Issue #1300, opened 2016-09-01
Add GPL to license_family


jakirkham:
Apparently one must specify `GPL2` or `GPL3` in `license_family` instead of using `GPL`. Such a constraint does not appear for `LGPL` or `AGPL`. Also, this defeats the point of `license_family` IMHO as it is now duplicating the `license` field which says as much. In other words, this is no longer a family and just the license. Could we please add `GPL` to the list of supported values for `license_family`?

xref: https://github.com/conda-forge/staged-recipes/pull/1378#issuecomment-244108997

jakirkham:
Please see PR ( https://github.com/conda/conda-build/pull/1340 ).

PR #1340, opened 2016-09-12
Add `GPL` as a license family

jakirkham:
Closes https://github.com/conda/conda-build/issues/1300

As stated, adds `GPL` as a license family. This seems more appropriate as the `license` itself should have license version; whereas the license family should merely note the style of license. It is also more consistent as no other license family listed requires version in this way.

Note: Would like to drop `GPL2` and `GPL3` from the list, but have kept them as that feels like a breaking change.

### DIFF
--- a/conda_build/metadata.py
+++ b/conda_build/metadata.py
@@ -140,6 +140,7 @@ allowed_license_families = set("""
 AGPL
 Apache
 BSD
+GPL
 GPL2
 GPL3
 LGPL


msarahan:
Thanks - makes sense to me.  Agree about keeping GPL2 and GPL3, unfortunately.

jakirkham:
Thanks. No worries. Completely understand.

PR #1341, opened 2016-09-12
WIP: Concatenate multiple license files

jakirkham:
Fixes https://github.com/conda/conda-build/issues/1247

This is still WIP. Though maybe this will help start the discussion and provide some useful feedback.

Basically this tries to write all the license files from a list into one `LICENSE.txt`. It seems useful to have one place to look at for this info even if there are multiple licenses. An alternative strategy would be to include multiple license files like `LICENSE.1.txt` and so on.

One thing I'm a little unsure of is we appear to have been using a util function that had an API lock previously. This is not doing that right now (though it probably should). Is there a write util function that is equivalent? If not, how should we proceed.

There could also be some nice tweaks like mention where each text blob came from.

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -56,6 +56,13 @@ from conda_build.create_test import (create_files, create_shell_files,
 from conda_build.exceptions import indent
 from conda_build.features import feature_list
 
+
+try:
+    basestring
+except NameError:
+    basestring = str
+
+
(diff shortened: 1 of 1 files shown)


jakirkham:
Closing this out. Had a hard time trying to get this to work and simply don't have the bandwidth to continue.

Issue #1007, opened 2016-06-07
During test phase cwd gets deleted


marscher:
For instance if one does a stack inspection during testing, the absolute path to the source files will be looked up. This will fail, because the _test directory is deleted during testing somehow and this is the current working directory.

Please note that this for now only popped up in conjunction with pytest. Fixing pytest-dev/pytest#1235 will only partially solve this problem, since every call to os.getcwd() will raise, no matter where it occurs.

This is a blocker for using pytest in testing phase - IMHO.

Thank you very much for taking a look.

msarahan:
Just a note to say I'm aware of this, and hope to work on it soon.

If you could post your use case, or a build log showing what you do, that would be helpful.

msarahan:
Never mind, just hit this myself while working on a conda-build recipe for conda-build.  I'll try to figure out where these temporary directories are misbehaving.  Thanks for the background info at pytest.

msarahan:
@marscher Can you explain why you think _test is cwd?  I see that _test is deleted at https://github.com/conda/conda-build/blob/master/conda_build/build.py#L665-L672

and then recreated at https://github.com/conda/conda-build/blob/master/conda_build/build.py#L693

but I don't see why you would be in the _test prefix, unless you are intentionally cd'ing there.  Do you mean the test-tmp folder?  In playing with conda-build, that seems to get me into hot water:

https://github.com/conda/conda-build/blob/master/conda_build/build.py#L644-L645

With conda-build making a conda-build package, I'm in that folder when it is removed, and I'm pretty sure this triggers the pytest issue you linked to.  Is there another way that conda-build trips over itself, or do you also have a way of being in test-tmp when conda-build tries to clean it up?

marscher:
Actually I'm not changing directory neither to _test nor test-tmp. So 
I'd guess that cwd is being set to one of these directories during 
test-phase.

When I print cwd within my run_test.py script, it points to
~/miniconda3/conda-bld/test-tmp_dir

So the culprit has to be the removal/recreation of this dir.

marscher:
Maybe it is sufficient to change directory to the newly created dir, since the inode changes during re-creation.

marscher:
Still present in 2.0.1.

msarahan:
Thanks for the update. I will try to reproduce and fix for the next
release.

marscher:
Am 09.09.2016 um 14:13 schrieb Mike Sarahan:
[...]

PR #1347, opened 2016-09-13
add test for pytest failing when tmpdir not present

msarahan:
Investigates #1007 

From what I can tell on my computer, this does not fail.  It is a recreation of the test at https://github.com/pytest-dev/pytest/issues/1235

If this passes, we need more information in #1007


msarahan:
@marscher - since this passes, I need more information from you.  I changed the test here to not fail on the assertion error, but otherwise this is the same test reported in the original issue.  Is this a valid test for your problem?

marscher:
The original issue is fixed in pytest-3+, however the recipe is still on 2.9.3. The fix circumvent pytest from crashing, but still causes problem due to the cwd removal. I'm not sure if the provided test will fail, since I get the error during a source code inspection (which will try to lookup the path).

The output can be found here:
https://travis-ci.org/marscher/PyEMMA/jobs/145554837#L1340

test-script:
https://github.com/marscher/PyEMMA/blob/pytest/devtools/conda-recipe/run_test.py

msarahan:
Thanks - I'm testing with conda's latest, which still seems to be 2.9.x.  I have cloned PyEMMA from you and am running the tests now.  Is there a simpler minimum reproducible test case, though?

The test case I picked up from that other issue on the pytest issue tracker seemed to reproduce the problem.  Is there something you would add to it?  Something like copying a file to the test folder in meta.yaml, and then running pytest with tests that depend on that file being there?

marscher:
On 14.09.2016 00:29, Mike Sarahan wrote:
> Thanks - I'm testing with conda's latest, which still seems to be 2.9.x.
> I have cloned PyEMMA from you and am running the tests now. Is there a
> simpler minimum reproducible test case, though?
Sorry, but is seems hard to isolate this...
>
> The test case I picked up from that other issue on the pytest issue
> tracker seemed to reproduce the problem. Is there something you would
> add to it? Something like copying a file to the test folder in
> meta.yaml, and then running pytest with tests that depend on that file
> being there?
Actually is is only a custom run_test.py and yes it copies the coverage 
report after pytest returns. But pytest-2.9.x crashes before that happens.

msarahan:
Sorry - I have yet to repeat this with either PyEMMA or my test cases.  It's definitely not crashing for you immediately - does it always crash on the same test, or is it random on where it does this internal error mess?  I have also seen internal errors sometimes with our test suite, but chalked them up to improper locking in parallel test running - not something that should be related to you here.

If we can't get this nailed down tonight, I'll tag 2.0.2 without it, but will keep working on it.

msarahan:
@marscher is this still an issue?

marscher:
No it has been fixed in pytest (or conda-build). Thank you very much for our effort!

PR #1356, opened 2016-09-15
Fix availability of CONDA_NPY when rendering recipe.

johanneskoester:
This fixes a bug where CONDA_NPY was not set when using the CONDA_NPY environment variable only.

### DIFF
--- a/conda_build/config.py
+++ b/conda_build/config.py
@@ -60,7 +60,10 @@ class Config(object):
         self.CONDA_PY = int(env('python', "%s%s" % (sys.version_info.major, sys.version_info.minor))
                         .replace('.', ''))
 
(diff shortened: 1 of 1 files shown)


johanneskoester:
@msarahan this is a final hotfix needed for bioconda. Without this fix, recipes containing `numpy x.x` fail to render when the `CONDA_NPY` environment variable is used instead of the `--numpy` CLI argument. Maybe it can still go into the next release?

msarahan:
Thanks.  To be clear, what is happening here is that in the single-line approach, we are getting a value for ```numpy```, but that value is None.  Your logic fixes that case.

The tests on Travis are failing to run pretty badly.  I'm OK merging this, but need to fix travis and appveyor before I can release.

PR #1364, opened 2016-09-15
Fix building when --skip-existing flag is set

johanneskoester:
This PR fixes a bug when using the --skip-existing flag: any package, regardless of whether it exists or not, was not being built, because the `return False` appeared at the wrong indentation level.

Further, this adds a dependency to requests, which was previously used but not mentioned in the setup.py.


johanneskoester:
@msarahan I have seen a commit with a message similar to what I propose her. I guess this is fixed in master then?

msarahan:
I think this is fine.  I do recall something else similar, but there would be a merge conflict if this were the same thing.  Merging.

PR #1366, opened 2016-09-16
Noarch python copy files

soapy1:


conda_build/build.py
```diff
@@ -236,7 +236,10 @@ def detect_and_record_prefix_files(m, files, prefix, config):
         else:
             files_with_prefix = []
 
-    if files_with_prefix and not m.get_value('build/noarch_python'):
+    is_noarch = m.get_value('build/noarch_python') or \
+        str(m.get_value('build/noarch')).lower() == "python"
```
msarahan: Stylistically, I prefer to start the statement after = with a (, so that the \ line continuation is unnecessary.  I find that this prevents errors later when editing these compound statements.

conda_build/build.py
```diff
@@ -336,6 +339,12 @@ def create_info_files(m, files, config, prefix):
     write_info_json(m, config, mode_dict)
     write_about_json(m, config)
 
+    if str(m.get_value('build/noarch')).lower() == "python":
```
msarahan: Since this expression is a bit ugly, and you use it multiple times, I recommend turning it into a function.

conda_build/build.py
```diff
@@ -336,6 +339,12 @@ def create_info_files(m, files, config, prefix):
...
+        import conda_build.noarch_python as noarch_python
```
msarahan: Unless you have very good reason, keep your imports at the top of the file.

conda_build/build.py
```diff
@@ -685,13 +694,16 @@ def build(m, config, post=None, need_source_download=True, need_reparse_in_env=F
 
         get_build_metadata(m, config=config)
         create_post_scripts(m, config=config)
-        create_entry_points(m.get_value('build/entry_points'), config=config)
+
+        if str(m.get_value('build/noarch')).lower() != "python":
```
msarahan: another one to move to a function (same one as above)

conda_build/build.py
```diff
@@ -685,13 +694,16 @@ def build(m, config, post=None, need_source_download=True, need_reparse_in_env=F
...
         files2 = prefix_files(prefix=config.build_prefix)
 
         post_process(sorted(files2 - files1),
-                        prefix=config.build_prefix,
-                        config=config,
-                        preserve_egg_dir=bool(m.get_value('build/preserve_egg_dir')))
+                     prefix=config.build_prefix,
+                     config=config,
+                     preserve_egg_dir=bool(m.get_value('build/preserve_egg_dir')),
+                     noarch= m.get_value('build/noarch'))
```
msarahan: extra space here between = and m.  Are you running flake8?  I think pycharm can do it on the fly for you.

conda_build/build.py
```diff
@@ -708,9 +720,13 @@ def build(m, config, post=None, need_source_download=True, need_reparse_in_env=F
                     croot=config.croot)
         create_info_files(m, sorted(files2 - files1), config=config,
                             prefix=config.build_prefix)
+
         if m.get_value('build/noarch_python'):
             import conda_build.noarch_python as noarch_python
             noarch_python.transform(m, sorted(files2 - files1), config.build_prefix)
+        elif str(m.get_value('build/noarch')).lower() == "python":
+            import conda_build.noarch_python as noarch_python
```
msarahan: move import to top

conda_build/build.py
```diff
@@ -708,9 +720,13 @@ def build(m, config, post=None, need_source_download=True, need_reparse_in_env=F
                     croot=config.croot)
         create_info_files(m, sorted(files2 - files1), config=config,
                             prefix=config.build_prefix)
+
         if m.get_value('build/noarch_python'):
             import conda_build.noarch_python as noarch_python
             noarch_python.transform(m, sorted(files2 - files1), config.build_prefix)
+        elif str(m.get_value('build/noarch')).lower() == "python":
```
msarahan: same function as above

conda_build/noarch_python.py
```diff
@@ -56,6 +56,13 @@ def rewrite_script(fn, prefix):
     return fn
 
 
+def create_entry_point_information(noarch_type, entry_points, config):
+    entry_point_information = {"type": noarch_type, "entry_points": entry_points}
+    file = os.path.join(config.info_dir, "noarch.json")
```
msarahan: Is @ilanschnell on board with this extra file?  It's fine with me.

conda_build/post.py
```diff
@@ -180,9 +187,12 @@ def compile_missing_pyc(files, cwd, python_exe):
             call([python_exe, '-Wi', '-m', 'py_compile', f], cwd=cwd)
 
 
-def post_process(files, prefix, config, preserve_egg_dir=False):
+def post_process(files, prefix, config, preserve_egg_dir=False, noarch=False):
     rm_pyo(files, prefix)
-    compile_missing_pyc(files, cwd=prefix, python_exe=config.build_python)
+    if not noarch:
```
msarahan: I'd flip this around.  Logic is simpler when not negated.

```
if noarch:
    rm_pyc()
else:
    compile_missing_pyc
```

The argument against might be that it is a less common code path, and is thus a little slower.  I don't think this is a critical speed path.

tests/test-recipes/metadata/noarch_python/meta.yaml
```diff
@@ -0,0 +1,28 @@
...
+    - noarch_python_test_package_script = noarch_python_test_package:main
+
+requirements:
+  build:
+    - python
+    - setuptools
+  run:
+    - python
+
+# Tests are commented out because they should fail since conda does not yet know how to install
```
msarahan: I think it's better to keep these enabled, but write the tests as xfail.  This means they'll be less likely to get lost.

tests/test-recipes/metadata/noarch_python/noarch_python_test_package/README
```diff
@@ -0,0 +1 @@
+Simple package to test noarch package building.
```
msarahan: not really necessary, unless you're going to go into more detail about what this test package does exactly

tests/test-recipes/metadata/noarch_python/run_test.py
```diff
@@ -0,0 +1,24 @@
+# Tests are commented out because they should fail since conda does not yet know how to install
+# noarch: python packages. Once it is implemented then this should be uncommented.
```
msarahan: Again, enable these, but make the test be an xfail.

tests/test_api_build.py
```diff
@@ -601,3 +601,13 @@ def test_noarch_foo_value():
     metadata = json.loads(package_has_file(fn, 'info/index.json').decode())
     assert 'noarch' in metadata
     assert metadata['noarch'] == "foo"
+
+
+def test_noarch_python():
+    recipe = os.path.join(metadata_dir, "noarch_python")
```
msarahan: Note that since noarch_python is a folder that is in ```metadata_dir``` and does not start with _, it will be built automatically as part of the test_recipe_builds test, which collects all such folders.  You probably want to rename your noarch_python folder to start with an underscore, and then also make this particular test be an xfail right now.

You might also want to break it up into more tests, so that you can only have smaller parts xfail (the parts you have commented now), while you test known-good parts in other tests.

conda_build/build.py
```diff
@@ -1140,3 +1138,7 @@ def is_package_built(metadata, config):
     # will be empty if none found, and evalute to False
     package_exists = [url for url in urls if url + '::' + metadata.pkg_fn() in index]
     return package_exists or metadata.pkg_fn() in index
+
+
+def is_noarch_python(build_noarch):
+    return str(build_noarch).lower() == "python"
```
msarahan: Good start.  I think you can make it even shorter by passing ```m``` directly, and then putting the ```get_value``` inside this function.

tests/test_api_build.py
```diff
@@ -608,3 +609,20 @@ def test_noarch_foo_value():
     metadata = json.loads(package_has_file(fn, 'info/index.json').decode())
     assert 'noarch' in metadata
     assert metadata['noarch'] == "foo"
+
+
+@pytest.mark.xfail
```
msarahan: Please add an explanation of why this is an xfail.  You can do this as a comment, or as the ```reason``` argument to xfail.

msarahan:
Thanks @soapy1

PR #1370, opened 2016-09-18
Problem: When using conda-build --skip-existing, packages won't get built if they are not available

gabm:
Cause: intendation of return false is wrong


msarahan:
Good catch - thanks for debugging this.  Is there a test that we can add to ensure that this does not regress again?

gabm:
Well I'm not sure about your test suite, but i think i remember that there is a dry-run flag .. but even if, the testcase would be quite sophisticated: 
- the package to be built needs to be available (local or remote)
- then the package needs to be built again using --skip-existing and --dry-run
- the output should indicate that something would have been built

msarahan:
Thanks. That outline is exactly what I wanted.

PR #1380, opened 2016-09-21
A couple of things.

mingwandroid:
Hi @msarahan, sorry about combining two unrelated things here, I'm rushing around to get things done.

Feel free to merge either or none until we can discuss in more detail, I guess neither is hugely urgent. Adding the build prefix during testing may not be wanted though, personally, I think it can be argued to make sense. I made it so that the test prefix appears first in PATH.


msarahan:
I disagree with adding the build env during testing.  My argument is that doing so can hide missing dependencies.  I welcome your thoughts, though.

PR #1381, opened 2016-09-21
Do not fatal on index if _license package in ${CONDA_PREFIX}/pkgs

mingwandroid:
As we know it is not a mirror.

@msarahan, @dsludwig, @ilanschnell, @quasiben please check that this looks reasonable.


ilanschnell:
Looks good to me.

mingwandroid:
I'm on vacation now so if people are OK with this then merge away!

msarahan:
Thanks @mingwandroid

PR #1388, opened 2016-09-26
don't index pkgs folder

### LINKED ISSUE #1339: Unable to acquire lock on Windows since 2.0
I'm currently unable to build on our build servers as conda-build just keeps timing out due to not being able to acquire the lock in time. I've changed down to only 1 build at a time and I'm still seeing the error. I run `conda clean --lock` before every build.

[Example 1](https://jenkins.menpo.org/job/menpo-pr/309/ARCH=x86,PYTHON_VERSION=3.4,label=windows/console)
[Example 2](https://jenkins.menpo.org/job/menpo-pr/309/ARCH=x86,PYTHON_VERSION=3.5,label=windows/console)

**Error**
```
File lock could on C:\miniconda\0\x86\pkgs\.conda_lock not be obtained.  You might need to try fewer builds at once.  Otherwise, run conda clean --lock
```

**System**
```
conda-build               2.0.1                    py35_0  
conda                     4.1.12                   py35_0
```
```
Windows 10 x64
NTFS File system - old 7200 RPM drive
```

This is likely because the machine is pretty slow. I'm wondering if it might be possible to have a flag for conda-build to ignore the locking mechanism as, in this case, I am absolutely certain only one build will be activated at a time. Or perhaps a flag to be able to customise the timeout interval?

msarahan:
Fixes #1339 

(maybe)

It seems like indexing the pkgs folder when it has never been indexed before takes quite a long time.  Since this was not done before, and because the index creation is not incremental - this can simply fail all builds because it never succeeds a first time (after which it would be fast).

CC @patricksnape

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -407,10 +407,10 @@ def create_env(prefix, specs, config, clear_cache=True):
                     if not os.path.isdir(folder):
                         os.makedirs(folder)
                     lock = filelock.SoftFileLock(join(folder, '.conda_lock'))
-                    update_index(folder, config=config, lock=lock, could_be_mirror=False)
-                    locks.append(lock)
-                for lock in locks:
                     lock.acquire(timeout=config.timeout)
+                    if not folder.endswith('pkgs'):
+                        update_index(folder, config=config, lock=lock, could_be_mirror=False)
+                    locks.append(lock)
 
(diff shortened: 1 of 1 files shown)


patricksnape:
I will try that either with the next release or when I can re-jig our build system to not overwrite a pip installed conda-build. Thanks!

PR #869, opened 2016-04-07
handle empty files. closes #868

brentp:
This adds an empty file to one of the recipes. Without this fix, the error reported in #868 manifests in tests/test_build_recipes.py

Any suggestions on how to test better? How/where to check if `empty.py` appears in the .bz2

PR #1389, opened 2016-09-26
try lstat and stat before memmapping

msarahan:
Curious if this is an answer to https://github.com/conda/conda-build/pull/869#pullrequestreview-1424290

@asmeurer is having trouble with just stat here - possibly with a broken symlink?


msarahan:
I'm not at all sure that this is the fix, but I'm tagging the next bugfix release, and I'm willing to give it a try.  Merging.

Issue #1379, opened 2016-09-21
Cannot set package name from load_setup_py_data()['name']


ma-ba:
`conda-build` fails if package name is set from `setup.py` data using Jinja. See:
```python
# setup.py
from setuptools import setup
setup(name='package_name')
```
```yaml
# meta.yaml
{% set data = load_setup_py_data() %}
package:
  name: {{ data['name'] }}
source:
  path: ./
```
```bash
+(root) $ conda build --no-anaconda-upload . 1>/dev/null
Error: package/name missing in: '/home/maba/dev/conda/conda-build/bugs/jinja-setup-py-name/meta.yaml'
```
Concerning the inaccurate error message see #1378.

However, loading using `from_recipe_dir=True` works:
```yaml
# meta.yaml
{% set data = load_setup_py_data(from_recipe_dir=True) %}
package:
  name: {{ data['name'] }}
source:
  path: ./
```
```bash
+(root) $ conda build --no-anaconda-upload . 1>/dev/null
```

ma-ba:
BTW: This also breaks `conda_build/templates/setuptools.yaml` which uses:
```jinja
{% set data = load_setup_py_data()%}
{% block body -%}
{% block package -%}
package:
  name: {{data.get('name').lower().replace(' ', '_')}}
...
```
In consequence `example_packages/setuptools/conda.yaml` is broken since it consists of:
```jinja
{% extends "setuptools.yaml" %}
```

ma-ba:
This issue will be resolved once #1379 is resolved by #1395.

lijunzh:
This problem re-occurred for me.

I have `meta.yaml` as following:

```
{% set data = load_setup_py_data() %}
package:
    name: {{ data['name'] }}
    version: {{ data['version'] }}

source:
    path: .
```

It throws error:

```
Error: package/name missing in: '/Users/lijun/git/vision/meta.yaml'
```

but when I type in the package name explicitly as 
```
{% set data = load_setup_py_data() %}
package:
    name:  vision
    version: {{ data['version'] }}

source:
    path: .
```

It works.


My `conda build` version is as follows:
```
$ conda build --version
conda-build 3.12.0
```

msarahan:
I don't find it too much to ask that the name be set from data in meta.yaml.  You can use variables defined in meta.yaml, but you can't depend on external definitions.  Sorry, but I have no plans to fix this.  If you want it to work, please consider fixing it yourself and submitting a PR.

[...]

PR #1395, opened 2016-09-26
Fix #1378 and #1379

### LINKED ISSUE #1378: Jinja errors in package name evaluation are masked
Jinja errors inside package name expression are overridden / masked by `Error: package/name missing in ...`. See:
```yaml
package:
  name: {{ some_erroneous_expression }}
```
```bash
+(root) $ conda build --no-anaconda-upload . 1>/dev/null 
Error: package/name missing in: '/home/maba/dev/conda/conda-build/bugs/jinja-package-name-error-masked/meta.yaml'
```
Notice that errors for other fields are not masked:
```yaml
package:
  name: some_package
  version: {{ some_erroneous_expression }}                                                                               
```
```bash
+(root) $conda build --no-anaconda-upload . 1>/dev/null 
Undefined Jinja2 variables remain (['some_erroneous_expression']).  Please enable source downloading and try again.
```

ma-ba:
#1378 boils down to `environ.meta_vars` calling `MetaData.name` which `sys.exit`s because the package's name is not set during rendering. `MetaData.name` gets called due to the following call stack:
```python
render.parse_or_try_download
  metadata.MetaData.parse_again
    metadata.MetaData._get_contents
      jinja_context.context_processor
        environ.get_dict
          environ.meta_vars
            metadata.MetaData.name
```
I did **not** check the implications of replacing `meta.name()` by `meta.get_value('package/name')` for other code paths. If the patch is thus not applicable, other potential solutions are:
- either guard the `meta.name()` call inside `meta_vars` by some optional Boolean argument which has to be handed down all the way from `parse_again` down to `meta_vars`,
- or simply artificially fulfil `MetaData.name`'s precondition by setting `MetaData.meta['package']['name'] = some_random_string` inside `MetaData.parse_again` prior to calling `MetaData._get_contents` and removing it after the call from the new `meta` if it did not change.

### DIFF
--- a/conda_build/environ.py
+++ b/conda_build/environ.py
@@ -335,7 +335,8 @@ def meta_vars(meta, config):
     elif external.find_executable('hg', config.build_prefix) and os.path.exists(hg_dir):
         d.update(get_hg_build_info(hg_dir))
 
-    d['PKG_NAME'] = meta.name()
+    # use `get_value` to prevent early exit while name is still unresolved during rendering
+    d['PKG_NAME'] = meta.get_value('package/name')
     d['PKG_VERSION'] = meta.version()
     d['PKG_BUILDNUM'] = str(meta.build_number())
     d['PKG_BUILD_STRING'] = str(meta.build_id())


msarahan:
Great idea here. I think a sane default may be to fall back to the recipe path dirname, then overwrite that if / when some new value is actually read. Maybe that should happen in the name() function? 

Thanks for thinking about this.

msarahan:
I'm going to merge this, then play with keeping the functionality in the name() function for the sake of consistency.  Thanks @ma-ba

ma-ba:
I'm not sure what use are exactly referring to. Do you mean simply replacing the call with `meta.get_value('package/name', recipe_path_dirname, False)`?
Or are you referring to the third option of changing `parse_again`? What I had in mind for this is some thing like:
```python
if 'package' not in self.meta:
    package = self.meta['package'] = {}
else:
    package = self.meta['package']
if not package.get('name', ''):
    package['name'] = some_random_string
meta = parse(self._get_contents(permit_undefined_jinja, config=config),
             config=config, path=self.meta_path)
if meta.get('package', {}).get('name') == some_random_string:
    del meta['package']['name']
self.meta = meta
```
where `some_random_string` should not be the real package name, i.e. something à la
```python
some_random_string = 'uninitialized-' + ''.join(random.sample(string.ascii_lowercase + string.digits, 20))
```
but still is a valid package name since calls to `MetaData.name` checks its validity. Of course it can also be any static string for which we can be certain no user would every choose it as a package name.
As for `MetaData.name` it do not think there should happen any substitutions / fallbacks.

PR #1407, opened 2016-09-27
add conda_versions info file to record build-time vers of conda, c-b

### LINKED ISSUE #1419: Conda-build and conda-versions should be included in package meta-data
I'm trying to update and build an older package for a client and I can not because of errors. I need to figure out the context in which this package was originally built.

cc @csoja  re Anaconda-Platform/anaconda-server#2949

msarahan:
Fixes #1419 

Just for information and perhaps reproducibility purposes.

Recommended by @kalefranz


srossross:
As much metadata as possible to make the build reproducible would be great. e.g. I would list the whole root environment, and the    some or all of the condarc information like channels, conda environment vars etc.

msarahan:
@srossross please review.  I have added all of the information that I though would be helpful, but tried to avoid anything that might introduce sensitive information (tokens?)  I trust your knowledge on that particular area more, so please pay special attention to that aspect.

This PR depends on https://github.com/conda/conda/issues/3588

conda_build/build.py
```diff
@@ -288,6 +290,21 @@ def write_about_json(m, config):
             value = m.get_value('about/%s' % key)
             if value:
                 d[key] = value
+        # for sake of reproducibility, record some conda info
+        conda_info = json.loads(subprocess.check_output(['conda', 'info', '--json', '-s']))
+        d['conda_version'] = conda_info['conda_version']
+        d['conda_build_version'] = conda_info['conda_build_version']
+        d['conda_env_version'] = conda_info['conda_env_version']
+        d['offline'] = conda_info['offline']
+        d['channels'] = conda_info['channels']
```
srossross: This will contain channels with the private token will it not?

msarahan: I don't think so.  There is an extra option for that:

```
--unsafe-channels  Display list of channels with tokens exposed.
```

This seems to have been added in 2015: https://github.com/conda/conda/blame/master/conda/cli/main_info.py#L74 so I think it's safe in this regard.

msarahan: I tested that, and it put the token in the output.  Yikes.  I have added a commit that scrubs URLs in a pretty rudimentary way.

```
re.sub('\/t\/[a-zA-Z0-9\-]*\/', '/t/<TOKEN>/', channel)
```

This is tested with:
```
def test_sanitize_channel():
    test_url = 'https://conda.anaconda.org/t/ms-534991f2-4123-473a-b512-42025291b927/somechannel'
    assert build.sanitize_channel(test_url) == 'https://conda.anaconda.org/t/<TOKEN>/somechannel'

```

hugobuddel: It will contain http credentials of the channels, see #3991, e.g. 
```
{
  "channels": [
    "https://USER:XXXXXXXX@example.com/conda",
    "conda-forge",
    "defaults"
  ],
```

Issue #1411, opened 2016-09-28
Crash in post-processing `patchelf`


bgruening:
The latest release crashes with the following error for some (pre-compiled) binaries.

```
Fixing permissions
cannot find section .dynamic
Traceback (most recent call last):
  File "/home/bag/miniconda2/bin/conda-build", line 6, in <module>
    sys.exit(conda_build.cli.main_build.main())
  File "/home/bag/miniconda2/lib/python2.7/site-packages/conda_build/cli/main_build.py", line 246, in main
    execute(sys.argv[1:])
  File "/home/bag/miniconda2/lib/python2.7/site-packages/conda_build/cli/main_build.py", line 238, in execute
    already_built=None, config=config)
  File "/home/bag/miniconda2/lib/python2.7/site-packages/conda_build/api.py", line 83, in build
    need_source_download=need_source_download, config=config)
  File "/home/bag/miniconda2/lib/python2.7/site-packages/conda_build/build.py", line 1004, in build_tree
    config=recipe_config)
  File "/home/bag/miniconda2/lib/python2.7/site-packages/conda_build/build.py", line 714, in build
    croot=config.croot)
  File "/home/bag/miniconda2/lib/python2.7/site-packages/conda_build/post.py", line 399, in post_build
    mk_relative(m, f, prefix)
  File "/home/bag/miniconda2/lib/python2.7/site-packages/conda_build/post.py", line 366, in mk_relative
    mk_relative_linux(f, prefix=prefix, rpaths=m.get_value('build/rpaths', ['lib']))
  File "/home/bag/miniconda2/lib/python2.7/site-packages/conda_build/post.py", line 315, in mk_relative_linux
    existing = check_output([patchelf, '--print-rpath', elf]).decode('utf-8').splitlines()[0]
  File "/home/bag/miniconda2/lib/python2.7/subprocess.py", line 574, in check_output
    raise CalledProcessError(retcode, cmd, output=output)
subprocess.CalledProcessError: Command '['/home/bag/miniconda2/bin/patchelf', '--print-rpath', '/home/bag/miniconda2/conda-bld/searchgui_1475044759195/_b_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_pla/share/searchgui-3.1.1-0/resources/Comet/linux/comet.exe']' returned non-zero exit status 1
```

[...]

PR #1418, opened 2016-09-30
Patchelf

mingwandroid:
Intended to address https://github.com/conda/conda-build/issues/1411 as well as another issue with handling rpaths (remove rather than assert on RPATHs outside the build prefix)


msarahan:
Thanks.  Is there a recipe that you would suggest building so that we can put in a regression test for this?

mingwandroid:
Not off the top of my head. I can add some more tests though next week?

msarahan:
Sure, that would be helpful.  I'm just going to merge this for now.

PR #1437, opened 2016-10-06
post: Fixes and a test for fix_permissions

mingwandroid:
We had a bug report https://github.com/ContinuumIO/anaconda-recipes/issues/52
where for some unknown reason qt.conf has no read permissions for
other, and our testing procedure did not spot this.

Our fix_permissions() function wasn't forcing read access. I took
the opportunity to change it in other ways too. Now, in total, it:

    1. Directories are all set to 0o755.
    2. Ensures the user can write to files.
    3. Broadcasts user execute to group and other for files.
    4. Ensures that user, group and other can read files.

### DIFF
--- a/conda_build/_link.py
+++ b/conda_build/_link.py
@@ -82,7 +82,7 @@ def create_script(fn):
         with open(dst, 'w') as fo:
             fo.write('#!%s\n' % normpath(sys.executable))
             fo.write(data)
-        os.chmod(dst, 0o755)
+        os.chmod(dst, 0o775)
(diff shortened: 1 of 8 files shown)


msarahan:
How about 775 for dirs and write for group too?

The use case being a central install that more than one person can admin,
but only those in a particular group.

mingwandroid:
Ok 0o775 it is.

msarahan:
Thanks.  If it ends up being a terrible idea, we'll go back to more restrictive permissions, per your scheme.

Issue #1316, opened 2016-09-02
conda build fails when multiple python/numpy versions are specified


bryanwweber:
Hi! I'm using conda-build 2.0.0 for my package, and I'm trying to build for multiple Python/Numpy versions in a single command. I didn't see anything in the CHANGELOG that this functionality was removed, but if it was, my mistake. The specific error messages are
```
conda build cantera --python=2.7 --python=3.4 --python=3.5 --numpy=1.11
Traceback (most recent call last):
  File "/home/bryan/software/anaconda3/bin/conda-build", line 6, in <module>
    sys.exit(conda_build.cli.main_build.main())
  File "/home/bryan/software/anaconda3/lib/python3.5/site-packages/conda_build/cli/main_build.py", line 234, in main
    execute(sys.argv[1:])
  File "/home/bryan/software/anaconda3/lib/python3.5/site-packages/conda_build/cli/main_build.py", line 180, in execute
    config = Config(**args.__dict__)
  File "/home/bryan/software/anaconda3/lib/python3.5/site-packages/conda_build/config.py", line 36, in __init__
    self.set_keys(**kwargs)
  File "/home/bryan/software/anaconda3/lib/python3.5/site-packages/conda_build/config.py", line 58, in set_keys
    self.CONDA_PY = int(env('python', "%s%s" % (sys.version_info.major, sys.version_info.minor))
AttributeError: 'list' object has no attribute 'replace'
```
and for Numpy:
```
conda build cantera --python=3.5 --numpy=1.11 --numpy=1.10
Traceback (most recent call last):
  File "/home/bryan/software/anaconda3/bin/conda-build", line 6, in <module>
    sys.exit(conda_build.cli.main_build.main())
  File "/home/bryan/software/anaconda3/lib/python3.5/site-packages/conda_build/cli/main_build.py", line 234, in main
    execute(sys.argv[1:])
  File "/home/bryan/software/anaconda3/lib/python3.5/site-packages/conda_build/cli/main_build.py", line 195, in execute
    set_language_env_vars(args, parser, config=config, execute=execute)
  File "/home/bryan/software/anaconda3/lib/python3.5/site-packages/conda_build/render.py", line 43, in set_language_env_vars
    execute(args, parser, config)
TypeError: execute() takes 1 positional argument but 3 were given
```
I get the same error on Linux (Ubuntu) and OS X. My conda-info:
```
conda info
Current conda install:

             platform : linux-64
        conda version : 4.1.11
[...]

PR #1443, opened 2016-10-09
Add regex Jinja preprocessor

bryanwweber:
This PR adds a preprocessor function in the same vein as `load_setup_py_data` for the meta.yaml file. It takes a file and regex string as arguments, and returns the `match` object from a `search` of the given file. This is related to #1316 

The use case for this is to grab version and other information that is specified in files other than `setup.py`. This may be common, for instance, for packages that have multiple language interfaces and want to specify the version number in a Makefile or its equivalent so that the same version is used for all the interfaces during a build.

The only issue that I had in implementing this was that Jinja2 doesn't like Python raw strings, so the backslashes in the regex string have to be escaped by doubling where they're specified in the meta.yaml file. The specific error was `jinja2: TemplateSyntaxError: expected token ',', got 'string'` I'm not sure if this is an issue with my implementation, or a general issue with Jinja2.

If I should add or remove anything from this PR (such as the link fix in the README), please let me know.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -95,7 +95,7 @@ Testing
 -------
 
 Running our test suite requires cloning one other repo at the same level as conda-build:
-https://github.com/conda/conda_build_test_repo - this is necessary for relative path tests
+https://github.com/conda/conda_build_test_recipe - this is necessary for relative path tests
 outside of conda build's build tree.
 
 The test suite runs with py.test. Some useful commands to run select tests,
--- a/conda_build/jinja_context.py
+++ b/conda_build/jinja_context.py
@@ -167,15 +167,12 @@ def load_npm():
         return json.load(pkg)
 
 
-def load_file_regex(config, load_file=None, regex_pattern=None, from_recipe_dir=False,
+def load_file_regex(config, load_file, regex_pattern, from_recipe_dir=False,
                     recipe_dir=None, permit_undefined_jinja=True):
     import re
     match = False
 
     cd_to_work = False
-
-    if load_file is None:
-        raise RuntimeError('File to be searched is not specified.')
 
     if from_recipe_dir and recipe_dir:
         load_file = os.path.abspath(os.path.join(recipe_dir, load_file))
--- a/conda_build/jinja_context.py
+++ b/conda_build/jinja_context.py
@@ -1,8 +1,3 @@
-'''
-Created on Jan 16, 2014
-
-@author: sean
-'''
 from __future__ import absolute_import, division, print_function
 
 from functools import partial
@@ -138,6 +133,9 @@ def load_setup_py_data(config, setup_file='setup.py', from_recipe_dir=False, rec
     if os.path.isfile(setup_file):
         code = compile(open(setup_file).read(), setup_file, 'exec', dont_inherit=1)
         exec(code, ns, ns)
+    else:
+        if not permit_undefined_jinja:
+            raise TypeError('{} is not a file that can be read'.format(setup_file))
 
     sys.modules['versioneer'] = versioneer
 
@@ -194,7 +192,8 @@ def load_file_regex(config, load_file, regex_pattern, from_recipe_dir=False,
     if os.path.isfile(load_file):
         match = re.search(regex_pattern, open(load_file, 'r').read())
     else:
-        raise TypeError('{} is not a file that can be read'.format(load_file))
+        if not permit_undefined_jinja:
+            raise TypeError('{} is not a file that can be read'.format(load_file))
 
     # Reset the working directory
     if cd_to_work:
--- a/conda_build/jinja_context.py
+++ b/conda_build/jinja_context.py
@@ -193,6 +193,8 @@ def load_file_regex(config, load_file, regex_pattern, from_recipe_dir=False,
 
     if os.path.isfile(load_file):
         match = re.search(regex_pattern, open(load_file, 'r').read())
+    else:
+        raise TypeError('{} is not a file that can be read'.format(load_file))
 
     # Reset the working directory
     if cd_to_work:
--- a/conda_build/jinja_context.py
+++ b/conda_build/jinja_context.py
@@ -167,6 +167,43 @@ def load_npm():
         return json.load(pkg)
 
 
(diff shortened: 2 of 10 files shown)


README.rst
```diff
@@ -95,7 +95,7 @@ Testing
 -------
 
 Running our test suite requires cloning one other repo at the same level as conda-build:
-https://github.com/conda/conda_build_test_repo - this is necessary for relative path tests
+https://github.com/conda/conda_build_test_recipe - this is necessary for relative path tests
```
msarahan: Thanks for fixing this.

conda_build/jinja_context.py
```diff
@@ -167,6 +167,43 @@ def load_npm():
...
 
 
+def load_file_regex(config, load_file=None, regex_pattern=None, from_recipe_dir=False,
+                    recipe_dir=None, permit_undefined_jinja=True):
+    import re
+    match = False
+
+    cd_to_work = False
+
+    if load_file is None:
```
msarahan: This should maybe be an assert instead.  I'd also change load_file to not have a default argument, since it really is required.

```
assert load_file, 'File to be searched is not specified.'
```

On the other hand, raising the right exception is also important, and for this I'd probably think more about a ValueError than a RuntimeError.  This may or may not gel well will conda-build's existing exception handling.  My opinion is more to do things right here, and fix whatever else conda-build may not have done right elsewhere.

conda_build/jinja_context.py
```diff
@@ -167,6 +167,43 @@ def load_npm():
...
+        raise RuntimeError('File to be searched is not specified.')
+
+    if from_recipe_dir and recipe_dir:
+        load_file = os.path.abspath(os.path.join(recipe_dir, load_file))
+    elif os.path.exists(config.work_dir):
```
msarahan: It looks like you have no test coverage here for the default execution path.  Can you add a test, please?

I say default execution path, because this is the path taken by the default arguments.

conda_build/jinja_context.py
```diff
@@ -167,6 +167,43 @@ def load_npm():
...
+    else:
+        message = ("Did not find {} file in manually specified location, and source "
+                  "not downloaded yet.".format(load_file))
+        if permit_undefined_jinja:
+            log.debug(message)
+            return {}
+        else:
+            raise RuntimeError(message)
+
+    if os.path.isfile(load_file):
```
msarahan: Probably want to throw some kind of error if it is not a file.  As it is, I think this will fail silently, which would make debugging hard.

conda_build/render.py
```diff
@@ -78,7 +78,8 @@ def parse_or_try_download(metadata, no_download_source, config,
 
     need_reparse_in_env = False
     if (force_download or (not no_download_source and (metadata.uses_vcs_in_meta or
-                                                       metadata.uses_setup_py_in_meta))):
+                                                       metadata.uses_setup_py_in_meta or
```
msarahan: This is fine for now.  I'll probably refactor it later to be a collection outside of the if statement, because it's getting pretty long.  It would probably also be even nicer to refactor it so that the only changes necessary here would be in metadata.py, rather than also here.

No changes necessary on your part, just thinking here.

conda_build/jinja_context.py
```diff
@@ -167,6 +167,43 @@ def load_npm():
         return json.load(pkg)
 
 
+def load_file_regex(config, load_file=None, regex_pattern=None, from_recipe_dir=False,
```
msarahan: Please make regex_pattern a non-default argument.  People really do need to specify it.

conda_build/render.py
bryanwweber: :+1: to making the changes in fewer places. I only found this by `grep`ping for `load_setup_py_data` to basically duplicate its interface.

bryanwweber:
Thanks for the comments! I fixed the few that I knew what to do and left 👍  on those. Please feel free to submit a PR to fix the testing, since I'm genuinely lost on that. Otherwise, if you give a bit of guidance, I can probably figure it out, but it might not be until next weekend.

conda_build/jinja_context.py
bryanwweber: Should this also be fixed in `load_setup_py_data`?

conda_build/jinja_context.py
bryanwweber: I should add as well that the default execution path works well for the actual use case I have in #1316, so I know it works in at least one case, its just a matter of writing the test for it.

conda_build/jinja_context.py
msarahan: Yes, definitely.  Thanks for pointing that out.

conda_build/jinja_context.py
msarahan: This error was because the rendering happens in multiple passes.  In the first pass, the source has not been downloaded, so your exception gets raised.  It's import to not raise unless undefined jinja is not allowed.

Fixing this in incoming PR.

msarahan:
Changes submitted at https://github.com/bryanwweber/conda-build/pull/1 - please review and merge when ready.  I think this PR is ready with those in place.

bryanwweber:
OK I merged your PR. Thanks for the help! One last question: At this line, https://github.com/bryanwweber/conda-build/blob/ab84fe512893ae2713d1962275582e3e3e694f46/conda_build/jinja_context.py#L188 I'm returning an empty dictionary, but this is left over from the `load_setup_py_data` heritage of this function. It seems like it might be more appropriate to `return None` at that location, to match what `re` returns when there's no search match. I can push a commit for that, or not, your call. Thanks again!

msarahan:
I think it's best to leave the empty dictionary.  Since we're storing that in meta.yaml, and then trying to do dictionary-style access, it'll be perhaps better to see a key error than the alternative.  I don't really know, though.  What do you think?  What's more clearly indicative of the problem?

bryanwweber:
OK then I think we can leave the empty dictionary. None of the methods of an `SRE_Match` object are also methods of the dictionary, so it shouldn't lead to any weird and hard to debug situations, it should just raise an AttributeError. This is good to go on my end!

msarahan:
Not sure what the Travis errors are about.  May be transient.  Didn't see them locally.  Merging, and will fix if necessary.

PR #1461, opened 2016-10-14
Run `ldd` after `patchelf`

jakirkham:
On Mac, when running `install_name_tool`, `conda-build` often provides some very helpful information about what things are linked to for free thanks to this [line](https://github.com/conda/conda-build/blob/714c1890c0ef0807648d52b6362f750baa539e54/conda_build/post.py#L252). However, we don't do a similar thing on Linux when running `patchelf`. This change runs `ldd` on each ELF after running `patchelf`. That way users have equivalent info in their logs about what things are linked to.

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -1026,6 +1026,16 @@ def tests_failed(m, move_broken, broken_dir, config):
 
 def check_external():
     if sys.platform.startswith('linux'):
+        ldd = external.find_executable('ldd')
+        if ldd is None:
+            sys.exit("""\
+Error:
+    Did not find 'ldd' in: %s
+    'ldd' is necessary for building conda packages on Linux with
+    relocatable ELF libraries.  Please install with your system package
+    manager.
+""" % (os.pathsep.join(external.dir_paths)))
+
         patchelf = external.find_executable('patchelf')
         if patchelf is None:
             sys.exit("""\
--- a/conda_build/post.py
+++ b/conda_build/post.py
@@ -329,6 +329,7 @@ def mk_relative_linux(f, prefix, rpaths=('lib',)):
     elf = join(prefix, f)
     origin = dirname(elf)
 
+    ldd = external.find_executable('ldd')
     patchelf = external.find_executable('patchelf', prefix)
     try:
         existing = check_output([patchelf, '--print-rpath', elf]).decode('utf-8').splitlines()[0]
@@ -366,6 +367,8 @@ def mk_relative_linux(f, prefix, rpaths=('lib',)):
     rpath = ':'.join(new)
     print('patchelf: file: %s\n    setting rpath to: %s' % (elf, rpath))
     call([patchelf, '--force-rpath', '--set-rpath', rpath, elf])
+    print('ldd: file: %s' % (elf,))
+    call([ldd, elf])
 
 
 def assert_relative_osx(path, prefix):


conda_build/build.py
```diff
@@ -1026,6 +1026,16 @@ def tests_failed(m, move_broken, broken_dir, config):
 
 def check_external():
     if sys.platform.startswith('linux'):
+        ldd = external.find_executable('ldd')
+        if ldd is None:
+            sys.exit("""\
+Error:
+    Did not find 'ldd' in: %s
+    'ldd' is necessary for building conda packages on Linux with
+    relocatable ELF libraries.  Please install with your system package
```
msarahan: Is ldd something that we can build and add as a conda-build dependency?  I really don't like introducing these ties to system packages.

jakirkham: > The standard version of ldd comes with glibc2.

xref: https://linux.die.net/man/1/ldd

jakirkham: In other words, I suppose we could build Glibc if we wanted to, but do we really want to go down that road? Even if we did, how do we ensure we install the right version `ldd` for that Linux system? It is [possible]( http://www.linuxfromscratch.org/lfs/view/6.4/chapter06/glibc.html ) though since you were curious. Just unconvinced that using `ldd` from anything but the system is a good idea.

jakirkham: Also if you were curious about musl, they provide their own `ldd` as well. Alpine Linux will expose it in the same standard way too.

xref: http://wiki.musl-libc.org/wiki/FAQ#Q:_where_is_ldd_.3F
xref: https://pkgs.alpinelinux.org/contents?branch=edge&name=musl-utils&arch=x86_64&repo=main

jakirkham: There are also implementations of `ldd` for [FreeBSD]( https://www.freebsd.org/cgi/man.cgi?query=ldd&sektion=1&manpath=FreeBSD+5.0-RELEASE ), [OpenBSD]( http://man.openbsd.org/OpenBSD-current/man1/ldd.1 ), [DragonFlyBSD]( https://www.dragonflybsd.org/cgi/web-man?command=ldd&section=ANY ), and most other Unix FWICT.

jakirkham: Interesting read about `ldd` and what it actually does.

xref: http://www.catonmat.net/blog/ldd-arbitrary-code-execution/

msarahan:
I am also concerned about output length.  Some of the time, we come close or exceed the log length limits of various CI's.  Making this output automatic seems like it could cause trouble.  

Is there a reason why conda-build needs this, specifically, and not conda-forge build scripts?  I understand it would be convenient, I'm just not convinced that conda-build is the right place.

jakirkham:
So I don't understand the concern. It sounds like you are saying (feel free to correct me that) you are concerned about how this affects log length in CIs. The suggestion I am hearing is conda-forge does this in the build scripts. There are several things that feel off to me about this statement and they are as follows.

First, it is unclear to me how moving these lines somehow from `conda-build` to conda-forge has changed the log length. Related to that it is unclear how this would be more easily or cleanly solved at conda-forge vs this solution in `conda-build`. Second, I'm unclear as to why `conda-build`, a command line tool that can be run anywhere, is getting unnecessarily constrained by one of its use cases (albeit a prominent one); namely building on CIs. Certainly this functionality would be handy anywhere including VMs that will inevitably replaces the CIs. Third, the log length problem has been one seen primarily (if not only) with Travis CI which has a 4MB log limit (though time limit is more often the problem). However, `ldd` is not run on Travis CI, but would be run on CircleCI where we have a 4GB limit. TBH I don't know that we have ever hit the CircleCI limit, but I doubt these few lines would be the reason for it.

jakirkham:
This is now irrelevant as `conda-build` 3 does this and in a much cleaner way.

PR #1477, opened 2016-10-24
Binary prefix

### LINKED ISSUE #1475: conda-build 2.0 is ignoring detect_binary_files_with_prefix
After updating to conda-build 2.0 I noticed that some packages were being build with prefix detection even though they were setting ```detect_binary_files_with_prefix: false```

Checking out the code, I found that this commit removed the handling of this: https://github.com/conda/conda-build/commit/2b611bc4bb58f354abb5c5e1f52c6f75ac02380b#diff-ac55d744935b1f37a3369258b88e8f1fL278

Was this intentional? The documentation stills refers to this setting

In this package specifically (```gcc``), I don't want conda to replace prefixes in binaries because basically all gcc binaries have them. So when installing gcc in an environment conda needs to make a copy of every binary and this occupy a lot of space

mingwandroid:
For your consideration @msarahan.

fixes #1475


msarahan:
Good call. I'm out at training. Will look at this in detail later today.

msarahan:
Seems like some pythons have different extractfile implementations?  Awfully screwy, but CIs seem happy now...

PS: go to bed.

msarahan:
Appveyor fail is a lock error.  All good here.  Merging.

PR #1479, opened 2016-10-25
otool -h header changed

minrk:
otool -h output starts with "Mach header" not the file path

Xcode 8, macOS 10.12

Also expanded the error message to show the output that didn't match, rather than just the path that it didn't match.


msarahan:
I wish this stuff was easier to test.  I hope we'll be able to have a multi-platform CI system soon - working on it.  For now, thanks much for this fix.

PR #1480, opened 2016-10-25
retry clone with local mirror as reference if first from-mirror clone fails

### LINKED ISSUE #1462: Clone fails if source has LFS files
Due to https://github.com/github/git-lfs/issues/1207 , cloning from a local repo fails if it contains files stored in git-lfs. Since conda-build seems to always clone to a cache first, then clone from there for the build (logical), repos storing files in LFS fail with errors like:

```
Cloning into '/Users/benjaminrk/conda/conda-bld/fenics_1476434449181/work'...
done.
Downloading demo/documented/bcs/aneurysm.xml.gz (686.73 KB)
Error downloading object: demo/documented/bcs/aneurysm.xml.gz (95ece98fe8dbf0e3ecc46cdf01cf0a9cbd65da1f1ebc4a27eebc564dad6f978c)

Errors logged to /Users/benjaminrk/conda/conda-bld/fenics_1476434449181/work/.git/lfs/objects/logs/20161014T104139.022737313.log
```

where the error in the log has a key phrase, such as `unsupported protocol scheme ""`.

If there were a way to mark these sources as not-cacheable (or if anyone has an idea for how to make the lfs fetches work), that would help me.

minrk:
While from-local clone will fail if the repo contains files in LFS, the local mirror can still be used as a reference to avoid redownloading git objects with a command like:

```
git clone --reference /local/mirror git@url destination
```

It would be a little simpler to _always_ use the `clone --reference`,
but I went with it as only a fallback if the regular clone fails because:
1. it's slower (a second or two, vs ~instant)
2. it won't work offline
3. it will hopefully become obsolete if/when upstream adds support for local clones with lfs

closes #1462

[ref for --reference](https://github.com/github/git-lfs/issues/1207#issuecomment-217455331)

### DIFF
--- a/conda_build/source.py
+++ b/conda_build/source.py
@@ -5,6 +5,7 @@ import logging
 import os
 from os.path import join, isdir, isfile, abspath, basename, exists, normpath
 import re
+import shutil
 from subprocess import Popen, PIPE, CalledProcessError
 import sys
 import time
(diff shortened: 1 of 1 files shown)


msarahan:
LGTM.  What would be necessary to add a test?  Is there a new dependency involved for LFS?

minrk:
To add a test, I think all that's needed would be to call `git_mirror_checkout_recursive` on a repo that has LFS files.

PR #1489, opened 2016-10-28
Verify packages

soapy1:
This PR will enable users to verify recipes and packages when they build packages;
- verify scripts (defined by the user) will be run by default but this can be disabled by using the `--no-verify` parameter
- the location of the verify scripts is assumed to be `$PREFIX/verify` but can be overridden by setting `conda_build_verify_scripts` in `.condarc`

Inside the verify folder
- the verify folder should have a `package` and `recipe` folder. The scripts in the folders will be run to verify packages and recipes respectively
- a package verify script should have a function `verify` that takes a path to the package (.tar.bz2 file)
- a recipe should have a function `verify` that takes the rendered `meta.yaml` and a path to the recipe dir


conda_build/build.py
```diff
@@ -599,7 +600,7 @@ def warn_on_old_conda_build(index=None, installed_version=None, available_packag
 """ % (installed_version, available_packages[-1]), file=sys.stderr)
 
 
-def build(m, config, post=None, need_source_download=True, need_reparse_in_env=False):
+def build(m, config, post=None, need_source_download=True, need_reparse_in_env=False, verify=True):
```
msarahan: verify should probably be a config parameter, rather than a new argument here.  Take a look at how test is done for reference.

conda_build/api.py
```diff
@@ -57,7 +57,7 @@ def check(recipe_path, no_download_source=False, config=None, **kwargs):
 
 
 def build(recipe_paths_or_metadata, post=None, need_source_download=True,
-          build_only=False, notest=False, config=None, **kwargs):
+          build_only=False, notest=False, config=None, verify=True, **kwargs):
```
msarahan: I'm a little bit ambivalent on this, but for the sake of consistency, maybe this should be defined as noverify, to match notest.  Keeping all logic behaving the same like this makes things easier to keep in line.

conda_build/build.py
```diff
@@ -1039,7 +1043,7 @@ def check_external():
 
 
 def build_tree(recipe_list, config, build_only=False, post=False, notest=False,
-               need_source_download=True, need_reparse_in_env=False):
+               need_source_download=True, need_reparse_in_env=False, verify=True):
```
msarahan: same as above: verify should probably be a config parameter, rather than a new argument here. Take a look at how test is done for reference.

conda_build/cli/main_build.py
```diff
@@ -153,6 +153,11 @@ def parse_args(args):
         help=("Build root folder.  Equivalent to CONDA_BLD_PATH, but applies only "
               "to this call of conda-build.")
     )
+    p.add_argument(
+        "--no-verify",
+        action="store_false",
+        help=("do not run verification on recipes or packages when building")
```
msarahan: This is very unintuitive to me.  Since you have some implicit boolean logic inversion going on here, it's hard for me to figure out.

What this code says to me right now is:
- if you pass --no-verify, then it sets no_verify to false
- if you don't pass it, no_verify is true

What might be clearer would be:

```
 p.add_argument(
         "--no-verify",
         action="store_false",
         dest="verify",
         help=("do not run verification on recipes or packages when building")
```

and then change the reference below to args.verify.

conda_build/conda_interface.py
```diff
@@ -22,7 +22,7 @@
 from conda.signature import KEYS, KEYS_DIR, hash_file, verify  # NOQA
 from conda.utils import human_bytes, hashsum_file, md5_file, memoized, unix_path_to_win, win_path_to_unix, url_path  # NOQA
 import conda.config as cc  # NOQA
-from conda.config import sys_rc_path  # NOQA
+from conda.config import rc_path  # NOQA
```
msarahan: Is this a change between conda 4.1 and 4.2?  Are you sure that this import is the same on both?  If not, please do the import differently in each section below.

conda_build/verify.py
```diff
@@ -0,0 +1,44 @@
...
+    run_scripts(verify_scripts, verify_path, *args)
+    print("All scripts passed")
+
+
+def verify_package(path_to_package, config):
+    verify_path = os.path.join(config.verify_scripts_path, "package")
+    verify(verify_path, path_to_package)
+
+
+def verify_recipe(recipe, config):
```
msarahan: As we discussed, the thing that will probably actually be passed here is not the recipe itself, but rather the rendered metadata.  That means that instead of doing a subprocess to run each script, it would be better to lay out some standard function that each verification script must implement.  Then, import each verify script and call that particular function, passing in the metadata with it.

For an example of how to do imports dynamically, see https://github.com/conda/conda-build/blob/master/conda_build/api.py#L144-L182

The alternative is to write the rendered yaml to a file, and then run the verification in subprocesses like you do here.  It's not as nice.

conda_build/conda_interface.py
soapy1: Both 4.1 and 4.2 have `rc_path` and `sys_rc_path`. In both `rc_path` is populated by calling `get_rc_path()` which looks for all possible condarc files and returns the first one that exists.

conda_build/verify.py
soapy1: Cool 🌵 I got this to work using importlib.

conda_build/cli/main_build.py
```diff
@@ -231,7 +237,7 @@ def execute(args):
     else:
         api.build(args.recipe, post=args.post, build_only=args.build_only,
                    notest=args.notest, keep_old_work=args.keep_old_work,
-                   already_built=None, config=config)
+                   already_built=None, config=config, noverify=args.verify)
```
msarahan: OK, you flipped both of these, so I have my same complaint as before.  The logic flip should only happen in one place - ideally in the argument.  To me, that means that you should always have the keyword and the argument agree (verify=args.verify or noverify=args.no_verify)  Pick one.

conda_build/exceptions.py
```diff
@@ -48,3 +48,11 @@ def error_body(self):
 
 class UnableToParseMissingSetuptoolsDependencies(CondaBuildException):
     pass
+
+
+class VerifyError(RuntimeError):
+    def __init__(self, error, script, *args):
+        self.error = error
```
msarahan: Hmmm - why is this not being tested?  Can we write a test case for a VerifyError and make sure that it is thrown and has the info you want?

conda_build/verify.py
msarahan: I don't understand why you did things this way with such different code between python 2 and 3.  I read that my use of ```__import__``` may not be ideal, but I'd like to understand why you chose to do what you did here.

conda_build/cli/main_build.py
soapy1: Oh ya, I missed this. To be consistent with `notest` I'll stick with `noverify`

conda_build/verify.py
msarahan: I think the tricky thing is that for ```__import__``` to work, things must be locatable with python's path.  That means that things probably need to be part of a package that is installed on the system.  Do you already have a package for these scripts, or not?

soapy1: I'm not currently using a package for these scripts. They are just sitting in a folder

conda_build/config.py
```diff
@@ -408,5 +412,19 @@ def show(config):
 # legacy exports for conda
 croot = Config().croot
 
+
+class Context(Configuration):
```
msarahan: Should the whole Config class become a Configuration subclass?  I'm thinking towards allowing people to set more permanent defaults for conda-build's other settings.

conda_build/verify.py
```diff
@@ -0,0 +1,41 @@
...
+        files.append(name)
+    return files
+
+
+def verify(verify_type, ignore_scripts, **kargs):
+    import_type = "verify.%s" % verify_type
+    try:
+        verify_module = __import__(import_type)
+    except ImportError as e:
+        print("can't find verify.recipe module, skipping verification")
```
msarahan: might need a different name for the other project than "verify" - but we'll see.

msarahan:
I think it's on the right track.  Any config file is the right track, but the actual use of that config file is probably part of the other project.  Conda-build should draw the line at knowing which config file to pass to the other project, or whatever data structure the other project is expecting.

soapy1:
Moving forward with this pr requires https://github.com/ContinuumIO/anaconda-verify/pull/10 to be merged

soapy1:
Looks like the same appveyor tests are failing on master 😞

tests/utils.py
```diff
@@ -55,7 +56,8 @@ def return_to_saved_path():
 
 @pytest.fixture(scope='function')
 def test_config(testing_workdir, request):
-    return Config(croot=testing_workdir, anaconda_upload=False, verbose=True, activate=False)
+    return Config(croot=testing_workdir, anaconda_upload=False, verbose=True, activate=False,
+                  noverify=False)
```
msarahan: why is it necessary to pass noverify here?  For the test configuration, things should be mostly defaults.  The stuff I have overridden here is just some tweaks to make the test stuff not muck up the system they're running on.

msarahan:
Looks good.  Thanks for all your work on this.

PR #1492, opened 2016-10-31
Support for absolute submodules, cached

neok-m4700:
I don't know if this is relevant, but it is related to [#753](https://github.com/conda/conda-build/pull/753).

This adds support for git submodules that are not relative. It will recursively download them into the cache as regular repositories.

Tested for https://github.com/Kitware/ParaView.git with .gitmodules file:
[submodule "VTK"]
	path = VTK
	url = https://gitlab.kitware.com/vtk/vtk.git
[submodule "Utilities/VisItBridge"]
	path = Utilities/VisItBridge
	url = https://gitlab.kitware.com/paraview/visitbridge.git
[submodule "ThirdParty/IceT/vtkicet"]
	path = ThirdParty/IceT/vtkicet
	url = https://gitlab.kitware.com/icet/icet.git
[submodule "ThirdParty/protobuf/vtkprotobuf"]
	path = ThirdParty/protobuf/vtkprotobuf
	url = https://gitlab.kitware.com/paraview/protobuf.git
[submodule "ThirdParty/QtTesting/vtkqttesting"]
	path = ThirdParty/QtTesting/vtkqttesting
	url = https://gitlab.kitware.com/paraview/qttesting.git

Testing against other examples will be needed.

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -65,6 +65,8 @@ from conda_build.features import feature_list
 import conda_build.noarch_python as noarch_python
 from conda_verify.verify import Verify
 
+from conda import __version__ as conda_version
+from conda_build import __version__ as conda_build_version
 
 if 'bsd' in sys.platform:
     shell_path = '/bin/sh'
@@ -392,8 +394,8 @@ def write_about_json(m, config):
         if hasattr(conda_info, 'decode'):
             conda_info = conda_info.decode(utils.codec)
         conda_info = json.loads(conda_info)
-        d['conda_version'] = conda_info['conda_version']
-        d['conda_build_version'] = conda_info['conda_build_version']
+        d['conda_version'] = conda_version
+        d['conda_build_version'] = conda_build_version
         # conda env will be in most, but not necessarily all installations.
         #    Don't die if we don't see it.
         if 'conda_env_version' in conda_info:


conda_build/source.py
```diff
@@ -182,18 +182,33 @@ def git_mirror_checkout_recursive(git, mirror_dir, checkout_dir, git_url, config
...
+                submod_url = matches.group(2)
+                git_dn = submod_url.split('://')[-1].replace('/', os.sep)
+                if git_dn.startswith(os.sep):
+                    git_dn = git_dn[1:]
+                git_dn = git_dn.replace(':', '_')
+                submod_mirror_dir = join(config.git_cache, git_dn)
+                if config.verbose:
+                    print('Absolute submodule %s found: url is %s, submod_mirror_dir is %s' % (
+                          submod_name, submod_url, submod_mirror_dir))
+                git_mirror_checkout_recursive(
```
msarahan: Is there a reason you avoid the TemporaryDirectory here?

It seems like the only really major change is that submod_mirror_dir differs in the two cases.  Can we reduce the code so that the if/else handles that, but the rest of the code is uniform (and not in the if/else?)

msarahan:
Thanks for this.  Minor comment in the code.  I think overall this is a good thing, but I want to understand how we can add a test case for this.

conda_build/source.py
neok-m4700: Hi, the purpose of the patch is to download the submodule to cache and checkout into the working directory not into a thrown away directory ? Else calling $ submodule update --init --recursive will download again the submodules ...

msarahan:
@mingwandroid (the author of much of that existing code) and I discussed this PR.  Here are some explanations and suggestions that we hope are helpful:

1) the main purpose of the recursion and checkout into temporary folders is that any relative-pathed submodule needs to be cached before the top-level git checkout (with submodules) can successfully occur.  Keep in mind that any given absolute submodule *may* have component relative submodules.  We don't think that the temporary folder breaks your absolute submodule checkout, and would prefer that you keep the temporary directory approach to consolidate code.
2) we need a test case before merging.  We recommend basing it on the one at https://github.com/conda/conda-build/blob/master/tests/test_api_build.py#L441 - and it should cover the case where an absolute submodule then has relative submodules.
3) we recommend defining a helper function that given some kind of git url (absolute or relative), returns the appropriate git_dn.  This would be cleaner code separation of this logic.

neok-m4700:
Hi @msarahan @mingwandroid,

Thank you for the dev hints.
1. I most certainly misunderstood the temporary directory pass, I will check this again
2. Perfect, I'll try it with a dummy repo, and try to come up with a robust test case
3. I agree with the minimalistic/reuse approach

PR #1500, opened 2016-11-01
clarify error message when extracting egg fails due to existing files

msarahan:
This is meant to address #1463 

I don't consider this a proper fix, but it is the best I can do right now.  Conda-build really should not be left to decide which is the correct file to overwrite - this must be a user choice.  The best we can do is try to be clear about what is happening, and how to work around it.

CC @janssen @jreback

### DIFF
--- a/conda_build/post.py
+++ b/conda_build/post.py
@@ -119,8 +119,17 @@ def remove_easy_install_pth(files, prefix, config, preserve_egg_dir=False):
                     # so the package directory already exists
                     # from another installed dependency
                     if os.path.exists(join(sp_dir, fn)):
-                        utils.copy_into(join(egg_path, fn), join(sp_dir, fn), config.timeout)
-                        utils.rm_rf(join(egg_path, fn))
+                        try:
+                            utils.copy_into(join(egg_path, fn), join(sp_dir, fn), config.timeout)
+                            utils.rm_rf(join(egg_path, fn))
+                        except IOError as e:
+                            fn = os.path.basename(str(e).split()[-1])
+                            raise IOError("Tried to merge folder {egg_path} into {sp_dir}, but {fn}"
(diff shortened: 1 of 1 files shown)


jreback:
so for example this came up when I had a build of ``dask`` locally against master, and then I built ``distributed``. Though I removed the entire build directly and still was erroring until I used the work-around.

msarahan:
I will take a closer look at dask and distributed tomorrow.  My hunch is that they have a similarly named file, and that is causing issues.  If that's accurate, that's a change I'll need to recommend to the Dask team.

msarahan:
@jreback, which recipes are you using to build Dask and Distributed?  I don't see conda recipes in their repo.

jreback:
i just cribbed off of conda-forge ones
with a git_url instead of fn

msarahan:
Sorry @jreback, can't reproduce.  Here's what I did:

- clone https://github.com/conda-forge/dask-feedstock
- replace source with git_url (pointing to master)
- conda-build that recipe - works OK

- clone https://github.com/conda-forge/distributed-feedstock
- replace source with git_url (pointing to master)
- conda-build that recipe - works OK

One other potential workaround (or cause, depending on how you look at this) is that if someone forgets to do the whole ```--single-version-externally-managed --record=record.txt``` setuptools dance, then they end up with eggs.  Under "normal" conditions, everyone should be using those flags when using setuptools in setup.py, so that eggs are not created in the first place.

janssen:
Or, didn't even know about --single-version-externally-managed --record=record.txt dance!  I'll try it with my pyobjc problem.  Is "record.txt" generic, or a specific name?

msarahan:
Generic. You can keep it around as a list of files that the setup process
installed, but I haven't heard of anyone actually using it. You can name it
anything.

PR #1503, opened 2016-11-02
Relax policing license_family to ensure existing recipes don't break

caseyclements:


msarahan:
Looks good. Thanks for the fast response.

msarahan:
Failures are test race conditions.  Merging.

PR #1511, opened 2016-11-04
Add creation of intermediate folders for `utils.copy_into`

### LINKED ISSUE #1510: `test source_files` does not create intermediate directories
I have the following part in my `meta.yml`

```yml
test:
  source_files:
    - ipynbtest/examples/ipynbtest_tutorial.ipynb
```

This fails with an error like: (I replaced my homefolder with `~` there are actually absolute paths)
```
IOError: [Errno 2] No such file or directory: u'~/anaconda/conda-bld/conda-recipe_1478272007306/test_tmp/ipynbtest/examples/ipynbtest_tutorial.ipynb'
```

After some digging it turns out that the problem is that in `test_tmp` there is no `examples` directory and hence the 

```
  File "~/anaconda/lib/python2.7/site-packages/conda_build/create_test.py", line 69, in create_files
    copy_into(f, f.replace(config.work_dir, config.test_dir), config.timeout)
```

can find the source but not open the destination

```
  File "~/anaconda/lib/python2.7/shutil.py", line 83, in copyfile with open(dst, 'wb') as fdst:
```

When change the part in `meta.yaml` to 

```yml
test:
  source_files:
    - ipynbtest/examples/*.ipynb
```

it also _fails_ whereas 

```yml
test:
  source_files:
    - ipynbtest/examples
```

it works. 

My conclusion is that this needs to be either well documented or better the intermediate folders needs to be created. This should be an easy fix if so desired.

jhprinz:
Resolves #1510 .

If `copy_into` is now used and you want to copy a file in a (nested) folder the intermediate folders for the destination are created. In case you were copying folders this was already the default. This enables the same for single file copies.

### DIFF
--- a/conda_build/utils.py
+++ b/conda_build/utils.py
@@ -115,8 +115,11 @@ def copy_into(src, dst, timeout=90, symlinks=False, lock=None):
         with lock:
             # if intermediate folders not not exist create them
             dst_folder = os.path.dirname(dst)
-            if not os.path.exists(dst_folder):
-                os.makedirs(dst_folder)
+            if dst_folder and not os.path.exists(dst_folder):
+                try:
+                    os.makedirs(dst_folder)
+                except OSError:
+                    pass
 
             # with each of these, we are copying less metadata.  This seems to be necessary
             #   to cope with some shared filesystems with some virtual machine setups.
--- a/conda_build/utils.py
+++ b/conda_build/utils.py
@@ -113,6 +113,11 @@ def copy_into(src, dst, timeout=90, symlinks=False, lock=None):
         if not lock:
             lock = get_lock(src_folder, timeout=timeout)
         with lock:
+            # if intermediate folders not not exist create them
+            dst_folder = os.path.dirname(dst)
+            if not os.path.exists(dst_folder):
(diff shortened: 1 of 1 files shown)


msarahan:
Thanks, this looks good to me.  I'll wait on the CI's, and merge assuming they come out OK.

jhprinz:
Great. Thanks for the quick response.

msarahan:
We have some tests that fail right now due to race conditions.  Among the failures on Travis, I think these are real, and need your attention:

https://travis-ci.org/conda/conda-build/jobs/173305834#L1057
https://travis-ci.org/conda/conda-build/jobs/173305834#L1031

The latter one is pretty weird though.  It may be that some exception is masking a copy_into error, or it may be that it's totally unrelated.  Please fix the first, but I don't think you need to fix the second unless you know what went wrong.

jhprinz:
@msarahan The problems you mentioned are fixed now and the remaining seem not (at least directly) originate from this PR. Let me know, when I should update the PR.

msarahan:
Confirming that all CI issues here are pre-existing.  This PR is good to go.  Thanks @jhprinz

PR #1516, opened 2016-11-07
Use version info from config if exists

soapy1:
fix for test_cli.py::test_skeleton_pypi_arguments_work test failure


soapy1:
tests/test_api_build.py::test_relative_git_url_submodule_clone failed on appveyor, though this is probably not related

msarahan:
Confused on how this broke, but that's not important.  Thanks for fixing.

PR #1518, opened 2016-11-09
FIX: args.version needs to be passed to skeleton api

cmorgan:


msarahan:
I think this is another approach to #1516 - have you tried master?  Does it work without your PR here?

The somewhat tricky thing is that because we have flexible input possibilities for each skeleton module, we can't know all of the input arguments at the API layer.  What we do instead is to store them as attributes on the config object, then unpack them later.  That's what is happening here: https://github.com/conda/conda-build/pull/1518/files#diff-523be2c11ece2c5147aacf823f87d624R54

#1516 fixed the fact that the default argument for version in the API was clobbering the config setting.  Your PR here has the same effect, but I don't think it should be necessary.

msarahan:
Closing this as duplicate of #1516 - if further work is needed, please reopen.

PR #1521, opened 2016-11-11
break symlink and copy file if symlink points to executable

msarahan:
@mingwandroid is this a reasonable approach to https://github.com/conda/conda/issues/3308 ?

I'm not sure that this will work as-is - if we copy the binary around, then the RPATH stuff will be wrong, correct?  Can we just shuffle this function to be called earlier and use the RPATH replacement stuff without modification?

### DIFF
--- a/conda_build/post.py
+++ b/conda_build/post.py
@@ -432,14 +432,14 @@ def post_build(m, files, prefix, build_python, croot):
         print("Skipping binary relocation logic")
     osx_is_app = bool(m.get_value('build/osx_is_app', False))
 
+    check_symlinks(files, prefix, croot)
+
     for f in files:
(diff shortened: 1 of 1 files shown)


conda_build/post.py
```diff
@@ -452,7 +452,13 @@ def check_symlinks(files, prefix, croot):
         if islink(path):
             link_path = readlink(path)
             real_link_path = realpath(path)
-            if real_link_path.startswith(real_build_prefix):
+            # symlinks to binaries outside of the same dir don't work.  RPATH stuff gets confused.
```
mingwandroid: Maybe change:
> . RPATH stuff gets confused
to:
> because ld.so follows symlinks in RPATHs

PR #1538, opened 2016-11-16
fix directly testing packages; freeze more meta.yaml data

### LINKED ISSUE #1534: list of used dependency versions for each build..
_From @martin-mann on November 14, 2016 13:54_

Hi,

we just discussed for bioconda the problem of updated runtime dependencies that break existing downstream packages (eg for perl etc..).

eg. think about the following situation: given a (perl) package A and a dependent package B, which only provides a minimal required version for A. Now A is updated and rebuild to a new version which drops the interface used by B. So the build of A is successful, but as soon as B is installed, the recent A is installed too and B fails on runtime.

Is is planned/possible/useful to add to each package also the *full list* of packages including version number used for the build?

This would very much ease 
- debugging for the described scenario and 
- enable the identification of outdated packages to be rebuild.

Thanks for feedback on that,
Martin

_Copied from original issue: conda/conda#3872_

msarahan:
This fixes ```conda build --test``` so that the actual package provided is the one that gets installed for the test.  It no longer tries to go download it from somewhere, based on metadata provided by rendering the recipe in the package.

This hard-codes important information in the rendered meta.yaml file:
* exact pins for build dependencies (even those not explicitly listed - it is whatever is present in the build env)
* absolute path substituted for relative source/path or source/git_url entries (necessary for testing)
* the build string, so that no further rendering is necessary to come up with a similar package spec.

Finally, this adds one option to conda build, ```--output-folder```, which dumps the output package to some folder.  This is useful for CI (particularly concourse CI), where output folder affects how artifacts are carried to the next stage.

closes #1534

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,11 +33,11 @@ install:
       conda config --set always_yes yes;
     else
       if [ -n "$CONDA" ]; then
-        conda install -y --no-deps conda=${CONDA} requests;
+        conda install -y --no-deps conda=${CONDA};
       fi
     fi
 
-  - conda install -q anaconda-client requests filelock contextlib2 jinja2 patchelf python=$TRAVIS_PYTHON_VERSION pyflakes=1.1 conda-verify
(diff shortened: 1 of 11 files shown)


msarahan:
CC @groutr - I used some of your code from conda-tools here.  Please speak up if any part of that is not OK.

PR #1539, opened 2016-11-17
package_metadata.json package-level data

kalefranz:
This PR ditches the `noarch.json` file in lieu of an `extra.json` file.  The purpose is to, in a forward-looking way, help limit the proliferation of files in the `info` directory.

The purpose for this file is to contain package-level information that is actionable by conda.  It distinguishes itself from `index.json` as this information will *not* be included in `repodata.json`.  It also should not contain per-path/per-file information.

To my knowledge, we have one conda-build release that writes the `noarch.json` file.  Given the recency of this release, and that conda isn't yet using it, I am comfortable backtracking to this alternate strategy.

`extra.json` may not be the appropriate file name.  I really don't care what the file name is.


kalefranz:
CC @soapy1

kalefranz:
Definitely not a fan of the `extra.json` file name.  For comparison, wheels use `metadata.json` for this (`metadata.json` also has a few other roles).

kalefranz:
Conda 4.3 will understand the following `package_metadata.json` example

```
{
    "version": 1,
    "noarch": {
        "type": "generic|python",
        "entry_points": [
            "conda = conda.cli:main",
            "conda-env = conda_env.cli.main:main"
        ]
    },
    "preferred_env": {
        "name": "conda|_conda_",
        "executable_paths": [
            "bin/conda",
            "bin/activate",
            "bin/deactivate"
        ]
    }
}
```

If the `info/packages_metadata.json` file exists, the only required top-level field is `version`.

If the `noarch` field exists, the `type` field is required, and the `entry_points` field is optional.  The only valid values for `type` are `generic` and `python`.  The `entry_points` field is only valid when `type` is `python`, and it must be a list of strings in the standard python entry point style.

If the `preferred_env` field is given, the `name` field is required.  The name is the requested name for the private env.  It must match the `preferred_env` field in `index.json` (the denormalization of this data is recognized and intentional).  The `executable_paths` field, if given, must be a list of strings.  The strings represent prefix-relative paths that should have application entry points created in the root env.

conda_build/build.py
```diff
@@ -302,6 +303,44 @@ def sanitize_channel(channel):
...
+    extra = dict()
+
+    noarch_type = m.get_value('build/noarch')
+    if noarch_type:
+        noarch_dict = dict(type=text_type(noarch_type))
+        if text_type(noarch_type).lower() == "python":
+            entry_points = m.get_value('build/entry_points')
+            if entry_points:
+                noarch_dict['entry_points'] = entry_points
+        extra['noarch'] = noarch_dict
```
soapy1: Should this also include a `version` and `preferred_envs` keys as per https://github.com/conda/conda-build/pull/1539#issuecomment-266322808 ?

conda_build/noarch_python.py
```diff
@@ -56,13 +56,6 @@ def rewrite_script(fn, prefix):
     return fn
 
 
-def create_entry_point_information(noarch_type, entry_points, config):
-    entry_point_information = {"type": noarch_type, "entry_points": entry_points}
-    file = os.path.join(config.info_dir, "noarch.json")
```
soapy1: 😄 👌

soapy1:
NOTE: this requires a change to conda-verify's recipe test `check_fields` to include allowing the fields `requirements/preferred_env` and `requirements/preferred_env_executable_paths`

conda_build/build.py
```diff
@@ -329,6 +330,53 @@ def sanitize_channel(channel):
...
+
+    mode_dict = {'mode': 'w', 'encoding': 'utf-8'} if PY3 else {'mode': 'wb'}
+    with open(join(config.info_dir, 'files'), **mode_dict) as fo:
+        if m.get_value('build/noarch_python'):
+            fo.write('\n')
+        elif is_noarch_python(m):
+            for f in files:
+                if f.find("site-packages") > 0:
+                    fo.write(f[f.find("site-packages"):] + '\n')
+                elif f.startswith("bin") and (f not in entry_point_script_names):
```
msarahan: Please add a test for these cases.

conda_build/build.py
```diff
@@ -428,42 +477,19 @@ def create_info_files(m, files, config, prefix):
     :type files: list of str
     '''
 
+    if on_win:
+        # make sure we use '/' path separators in metadata
+        files = [_f.replace('\\', '/') for _f in files]
```
msarahan: I think this coverage info just isn't being reported because of Appveyor build failures.  Looking past this gap in coverage.

conda_build/build.py
msarahan: @kalefranz please chime in on this.

kalefranz:
We're working on this Mike.  Trying to get a particular test failure quashed.

msarahan:
OK.  I'll wait.

kalefranz:
OK @msarahan, just pushed with Sophia.  This should solve what we need for the noarch part of `package_metadata.json` in conda-build 2.1.  The `preferred_env` stuff isn't critical for 2.1.  We need your help possibly, and at a minimum a little more time.  But I know you want to get 2.1 out, so I'd offer this as compromise.

msarahan:
OK, sounds good.  Will wait on CI and take another look.

kalefranz:
Sorry I fell off for a bit.  I had to get some sleep.  I was exhausted.  I just pushed a commit that should fix the tests.

kalefranz:
The canary failures coming from Travis-CI are all an import error from `conda/exports.py` that we've fixed, merged, and will be in 4.3.1 on Monday.

kalefranz:
Ok looks like appveyor passed. And the only failures with Travis-CI were with the canary flag, it was a conda issue, and we've already fixed it.

msarahan:
Cool.  Thanks for including the timed fail to remind us to discuss the other thing.  Merging.

PR #1542, opened 2016-11-21
Allow `!` in package version

mingwandroid:
It is used to represent 'epoch'.

### DIFF
--- a/conda_build/metadata.py
+++ b/conda_build/metadata.py
@@ -295,9 +295,11 @@ FIELDS = {
 
 
 def check_bad_chrs(s, field):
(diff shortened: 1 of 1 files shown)


msarahan:
How does this affect version sorting?  Is there a corresponding change needed in Conda to handle this properly?  Specifically, probably this class: https://github.com/conda/conda/blob/master/conda/version.py#L25

mingwandroid:
https://github.com/conda/conda/blob/master/conda/version.py#L31-L34

msarahan:
Of course, sorry I didn't look harder.

mingwandroid:
no problem!

PR #1554, opened 2016-11-28
support globs in prefix file functions

### LINKED ISSUE #1551: Bus error with conda-build 2.0 in Constructor installation
We are building our own Conda distribution (similar to miniconda) using [constructor](https://github.com/conda/constructor/).

When we build *Python* (3.5.2 or 3.6.0b4) with *conda-build 1.21.11*, everything works fine and as expected.  When we build it with *conda-build 2.0.x*, the Conda installer generated by Constructor fails with:

```
./emsconda.sh: Zeile 161: 23713 Bus error $PYTHON -E -s $PREFIX/pkgs/.install.py $INST_OPT
```

Context from the installer:

```bash
install_dist()
{
    # This function installs a conda package into prefix, but without linking
    # the conda packages.  It untars the package and calls a simple script
    # which does the post extract steps (update prefix files, run 'post-link',
    # and creates the conda metadata).  Note that this is all done without
    # conda.
    echo "installing: $1 ..."
    PKG=$PREFIX/pkgs/$1.tar.bz2
    tar xjf $PKG -C $PREFIX --no-same-owner || exit 1
    if [[ $1 == 'python-3.5.2-6' ]]; then
        $PYTHON -E -V
        if (( $? )); then
            echo "ERROR:
cannot execute native linux-64 binary, output from 'uname -a' is:" >&2
            uname -a
            exit 1
        fi
    fi
    $PYTHON -E -s $PREFIX/pkgs/.install.py $INST_OPT || exit 1
    rm $PKG
}

install_dist python-3.5.2-6
```

The error happens when `.install.py:update_prefix()` tries to open `lib/libpython3.5m.so.1.0` for writing (obviously, Python does not not like its lib being edited while it is running).

The reason why it works with conda-build 1.21 but not with 2.0 is that 1.21 inserts prefixes only in the executable files, but not in the `*.so` files.  2.0 also inserts prefixes in `*.so` which the installer then tries to fix (and then fails).

Output of `conda info`:
```
Current conda install:

               platform : linux-64
          conda version : 4.2.12
       conda is private : False
      conda-env version : 2.5.1
    conda-build version : 2.0.0 (or 1.21.11 in the working case)
         python version : 3.5.2.final.0
       requests version : 2.11.1
       root environment : /tmp/py3  (writable)
    default environment : /tmp/py3
       envs directories : /tmp/py3/envs
          package cache : /tmp/py3/pkgs
           channel URLs : file:///tmp/py3/conda-bld/linux-64
                          file:///tmp/py3/conda-bld/noarch
            config file : /tmp/py3/.condarc
           offline mode : False
```

I can provide the installers (a working one and a failing one) and also the recipes if necessary.

msarahan:
fixes #1551

### DIFF
--- a/conda_build/metadata.py
+++ b/conda_build/metadata.py
@@ -176,11 +176,12 @@ def parse(data, config, path=None):
     return sanitize(res)
 
 
-def expand_globs(path_list):
+def expand_globs(path_list, root_dir):
     files = []
     for path in path_list:
-        files.extend(glob.glob(path))
-    return files
+        files.extend(glob.glob(os.path.join(root_dir, path)))
+    # list comp is getting rid of absolute prefix, to match relative paths used in file list
+    return [f.replace(root_dir + os.path.sep, '') for f in files]
 
 
 trues = {'y', 'on', 'true', 'yes'}
@@ -693,7 +694,7 @@ class MetaData(object):
             if any('\\' in i for i in ret):
                 raise RuntimeError("build/has_prefix_files paths must use / "
                                    "as the path delimiter on Windows")
-        return expand_globs(ret)
+        return expand_globs(ret, self.config.build_prefix)
 
     def ignore_prefix_files(self):
         ret = self.get_value('build/ignore_prefix_files', False)
@@ -704,7 +705,7 @@ class MetaData(object):
             if type(ret) is list and any('\\' in i for i in ret):
                 raise RuntimeError("build/ignore_prefix_files paths must use / "
                                    "as the path delimiter on Windows")
-        return expand_globs(ret) if type(ret) is list else ret
+        return expand_globs(ret, self.config.build_prefix) if type(ret) is list else ret
 
     def always_include_files(self):
         files = ensure_list(self.get_value('build/always_include_files', []))
@@ -714,7 +715,7 @@ class MetaData(object):
         if on_win:
             files = [f.replace("/", "\\") for f in files]
 
-        return expand_globs(files)
+        return expand_globs(files, self.config.build_prefix)
 
     def include_recipe(self):
         return self.get_value('build/include_recipe', True)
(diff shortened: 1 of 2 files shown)


msarahan:
@sscherfke, please test this and make sure it does what you want.

sscherfke:
Seems to work as expected and the recipe looks much cleaner (and more robust) now:
```diff
diff --git a/python-3.6/meta.yaml b/python-3.6/meta.yaml
index 23881aa..2adb0dd 100644
--- a/python-3.6/meta.yaml
+++ b/python-3.6/meta.yaml
@@ -8,19 +8,10 @@ source:
 build:
-  number: 1
+  number: 2
   ignore_prefix_files:
     - lib/libpython3.6m.so.1.0
-    - lib/python3.6/lib-dynload/_bz2.cpython-36m-x86_64-linux-gnu.so
-    - lib/python3.6/lib-dynload/_curses.cpython-36m-x86_64-linux-gnu.so
-    - lib/python3.6/lib-dynload/_curses_panel.cpython-36m-x86_64-linux-gnu.so
-    - lib/python3.6/lib-dynload/_hashlib.cpython-36m-x86_64-linux-gnu.so
-    - lib/python3.6/lib-dynload/_lzma.cpython-36m-x86_64-linux-gnu.so
-    - lib/python3.6/lib-dynload/_sqlite3.cpython-36m-x86_64-linux-gnu.so
-    - lib/python3.6/lib-dynload/_ssl.cpython-36m-x86_64-linux-gnu.so
-    - lib/python3.6/lib-dynload/binascii.cpython-36m-x86_64-linux-gnu.so
-    - lib/python3.6/lib-dynload/readline.cpython-36m-x86_64-linux-gnu.so
-    - lib/python3.6/lib-dynload/zlib.cpython-36m-x86_64-linux-gnu.so
+    - lib/python3.6/lib-dynload/*.so
```
Thank you very much! :)

PR #1560, opened 2016-11-29
Copy CONDA_PATH_BACKUP into build/test env.

mingwandroid:
Hi @msarahan, this needs to be copied over to ensure that PATH is manipulated correctly (i.e. that the root env gets removed from it). Conda 4.3 will necessitate changing this again.

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -32,7 +32,7 @@ from conda_verify.verify import Verify
 
 # used to get version
 from .conda_interface import cc
-from .conda_interface import envs_dirs, root_dir
+from .conda_interface import envs_dirs, env_path_backup_var_exists, root_dir
 from .conda_interface import plan
 from .conda_interface import get_index
 from .conda_interface import PY3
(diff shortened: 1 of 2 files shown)


conda_build/build.py
```diff
@@ -855,6 +855,8 @@ def build(m, config, post=None, need_source_download=True, need_reparse_in_env=F
                     with path_prepended(config.build_prefix):
                         env = environ.get_dict(config=config, m=m)
                     env["CONDA_BUILD_STATE"] = "BUILD"
+                    if sys.platform != 'win32':
+                        env["CONDA_PATH_BACKUP"] = os.environ["CONDA_PATH_BACKUP"]
```
mingwandroid: Yeah since it was removed on conda master I'm hitting that issue. I was thinking to set s variable in conda_interface to the name of this env var actually? New version coming a little later..

mingwandroid:
The travis failure is a locking issue and does not seem related. I think this is ready to go now.

msarahan:
Yep, I need to fix that lock error.  It seems to be deterministic on Travis, but I can't repeat it locally yet.

Thanks for the other fix here.

PR #1575, opened 2016-12-14
Call make_hardlink_copy() on Windows too

mingwandroid:
Because otherwise conda fails to install and gives up the ghost:
```
  conda.exceptions.CondaOSError: OS error:
  failed to link (src='.../Tcl_FSGetCwd.3.gz', dst='.../Tcl_FSGetCwd.3.gz', type=1, error=CondaOSError:
  OS error: win32 hard link failed
```


msarahan:
Failures are unrelated.  Merging.  Thanks @mingwandroid

PR #1410, opened 2016-09-28
WIP: Support split packages

msarahan:
As discussed at #1338

PR #1576, opened 2016-12-15
Implement split packages and wheel building

### LINKED ISSUE #1338: Split packages proposal
@mingwandroid and I have been discussing split packages as an urgent prerequisite to enabling easier construction of build toolchains (perhaps with crosstools-ng, http://crosstool-ng.org/)

In coming up with ideas for implementation, we're looking at precedent set by Linux distributions.
- Arch: Defines list of packages to be built in pkgname field (bash array).  Bash functions have a specific form that connects them with package names.  Example: https://git.archlinux.org/svntogit/packages.git/tree/trunk/PKGBUILD?h=packages/gcc#n7
- Red Hat: "subpackages" - http://www.rpm.org/max-rpm/s1-rpm-subpack-spec-file-changes.html - explicit lists of files.
- Debian: Makefile based.  Additional targets specified in makefile and in helper files. https://wiki.debian.org/PkgSplit and http://php-java-bridge.cvs.sourceforge.net/viewvc/php-java-bridge/php-java-bridge/debian/rules?view=markup and http://php-java-bridge.cvs.sourceforge.net/viewvc/php-java-bridge/php-java-bridge/debian/

What we have as initial design ideas for conda-build are:
- Change the name field to allow lists.  Each entry is a separate package.  These entries would be used to select logic for installation in the shell scripts.

```
package:
  name:
    - somename_somefeature1
    - somename_somefeature2
```
- Alternatively, Keep the name field the same, but add an "outputs" field that lists other outputs.  If not specified, defaults to normal conda package output.

```
package:
  name: somename
  outputs:
    - somefeature1
    - somefeature2
```
- Prefer automatic collection of files, as is done now, over explicit lists of files.  After each install+packaging step, the files would be removed from the build prefix.
- Prefer functions defined in central shell/batch scripts (as opposed to several scripts), but have conda-build call the correct function/subroutine based on output name.
- Alternatively, have outputs list as dictionary instead, with values being scripts to run for packaging.

CC @pelson @ocefpaf @janschulz - we wanted to involve you, given your involvement with some of these systems.  Do you have opinions or battle scars to share?  Naming is all negotiable here at this point - if something like the "outputs" field seems like a good idea with a bad name.

msarahan:
closes #1338 

supersedes #1410


hainm:
fantastic. thanks.

hmaarrfk:
Is there documentation about how to do this well?

The documentation on [conda.io](https://conda.io/docs/user-guide/tasks/build-packages/define-metadata.html#outputs-section) only gives examples of really easy package that probably don't benefit from split packages. 

I'm trying to use this to split sqlite https://github.com/conda-forge/sqlite-feedstock/pull/31 but I don'tl ike the fact that now we have to specify each files section. 

Without a split package, conda-build would automagically find the files that were installed.

Are there better ways to do it?

I kinda wish there was a way for me to say: `files 1 and 2 go in this sub package`, `everything else go in this`.

PR #1579, opened 2016-12-16
add prefix length arg.  Allow disabling of prefix length fallback

msarahan:
With Conda currently silencing conda-build's logger output, the warnings about prefix length fallback have been hidden, leading to some packages being built unintentionally with short prefixes.  This adds a flag to disable that fallback behavior.  This also adds the longstanding community request that prefix length be settable, but with the warning that they should really prefer to set croot to non-encrypted locations instead.

CC @mingwandroid @ilanschnell

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -649,21 +649,24 @@ def create_env(prefix, specs, config, clear_cache=True):
                         'post-link failed for: openssl' in str(exc) or
                         isinstance(exc, PaddingError)) and
                         config.prefix_length > 80):
-                    log.warn("Build prefix failed with prefix length %d", config.prefix_length)
-                    log.warn("Error was: ")
-                    log.warn(str(exc))
-                    log.warn("One or more of your package dependencies needs to be rebuilt "
-                            "with a longer prefix length.")
-                    log.warn("Falling back to legacy prefix length of 80 characters.")
-                    log.warn("Your package will not install into prefixes > 80 characters.")
-                    config.prefix_length = 80
+                    if config.prefix_length_fallback:
+                        log.warn("Build prefix failed with prefix length %d", config.prefix_length)
+                        log.warn("Error was: ")
+                        log.warn(str(exc))
+                        log.warn("One or more of your package dependencies needs to be rebuilt "
+                                "with a longer prefix length.")
+                        log.warn("Falling back to legacy prefix length of 80 characters.")
+                        log.warn("Your package will not install into prefixes > 80 characters.")
+                        config.prefix_length = 80
 
-                    # Set this here and use to create environ
-                    #   Setting this here is important because we use it below (symlink)
-                    prefix = config.build_prefix
+                        # Set this here and use to create environ
+                        #   Setting this here is important because we use it below (symlink)
+                        prefix = config.build_prefix
 
-                    create_env(prefix, specs, config=config,
-                                clear_cache=clear_cache)
+                        create_env(prefix, specs, config=config,
+                                    clear_cache=clear_cache)
+                    else:
+                        raise
         warn_on_old_conda_build(index=index)
 
     # ensure prefix exists, even if empty, i.e. when specs are empty
--- a/conda_build/cli/main_build.py
+++ b/conda_build/cli/main_build.py
@@ -187,7 +187,33 @@ different sets of packages."""
         help=("folder to dump output package to.  Package are moved here if build or test succeeds."
               "  Destination folder must exist prior to using this.")
     )
-
+    p.add_argument(
+        "--no-prefix-length-fallback", dest='prefix_length_fallback',
+        action="store_false",
+        help=("Disable fallback to older 80 character prefix length if environment creation"
+              " fails due to insufficient prefix length in dependency packages"),
+        default=True,
+    )
+    p.add_argument(
+        "--prefix-length-fallback", dest='prefix_length_fallback',
+        action="store_true",
+        help=("Disable fallback to older 80 character prefix length if environment creation"
+              " fails due to insufficient prefix length in dependency packages"),
+        # this default will change to false in the future, when we deem that the community has
+        #     had enough time to build long-prefix length packages.
+        default=True,
+    )
+    p.add_argument(
+        "--prefix-length", dest='_prefix_length',
+        help=("length of build prefix.  For packages with binaries that embed the path, this is"
+              " critical to ensuring that your package can run as many places as possible.  Note"
+              "that this value can be altered by the OS below conda-build (e.g. encrypted "
+              "filesystems on Linux), and you should prefer to set --croot to a non-encrypted "
+              "location instead, so that you maintain a known prefix length."),
+        # this default will change to false in the future, when we deem that the community has
+        #     had enough time to build long-prefix length packages.
+        default=255, type=int,
+    )
     add_parser_channels(p)
 
     args = p.parse_args(args)
--- a/conda_build/config.py
+++ b/conda_build/config.py
@@ -18,12 +18,12 @@ from .utils import get_build_folders, rm_rf
 
 log = logging.getLogger(__file__)
 on_win = (sys.platform == 'win32')
+DEFAULT_PREFIX_LENGTH = 255
 
 # Don't "save" an attribute of this module for later, like build_prefix =
 # conda_build.config.config.build_prefix, as that won't reflect any mutated
 # changes.
 
-DEFAULT_PREFIX_LENGTH = 255
 conda_build = "conda-build"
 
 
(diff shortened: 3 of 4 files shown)


ilanschnell:
+1

mingwandroid:
> With Conda currently silencing conda-build's logger output

@kalefranz would it be possible to fix this?

kalefranz:
> would it be possible to fix this?

Absolutely. At one point I remember conda-build doing some things with conda's loggers, and if I'm now doing it with conda-build's, then that needs fixed too.  I hadn't realized it was an issue until you pointed it out.

kalefranz:
Digging in and investigating.  Here's what I've found so far.

First, python's logging is robustly configurable, but with a *huge* learning curve.  Doing it right is non-trivial.

I don't think it's accurate to say that conda silences conda-build's logger output.  Rather, conda--by default--sets the root logger level to WARN.

For reference, all of conda's logging configuration is in two places--and two places only.

* https://github.com/conda/conda/blob/master/conda/gateways/logging.py
* https://github.com/conda/conda/blob/master/conda/console.py

The first is the "new," and the second, the "old."  Eventually the old will be completely migrated.

I think the extent of code that's important here is this

```
def set_verbosity(verbosity_level):
    try:
        set_all_logger_level(VERBOSITY_LEVELS[verbosity_level])
    except IndexError:
        raise CondaError("Invalid verbosity level: %(verbosity_level)s",
                         verbosity_level=verbosity_level)

def set_all_logger_level(level=DEBUG):
    formatter = Formatter("%(message)s\n") if level >= INFO else None
    attach_stderr_handler(level, formatter=formatter)
    attach_stderr_handler(level, 'conda', formatter=formatter)
    attach_stderr_handler(level, 'binstar', formatter=formatter)
    attach_stderr_handler(level, 'requests')
    attach_stderr_handler(level, 'requests.packages.urllib3')

def attach_stderr_handler(level=WARN, logger_name=None, propagate=False, formatter=None):
    # get old stderr logger
    logr = getLogger(logger_name)
    old_stderr_handler = next((handler for handler in logr.handlers if handler.name == 'stderr'),
                              None)

    # create new stderr logger
    new_stderr_handler = StreamHandler(sys.stderr)
    new_stderr_handler.name = 'stderr'
    new_stderr_handler.setLevel(NOTSET)
    new_stderr_handler.setFormatter(formatter or _FORMATTER)

    # do the switch
    with _logger_lock():
        if old_stderr_handler:
            logr.removeHandler(old_stderr_handler)
        logr.addHandler(new_stderr_handler)
        logr.setLevel(level)
        logr.propagate = propagate
```

There isn't a single place in conda (as it turns out, but I had to verify to make sure) that conda-build's logger is being configured.  If it were, that here would look something like `attach_stderr_handler(level, 'conda_build', formatter=formatter)`.  I do, however, set the root logger level to WARN by default, and make sure that it is properly configured.  That's this `attach_stderr_handler(level, formatter=formatter)`.  This has implications. If you've configured logging to propagate to the root logger and you're relying only on that root logger configuration, then you'll likely have problems. 

It turns out conda-build has quite a few places logging configuration is touched.  There are two varieties.  The first looks similar to what I've done with conda, in so far as being explicit about the loggers being configured.

https://github.com/conda/conda-build/blob/d6a95b51fa595fb6a900434f61a8279b98efbefe/conda_build/build.py#L595-L607
loggers affected:
  * conda
  * binstar
  * install
  * conda.install  (this one is redundant)
  * fetch
  * print
  * progress
  * dotupdate
  * stdoutlog
  * requests

https://github.com/conda/conda-build/blob/d6a95b51fa595fb6a900434f61a8279b98efbefe/conda_build/utils.py#L451-L467
loggers affected:
  * `os.path.dirname(__file__)` ( ¯\\\_(ツ)\_/¯)
  *  conda
  * binstar
  * install (conda doesn't use an install logger, not sure if it's meant to affect conda or not)
  * conda.install  (this time not redundant)
  * fetch
  * print
  * progress
  * dotupdate
  * stdoutlog
  * requests

https://github.com/conda/conda-build/blob/d6a95b51fa595fb6a900434f61a8279b98efbefe/conda_build/environ.py#L51-L58
loggers affected:
  * conda_build.environ

https://github.com/conda/conda-build/blob/d6a95b51fa595fb6a900434f61a8279b98efbefe/conda_build/environ.py#L142-L147
loggers affected:
  * conda_build.environ

https://github.com/conda/conda-build/blob/d6a95b51fa595fb6a900434f61a8279b98efbefe/conda_build/cli/main_build.py#L27
loggers affected:
  * root

https://github.com/conda/conda-build/blob/d6a95b51fa595fb6a900434f61a8279b98efbefe/conda_build/cli/main_build.py#L281
loggers affected:
  * root

https://github.com/conda/conda-build/blob/d6a95b51fa595fb6a900434f61a8279b98efbefe/conda_build/cli/main_convert.py#L17
loggers affected:
  * root

https://github.com/conda/conda-build/blob/d6a95b51fa595fb6a900434f61a8279b98efbefe/conda_build/cli/main_develop.py#L16
loggers affected:
  * root

https://github.com/conda/conda-build/blob/d6a95b51fa595fb6a900434f61a8279b98efbefe/conda_build/cli/main_index.py#L12
loggers affected:
  * root

https://github.com/conda/conda-build/blob/d6a95b51fa595fb6a900434f61a8279b98efbefe/conda_build/cli/main_inspect.py#L15
loggers affected:
  * root

https://github.com/conda/conda-build/blob/d6a95b51fa595fb6a900434f61a8279b98efbefe/conda_build/cli/main_metapackage.py#L17
loggers affected:
  * root

https://github.com/conda/conda-build/blob/d6a95b51fa595fb6a900434f61a8279b98efbefe/conda_build/cli/main_render.py#L131-L137
loggers affected by first conditional branch:
  * root
  * `os.path.dirname(__file__)`
  *  conda
  * binstar
  * install (conda doesn't use an install logger, not sure if it's meant to affect conda or not)
  * conda.install  (this time not redundant)
  * fetch
  * print
  * progress
  * dotupdate
  * stdoutlog
  * requests
loggers affected by second conditional branch:
  * root

https://github.com/conda/conda-build/blob/d6a95b51fa595fb6a900434f61a8279b98efbefe/conda_build/cli/main_sign.py#L13
loggers affected:
  * root

https://github.com/conda/conda-build/blob/d6a95b51fa595fb6a900434f61a8279b98efbefe/conda_build/cli/main_skeleton.py#L21
loggers affected:
  * root

After documenting all of this now, it strikes me that conda-build actually isn't configuring anywhere it's own `conda_build` logger, except for a couple of times and via duplicated logic configuring the `conda_build.environ` logger.  Everywhere that conda-build attempts to configure it's own logger, it's using something like `logging.basicConfig(level=logging.INFO)`.  This configures the root logger, not the `conda_build` logger.

So it turns out that conda-build mutates conda's logger configuration in several places, and without studying the code paths, in ways that aren't clear to me what the implications are.  Conda never touches conda-build's logger configuration though.

I think the fix here is more with conda-build code, not conda.  The only thing I could really do to help in conda is to relax the "set root logger to WARN by default" part, but the implications of that would be dangerous.  Instead, conda-build should (1) consider consolidating all of it's logger configuration to one place, and (2) actually configure it's own logger, by attaching a handler to the `conda_build` logger and not relying on the root logger handler to output entries.

kalefranz:
Just want to reiterate again that python's logging configuration is really hard to get right.  Take a a minute or two to read through https://github.com/conda/conda/blob/master/conda/common/io.py.  It's helper functions mostly for dealing with logging (and some other io-related context managers). To fully configure a logger, there are a *lot* of attributes you have to set.  Each one does something different, even though some of them seem like they do the same thing.  It's all really confusing.  But maybe some of that code can help.

msarahan:
Thanks, I will look into configuring a package-local logger.  However, even with Conda configuring the root logger level to warn, why doesn't this message show up? https://github.com/conda/conda-build/blob/d6a95b51fa595fb6a900434f61a8279b98efbefe/conda_build/build.py#L654-L660

kalefranz:
This: https://github.com/conda/conda-build/blob/d6a95b51fa595fb6a900434f61a8279b98efbefe/conda_build/utils.py#L456

I actually sort of misinterpreted it at first.  But that looks suspicious to me.  The value there of `os.path.dirname(__file__)` depends on whatever the cwd is.  Sometimes the value might be just `conda_build`.  Other times the value is a relative path from the cwd.

I'm guessing what is meant there is actually just `'conda_build'`, right?

That's where I'd start, especially because it looks like it's getting invoked via https://github.com/conda/conda-build/blob/d6a95b51fa595fb6a900434f61a8279b98efbefe/conda_build/build.py#L607

kalefranz:
Oh, but there's `show_warnings_and_errors=True` in that line I referenced.  Here it's false though
https://github.com/conda/conda-build/blob/d6a95b51fa595fb6a900434f61a8279b98efbefe/conda_build/cli/main_render.py#L133

I'm just not quite sure.  If you can give me a command to run where I can set a break point there and I'll hit it, I'll take a look at the state of the logger config there.

PR #1581, opened 2016-12-16
Delete the work directory before running the tests

### LINKED ISSUE #1580: work directory not deleted for test stage
The work directory is not deleted when the recipe tests are run. This has caused a problem where paths from the source directory were hardcoded in the executable, but I did not notice because the test passed. See https://github.com/conda-forge/emacs-feedstock/issues/2#issuecomment-267542043.

I don't remember if this has always been this way, or if there was a good reason for it if it was. I agree that the work directory should stay intact if there are no tests, or if the build fails and the tests are not gotten to, but it should be deleted (or if you're worried about deletion, at least set aside) during the tests, for the same reason that the build environment is deleted for the tests.

asmeurer:
Fixes #1580.

Let me know if you want me to add some kind of test.

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -1166,6 +1166,9 @@ def test(recipedir_or_package_or_metadata, config, move_broken=True):
         print("Nothing to test for:", metadata.dist())
         return True
 
+    # Needs to come after create_files in case there's test/source_files
+    rm_rf(config.work_dir)
+
     print("TEST START:", metadata.dist())
 
     get_build_metadata(metadata, config=config)
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -1166,10 +1166,11 @@ def test(recipedir_or_package_or_metadata, config, move_broken=True):
         print("Nothing to test for:", metadata.dist())
         return True
 
-    # Needs to come after create_files in case there's test/source_files
-    rm_rf(config.work_dir)
-
     print("TEST START:", metadata.dist())
+
+    # Needs to come after create_files in case there's test/source_files
+    print("Deleting work directory,", config.work_dir)
+    rm_rf(config.work_dir)
 
     get_build_metadata(metadata, config=config)
     specs = ['%s %s %s' % (metadata.name(), metadata.version(), metadata.build_id())]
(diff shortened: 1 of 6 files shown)


msarahan:
Yes, please add a test.  Whatever best reflects the situation that caused you the headache with the emacs feedstock.

asmeurer:
I guess a recipe test? Or is there a good mechanism for Python tests of this sort of thing now?

asmeurer:
Is this Travis test failure ignorable?

asmeurer:
Ugh you guys are doing CLAs now. 👎

msarahan:
The Travis test failure is ignorable.  Conda 4.3 is in early release, and there's a known issue.

Recipes are often the easiest way to test this, but they're not the only way any more.  See tests/test_build.py for some lower-level tests, and tests/test_api_build.py for higher-level tests that mostly use recipes, but do so via a Python API so that we get coverage info.

As a former Continuum employee, you're not required to sign the CLA.  If they're really that upsetting to you, please email Travis.  I'm sure he'd appreciate your take on the matter.

asmeurer:
Weird, `rm_rf` is deleting the contents of the `work` directory, but not the directory itself. Any idea why that might be happening?

msarahan:
Regarding rm_rf, it probably is deleting it, but it is being re-created if you ever access that variable again.  All folder variables create the folders that they represent if they do not exist when the variable is accessed.  I got sick of having makedirs everywhere, and consolidated it.

I don't see the later access of that variable in the diff here, but that's probably what is happening.

asmeurer:
Ah that makes sense. Trying to completely delete it would be fragile and unnecessary, so I've changed the test to only check that it is empty.

msarahan:
Travis is as happy as it is going to be right now.  Waiting on appveyor, then will merge.  Thanks for the PR.

msarahan:
Fluke on appveyor - seems you are missing pytest-mock, which I just have added to master.  Merging.

dgrigonis:
Once `work` is not available anymore...

Could someone help me out to run pytest and make sure that conftest with custom options is properly picked up?

I was always using `work` directory for running tests on it...

asmeurer:
@dgrigonis you can use the [source files](https://conda.io/docs/building/meta-yaml.html#source-files) section of the meta.yaml to copy your tests from the source to the tests directory.

PR #1583, opened 2016-12-17
logger rework.  Package local logger.

msarahan:
Fixes logger messages not showing up due to root logger being configured at higher logging level.  Localizes logger configuration to conda-build.

This probably trims things back too far.  Running conda-build in debug mode should probably also run conda in debug mode, for example.  I leave that for a future PR.

CC @mingwandroid @kalefranz

### DIFF
--- a/conda_build/__init__.py
+++ b/conda_build/__init__.py
@@ -8,6 +8,8 @@ from ._version import get_versions
 __version__ = get_versions()['version']
 del get_versions
 
+import logging
+
 # Sub commands added by conda-build to the conda command
 sub_commands = [
     'build',
@@ -21,3 +23,13 @@ sub_commands = [
     'sign',
     'skeleton',
 ]
+
+# Set default logging handler to avoid "No handler found" warnings.
+try:  # Python 2.7+
+    from logging import NullHandler
+except ImportError:
+    class NullHandler(logging.Handler):
+        def emit(self, record):
+            pass
+
+logging.getLogger(__name__).addHandler(NullHandler())
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -55,7 +55,7 @@ import conda_build.os_utils.external as external
 from conda_build.post import (post_process, post_build,
                               fix_permissions, get_build_metadata)
 from conda_build.utils import (rm_rf, _check_call, copy_into, on_win, get_build_folders,
-                               silence_loggers, path_prepended, create_entry_points,
+                               path_prepended, create_entry_points,
                                prepend_bin_path, codec, root_script_dir, print_skip_message,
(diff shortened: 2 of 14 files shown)


conda_build/__init__.py
```diff
@@ -21,3 +23,13 @@
...
+
+# Set default logging handler to avoid "No handler found" warnings.
+try:  # Python 2.7+
+    from logging import NullHandler
+except ImportError:
+    class NullHandler(logging.Handler):
+        def emit(self, record):
+            pass
+
+logging.getLogger(__name__).addHandler(NullHandler())
```
kalefranz: Maybe you're not done with this PR yet, so this could be a premature comment.
Using NullHandler is exactly the right thing to do for a library.  It makes your library a good logging citizen.  Also, the package's root-level `__init__.py` file is almost always the right place to invoke the basic logging configuration.  (In conda's case I'm not, because I don't want to bog down the `..activate` etc cli stuff with logger initialization.  It's way too slow already.)

However, unless you add another handler somewhere else, the problem with conda-build logs will remain.  I'm guessing you're just not done with the PR though.

msarahan: Thanks for the comment.  I'm not sure why, but things seem to be working with the code as is.  Any insight?

I have a package on my channel, intentionally built with an 80-character prefix.  This recipe will try to bring it in as a dependency:

```
package:
  name: test-pkg
  version: 1.0

requirements:
  build:
    - conda-build-test-has-prefix-files
```

Now, if you build that recipe (``conda build -c msarahan .``), you see the logger output, as I would hope.  is there something else that isn't showing because of the way I have things now?

```
 msarahan@0109-msarahan-2  ~/code/conda-build/tests/test_pkg   logger  conda build -c msarahan .
BUILD START: test-pkg-1.0-0
updating index in: /Users/msarahan/miniconda2/conda-bld/osx-64
updating index in: /Users/msarahan/miniconda2/conda-bld/noarch

The following packages will be downloaded:

    package                    |            build
    ---------------------------|-----------------
    conda-build-test-has-prefix-files-1.0|                0           3 KB  local

    conda-build-test-has-prefix-files: 1.0-0 local

WARNING conda_build.build:create_env(648): Build prefix failed with prefix length 255
WARNING conda_build.build:create_env(649): Error was:
WARNING conda_build.build:create_env(650): Placeholder of length '80' too short in package local::conda-build-test-has-prefix-files-1.0-0.
The package must be rebuilt with conda-build > 2.0.
WARNING conda_build.build:create_env(651): One or more of your package dependencies needs to be rebuilt with a longer prefix length.
WARNING conda_build.build:create_env(653): Falling back to legacy prefix length of 80 characters.
WARNING conda_build.build:create_env(654): Your package will not install into prefixes > 80 characters.
```

Issue #1142, opened 2016-07-23
Recipe customization / variants / better pinning


msarahan:
This issue is to track progress on an issue that has already spanned several PRs and issues.  Since this issue has already spread across several PRs, this issue is being proposed to be the central discussion point, and any discussion on PRs should strictly be review of code, more than discussion of concepts.  

The idea is that we need a better way of customizing recipes.  When the numpy x.x style of pinning was added ( #573 ), people appreciated it, but wanted to generalize it.  Jinja2 is already flexible enough, but we lack configuration schemes for reusable data, as well as some syntactic sugar to make configuration sufficiently easy.
## Design goals:
- support config files at 3 levels, with lower levels overriding the upper ones:
  - included with conda-build
  - user's install-wide
  - per-package
- Identify key confiugration types:
  - pinning versions (versions that must match between build and run)
  - variants of a particular feature (BLAS: MKL/OpenBLAS/ATLAS/etc.)
  - Specifying compiler / toolchain options
- Specify how to distinguish builds: how do different variations end up with different filenames?
  - Probably can't just have all variations contribute to build string directly.
  - Hash instead?  Installation would need to access metadata, not rely on build string

Current PR for implementation: #966 

Past issues/PRs where much discussion has occurred: 
- #911: @mwiebe suggests a batch/bash file-based approach to lining up compiler
- #908: @msarahan added conda render command, moving towards all version and recipe metadata being required to be present pre-build
- #857 @kalefranz's proposal to add a separate jinja2 syntax for each jinja pass
- #848: @ukoethe's proposal to improve pinning
- https://github.com/conda/conda/issues/1959: @ukoethe's proposal of customization of compiler, to enable non-standard Python distributions on Windows (to support C++11)
- #747 @pelson's recommendation that build dependencies be resolved before build id is computed
- #728 @johanneskoester's generalization of the x.x syntax to all packages
- #653 @mwcraig's 
- #650 @pelson's proposal that the x.x numpy pinning syntax need only be in the run requirements, allowing the build requirements to be more meaningful (such as lower/upper bounds)
- #573 @ilanschnell's original inclusion of the numpy x.x syntax

pitrou:
Is there an explanation of what the "config file" contents would be?

msarahan:
That's up for discussion.  My thoughts thus far have been to support sections with some meaning, such as:

```
pinned_packages:
  - hdf5
```

[...]

PR #1585, opened 2016-12-18
WIP Build customization

msarahan:
Implements #1142 

Supersedes #966 

Copying design goals from #1142:

## Design goals:

- [x] support config files at 3 levels, with lower levels overriding the upper ones:
  - [x] included with conda-build
  - [x] user's install-wide
  - [x] per-package
  - [x] document how older options, such as CONDA_PY and --python interact with/override config files
- [x] Identify key confiugration types:
  - [x] pinning versions (versions that must match between build and run)
  - [x] variants of a particular feature (BLAS: MKL/OpenBLAS/ATLAS/etc.)
  - [x] Specifying compiler / toolchain options
- [x] Specify how to distinguish builds: how do different variations end up with different filenames?
  - [x] Probably can't just have all variations contribute to build string directly.
  - [x] Hash instead?  Installation would need to access metadata, not rely on build string


msarahan:
@groutr @ukoethe @stuarteberg - finally, I think this is very close to merging.  Please take a look.

Docs proposed at https://github.com/conda/conda-docs/pull/414 - almost certainly not complete, but hopefully a useful start.

stuarteberg:
Thanks for the ping.  I'll try to offer useful comments.  I'll need some time, of course -- this covers a lot of ground! -- and I can't start today.

@ukoethe: FWIW, I built html for @msarahan's new documentation**  here:
http://stuarteberg.github.io/conda-docs/building/variants.html

** (Built from msarahan/conda-docs@bc1f2e20b9aa1653574a30f3762d0471313bc3f4)

msarahan:
Yes, @stuarteberg, this spun out of control.  You guys pointed me in the right direction, though.  Thanks for taking a look!

mingwandroid:
> http://stuarteberg.github.io/conda-docs/building/variants.html

That's fortuitous as @msarahan asked me to review the docs not long ago :-)

jjhelmus:
@msarahan I'm a bit confused on the behavior of the `pin_compatible` function which seems to be one of the key components of this change.  Specifically the text:

---

Looks up compatibility of named package installed in the build environment, and writes compatible range 
pin for run and/or test requirements. Presently primarily only a semver-based assumption: >=(current
 version),<(next minor version).

---

This seems to indicate that if the `foobar` package was included in the `pin_compatible` section and version 1.2.3  was used in the build, the resulting pinning would be `foobar >= 1.2.3, <1.3`.  

This seems overly strict based upon my understanding of semver.  Packages following semver should guarantee that the public API does not change without the major version (X.y.z) being incremented.  Based on this the pinning can safely be set to `foobar >= 1.2.3, <2`.  

That said, many software projects do not follow semver (including numpy) and therefore pinning to the minor version, or even patch version, may be desired.  Also many packages use alternative versioning schemes which should be at least partially supported.

The text seems to indicate that this compatibility information will be determined from ABI Laboratory with some yet to be determined manner.  This idea has merit but I think a mechanism to override these automatically determined compatibly ranges would be prudent.  

Perhaps using and expanding the already existing `x.x` syntax would be possible as a method for defining version level compatibility where the number of `x`'s define the compatibility.  

I'm imaging the follow syntax for specifying package compatibility:

```
pin_compatible:
  - spam x
  - eggs x.x
  - foo x.x.x
```

If the build environment had version 1.2.3 of all three packages the range pins created would be:

```
spam >= 1.2.3, <2
eggs >=1.2.3, <1.3
foo >=1.2.3, <1.2.3
```

msarahan:
@jjhelmus, I really don't like the x.x syntax, but you make a pretty compelling case for it, and you've given it a lot more meaning than it used to have.  I'll think about to code it, but I think you are absolutely right that I need to provide a way to specify/tweak the compatibility behavior.

ABI Laboratory reports are at https://github.com/lvc/abi-reports/tree/master/report - they express ABI compatibility from one version to another.  In order to span large version ranges, we'll have to process their files somehow.

Thanks for looking at it!

mwiebe:
A question about the pinning - is there a way for the dependency package (e.g. custom build of Python against a custom compiler with non-standard settings like enabling thread-sanitizer) to specify the default way it gets pinned in dependent recipes (e.g. NumPy built against that Python)? It seems to me that it knows more about how its ABI will change than other packages which include it, and should have the main say about it.

msarahan:
@mwiebe thanks for the comment.  Let's talk in terms of use cases.  I'll name the ones I've thought of, but please add to this list.

* packages utilizing a C/C++ API (the numpy x.x use case).  This is handled with the ``pin_run_as_depends`` feature of this PR.  It is not as fine grained as it could be.  If the dependency is in that list, and in the build requirements, and in the run requirements, then it gets pinned.  Here, numpy could be pinned needlessly.  The workaround is to only put numpy in ``pin_run_as_depends`` in the variant config file in the recipe itself.  That should localize its effects to that particular build.
* Downstream compiled things, like you mention: I think these are handled with the ``{{ compiler('language') }}`` and ``{{ runtime('language') }}`` concepts.  In order to be consistent, every package in the stack would need to have that compiler jinja2 function used as a build requirement.    This would get you build-time consistency with your options.  Doing so means that the specific variant actually filled into that template variable gets recorded as a build requirement, but this alone is not enough to get you consistency at runtime.  The key to the ``{{ runtime('language') }}`` jinja2 templating is that whatever values are possible for it need to be mutually exclusive (i.e. they all need to be different versions of the same package) so that only one can be installed at a time.  This uses conda's solver to enforce the consistency.  With the corresponding (and mutually exclusive) ``{{ runtime('language') }}`` jinja2 templating in the runtime requirements, the set should be consistent at runtime.

This is somewhat conjecture right now, as I haven't actually created such a set of software.  However, I think it should work...

mwiebe:
In how we're using conda, there are many packages representing C/C++ APIs, a majority of packages fit into both cases simultaneously. It feels funny to put a list of say 20 packages in the build, run, and pin_run_as_depends list, it would make more sense that those 20 packages indicate in their own metadata that they should be pinned by default, and the recipes using them all just says it's using those library and doesn't have to worry about pinning.

As an aside, the current build/run distinction already works poorly for libraries which have variants built as dynamically vs statically linked. For a dynamic library, it should be both build/run, and for a static library it should just be build. The same exact recipe will be built in both ways, and with 20 dependencies following this, that's 2^20 possible configurations of static/dynamic dependencies, it's not something to manage by hand.

mwiebe:
For your numpy use case, where dependent libraries might use the numpy API or not, could doing split packages solve it more cleanly? E.g. a numpy build would split into `numpy` and `numpy_c_api` packages, and all the .lib and .h files would just go into the latter. The `numpy_c_api` would insert the additional pinning necessary for `numpy` under C API usage.

msarahan:
I see, yes.  I think pin_run_as_depends should probably be extended into the recorded package metadata, so that just using a package means that if it defines pin_run_as_depends, then anything that depends on the package will also have that.

For example, something like:

```
package a:
  pin_downstream: True
```

```
package b:
  depends on package a, will have pin_run_as_build = [(defaults) + a]
```

For your numpy and numpy_c_api, that's a great way to do it.  Unfortunately, it requires everyone to change their recipes.  I think we'd probably need to support both ways for quite a while, but moving longer-term to your idea solely would make life a lot simpler.

mwiebe:
That sounds good, yeah. One possible tweak to your example syntax would be to include the nature of the pinning, because the package itself knows whether it is following semantic versioning or some other scheme:

```
package a:
  pin_downstream: a >= {{ version }}, a < {{ max_version }}
```

mwiebe:
One thought on the `x.x` syntax, could that be done via jinja2 to avoid special cases? E.g. `pkg_a x.x` be written as `pkg_a {{ verconstraint(version, 'x.x') }}`.

msarahan:
Thanks @jjhelmus @mwiebe I have implemented both of your suggestions, I think, in https://github.com/conda/conda-build/pull/1585/commits/cb674baefd917e2a1f71171f8970451d2e00a2ff

Docs at https://github.com/msarahan/conda-docs/blob/document_conda_build_variants/docs/source/building/variants.rst#customizing-compatibility

msarahan:
@mwiebe, I'll look into the idea of persistent downstream pinning with upper and lower bound constraints tomorrow.

minrk:
How precisely can `pin_run_as_build` constraints be specified? Most of the time, I'd like to pin major.minor, but I've recently discovered that swig has to be pinned all the way to major.minor.patch for fenics, as patch releases break things.

msarahan:
@minrk, that option is an exact pin - build number and everything.  The pin_compatible function is what gives you flexibility,  and that's what @jjhelmus was asking about, and what I hope to have implemented last night with https://github.com/msarahan/conda-build/commit/24a916e397cfe517825ee697cd320f221b8e9c1c

minrk:
Excellent, thanks @msarahan!

ukoethe:
Awesome work, @msarahan! I think you addressed all my suggestions in spirit, if not in letter. Since I'm not creating many new recipes at the moment (probably again in spring), I can't comment on any subtleties I might have encountered. Still, some things popped up upon reading your explanations:

> from https://github.com/conda/conda-build/pull/1585#issuecomment-276209701: Is there a way for the dependency package (e.g. custom build of Python against a custom compiler with non-standard settings like enabling thread-sanitizer) to specify the default way it gets pinned in dependent recipes (e.g. NumPy built against that Python)?

I understand that you've implemented this in the meantime, and I really like the functionality. However, my feeling is that the current rules involving `pin_run_as_build`, `pin_downstream` etc. (what takes precedence? what are the defaults? how are customizations specified?) have become a bit more complicated and less consistent than they should be. Perhaps you can come up with some simplifications by looking at the big picture of all the supported options?

> from the documentation: Pinning expressions are really only counting the number of things separated by the . character. What you put as the actual characters doesn't matter.

I strongly recommend to be less liberal here and fix just a single permissible letter (`p` or `x`). Otherwise, you needlessly constrain your generalization possibilities later, when someone proposes a creative syntax extension. Under the current anything-goes scheme, you would have to come up with complex backwards compatibility rules to support even the most minor change. (I think I'd go with `x` because `p` is more common in English words.)

groutr:
I agree with @ukoethe on the syntax of pinning expressions.  I don't see the common usage deviating from a single letter.  If it really is arbitrary, and the letters don't contribute to anything meaningful, why use letters at all?  `.` would be equivalent to `p.p` (or `..` for `p.p.p`)

msarahan:
Thanks @ukoethe and @groutr for the feedback.  Very sane suggestions.  We're in the middle of AnacondaCON right now, but I'll revisit and make your suggested changes probably late this week or this weekend.

conda_build/utils.py
```diff
@@ -718,7 +721,128 @@ def __exit__(self, et, ev, tb):
...
+
+class HashableDict(dict):
+    """use hashable frozen dictionaries for resources and resource types so that they can be in sets
+    """
+    def __init__(self, *args, **kwargs):
+        super(HashableDict, self).__init__(*args, **kwargs)
+        self = _convert_lists_to_sets(self)
+
+    def __hash__(self):
+        return hash(json.dumps(self, sort_keys=True))
```
groutr: Does it matter that hashes of strings are not the same between different instances of the Python interpreter?  Do we want to be able to reproduce the hashes?

conda_build/jinja_context.py
```diff
@@ -203,14 +205,154 @@ def load_file_regex(config, load_file, regex_pattern, from_recipe_dir=False,
...
+    },
+    'linux': {
+        'c': 'gcc',
+        'cxx': 'g++',
+        'fortran': 'gfortran',
+    },
+    # TODO: Clang?  System clang, or compiled package?  Can handle either as package.
+    'osx': {
+        'c': 'gcc',
+        'cxx': 'g++',
```
jakirkham: IMHO this should be the system toolchain, which would be `clang` and `clang++`.

Apple has been moving towards the elimination of `gcc` for some time. 10.6 was the last version that actually had these as the standalone default system toolchain. By 10.7 they were merely frontends to llvm. By 10.9, `gcc` merely redirects to `clang`, but links with `stdlibc++` for C++ code. It should also be noted that all of the libraries and includes are from 4.2.1 as it was the last version under GPL2 (4.2.2 was under GPL3). So if one uses this as the system compiler, they inherit all the bugs from working with this old version.

ref: http://stackoverflow.com/a/8707446
ref: http://stackoverflow.com/q/19535422
ref: http://unix.stackexchange.com/a/49970

Note: Am not saying people can't install something else (e.g. the `gcc` package) and use that. Just that the default case should assume the user has the system toolchain available and make no other assumptions.

msarahan: Agree completely.  I'll change it.  I need to create packages for clang and clang++ - even if they are nothing more than setting the appropriate environment variables.  Same thing for the vs20?? compilers.

conda_build/utils.py
jakirkham: That's a good point. Doesn't Python 3 add some random value to these?

I suppose one could pick a particular type of hash from `hashlib` (e.g. `sha256`). Not sure if that is needed or not though.

msarahan: Yep, going with sha1.  Starting with 4 characters.  We'll see...

groutr: @jakirkham Python 3.3+ randomizes string hashing by default.   Recent versions of Python 2.7 appear to do the same.

jakirkham: Yep, that's what I was meaning. Didn't realize this got backported to Python 2. Do you know what release it showed up in?

groutr: For Python 2, I'm mistaken.  The default behavior is not to randomize string hashes.  It is enabled with the -R flag.  The change was added in for 2.7.3rc1.
https://hg.python.org/cpython/raw-file/v2.7.3/Misc/NEWS

Python 3.3+ is when the behavior became default.  I guess this is yet another *very subtle* thing to check in Python 2 -> Python 3 conversions.

groutr: @msarahan 4 hex characters will contain 16 bits of the 160 bit sha1 hash (4 bits per hex character).  Fortunately, we should be able to estimate if its good enough by calculating the probability of a hash collision.  If we have 20 variants, the probability of a collision is ~2.9%.   If we do those same 20 variants with 20 bits (5 hex chars), the probability of a collision drops to ~.02%.  How many variants of the same package do we want to be able to handle?

Method described here: http://preshing.com/20110504/hash-collision-probabilities/
Convenient calculator here :) http://everydayinternetstuff.com/2015/04/hash-collision-probability-calculator/

msarahan: Fantastic!  Thanks for doing this.  IMHO, 5 chars is a better choice.  Maybe 6 to be really conservative.  I'll update the code to 6 chars.

jakirkham: > It is enabled with the -R flag. The change was added in for 2.7.3rc1.
https://hg.python.org/cpython/raw-file/v2.7.3/Misc/NEWS

Didn't know setting `PYTHONHASHSEED` would enable this on Python 2. Good to know. Nice little check for moving to Python 2/3 or Python 3 compatible code. Thanks.

msarahan:
@mwiebe I have now implemented pin_downstream.  There's examples at

https://github.com/conda/conda-build/pull/1585/files#diff-ebfc8e2b68ef082929da0400b7dddbabR6

https://github.com/conda/conda-build/pull/1585/files#diff-ea35225173773a853749849d5deffb05R61

you can specify this either for the top-level recipe or any number of the split package outputs.  If the split packages are specified, they override the top-level recipe.

Still need to fix a few tests, but this is getting close.

jakirkham:
Putting this on your radar @pelson as I expect you will be pretty interested in this.

msarahan:
Merging.  Outstanding issues:

- retry logic for compensating for conda's handling of the package cache fails builds on Travis.  Working with conda team to come up with a proper fix for the package cache problems.
- Windows builds failing noarch tests due to conda packaging error.  Being tracked at https://github.com/conda/conda/issues/4715 - should be fixed in conda 4.3.14.

jasongrout:
Wow, this is huge. Thanks! Is there a high-level summary added to the docs that I can read?

CC @aburgm, @SylvainCorlay

jakirkham:
FYI apparently this added `six` as a requirement, but it doesn't appear in `run` currently.

msarahan:
@jasongrout docs are at https://github.com/conda/conda-docs/pull/414

@jakirkham thanks - fixed in #1766

Issue #1588, opened 2016-12-19
conda build produces a corrupted output rendered meta.yaml


mwiebe:
What appears to be happening is the code https://github.com/conda/conda-build/blob/master/conda_build/build.py#L221 is overwriting perfectly fine build deps with a function object that gets rendered in some yaml escape syntax. The repr of the bad `build_deps` is `<bound method Environment.package_specs of <conda_build.environ.Environment object at 0x7fafdb8c6050>>`

If I fix up that line 221 to this (adding the function call):

```
    build_deps = environ.Environment(m.config.build_prefix).package_specs()
```

the corrupted yaml shown below becomes

```
   build:
       - toolchain 1.0 vfxp2016
       - cmake 3.7.1 vfxp2016_1
       - vfxp2016 2.0 4
       - zlib 1.2.8 vfxp2016_9
   run:
       - toolchain * vfxp2016
```

The output from `conda render` which is as expected is:

```
requirements:
    build:
        - vfxp2016
        - cmake
        - zlib
    run:
        - toolchain * vfxp2016
```

The output from `conda build` when it produces its final rendered output gets corrupted to:

```
requirements:
    build: !!python/object:__builtin__.instancemethod {}
    run:
        - toolchain * vfxp2016
```

A corrupted output from another platform is:

```
requirements:
    build: !!python/object/apply:builtins.getattr
        - !!python/object:conda_build.environ.Environment
            _meta: C:\Anaconda3\conda-bld\ptex_1482084865450\_b_env\conda-meta
            _packages: {}
            path: C:\Anaconda3\conda-bld\ptex_1482084865450\_b_env
        - package_specs
    run:
        - toolchain * msvc2015rel
```

Curiously the vfxp2016 one fails to parse, while the msvc2015rel one succeeds.

msarahan:
Reopening to remind me to write a test for this.  Thanks @mwiebe

sscherfke:
Current conda-build versions now produce an empty list for build requirements.
Input recipe:
```
requirements:
  build:
    - python
    - setuptools
```
built package -> info/recipe/meta.yaml:
```
requirements:
  build: []
```

msarahan:
Thanks, I'll look into this ASAP.

msarahan:
fixed and tested in #1585

PR #1589, opened 2016-12-19
Fix bug corrupting rendered meta.yaml from conda build

mwiebe:
Fixes https://github.com/conda/conda-build/issues/1588. Any ideas how a test for this would be structured?

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -219,7 +219,7 @@ def copy_recipe(m, config):
         # we only care if we actually have build deps.  Otherwise, the environment will not be
         #    valid for inspection.
         if m.meta.get('requirements') and m.meta['requirements'].get('build'):
-            build_deps = environ.Environment(m.config.build_prefix).package_specs
+            build_deps = environ.Environment(m.config.build_prefix).package_specs()
 
         if not rendered_metadata.meta.get('build'):
             rendered_metadata.meta['build'] = {}


msarahan:
Thanks!  I'll look into a test tomorrow and submit a PR on your branch.

I'm also pretty close to working out the build customization stuff, finally.  See #1585 - but I'll also ping you when I think it's really ready to be honestly evaluated.

I think where I'm headed with it is that compilers are going to be conda packages themselves.  Things like VS may not actually contain the compilers, but will contain the code to put the necessary compilers on PATH.  This is going to depend on using conda-build to activate the build environment, as I'm thinking the compiler activation stuff will be scripts that run when an environment is activated.  Still about a day or two's work to get there.

mwiebe:
Cool! The way our stuff is set up right now is with a `toolchain` variant package the way @mcg1969 gisted, to represent whether it's msvc 2015 release, msvc 2015 debug, etc, and then a compiler configuration package on top of that which brings in the actual compiler via a script that all our recipes source when they start. Currently we're not using multiple compilers for a given `toolchain`,  but the structure will be useful for packages built with clang or mingw, but matching the runtime of a particular msvc.

mwiebe:
We've also started using the `script_env` for parameterizing packages, and create build matrices based on all the possible values of these variables (with the `build/skip` metadata pruning the explosion of possible combinations).

msarahan:
Not sure how to test.  Probably should build a package and look at the generated meta.yaml file in that built package.  Merging for now, will leave your issue open to remind me to test.  Thanks @mwiebe

mwiebe:
If you want the issue to stay open you'll have to reopen it, I put the "closes" meta-phrase in the merge request. :)

PR #1592, opened 2016-12-19
Support for python 3.6

hajs:
Currently conda-build does not understand recipes with
```yaml
build:
  skip: true  # [py36]
```
This patch should add the missing variable for the upcoming Python version.


msarahan:
Thanks!

PR #1598, opened 2016-12-22
remove packaged files from build prefix, to give next script fresh start

msarahan:
Fixes a couple of minor issues with split packages:

1. remove files in build prefix prior to next script, so that each script starts "fresh" and duplicates in packages are OK (though discouraged).
2. only make a metapackage if an output with the same name is not explicitly listed in the outputs
3. add cmd.exe execution flags

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -785,7 +785,9 @@ def bundle_conda(output, metadata, config, env, **kw):
         if f not in files:
             files.append(f)
     files = filter_files(files, prefix=config.build_prefix)
-    output_folder = os.path.join(config.output_folder, config.subdir) if config.output_folder else None
+    output_folder = None
+    if config.output_folder:
+        output_folder = os.path.join(config.output_folder, config.subdir)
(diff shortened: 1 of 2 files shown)


patricksnape:
@msarahan this commit (specifically [these lines](https://github.com/conda/conda-build/pull/1598/files#diff-ac55d744935b1f37a3369258b88e8f1fR829) as found by running `git bisect`) - causes some odd behaviour - seen [here](https://travis-ci.org/menpo/lsfm/jobs/191635715#L734). It seems that if you have a script in your project then the `rm_rf` call traverses the possibly? sym-linked script file and thus deletes the script in the original repository. This also may only be an issue for recipes built where the source is next to the recipe folder and thus copied using a relative path:

```
source:
  path: ../
```

To be clear:

After this commit, after the test phase the `bin/lsfm` script is deleted from inside the repository and not just from the work/build environment.

msarahan:
That's bad. Thanks for raising. Will fix in next bugfix release.

PR #1599, opened 2016-12-22
rename two instances of conda.export to conda.exports

kalefranz:


kalefranz:
ping me directly @msarahan if you encounter anymore problems that could block your release

msarahan:
Thanks.  This gets things further, but it seems that there are issues with conda 4.3: https://travis-ci.org/conda/conda-build/jobs/186129726

I think the critical detail here is that I'm adding a channel dynamically at test time: https://github.com/conda/conda-build/blob/master/conda_build/build.py#L1177-L1181

msarahan:
Merging this, assuming we can work out the 4.3 issues separately.

msarahan:
I suspect that the issue might be caching of the index, but I'm not sure.

msarahan:
I am also concerned that conda 4.3 makes my test suite take twice as long: https://travis-ci.org/conda/conda-build/builds/186129723 - any comment on why that might be?

kalefranz:
Yeah, 4.3.0 and 4.3.1 are slower.  There's a lot of new object creation that is on my list to profile and knock out the hot spots.  There's also a TON of asserts from getting the `Dist` object stuff worked out.  If you invoke conda's python interpreter with `-O`, I'm pretty sure there's a substantial speed up right now.  After we're sure we have the dist stuff settled, I'm also pulling out all of those asserts.

kalefranz:
Yeah also might have to tell conda to clear some of its in-memory caches between tests.  I'll try to look at that if I get a sec.

Issue #712, opened 2015-12-22
Conda convert misses some files


gomyhr:
When running "conda convert <package> -p win-64" on linux-64, one of my entry-points were missing in the windows-package. It turns out that the problematic code line is

```
files = files.replace(oldpath, newpath)
```

in `conda-build/convert.py`where files is a string containing the whole info/files file. When I have two entry points (scripts) where one is a the first part of the other, e.g. `bin/foo` and `bin/foo-bar`, the latter may be converted to Scripts/foo-script.py-bar in the same round that the first is converted to Scripts/foo-script.py (this only happens if bin/foo comes before bin/foo-bar). This makes the wrong file name end up in info/files, and therefore linking of that file does not work.

A minimal fix is probably to add a newline to each of the strings in the call to replace(). I have suggested a slightly more elaborate fix at https://github.com/gomyhr/conda-build/commit/a9e987e79cfb3f02918ca0b5b2ac1cd5cb29216e, where files is a list of strings (one per line) and is then converted back to a string once the comparison is done. The large string is anyway converted to a list, sorted and converted to a string again, so this only does the conversion to list earlier in order to simplify the manipulation. At the same time, it prints a warning if the file name it tries to replace is not present in the file (which I guess means that the package is corrupt). Feel free to use all or parts of that commit to fix the issue, or to fix it in an entirely different manner.

groutr:
Thanks for the report and suggestion.  We will investigate this further.

gomyhr:
This seems to have been fixed with commit dd04ea90b1f740a928b7cc8304c847910efa2c47

msarahan:
Thanks for following up on this!

gomyhr:
It turns out that the bug was not fixed, only that the consequences changed. The bug in generation of info/files previously meant that the files were not copied into the package. Now, the files are generated correctly, but info/files is still wrong, making the converted packages uninstallable.
I have submitted PR #1609 to change a test so that it will detect this issue, along with a fix.

gomyhr:
Fixed in PR #1609.

PR #1609, opened 2017-01-01
Package consistency

gomyhr:
This is really a bugfix for issue #712. 

The actual fix is in commit 
https://github.com/gomyhr/conda-build/commit/23230c65c0bfad50f4bca755603514ed56bfbe51. It is a bit more elaborate than the simplest solution in the issue, in that info/files is read as a list instead of a string straight away. This should tie better in with the change when info/paths.json will be used instead of info/files later.

The first two commits (
https://github.com/gomyhr/conda-build/commit/8f5b3d19e2e6a168baf060e08b7af9a5deb8f393 and 
https://github.com/gomyhr/conda-build/commit/eb459b13243714e66caf5178fe14f598e0e84c02) add a package consistency check to the convert test with entry points.  I wasn't quite sure tests/utils.py was the right place to add a function for checking package consistency, but I could not see any better place.

The third commit 
(https://github.com/gomyhr/conda-build/commit/ccb82c54698c68f133fbf00940eec504acbe403a) adds another entry point to the test, which generates an inconsistent package, and this is picked up by the newly added test.

The last commit (
https://github.com/gomyhr/conda-build/commit/8adc066e415a4dfdcfb3d9331e5650eea0552c97) adds a warning if things go wrong. I thought it was more transparent to split this off from the bugfix itself.

### DIFF
--- a/conda_build/convert.py
+++ b/conda_build/convert.py
@@ -199,7 +199,7 @@ def get_pure_py_file_map(t, platform):
     dest_plat, dest_arch = platform.split('-')
     dest_type = 'unix' if dest_plat in {'osx', 'linux'} else 'win'
 
-    files = t.extractfile('info/files').read().decode("utf-8")
+    files = t.extractfile('info/files').read().decode("utf-8").splitlines()
 
     if source_type == 'unix' and dest_type == 'win':
         mapping = path_mapping_unix_windows
@@ -268,7 +268,8 @@ def get_pure_py_file_map(t, platform):
                 assert member.path == oldpath
                 file_map[oldpath] = None
                 file_map[newpath] = newmember
-                files = files.replace(oldpath, newpath)
+                loc = files.index(oldpath)
+                files[loc] = newpath
                 break
         else:
             file_map[oldpath] = member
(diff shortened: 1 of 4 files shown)


gomyhr:
The ValueError caught in ``conda_build/convert.py`` line 274 is first and foremost triggered by a corrupt package - i.e. the info/files are missing some files, but also if ``files`` is somehow wrongly updated due to a similar bug as the one that is fixed. 

So on second thought, it might be better to not only issue a warning, but rather issue an error. The reason I ended up with this was that I wanted to to be as little invasive as possible and not change more than fixing the bug. I also saw how other warnings were issued, and tried to implement it similarly. With the old string replace approach, there was no check that anything was actually replaced (and when the bug triggered, some paths in file would have been corrupted, so no replacement was made)

I could change this part by either not caching the exception or raising an error in another way. It seems to me that most places in ``convert.py``, exceptions are raised when an error condition is detected. How is the best way to do this here?

msarahan:
IMHO, it's fine to just not catch the exception and let it bubble up and kill the build.

gomyhr:
I have replaced the last three commits with two new ones:

Instead of adding a single entry point I added two new ones in 
https://github.com/conda/conda-build/pull/1609/commits/1eab14b766ced6309ccd7be91ba2a74dd3076f07. I discovered that the version committed did not actually trigger the bug on my system after I (before committing) changed the name from test-script-manual-extra to test-script-manual-postfix. The reason is that the file order in the package is undefined, and with -extra it ended up after the other and with -postfix it ended up before. The bug only triggers when the shorter version comes before a longer version in the tar file. With the additional entry point the probability for triggering increases to "most likely".

The possible ValueError from files.index(oldpath) is no longer caught, since it is an error in the package (or a bug) if oldpath is not is files.

msarahan:
Thanks.  Really nice work here.  Merging.

PR #1612, opened 2017-01-02
WIP: support for build.sh on win

### LINKED ISSUE #1605: packages without "Lib" on windows, but not on Mac or Linux
I have a set-up where I build packages on three different CI servers for different platforms:

* DroneIO and CircleCI builds linux packages
* Travis builds MacOSX packages
* Appveyor builds windows packages

The set up is using Miniconda for each platform and the same meta.yaml is used.

The problem is that the windows packages lack the Lib directory and jusat contain the info directory. The Mac and Linux packages seem ok. The weird thing is that the packages pass all the tests during the build phase. Trying to install them give no error until trying to import the packages.

The packages are [here](https://anaconda.org/BjornFJohansson/pydna/files) . The version 1.0.2 is the last good one and the windows packages lack the Lib dir for 1.1.1, 1.1.4, 1.1.5 and 1.1.6a4. 

The last good one was built on linux and then converted. conda-build 2.0.4 was used for this.

The bad builds were done on:

- conda-build 2.0.10              pydna 1.1.1 [build](https://ci.appveyor.com/project/BjornFJohansson/pydna/build/1.0.412)
- conda-build 2.0.12              pydna 1.1.4 [build](https://ci.appveyor.com/project/BjornFJohansson/pydna/build/1.0.486)
- conda-build 2.0.12              pydna 1.1.5 [build](https://ci.appveyor.com/project/BjornFJohansson/pydna/build/1.0.489)
- conda-build 2.1.0beta1       pydna 1.1.6a4 [build](https://ci.appveyor.com/project/BjornFJohansson/pydna/build/1.0.511)

The last conda-build was pip installed from the conda-build git repo to see if it makes a difference (does not seem like it). The builds are run by this [script ](https://github.com/BjornFJohansson/pydna/blob/py3dev/ci.sh) on all platforms.

I have experienced problems before uisng conda-build, but they did not pass silently.
I can at least not tell from the build outputs outputs (see build links above) that the builds failed 
so catastrophically. Here is my [meta.yaml](https://github.com/BjornFJohansson/pydna/blob/py3dev/meta.yaml)

At this point I do not know what to do except build on linux and then convert. Can someone knowledgeable show me the the right direction.

I would be very grateful for help on this matter!

msarahan:
fixes #1605 

This presently runs build.sh and run_test.sh.  It does so by preferentially running those, and putting in shims to run the older bat files from within a .sh file, calling out to cmd.exe.  This is a philosophical shift to having sh files be the standard everywhere.

This is not done.  We still have gaping questions to answer about how the Visual Studio variable activation should happen.  I think the best way forward is to move all of that activation logic out of conda-build and into conda packages that set the appropriate variables in their activate.d scripts.

CC @mingwandroid

### DIFF
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -31,6 +31,7 @@ requirements:
     - jinja2
     - patchelf      [linux]
     - pkginfo
+    - posix         [win]
     - pycrypto
     - python
     - pyyaml
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -67,9 +67,9 @@ from conda_verify.verify import Verify
 
 
 if 'bsd' in sys.platform:
-    shell_path = '/bin/sh'
+    shell_path = 'sh'
 else:
-    shell_path = '/bin/bash'
+    shell_path = 'bash'
 
 
 def prefix_files(prefix):
@@ -987,8 +987,7 @@ def build(m, config, post=None, need_source_download=True, need_reparse_in_env=F
                 script = '\n'.join(script)
 
         if isdir(src_dir):
-            if utils.on_win:
-                build_file = join(m.path, 'bld.bat')
+            if utils.on_win and os.path.isfile(join(m.path, 'bld.bat')):
                 if script:
                     build_file = join(src_dir, 'bld.bat')
                     with open(build_file, 'w') as bf:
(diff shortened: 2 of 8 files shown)


msarahan:
Punting.  I think the new idea of putting compiler configuration into packages will make this work well, but I don't want this particular issue to hold up the 3.0.0 release.

jakirkham:
Is there a way for the user to disable this if they don't want this to happen? It could make debugging failed Windows builds that much more complicated to debug if this is happening behind the scenes (e.g. not explicitly requested).

> This is not done. We still have gaping questions to answer about how the Visual Studio variable activation should happen. I think the best way forward is to move all of that activation logic out of conda-build and into conda packages that set the appropriate variables in their activate.d scripts.

Could you please expand on this? Are you meaning in `conda` itself or are you meaning recipes take this on?

If the latter, I'm a little nervous about this. It has been a hard sell for people to use things like the `toolchain` package in conda-forge (though some definitely do love it). There are also cases where it does not work as smoothly as we would like. I'm not convinced this is a good idea to try on Windows as well.

cc @patricksnape

patricksnape:
Sounds a bit scary - VS activation is a huge huge pain - one of the great things about conda-build is that it usually 'just works' with VS.

beeankha:

PR #1628, opened 2017-01-05
test against any conda tag

kalefranz:
@msarahan I think this is a better alternative to https://github.com/conda/conda-build/pull/1627


msarahan:
Thanks.  Much nicer.

PR #1635, opened 2017-01-09
add 'aggressive_update' field to index.json

kalefranz:
Once an environment is created, the conda solver optimizes for minimum changes to packages on each successive operation.  @mcg1969 and I have iterated a lot on this behavior, and found the 'minimum change' case to generally be what users expect.  However there are certain packages you want to always be up-to-date, i.e. aggressively updated.  This adds a field `aggressive_update` to index.json that takes a version specification string.  When `aggressive_update` is specified in the recipe meta.yaml, the conda solver will override the minimum change behavior.

This will replace the need for the hard-coded `add_defaults_to_specs()` function in conda.

An example use of this is for python.  The `aggressive_update` field for a `2.7.x` python package should be

    requirements:
      aggressive_update: 2.7.*

In conda, we will also create a new configuration parameter to ignore aggressive update directives.  This will be useful (and perhaps needed) for package builders and recipe maintainers.

CC @mcg1969

### DIFF
--- a/conda_build/metadata.py
+++ b/conda_build/metadata.py
@@ -291,13 +291,13 @@ default_structs = {
     'build/script_env': list,
     'build/features': list,
     'build/track_features': list,
+    'build/aggressive_update': text_type,
     'requirements/build': list,
     'requirements/host': list,
     'requirements/run': list,
     'requirements/conflicts': list,
     'requirements/preferred_env': text_type,
     'requirements/preferred_env_executable_paths': list,
-    'requirements/aggressive_update': text_type,
     'test/requires': list,
     'test/files': list,
     'test/source_files': list,
@@ -387,7 +387,7 @@ FIELDS = {
                'svn_url', 'svn_rev', 'svn_ignore_externals',
                'patches'
                ],
-    'build': ['number', 'string', 'entry_points', 'osx_is_app',
+    'build': ['number', 'string', 'entry_points', 'osx_is_app', 'aggressive_update',
               'features', 'track_features', 'preserve_egg_dir',
               'no_link', 'binary_relocation', 'script', 'noarch', 'noarch_python',
               'has_prefix_files', 'binary_has_prefix_files', 'ignore_prefix_files',
@@ -396,7 +396,7 @@ FIELDS = {
               'pin_depends', 'include_recipe',  # pin_depends is experimental still
               'preferred_env', 'preferred_env_executable_paths',
               ],
-    'requirements': ['build', 'run', 'conflicts', 'aggressive_update'],
+    'requirements': ['build', 'run', 'conflicts', ],
     'app': ['entry', 'icon', 'summary', 'type', 'cli_opts',
             'own_environment'],
     'test': ['requires', 'commands', 'files', 'imports', 'source_files'],
@@ -933,8 +933,8 @@ class MetaData(object):
                 d['noarch'] = build_noarch
         if self.is_app():
             d.update(self.app_meta())
-        if self.get_value('requirements/aggressive_update'):
-            d['aggressive_update'] = self.get_value('requirements/aggressive_update')
+        if self.get_value('build/aggressive_update'):
+            d['aggressive_update'] = self.get_value('build/aggressive_update')
         return d
 
     def has_prefix_files(self):
--- a/conda_build/metadata.py
+++ b/conda_build/metadata.py
@@ -295,6 +295,9 @@ default_structs = {
     'requirements/host': list,
     'requirements/run': list,
     'requirements/conflicts': list,
+    'requirements/preferred_env': text_type,
+    'requirements/preferred_env_executable_paths': list,
+    'requirements/aggressive_update': text_type,
     'test/requires': list,
     'test/files': list,
     'test/source_files': list,
@@ -393,7 +396,7 @@ FIELDS = {
               'pin_depends', 'include_recipe',  # pin_depends is experimental still
               'preferred_env', 'preferred_env_executable_paths',
               ],
-    'requirements': ['build', 'host', 'run', 'conflicts'],
+    'requirements': ['build', 'run', 'conflicts', 'aggressive_update'],
     'app': ['entry', 'icon', 'summary', 'type', 'cli_opts',
             'own_environment'],
     'test': ['requires', 'commands', 'files', 'imports', 'source_files'],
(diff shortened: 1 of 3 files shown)


msarahan:
So what exactly does this ```aggressive_update``` key apply to?  Whatever the parent package is?  But only the matching version of the parent package?  I see the reason for this ```aggressive_update``` option, but don't understand what the version number is for.

mcg1969:
The value of `aggressive_update` specifies how granular the version check should be. We should either assume, or enforce, the requirement that the value matches the current version. So for instance, for a package with an `aggressive_key` value `2.7.*`, if `version` is `2.8.1`, that would be an error.

I'm a little concerned that this is going to be hard to do on a truly generic basis. I'll need to think about what this looks like in the solver.

msarahan:
OK, thanks.  So this particular example would be for a Python recipe?  Or something with version 2.7.*?  The thing I'm confused about is which exact version this applies to. I'm getting more clear that it is the version of the recipe - the thing being built.

And this will free up the solver for the top-level recipe, right?  Not for the dependencies of a recipe?

To generalize it a bit, you might be able to do some jinja2 stuff.  Split the version as set in a jinja2 variable up top, and then join your desired versions, leaving the last one to be *.

Since not everything follows semver, it probably doesn't make sense to have named aliases, like

```aggressive_key: minor```
or
```aggressive_key: bugfix```

or maybe enough packages do use that that it does make sense?

mcg1969:
OK, this is what I think needs to happen here. Let's focus on the three main conda optimization passes:

1. Maximize the versions of the _requested_ packages
2. Minimize the _number_ of other packages (besides those in #1) that need upgrading.
3. Maximize the versions of the aforementioned packages.

What we are talking about here affects #2 and #3. I argue they need to be replaced with:

2. Minimize the number of other packages (besides those in #1) that need adjustment _beyond their `aggressive_upgrade` requirement.
3. Maximize the versions of all packages in #2, subject to the limit imposed there.

It turns out that it is _pass #2_ that needs fixing in this scenario.

Let me explain. Consider a package `foo` with 6 versions: `1.4`, `1.5`, `1.6`, `1.7`, `2.0`, `2.1`. Suppose `1.5` is the installed version. In pass 2, we give `A-1.5` a score of 0, and all other versions a score of 1. This forces the optimizer to try _not_ to upgrade `A` at all.

But now suppose `aggressive_upgrade` has a value of `1.*`. In that case, the solver should give `1.5`, `1.6`, and `1.7` _all_ a score of 0. Only `1.4` and `2.1` should be given a score of 1. Pass 2 is now _less_ aggressive, and it frees Pass 3 to try and upgrade package `A` all the way to 1.7.

As I formulated this, I realized something: what should the "default" value of `aggressive_upgrade` be? I argue that it should be the _build number_. That is, perhaps conda should _always_ try to upgrade to the latest builds of all packages in an environment, unless `aggressive_upgrade` tells it to go further.

Thoughts? Does this make sense?

mcg1969:
Sorry, had to edit my thoughts above, so what you got in your email should be replaced with the above.

mcg1969:
One more thing: if you're wonderinng if this could get us out of the hotfixing business---because it would allow us to bump the build number of a package---I'm afraid that's not correct. Because unless the dependencies are hotfixed, they might _prevent_ a build upgrade.

However, there's one thing we could do to absolutely ensure it: we could impose a rule allows metadata from a given build to supercede _all previous builds_ of a package, for the same version and channel.

Suppose we have A-1.5-0 and A-1.5.1, then we discover a dependency problem. We issue a fix, bump the build number, and post A-1.5-2. With this rule, the new metadata for A-1.5-2 would be applied to _all three buids_. This makes all three builds interchangeable with one another, and conda would then be free to upgrade to the latest build.

No hotfixing required. Something to think about. It might break for other reasons, so we'd need to be careful here. In particular, this would be an issue if the dependencies for py27 differ from py35, so we'd have to determine how to partition those.

mcg1969:
Ok, still one more thing.

This discussion was prompted by a desire to fix the OpenSSL upgrade challenge. If we wanted _simply_ to address that issue, there is a much simpler way to do it. Basically, we add a field to the metadata that tells `conda` _not_ to try and hold that version of the package in place, so that Pass #3 above gets a chance to upgrade it.

A single bit gets us a fix for the OpenSSL issue.

kalefranz:
As in https://github.com/conda/conda-build/pull/1676#discussion_r96671338, I had to move `aggressive_update` in `meta.yaml` to the `build` section.  It really belongs in a `run` section.  But `build` seems to be the only option right now.

kalefranz:
> what should the "default" value of `aggressive_upgrade` be? I argue that it should be the _build number_. That is, perhaps conda should _always_ try to upgrade to the latest builds of all packages in an environment, unless `aggressive_upgrade` tells it to go further.

This discussion is more on conda now, but I agree that I think conda should always bump to the latest build number if the solve finds a newer build number available.  But then, if we're touching the package anyway, why wouldn't we bump to the absolute latest version that is compatible with the environment?

kalefranz:
> Suppose we have A-1.5-0 and A-1.5.1, then we discover a dependency problem. We issue a fix, bump the build number, and post A-1.5-2. With this rule, the new metadata for A-1.5-2 would be applied to all three buids. This makes all three builds interchangeable with one another, and conda would then be free to upgrade to the latest build.

That's really interesting.

kalefranz:
> This discussion was prompted by a desire to fix the OpenSSL upgrade challenge. If we wanted simply to address that issue, there is a much simpler way to do it. Basically, we add a field to the metadata that tells conda not to try and hold that version of the package in place, so that Pass #3 above gets a chance to upgrade it.

Sort of.  Except I think the reason we wanted not to just do `aggressive_upgrade: True` but instead be able to put a matchspec string there was to able to say something like `aggressive_upgrade: 1.0.2*` for openssl.

mcg1969:
> But then, if we're touching the package anyway, why wouldn't we bump to the absolute latest version that is compatible with the environment?

Because that's way too risky. Who's to say a later version won't introduce a breaking change? No, a build bump is the only safe thing we should do by default, if anything at all.

msarahan:
since the build_customization branch is merged, you might also want to consider something like "pinning expressions" - documented at https://github.com/conda/conda-docs/pull/414/files#diff-d1a5dbb2f223de4935d93990676c7074R466

beeankha:

PR #1639, opened 2017-01-10
use conda's cli to create envs, rather than internals

msarahan:


msarahan:
Closing.  This was way too slow, and we're working on a proper public-facing api in the near term.

PR #1641, opened 2017-01-10
utilize chardet and bs4 to guess file encoding

### LINKED ISSUE #1632: Encoding issues on Windows build in the fix_staged_scripts step
Hello.

I'm trying to build the [waf package on windows](https://github.com/conda-forge/waf-feedstock/pull/4), but I'm having some trouble with ```conda_build``` in the ```fix_staged_scripts``` step. I couldn't find any similar issue. I'll follow with some details:

While building, there is a file named ```waf``` that is meant to be moved to the ```%SCRIPTS%``` folder. The problem seems to be that ```conda-build``` tries to read the file to do some checking, but fails to do so because of the file encoding (https://github.com/conda/conda-build/blob/master/conda_build/windows.py#L45)

I believe the waf file should be opened with encoding='iso-8859-1'. Making a local change replacing with the code bellow builds the package successfully, but is not a solution:

```
with open(join(scripts_dir, fn), encoding='iso-8859-1') as f:
```

Any suggestions here? I'm unsure how this case should be handled.
I'll leave some suggestions that occurred me:

1- Fix this at package level, by avoiding running into the ```fix_staged_scripts``` function, somehow. (I don't like this solution because it'll not really fix anything besides this specific package)

2- Do some try...catch with several encodings near the presented code above

3- Have a metadata somewhere (maybe in the recipe?) to handle the file encoding. Fallback to the current case if none.

msarahan:
fixes #1632 

Chardet enhances bs4.  It is not directly imported.

Not sure if there's anywhere else that also needs this encoding guessing.  @jakirkham if you are aware of other recipes that are broken with encoding errors, please let me know.

@tarcisiofischer this works, sort of.  I can build the package, but the tests fail with some weird characters:

```(C:\MC2x64\conda-bld\recipe_1484008724922\_t_env) C:\MC2x64\conda-bld\recipe_1484008724922\test_tmp>waf --help
[91mError: run waf-light from a folder containing waflib[0m
```

Not sure if that's related, or just an artifact.  Sometimes that stuff shows up with bash involved.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,8 @@ install:
         hash -r;
         conda info;
     fi
-  - conda install -q anaconda-client requests=2.11.1 filelock contextlib2 jinja2 patchelf python=$TRAVIS_PYTHON_VERSION pyflakes=1.1 conda-verify
+  - conda install -q anaconda-client requests filelock contextlib2 jinja2 patchelf python=$TRAVIS_PYTHON_VERSION
+  - conda install -q pyflakes conda-verify beautifulsoup4 chardet
   - pip install pkginfo
   - if [[ "$FLAKE8" == "true" ]]; then
       conda install -q flake8;
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -52,8 +52,8 @@ install:
   - python -c "import sys; print(sys.executable)"
   - python -c "import sys; print(sys.prefix)"
   - conda update -q --all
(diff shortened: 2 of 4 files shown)


jakirkham:
The solution looks nice. Thanks for digging into this one.

This is the first case, at least that I can recall, of having encoding issues when building a recipe. Certainly will let you know if others crop up.

I think the weird characters are supposed to specify terminal colors FYI. So don't think encoding is the issue in that case, but something else.

PR #1667, opened 2017-01-17
Conda 4.3 compatibility - fix

msarahan:
fixes #1650 
fixes #1652
fixes #1654

PR #1673, opened 2017-01-18
fix noarch tests and builds

kalefranz:
There are a few logic fixes here.  These were real bugs.  There is also a hook to not create a `info/paths.json` file for the old-style python noarch packages.  They're just not compatible together.


msarahan:
Thanks! This should move us closer to things working. Hopefully later today...

msarahan:
closing - this is merged as part of #1667

PR #1676, opened 2017-01-18
fix preferred_env

kalefranz:


conda_build/metadata.py
```diff
@@ -213,6 +211,8 @@ def parse(data, config, path=None):
     'build/detect_binary_files_with_prefix': bool,
     'build/skip': bool,
     'build/skip_compile_pyc': list,
+    'build/preferred_env': text_type,
+    'build/preferred_env_executable_paths': list,
```
kalefranz: Had to move these flags to the `build` section because they don't work in `requirements`.  I guess `requirements` has extra logic that it applies.

Longer term,  it might be worth considering a "v2 format" where we have a `run` section instead of a `requirements` section.  Then the `build` and `run` sections could have a `requires`, which would then be consistent with the `requires` for the `test` section.

msarahan:
Closing - this is merged as part of #1667

PR #1680, opened 2017-01-19
build: Fix info file prefix replacement on Windows

mingwandroid:
The tar ended up with info files with full-paths because the replace did not happen
due to using the wrong slash.

I have no time to write a test for this. It seems like something that should fail an existing one though?


msarahan:
Good catch.  I'll try to think about a proper test here.

PR #1683, opened 2017-01-20
source: Catch CondaHTTPError so mutliple urls get checked

mingwandroid:
@msarahan, @kalefranz, please review. I'm not sure I need the old:

```
             except RuntimeError as e:
                 print("Error: %s" % str(e).strip(), file=sys.stderr)
```

.. still, but I kept it anyway.

This is needed for R packages where each meta.yaml lists two URLs:

```
  url:
    - http://cran.r-project.org/src/contrib/backports_1.0.4.tar.gz
    - http://cran.r-project.org/src/contrib/Archive/backports/backports_1.0.4.tar.gz
```

### DIFF
--- a/conda_build/source.py
+++ b/conda_build/source.py
@@ -13,7 +13,7 @@ from .conda_interface import download, TemporaryDirectory
 from .conda_interface import hashsum_file
 
 from conda_build.os_utils import external
-from conda_build.conda_interface import url_path
+from conda_build.conda_interface import url_path, CondaHTTPError
 from conda_build.utils import (tar_xf, unzip, safe_print_unicode, copy_into, on_win, ensure_list,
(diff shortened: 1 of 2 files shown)


conda_build/source.py
```diff
@@ -59,6 +59,8 @@ def download_to_cache(metadata, config):
             try:
                 print("Downloading %s" % url)
                 download(url, path)
+            except CondaHTTPError as e:
+                print("Error: %s" % str(e).strip(), file=sys.stderr)
```
msarahan: would like these to be calls to logging rather than print, but I'm going to merge this as-is and refactor it next time I see it.

msarahan:
LGTM.  All failures are known and part of the conda 4.3 stuff.  Merging.

PR #1684, opened 2017-01-20
add test for rm_rf invasively removing relative paths

### LINKED ISSUE #1661: Relative path in source causes rm_rf to remove linked source files
from https://github.com/conda/conda-build/pull/1598#issuecomment-272691858

@msarahan this commit (specifically these lines as found by running git bisect) - causes some odd behaviour - seen here. It seems that if you have a script in your project then the rm_rf call traverses the possibly? sym-linked script file and thus deletes the script in the original repository. This also may only be an issue for recipes built where the source is next to the recipe folder and thus copied using a relative path:

source:
  path: ../

To be clear:

After this commit, after the test phase the bin/lsfm script is deleted from inside the repository and not just from the work/build environment.

msarahan:
closes #1661 

@patricksnape this test is meant to diagnose your reported rm_rf issue.  I did duplicate the behavior once, but in the process of shooting myself in the foot with wrong config arguments (config not passed as a keyword, which is why I added https://github.com/conda/conda-build/compare/master...msarahan:test_rm_rf?expand=1#diff-e2183a8fc8686417bd5c8f72d886e4ffR65), I can no longer repeat the behavior.  Would you please look over this test and ensure that it is repeating enough of the lsfm behavior to be an adequate test?

### DIFF
--- a/conda_build/api.py
+++ b/conda_build/api.py
@@ -62,6 +62,9 @@ def build(recipe_paths_or_metadata, post=None, need_source_download=True,
     from conda_build.conda_interface import string_types
     from conda_build.utils import find_recipe
 
+    assert post in (None, True, False), ("post must be boolean or None.  Remember, you must pass "
+                                         "other arguments (config) by keyword.")
+
     config = get_or_merge_config(config, **kwargs)
 
     recipe_paths_or_metadata = _ensure_list(recipe_paths_or_metadata)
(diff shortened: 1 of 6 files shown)


msarahan:
@patricksnape I'm going to merge this, I need to cut the 2.1.2 release here.  Please let me know if the problem persists, and if so, how I can improve this test.

patricksnape:
Sorry for missing this - I'm finding it difficult to filter all the notifications on Github!

PR #1686, opened 2017-01-20
Fix some small issues with pytest configuration

nicoddemus:
I noticed that the current pytest configuration is emitting some warnings which are easily fixable:

```
WC1 None [pytest] section in setup.cfg files is deprecated, use [tool:pytest] instead.
WI1 /home/travis/miniconda/lib/python3.5/site-packages/pytest_capturelog.py:171 'pytest_runtest_makereport' hook uses deprecated __multicall__ argument
WC1 None pytest_funcarg__caplog: declaring fixtures using "pytest_funcarg__" prefix is deprecated and scheduled to be removed in pytest 4.0.  Please remove the prefix and use the @pytest.fixture decorator instead.
WC1 None pytest_funcarg__capturelog: declaring fixtures using "pytest_funcarg__" prefix is deprecated and scheduled to be removed in pytest 4.0.  Please remove the prefix and use the @pytest.fixture decorator instead.
```

* `[pytest]` should be renamed to `[tool:pytest]` in `setup.cfg` to avoid conflicts with other distutils commands (see pytest-dev/pytest#567 for details);
* The warnings about `pytest_funcarg` functions are due to deprecated hooks used by the [`pytest-capturelog`](https://pypi.python.org/pypi/pytest-capturelog) plugin, which is considered abandoned by the community; this PR replaces it with [`pytest-catchlog`](https://pypi.python.org/pypi/pytest-catchlog) which is a maintained fork and has the exact same functionality.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,7 @@ install:
     else
       conda install -c conda-forge -q perl;
       conda install -q pytest pip pytest-cov numpy mock;
-      $HOME/miniconda/bin/pip install pytest-xdist pytest-capturelog pytest-mock;
+      $HOME/miniconda/bin/pip install pytest-xdist pytest-catchlog pytest-mock;
       pushd .. && git clone https://github.com/conda/conda_build_test_recipe && popd;
     fi
   - pip install --no-deps .
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -75,9 +75,9 @@ test_script:
   - set "PATH=%CONDA_ROOT%;%CONDA_ROOT%\Scripts;%CONDA_ROOT%\Library\bin;%PATH%"
   - set PATH
   - mkdir C:\cbtmp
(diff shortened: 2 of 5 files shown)


msarahan:
Thank you!  I had done this, but then saw some failures either locally or in the tests, I don't remember any more which.  Basically, one or the other needed to be updated - probably locally, since Travis should be getting the current versions each time.  I'll wait on CI, just in case, but this is a great PR - thank you.

nicoddemus:
My pleasure, thanks for the great tool! 👍

nicoddemus:
Noticed that `caplog.text` is now recommended instead of `caplog.text()`; updated the calls and rebased on latest `master`.

msarahan:
That color is really, really nice to have.  I know about the test failures, and will work to fix them in another PR.  Thanks for this!

PR #1687, opened 2017-01-20
Use monkeypatch fixture in a few places of the test suite

nicoddemus:
The [monkeypatch](http://doc.pytest.org/en/latest/monkeypatch.html) fixture automatically reverts environment variables and cwd to their previous values.

I did not run the tests locally and the changes should be harmless, but if something fails I will take a look later. 😁


nicoddemus:
Yep, same cpan failure as in the other builds. ☹️

msarahan:
Thanks.  I'll work on the CPAN failure in another PR.  Merging.

PR #1690, opened 2017-01-20
add retry for conda version error that happens if info/files missing

msarahan:
depends on https://github.com/conda/conda/pull/4388


msarahan:
Merging.  This won't actually work unless conda/conda#4388 is merged, but I want to kick 2.1.2 out.  If that PR or its equivalent doesn't end up merged, I'll change this for the next conda-build release.

Issue #1697, opened 2017-01-26
conda-skeleton cannot jump start packages using `cythonize` in `setup.py`


rainwoodman:
conda-skeleton fails with an ImportError on Cython, if setup.py imports cythonize to build Cython Extensions.

This is because, `get_requirements` will try to run setup.py with an environment that doesn't have cython installed. 

cython is just a commonly used case. The setup.py of my packages needs mpi4py in order to find the mpicc compilers. There shall be a way to notify conda-skeleton about these packages -- as an commandline argument?

msarahan:
@rainwoodman is this fixed by #1698?  Can I close this issue?

rainwoodman:
Yes it is. May want to document this flag somewhere though.

Y

kheuton:
I'm still having a trouble even after [#1698](https://github.com/conda/conda-build/pull/1698), but maybe it's a slightly different issue.

I was able to successfully use the `--extra-specs` flag to use `conda skeleton` to create a recipe for my package that required Cython, but when attempting to `conda build` that recipe, it fails again with `ImportError: No module named Cython.Distutils`.

I guess there's a few options here:

1. Make the `--extra-specs` flag add the extra dependencies to the recipe's build requirements
2. Add an `--extra-specs` flag to `conda build`
3. Edit the recipe by hand

Or is this problem avoidable by some other route I'm not seeing?

rainwoodman:
Did you declare cython in your install_requirements?

kheuton:
What's install_requirements? I'm using `conda skeleton` to build a recipe from someone else's pypi package. I was able to successfully use `conda skeleton` with `--extra-specs cython`, but then `conda build` failed until I went into the `meta.yaml` created by conda-skeleton and added cython to the build requirements

rainwoodman:
I thought a proper setup.py shall declare the packages required for
building in install_requirements argument of the call to setup.

Does the package install from pip when cython is not present?

Yu

On Mar 10, 2017 10:36 AM, "kheuton" <notifications@github.com> wrote:

What's install_requirements? I'm using conda skeleton to build a recipe
from someone else's pypi package. I was able to successfully use conda
skeleton with --extra-specs cython, but then conda build failed until I
went into the meta.yaml created by conda-skeleton and added cython to the
build requirements

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/conda/conda-build/issues/1697#issuecomment-285748996>,
or mute the thread

[...]

PR #1698, opened 2017-01-26
WIP: --extra-specs for conda-skeleton.

rainwoodman:
Proposed solution to Issue #1697.

- the documentation / guides may need changes.
- we may be able to change the current numpy heuristics to the default value of --extra-specs.

### DIFF
--- a/conda_build/skeletons/pypi.py
+++ b/conda_build/skeletons/pypi.py
@@ -302,6 +302,7 @@ def skeletonize(packages, output_dir=".", version=None, recursive=False,
                 all_urls=False, pypi_url='https://pypi.python.org/pypi', noprompt=False,
                 version_compare=False, python_version=default_python, manual_url=False,
                 all_extras=False, noarch_python=False, config=None, setup_options=None,
+                extra_specs=[],
                 pin_numpy=False):
     client = get_xmlrpc_client(pypi_url)
     package_dicts = {}
@@ -404,7 +405,7 @@ def skeletonize(packages, output_dir=".", version=None, recursive=False,
 
         get_package_metadata(package, d, data, output_dir, python_version,
                              all_extras, recursive, created_recipes, noarch_python,
-                             noprompt, packages, config=config, setup_options=setup_options)
+                             noprompt, packages, extra_specs, config=config, setup_options=setup_options)
 
         if d['import_tests'] == '':
             d['import_comment'] = '# '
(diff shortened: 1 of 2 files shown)


msarahan:
I can't remember exactly how this works.  I kind of think that the root environment is the one used to run setup.py right now.  This is not desirable.  It should be a purpose-built environment.  I may be getting mixed up between this skeletonize operation and the reading of setup.py for Jinja2 templating.  Both should be using a build environment, and I think your approach here is a good generalization for numpy.  If there are few enough special cases to handle (numpy, cython, what else?) then maybe it makes sense to add your option, but keep the special cases to make life easier for people.

rainwoodman:
- In current master, The call to `create_env` in `pypi.py` appears to create a purpose-built env, not using `root`.

- I thought the sketonize has to run setup.py in the purpose-built env to extra values for 'template building'. But I never really read the portion of code about templates.

- `numpy, cython, mpi4py`, are those I know of; but we cannot really speak for other users. Potential other cases where setup.py may need other package are code generation with other conda packages, e.g. linking against GPU libraries or using LLVM? 

- One can use a static source code analyze tool to find all Python imports from setup.py and build a list from that -- but then it is going to be difficult to relate the module name to the conda package name. This is why I believe the proposed --extra_specs makes more sense.

- I don't know if this is possible with conda -- pinning against any package? We will run into issues when the binary extensions depend on each other in a binary incompatiable way. But this is probably a separate thread.

msarahan:
The Travis failure is just a race condition failure that I'm working on.  Appveyor is a separate unrelated issue that I think is fixed on master.  No need to rebase.  I'll work it out elsewhere.

Would you mind adding a test for this?  You can probably mostly copy the existing tests in test_api_skeleton.py, but point to recipe that needs cython or mpi4py, and demonstrate both that 1) the recipe fails without specifying your new option and 2) that it works when your option is specified.

Are you willing/able to sign the CLA?  If so, please do so.  Otherwise, please let me know who your employer is, and I'll need to review these changes with the higher ups.  Email me if you'd prefer: msarahan@continuum.io

rainwoodman:
I didn't demontrate 1) (not sure how to capture the exceptions), but I demonstrated 2). I made an attempt at the CLA. My employer is UC Berkeley and I could only find a FAQ page on IP and software which didn't really say much about contribution to other projects.

msarahan:
Thanks.  There were a couple of timeouts in the Travis tests.  No big deal.  Appveyor's in good shape.

For your future reference, you can catch exceptions in pytest like this:

```
def test_failed_tests_exit_build(testing_workdir, test_config):
    """https://github.com/conda/conda-build/issues/1112"""
    with pytest.raises(SystemExit) as exc:
        api.build(os.path.join(metadata_dir, "_test_failed_test_exits"), config=test_config)
        assert 'TESTS FAILED' in exc
```

(from https://github.com/conda/conda-build/blob/master/tests/test_api_build.py#L371-L375)

It is extremely handy for getting coverage of those annoying hard-to-cover bits of exception handling code.

PR #1700, opened 2017-01-26
allow wheels as source format

minrk:
- unzip wheel to enable including files in the wheel, such as `about.license_file`
- copy untouched wheel to allow `pip install --no-deps {{ fn }}` as install command


msarahan:
Thanks.  Test fixes seem unrelated.  Merging.

Issue #1693, opened 2017-01-24
conda build try to remove "dat" folder


hainm:
Hi,

I got very weird deletion when using version `2.1.1`

My root folder is something like

```bash
$ ls .
AmberTools/src/conda-recipe
dat
```

`dat` folder has a file, e.g: `hello.c`.

If I run to conda build at root folder, all of the files in `dat/` folder will be deleted.
```
conda build AmberTools/src/conda-recipe
```

But if I did below, there won't be issue
```bash
cd AmberTools/src
conda build conda-recipe
```

Please see the attached folder (Note: you need to rename `.txt` to `.tar`)
[try.txt](https://github.com/conda/conda-build/files/725895/try.txt)

cheers
Hai

msarahan:
I think this is the same as https://github.com/conda/conda-build/issues/1661, which I hope to address in https://github.com/conda/conda-build/pull/1715

Sorry this took a while to figure out...

PR #1715, opened 2017-01-31
ensure bundle_conda cleanup is acting on absolute paths

### LINKED ISSUE #1661: Relative path in source causes rm_rf to remove linked source files

msarahan:
Do not delete project files - which was possible if file paths were relative
paths.  Whoops.

Closes #1661 #1693 

CC @hainm @patricksnape @jabooth

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -881,6 +881,8 @@ def bundle_conda(output, metadata, config, env, **kw):
     # remove files from build prefix.  This is so that they can be included in other packages.  If
     #     we were to leave them in place, then later scripts meant to also include them may not.
     for f in files:
+        if not os.path.isabs(f):
+            f = os.path.abspath(os.path.normpath(os.path.join(metadata.config.build_prefix, f)))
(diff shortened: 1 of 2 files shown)


jabooth:
@msarahan right, makes total sense! 

Thanks for chasing this down!

msarahan:
Tested locally.  Adapted test to actually be meaningful.  Merging, I want to cut a 2.1.3 release ASAP.

hainm:
thanks.

PR #1721, opened 2017-02-03
test convert's handling of win entry points better.  Fix it.

### LINKED ISSUE #1718: conda-convert does not update entry points in paths.json
After PR #1701, issue #1678 is half-way fixed. However, the bat-file entry points created by linux-to-win conversion are not included in ``paths.json``. For those, all the keys (``"_path"``, ``"sha256"``, ``"path_type"``, ``"size_in_bytes"``, etc.) need to be set in the same way as conda-build does.

This means that packages with entry points converted from linux to windows are still uninstallable with conda 4.3.x. I just confirmed this with the package generated by 

    py.test tests/test_api_convert.py::test_convert_from_unix_to_win_creates_entry_points

(on conda-build commit e92e8fdab8a5bc3b47bd3bd947df8c17faeb8bf9 on linux), installing with conda 4.3.9 on Windows.

msarahan:
Fixes #1718 

also some flake8 in here.

CC @bryevdv @gomyhr


msarahan:
Sorry, just realized this won't quite work yet.  I have to be less lazy and actually compute the file attributes for the bat file, or else conda will probably barf.  I'll test that.

msarahan:
OK, should be working now.  I have tested that this really works by building the recipe for the test entry points on OS X.  I ran conda-convert in windows, targeting windows output.  I installed the package from the converted output (note that dependencies were not installed, because installing paths with conda doesn't install dependencies. =( )  The entry points all work.

@gomyhr, I'd feel even better about this if you tested it and made sure that it works for you.  Do you have time?

conda_build/convert.py
```diff
@@ -32,7 +32,7 @@
 @echo off
 set PYFILE=%~f0
 set PYFILE=%PYFILE:~0,-4%-script.py
-"%~f0\\..\\..\\python.exe" "%PYFILE%" %*
+"%~dp0\..\python.exe" "%PYFILE%" %*
```
msarahan: I'm not sure that the entry point scripts ever worked.  This expression was just wrong.

gomyhr:
@msaharan, I will have time to test - if not on the weekend, then probably on Monday.

gomyhr:
I have tested, and it works - both the test package described above and another package that I tried.

gomyhr:
I should also mention that I tried to review the commits, but I had to give up as most of them included additional changes, and e.g. what commit eb9627004d9e9166b5555340040294f5991d78d does, seems unrelated to its commit message "copy data into -script.py/pyc files"

msarahan:
@gomyhr, thanks for testing. I do sometimes put too much into commits. The one you point out is actually not a case of that, though. Before that commit, what was happening was that the bat file and the py file generated we both just the bat file stub. The actual script was getting lost in the conversion.

conda_build/convert.py
gomyhr: The entry points worked fine before, but the new version is more direct (and maybe more robust). The old one would be 

    C:\path\to\anaconda\Scripts\entrypoint.bat\\..\\..\\python.exe

which works, although somewhat convoluted (and the double slashes are unneccessary). The new one gives

    C:\path\to\anaconda\Scripts\..\python.exe

which is more direct.

Windows is actually very forgiving in evaluating paths with .., as this example shows:

    C:\Users\gom\AppData\Local\Continuum\Miniconda2>this\path\\does\not\\\exist\thisfileneither.exe\..\..\..\..\..\..\python.exe --version
    Python 2.7.12 :: Continuum Analytics, Inc.

msarahan: Wow! Makes sense, I guess. It must just do the parsing of all the .. before anything else. Thanks for teaching me.

PR #1736, opened 2017-02-07
Problem: Croot path cannot be relative

gabm:
Solution: make any path absolute

### DIFF
--- a/conda_build/config.py
+++ b/conda_build/config.py
@@ -82,7 +82,7 @@ class Config(object):
         self._build_id = kwargs.get('build_id', getattr(self, '_build_id', ""))
         croot = kwargs.get('croot')
         if croot:
-            self._croot = croot
+            self._croot = os.path.abspath(croot)
         else:
             # set default value (not actually None)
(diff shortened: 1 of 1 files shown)


msarahan:
I haven't played much with using solely abspath - I usually do a bit more with normpath, like https://github.com/conda/conda-build/blob/9a01114a990d1fa530f82f94191f3860b1823c48/conda_build/utils.py#L248

I don't think that's strictly necessary, but I'd like to hear your opinion on whether it's a good idea.  It keeps paths cleaner as they trickle through conda-build, though I don't think it affects the actual outcome.

gabm:
ok, I'm having a look - maybe that solves a problem that I just observed in windows with spaces in paths... I'll ping you back

gabm:
@msarahan I updated to use normpath.. as I read indeed both is needed. mm

I still have a problem on windows that git does not clone properly when the croot contains spaces - but i think it's unrelated to this PR... so ready to be merged!

msarahan:
Looks like the appveyor failures are there from before.  Merging.  Thanks!

PR #1744, opened 2017-02-14
unify allowed license families with anaconda-verify

### LINKED ISSUE #1649: wrong allowed license families
The license families defined in
https://github.com/conda/conda-build/blob/master/conda_build/license_family.py#L9
are not what we decided (in March 2015) on.  What we decided on is here:
https://github.com/ContinuumIO/anaconda-verify/blob/master/anaconda_verify/const.py#L1

This bug does not affect my work, as I use `anaconda-verify` to check both the recipes and the packages I work on (and in case of ambiguity, I always explicitly specify the license family).  However, it would be nice to fix this.

msarahan:
fixes #1649 

CC @ilanschnell

### DIFF
--- a/conda_build/license_family.py
+++ b/conda_build/license_family.py
@@ -1,6 +1,5 @@
 from __future__ import absolute_import, division, print_function
 
-from difflib import get_close_matches
 import re
 import string
 from conda_build import exceptions
@@ -14,11 +13,11 @@ GPL2
 GPL
 BSD
 MIT
-APACHE
+Apache
 PSF
-PUBLICDOMAIN
-PROPRIETARY
-OTHER
+Public-Domain
+Proprietary
+Other
 NONE
 """.split()
 
@@ -93,7 +92,7 @@ def guess_license_family(license_name=None,
     for family in recognized:
         if license_name in family:
             return family
-    return 'OTHER'
+    return 'Other'
 
 
 def ensure_valid_license_family(meta):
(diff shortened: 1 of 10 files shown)


ilanschnell:
+1

conda_build/license_family.py
```diff
@@ -14,11 +13,11 @@
 GPL
 BSD
 MIT
-APACHE
+Apache
 PSF
-PUBLICDOMAIN
-PROPRIETARY
-OTHER
+Public-Domain
```
jakirkham: Wouldn't this break anyone that was using `PUBLICDOMAIN`? Also do the case changes result in any problems?

mstimberg: Actually, this (and `'Other'` --> `'OTHER'`, etc.) break *all* recipes using these license families... The check tests
```Python
if remove_special_characters(normalize(license_family)) not in allowed_license_families
```
The function `normalize` changes the license family specified in the recipe to use "ALL CAPS", which therefore will *never* match `Public-Domain`, `Other`, `Apache`, or `Proprietary`.

msarahan: Thanks for your concerns @mstimberg and @jakirkham.  I will fix the normalize function. Unfortunately, the breakage with public domain is unavoidable. Internal recipes were broken. It's my fault for not reviewing the original license guessing PR more carefully.

I generally advise you to not set license family manually.  If you have strong reason to, we should talk more.

mstimberg: > I generally advise you to not set license family manually. If you have strong reason to, we should talk more.

Ah, thanks, I did not know that, I thought it was recommended to specify it (maybe because it is included in conda-forge's example `meta.yaml`). In my case, I was setting it to `Other` which `guess_license_family` will do automatically, anyway.

jakirkham: > I generally advise you to not set license family manually. If you have strong reason to, we should talk more.

We started doing it at conda-forge after multiple requests to do so. Are you saying we should stop doing this now? TBH I don't care either way. That's just going to be a lot of recipes to fix and people's behavior to correct.

jakirkham: Also another point on the `license_family`. Currently `conda skeleton pypi` seems to be filling out this information. If people shouldn't be setting it, maybe that tool should stop adding it.

Though better yet would be to get `conda-build` to issue some sort of warning that the license family field is deprecated and subject to removal.

msarahan: Both good ideas. I'll make a new issue for it.

jakirkham: xref: https://github.com/conda/conda-build/issues/1780

Issue #1646, opened 2017-01-10
Version 2.1 breaks git environment variables


Zaharid:
The following recipe:
https://github.com/NNPDF/reportengine/tree/master/conda-recipe

works with conda-build 2.0 when I type:

```
conda build conda-recipe
```

from the root folder of the project.

Instead with 2.1 the environment variables `GIT_DESCRIBE_TAG` and `GIT_DESCRIBE_NUMBER` are undefined and what I see is:

```
reportengine (master *%)$ conda build conda-recipe
Cloning into bare repository '/home/zah/anaconda3/conda-bld/git_cache/home/zah/phd/reportengine'...
done.
Cloning into '/home/zah/anaconda3/conda-bld/conda-recipe_1484075474810/work'...
done.
checkout: 'HEAD'
Your branch is up-to-date with 'origin/master'.
==> git log -n1 <==

commit bb51ceb565198c1859d0bdd38fc8f753d089c508
Author: Zahari <zaharid@gmail.com>
Date:   Mon Jan 2 15:16:02 2017 +0100

    Refactor templateparser
    
    Move parsing to individual function.


==> git describe --tags --dirty <==

v0.9-14-gbb51ceb


==> git status <==

On branch master
Your branch is up-to-date with 'origin/master'.
nothing to commit, working directory clean


Traceback (most recent call last):
  File "/home/zah/anaconda3/bin/conda-build", line 6, in <module>
    sys.exit(conda_build.cli.main_build.main())
  File "/home/zah/anaconda3/lib/python3.5/site-packages/conda_build/cli/main_build.py", line 312, in main
    execute(sys.argv[1:])
  File "/home/zah/anaconda3/lib/python3.5/site-packages/conda_build/cli/main_build.py", line 304, in execute
    already_built=None, config=config, noverify=args.no_verify)
  File "/home/zah/anaconda3/lib/python3.5/site-packages/conda_build/api.py", line 86, in build
    need_source_download=need_source_download, config=config)
  File "/home/zah/anaconda3/lib/python3.5/site-packages/conda_build/build.py", line 1413, in build_tree
    config=config)
  File "/home/zah/anaconda3/lib/python3.5/site-packages/conda_build/render.py", line 157, in render_recipe
    config=config)
  File "/home/zah/anaconda3/lib/python3.5/site-packages/conda_build/render.py", line 90, in parse_or_try_download
    metadata.parse_again(config=config, permit_undefined_jinja=False)
[...]

PR #1752, opened 2017-02-19
Conda cookiecutter for new python projects

postelrich:
This PR implements a new cookiecutter feature conda.  It add a single command `conda cookie cut PROJECT_NAME PROJECT_PATH` to setup a new directory with all the things needed for a python project.  This includes the directory structure, conda recipe, setup.py, setup.cfg, travis.yml, git, GitHub repo creation, versioneer, coverage and more. 

This is a rough POC though there really isn't much to it so opening the PR for discussion and feedback.  

I would still like to add in a simple test, initial tag, setting up travisci and clean up the options and output.  Still needs tests.


conda_build/templates/cookie/meta.yaml
```diff
@@ -0,0 +1,28 @@
+package:
+  name: {{ name }}
+{% raw %}
+  version: {{ environ.get('GIT_DESCRIBE_TAG', '') }}
```
msarahan: Please do not use this style of Jinja2 variables. Setting defaults kind of breaks the iterative parsing.  See https://github.com/conda/conda-build/issues/1646 for more info.

conda_build/cookie.py
```diff
@@ -0,0 +1,157 @@
...
+    r = subprocess.Popen(['git', 'config', '--get', 'user.{}'.format(field)], stdout=subprocess.PIPE).communicate()[0]
+    if r:
+        return r.decode('utf-8').strip()
+    else:
+        r = input("Enter {}: ".format(field)) or ""
+        return r.strip()
+
+
+def run_cmd(cmd):
+    """Run shell command and check if successful."""
```
msarahan: Please use the utils.cmd_in_env function rather than this new wrapper.

conda_build/api.py
```diff
@@ -281,3 +282,13 @@ def update_index(dir_paths, config=None, force=False, check_md5=False, remove=Fa
 
     for path in dir_paths:
         update_index(path, config, force=force, check_md5=check_md5, remove=remove)
+
+
+def cut_project(name, path, conf=None):
+    cookie = Cookie(name, path, conf=conf)
+    create_project_skeleton(cookie)
+    cookie.initial_commit()
+    python_ver = input("Enter python version [2.7]: ") or '2.7'
```
msarahan: Please never, ever require interactive input. Make it a CLI argument and pass it as a keyword arg here.

conda_build/cli/main_cookie.py
```diff
@@ -0,0 +1,34 @@
+import argparse
+import sys
+import conda_build.api as api
+from attrdict import AttrDict
+from conda_build.config import Config
+
+
+def parse_args(args):
+    p = argparse.ArgumentParser()
```
msarahan: Conda build's commands all extend conda. This is good and bad. It makes the system nice and uniform, but it is also very slow. Until we do something else, we should stick to the way things are done now.  However, I'm not at all clear on why this whole PR should be a new conda build feature and not just another cookiecutter recipe that we can point to.

msarahan:
@postelrich how do you want to proceed here?  I'd like to discuss this more.  The concept is very valuable, I'm just not sure whether it should belong in conda-build, or as a cookiecutter template repository.

postelrich:
@msarahan I would proceed by implementing your suggestions and expanding on the features.  As for the proper place this should go, my thought process for adding it to conda-build was that it would be one less package a user has to install but it certainly could be it's own `conda/cookie` repo. What are your thoughts?

Aside from adding additional project initialization features, I also wonder what could be done in terms of project management, if there is any utility to wrapping some common tasks.

msarahan:
I think it's feature creep for it to be in conda-build.  I agree that it would be convenient to dodge the extra package install, but it adds much more code to conda-build, code that is effectively duplicated from the cookiecutter project.  I recommend making a repo that cookiecutter can point to, and then conda-build should document and advertise its availability.

postelrich:
@msarahan pulled it out into its own repo. https://github.com/postelrich/conda-cookie

msarahan:
Dang, I wish I had been in communication with you better.  Our summer intern also pulled it out here:  https://github.com/conda/cookiecutter-conda-python/pull/1

Would you like to add to that PR at all?  Would you prefer to replace that PR?  Let me know.  Either way, I'm happy to host it on the conda repo so that you don't have to maintain it if you don't want to.

postelrich:
@msarahan Can we do a conda/cookie repo?

msarahan:
We can, but I'm not sure we should.  My line of thought in naming cookiecutter-conda-python was based on looking at https://github.com/audreyr/cookiecutter#a-pantry-full-of-cookiecutters - they are divided by language.  I thought we would eventually have things like:

* cookiecutter-conda-python
* cookiecutter-conda-r
* cookiecutter-conda-lua
* cookiecutter-conda-perl

Do you have thoughts on that?

postelrich:
Oh I wasn't even making this as a cookiecutter template, didn't even think about that.  This goes farther by setting up git, conda, github, travis, and more.

Issue #1723, opened 2017-02-03
Additional Architectures in Skip clause


gabm:
We have some custom builds of conda running that have additional architectures defined in conda. That works fine so far. Now we stumbled across a problem, that we want to skip some of the builds for some platforms. Namely we want to do this:

```
build:
  skip: True #[aarch64]
```

That works very well on our custom/patched aarch64 build as this architecture is added to `non_x86_linux_machines` and therefore known. However, it interferes with the other conda distributions on windows/linux that are supported by continuum itself. These platforms don't know about aarch64 and therefore produce a crash in `metadata.py:114`. 

As we don't really want to go into the process of shipping patched versions of these platforms as well, maybe we could provide a PR to solve this? I'd suggest to evaluate unknown selectors to `False` and letting the user know that there was an unknown selector. Additionally we could add a command-line flag (`--ignore-unknown-selectors`) to switch on the new behavior to not break the API..

@msarahan What do you think?

msarahan:
PR is welcome.  The selectors are very hard-coded and not terribly friendly.  I would really like any work that builds towards making them more flexible - especially if you can use Jinja2 expressions to dynamically code selectors.  Not sure if that's already possible - might be, but might not be due to order of operations (templating before or after selectors applied?)

gabm:
I'd almost say that templating is done before selectors, but I'll check that again and confirm.

Do you think we'd have to keep APIs stable? in this matter. I mean by default? Evaluating unknown selectors to `False` will change the current behavior. It will lead to side effects like typos (e.g. `win3` instead of `win32`} being evaluated to `False` instead of throwing an error...

msarahan:
@gabm is this resolved by #1753?  Can I close this issue?

gabm:
yes.. thats solved.. works as expected!

PR #1753, opened 2017-02-19
Problem: Unknown selectors crash conda-build

gabm:
Solution: evaluate them to false...

this is related to #1723 

It allows code like this, even if `linux-aarch64` is not known at the time of writing

```
build:
  skip: True #[linux-aarch64]
```

### DIFF
--- a/conda_build/metadata.py
+++ b/conda_build/metadata.py
@@ -106,6 +106,7 @@ def eval_selector(selector_string, namespace):
         return eval(selector_string, namespace, {})
     except NameError as e:
         missing_var = parseNameNotFound(e)
+        print("Warning: Treating unknown selector \'" + missing_var + "\' as if it was False.")
         next_string = selector_string.replace(missing_var, "False")
         return eval_selector(next_string, namespace)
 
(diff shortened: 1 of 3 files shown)


conda_build/metadata.py
```diff
@@ -88,6 +88,26 @@ def ns_cfg(config):
...
+# If we encounter a NameError (unknown variable in selector), then we replace it by False and re-run the evaluation
+def eval_selector(selector_string, namespace):
+    try:
+        # TODO: is there a way to do this without eval?  Eval allows arbitrary
+        #    code execution.
+        return eval(selector_string, namespace, {})
+    except NameError as e:
+        missing_var = parseNameNotFound(e)
+        next_string = selector_string.replace(missing_var, "False")
+        return eval_selector(next_string, namespace)
```
msarahan: Please show a warning log message at least, so misspellings are not just always silent failures.

gabm:
I added the warning - but the selectors get evaluated 3 times and so the output appears 3 times...

msarahan:
Thanks.  I'm not too worried about overly noisy warnings.  I'll work on fixing the logger configuration to avoid duplicate messages in the longer term, but this is ready to go right now.

PR #1767, opened 2017-02-24
add preferred_env to index.json

kalefranz:
The `preferred_env` field needs to be part of index.json for conda to properly plan installations.  That it hasn't been was an oversight that I just discovered as I'm now going back and writing extra integration tests for this feature.


msarahan:
Thanks

Issue #149, opened 2014-07-25
Issues building Erlang


tswicegood:
Seeing this when I try to build Erlang 17.1:

```
$ conda -V && conda build -V && conda build .
conda 3.5.5-99-gf8a075f
conda-build 1.5.0-41-g3d9418c
BUILD START: erlang-17.1-0
Fetching package metadata: .........
Solving package specifications: .
The following NEW packages will be INSTALLED:

    openssl:  1.0.1h-0
    readline: 6.2-2

Linking packages ...
[      COMPLETE      ] |################################################################################################################################| 100%
Source cache directory is: /Users/travis/anaconda/conda-bld/src_cache
Found source in cache: otp_src_17.1.tar.gz
An unexpected error has occurred, please consider sending the
following traceback to the conda GitHub issue tracker at:

    https://github.com/conda/conda/issues

Include the output of the command 'conda info' in your report.


Traceback (most recent call last):
  File "/Users/travis/anaconda/bin/conda-build", line 7, in <module>
    execfile(__file__)
  File "/Users/travis/Dropbox/work/continuum/conda-build/bin/conda-build", line 5, in <module>
    sys.exit(main())
  File "/Users/travis/Dropbox/work/continuum/conda-build/conda_build/main_build.py", line 109, in main
    args_func(args, p)
  File "/Users/travis/Dropbox/work/continuum/conda/conda/cli/main.py", line 182, in args_func
    args.func(args, p)
  File "/Users/travis/Dropbox/work/continuum/conda-build/conda_build/main_build.py", line 268, in execute
    build.build(m, verbose=not args.quiet, post=post)
  File "/Users/travis/Dropbox/work/continuum/conda-build/conda_build/build.py", line 274, in build
    source.provide(m.path, m.get_section('source'))
  File "/Users/travis/Dropbox/work/continuum/conda-build/conda_build/source.py", line 219, in provide
    unpack(meta)
  File "/Users/travis/Dropbox/work/continuum/conda-build/conda_build/source.py", line 68, in unpack
    tar_xf(src_path, WORK_DIR)
  File "/Users/travis/Dropbox/work/continuum/conda-build/conda_build/utils.py", line 45, in tar_xf
    t.extractall(path=dir_path)
  File "/Users/travis/anaconda/lib/python2.7/tarfile.py", line 2051, in extractall
    self.extract(tarinfo, path)
  File "/Users/travis/anaconda/lib/python2.7/tarfile.py", line 2088, in extract
[...]

PR #1779, opened 2017-03-01
fix error unpacking tarball containing unicode filenames

stuertz:
The problem may occur only on windows and only with py2.


msarahan:
LGTM.  Is there a test for a specific package that we can add to our test suite, so that we can make sure it's behaving like you want?

stuertz:
I had the issue with the currently uploaded pyslet package for conda-forge: 
https://github.com/conda-forge/staged-recipes/pull/2534

But it may fail with any package containing non ascii filenames.
In that case the filename was: "英国.xml" which is chinese for united kingdom.

stuertz:
I think this fixes #149, which can be closed then.

PR #1782, opened 2017-03-01
allow un-finalizable recipes (things that are not installable)

msarahan:
This is especially necessary for concourse's test modules, where they are just never meant to actually install/build packages.  The net result of this PR is that with api.render, you can now allow it to return un-finalized metadata.  For building, metadata will still be finalized, or will raise errors when it is not able to be finalized.


msarahan:
Merging.  CI failures are due to conda API breakage and a fluke on appveyor.

PR #1807, opened 2017-03-06
build: toposort split-packages, manage intradependencies in build_prefix

mingwandroid:
We fake install and uninstall packages so that only those that should
be present are present.

This should not change behaviour for packages without intradependencies.

When I say fake install and uninstall I mean I do actually perform the file system actions, just not using conda (which wouldn't work since they are not installed packages at this stage).


tests/test_subpackages.py
```diff
@@ -77,3 +78,12 @@ def test_hash_includes_recipe_files(testing_workdir, testing_config):
     a given output."""
     recipe = os.path.join(subpackage_dir, 'script_install_files')
     outputs = api.build(recipe, config=testing_config)
+
+
+@pytest.mark.skipif(sys.platform == "win32",
```
msarahan: I think you can allow it on win32 if you put in posix as a build-time requirement.

tests/test-recipes/split-packages/_intradependencies/meta.yaml
```diff
@@ -0,0 +1,41 @@
...
+      - r-base {{ r }}
+    script: install-py1-intradependencies.sh
+
+  - name: r1-intradependencies
+    requirements:
+      - {{ pin_subpackage('lib1intradependencies', exact=True) }}
+      - python {{ python }}
+    script: install-r1-intradependencies.sh
+
+  - name: lib2intradependencies
```
msarahan: it's not clear to me what the relationship is between lib1 and lib2 here.  Can you please add a comment?  I think you're just making sure that lib1 stuff doesn't show up in lib2, right?

mingwandroid: I have the following layout:
```
lib1intradependencies
┃  ┃
┃  ┗ py1-intradependencies
┃
┗ r1-intradependencies

lib2intradependencies
┃  ┃
┃  ┗ py2-intradependencies
┃
┗ r2-intradependencies
```

So there is no relationship, and I ensure that files from one of these islands are never present at during the installation of a package from the other island.

msarahan:
superseded by #1823 - closing

PR #1831, opened 2017-03-11
ENH: add extra and recipe-maintainers to metadata fields

CJ-Wright:


msarahan:
Sorry @CJ-Wright this particular field is part of conda-forge, and is a superset of conda-build's meta.yaml spec.  I don't think adding it here is correct.  Instead, I'd propose adding a new rule to conda-verify that can optionally be enabled.

CJ-Wright:
@msarahan That's ok I will try to put it there.

PR #1834, opened 2017-03-13
fix #1825 monkey patch ensure_use_local

### LINKED ISSUE #1825: conda install --yes flag ignored when --use-local set with conda-build 2.1.6
Trying to execute `conda install --yes --use-local {package}` from a package built with `conda-build` ignores the `--yes` flag and waits for user input. This [has](https://travis-ci.org/choderalab/openmmtools/jobs/209789665#L951) put a [halt](https://travis-ci.org/choderalab/pymbar/jobs/209453492#L646) to trying to do automatic builds as they time out, I am only using the 2 links as examples.

I can replicate the problem on my local machine and on downgrading to `conda-build` 2.1.5 the problem goes away, even without re-building the package.

kalefranz:
This is what I'd recommend, given how we also neutered the same function in the conda code via https://github.com/conda/conda/pull/4867.

### DIFF
--- a/conda_build/conda_interface.py
+++ b/conda_build/conda_interface.py
@@ -9,6 +9,17 @@ from enum import Enum
 
 from conda import __version__ as CONDA_VERSION
 
+
+try:
+    # This monkey patch is addressed at #1825. The ensure_use_local is an outdated vestige
+    #   and no longer has any relevant effect.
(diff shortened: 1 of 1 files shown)


msarahan:
Thanks for this.

PR #1837, opened 2017-03-14
Pass WINDIR env var through on win

mcs07:
This issue was encountered when building a recipe for cx_Oracle. It was failing on the test stage because `import cx_Oracle` results in setuptools `pkg_resources.resource_filename` being called, which in turn has a call to `os.environ['windir']` on Windows.

I guess that the `WINDIR` environment variable is widely assumed to always be set on Windows, like `SYSTEMROOT`.


msarahan:
Yikes, changing base branch to 2.1.x was bad - I'll cherry-pick this back to the 2.1.x branch.  Thanks.

msarahan:
Do you mind signing our CLA?  I can't merge your PR without it, unfortunately.

mcs07:
Signed

msarahan:
Great, thank you!  Merging.

PR #1855, opened 2017-03-17
Set permissions when extracting from a zipfile

### LINKED ISSUE #778: Permissions not set when extracting from zipfile.
When extracting source code from a `zipfile`, permissions are lost, and for example:

```
./configure
```

will fail, because the file is no more executable.

This was solved in [Distribute](https://pythonhosted.org/distribute/) (deprecated fork of the Setuptools project) with this [pull request](https://bitbucket.org/tarek/distribute/pull-requests/14/set-permissions-when-extracting-from/diff).

Currently, [Setuptools](http://pythonhosted.org/setuptools) has a [unpack_zip](https://bitbucket.org/pypa/setuptools/src/0d80c1398b1901def27014ec91230103d0830188/setuptools/archive_util.py?at=default&fileviewer=file-view-default#archive_util.py-88:122) function that manage permissions, while the `conda-build` [unzip](https://github.com/conda/conda-build/blob/7a6a0d3a791acd94c680769b6e470add7db5a462/conda_build/utils.py#L86-L97) function do not.

We may update the `conda-build` `unzip` function, or maybe remove it and use `archive_util.py` module of`Setuptools` instead?

jjhelmus:
closes #778

### DIFF
--- a/conda_build/utils.py
+++ b/conda_build/utils.py
@@ -345,7 +345,8 @@ unxz is required to unarchive .xz source files.
 
 def unzip(zip_path, dir_path):
     z = zipfile.ZipFile(zip_path)
-    for name in z.namelist():
+    for info in z.infolist():
+        name = info.filename
         if name.endswith('/'):
             continue
         path = join(dir_path, *name.split('/'))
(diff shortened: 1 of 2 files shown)


msarahan:
Nice, thanks.  I will cherry-pick this back onto 2.1.x.

jjhelmus:
@msarahan This new test should be skipped on Windows as `os.chmod` [does not effect the execution bit of the file.](https://docs.python.org/3/library/os.html#os.chmod).

PR #1859, opened 2017-03-20
Add more important Windows environment variables

### LINKED ISSUE #1857: Incomplete list of essential Windows environment variables
Hi, after one of my packages failed to build with conda-build 2.1.7 due to a missing environment variable (`USERNAME`, used by [`getpass.getuser`](https://github.com/python/cpython/blob/master/Lib/getpass.py#L162)), I had a look into `conda_build/environ.py` and noticed that `windows_vars(prefix)` is missing some important environment variables.

I will prepare a pull request which adds the [default Windows environment variables](https://en.wikipedia.org/wiki/Environment_variable#Default_values).

wulmer:
Fixes #1857 by adding essential* Windows environment variables.

(*) see https://en.wikipedia.org/wiki/Environment_variable#Default_values


Korijn:
This is HUGELY important. For example, our build fails, because the test command runs some matplotlib code, which calls `getpass.getuser()`, which fails because the `USERNAME` environment variable is not available.

msarahan:
Thanks. I will cherry pick this onto 2.1.x

Korijn:
This build actually doesn't fix the problem, as the values are already empty at the point this code is executed...

PR #1861, opened 2017-03-21
Python 2 os.environ Windows compat. fix

mingwandroid:
The current release doesn't work on Python 2.7 with conda >= 4.2 without this unfortunately:

```TypeError: environment can only contain strings```


msarahan:
Thanks Ray.

jakirkham:
Has this already been ported to `master`?

PR #1867, opened 2017-03-23
Use symlinks while copying from file sources to improve perf.

stuertz:
I wondered why using file sources, no symlinks were used. Seems to be forgotten.

I tried running the very long running mentioned tests, but they fail before and after my change :-( 

Running my local build worked fine for me with this change.


msarahan:
Sorry, master is in heavy construction.  Should be more stable soon.  Do you mind rebasing your work on the 2.1.x branch for now?  I will get it in more quickly that way, and the tests there should all be passing.

msarahan:
I don't know what's up with the CI's.  I'm just going to merge this and see how it does.  Thanks @stuertz

PR #1871, opened 2017-03-24
Improved cran skeleton

johanneskoester:
This PR further improves the cran skeleton.

* sha256 checksum is calculated automatically (unfortunately cran does not provide any checksums we could use directly)
* package name is no longer defined as jinja variable. This should improve readability.


msarahan:
@mingwandroid can you clarify what part(s) exactly on this PR make your job harder?  Are there any alterations that could be made to this PR to keep your life easy?

mingwandroid:
Since I do a diff on the generated recipes, any change makes that harder, but merge anyway, I'll just revert locally.

msarahan:
I'd really rather not make you maintain a fork.  Sorry @johanneskoester, I can't merge this, because it would mess up Ray's workflow too much.  Thanks for submitting, anyway.

johanneskoester:
So then conda forge and bioconda has to maintain a fork. Maybe there is a compromise? Is it only that you don't like the change? As far as I know, this PR even removes one Jinja variable, so that it becomes closer to the original format. Also, this adds a checksum, which should be beneficial for you as well.

mingwandroid:
I don't mind if we merge this, overall, I was making a more general point that changing skeletons does have hidden costs, but the changes here are good as you say. Thanks, and apologies for causing a fuss this time instead of last time :-)

msarahan:
@johanneskoester please rebase onto 2.1.x if you want this in the next 2.1 branch release.  There may not be any further 2.1.x releases after this next one.

johanneskoester:
Thanks @mingwandroid! And sorry for not really asking for your opinion last time.
@msarahan ok I'll do that.

johanneskoester:
Closing in favor of #1879.

PR #1878, opened 2017-03-24
read metadata / options from setup.cfg in load_setup_py_data

ma-ba:
`setuptools 30.3.0` added the ability to read metadata / options, i.e. keyword arguments for `setuptools.setup`, from configuration files. See https://setuptools.readthedocs.io/en/latest/setuptools.html#configuring-setup-using-setup-cfg-files.
This PR makes this functionality available in `conda_build.jinja_context.load_setup_py_data`.

I assumed that _setup.cfg_ resides next to _setup.py_ in the same directory. According to the doc-string of `distutils.dist.Distribution.find_config_files` (which is the function that provides the configuration filenames for `setup`) _setup.cfg_ is to be found "in the current directory". Thus maybe L142 should simply be `setup_cfg = 'setup.cfg'`. I am not sure whether the current working directory in `load_setup_py_data` would be the one `setuptools` / `distutils` assumes.
That aside, the reason for hard-coding the filename `'setup.cfg'` instead of making it an additional parameter to the function is that `find_config_files` uses this same hard-coded name.

Notice that the corresponding test is skipped for older versions of `setuptools`, e.g. the one available at https://repo.continuum.io/pkgs/free/. Conda-forge provides a more recent version of `setuptools`, though.

### DIFF
--- a/conda_build/jinja_context.py
+++ b/conda_build/jinja_context.py
@@ -134,6 +134,17 @@ def load_setup_py_data(config, setup_file='setup.py', from_recipe_dir=False, rec
         '__file__': setup_file,
     }
     if os.path.isfile(setup_file):
+        try:
+            from setuptools.config import read_configuration
+        except ImportError:
+            pass  # setuptools <30.3.0 cannot read metadata / options from 'setup.cfg'
+        else:
+            setup_cfg = os.path.join(os.path.dirname(setup_file), 'setup.cfg')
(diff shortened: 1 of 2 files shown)


ma-ba:
Test failures seem unrelated, though the added test was always skipped since `setuptools 27.2.0` was used.

msarahan:
Don't worry about the failures. I'll sort them out. Thanks for the PR.

PR #1886, opened 2017-03-28
TST: Add skip to test_unzip unit test

jjhelmus:
test_unzip should only be run on Unix system where os.chmod works on all file permissions.


msarahan:
Thanks!

PR #1888, opened 2017-03-29
avoid un-necessary rehashing in some cases

groutr:
Small restructuring of a loop to avoid a potentially costly second call to hashsum_file.

@msarahan, is there a better error to be raising here than RuntimeError?  What is exceptions.VerifyError intended for?


msarahan:
great, thanks!

I think the VerifyError is for when things fail the Conda-Verify checks.  I'd be happy to add a specific exception type for this - it would be better to throw something more specific than a RuntimeError.

PR #1899, opened 2017-04-04
check that files are executable in finding executables.

tatome:
`conda build` got hung up on a file called `git` in my `$PATH` that was not executable (by me).  This fixes it.

### DIFF
--- a/conda_build/os_utils/external.py
+++ b/conda_build/os_utils/external.py
@@ -36,5 +36,6 @@ def find_executable(executable, prefix=None):
                     return path
         else:
             path = join(dir_path, executable)
-            if isfile(expanduser(path)):
+            exp_path = expanduser(path)
+            if isfile(exp_path) and os.access(exp_path, os.X_OK):
                 return expanduser(path)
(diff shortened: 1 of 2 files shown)


msarahan:
Cool, thanks.  Do you know how this behaves on Windows?  Is this worth adding a test for, so that future work doesn't break it?

tatome:
It probably makes as much sense on Windows as on *nix to check whether a file with the correct name is actually executable, but I don't have access to a Windows machine, so I can't test that.

That's a good point about the regression test, though.  I'll see what I can do.  Is it OK if I make the test conditional on the OS?

tatome:
OK, I wrote a test for find_executable.

Since I can't test on Windows, I'm not changing the code for win32 in find_executable and the test does nothing if on that architecture.  Someone else with access to a Windows machine could fix find_executable and enable the test for win32.

msarahan:
Great, thanks!  I'll test on windows when I can.  Nice test!

The Travis failure looks like a fluke, so I'm going to merge this.

PR #1908, opened 2017-04-05
Create entry points scripts using pip template

jjhelmus:
When creating entry_points for entries in the build/entry_points section use
the same template as pip uses when creating entry points for wheels.


msarahan:
Travis failure is just a crashed worker.  This is good to go.  Thanks.

PR #1924, opened 2017-04-13
fix empty create_env

kalefranz:
a valid conda environment, at a minimum, contains a conda-meta
directory and conda-meta/history file


msarahan:
Cool, thanks.

PR #1927, opened 2017-04-14
fix for conda's new MatchSpec implementation

kalefranz:
This fixes a changed exception type associated with conda/conda#4158


kalefranz:
The failing test was `test_pip_in_meta_yaml_fail`.

Not exactly sure what the error was before, but the current error is now

    CondaValueError: Value error: Cannot match on field pip

msarahan:
Thanks!

PR #1930, opened 2017-04-17
don't set language env vars when lang package not installed

msarahan:
This removes a problem where, for example, the PERL env var is set to a path, even when that path does not exist.


conda_build/environ.py
```diff
@@ -257,39 +258,58 @@ def conda_build_vars(prefix, config):
...
 def python_vars(config, prefix):
-    d = {
-        'PYTHON': config.python_bin(prefix),
-        'PY3K': str(config.PY3K),
-        'STDLIB_DIR': utils.get_stdlib_dir(prefix),
-        'SP_DIR': utils.get_site_packages(prefix),
-        'PY_VER': get_py_ver(config),
-        'CONDA_PY': str(config.CONDA_PY),
-    }
+    vars = {}
```
goanpeca: It may be good to not use a python reserved global method for the variable name?

`env_vars` maybe?

PR #1932, opened 2017-04-17
Language version vars 3

mingwandroid:
Cherry picked from 2 branch.


mingwandroid:
Hey @msarahan, the failures here on Windows are the env. var case insensitivity thing ('perl') and on Linux, some sort of bad cached json files so I think this is ready to be merged now.

PR #1941, opened 2017-04-20
fix preferred_env in index.json in 2.1.x

kalefranz:
This patch is in conda-build 3, but I guess it didn't make it into the 2.1.x branch.

I'm running into some problems right now building conda with conda-build 3, so really need this in conda-build 2 to get the alpha release out today.  (Also looking into why conda won't build with conda-build 3.)


msarahan:
CI failures appear to be unrelated.  Merging.

PR #1931, opened 2017-04-17
Correct rendering with --dirty flag if recipe name appears as a substring of another's name

rendinam:
Delimit package_name to prevent list from being loaded with other build folder names that contain the package name as a substring.

This allows recipes that call for the population of git-related environment variables to be rendered correctly when the `--dirty` flag is used.

PR #1953, opened 2017-04-24
Match package_name plus delimiter exactly to avoid spurious matches...

rendinam:
...of prefix strings. I.e if package names "package" and "ABCpackage" exist, previous fix would still return list containing both. Covers cases missed by #1931.


msarahan:
Travis had a bad run.  Restarted, and will merge when one or more jobs have passed.

Thanks @rendinam

PR #1968, opened 2017-04-28
fix perl only creating run test py #1967

### LINKED ISSUE #1967: perl only test.imports is creating run_test.py
I'm on 3.0.0beta1-4-gec5416844 and a perl-only recipe is failing in test phase because the `test.imports` section is creating `run_test.py` that tries to `import LWP::UserAgent` and fails miserably because Python doesen't use `::`

I have a PR that I'm going to submit once I have this issue open.

timsnyder:
- [ ] fixes #1967 

This PR has 4 commits:

1. create a test to demonstrate what I ran into in #1967 
1. fix the test from 1
1. create another test for a case that I'm not sure conda-build will allow.  If test/imports takes a list of dicts, does anything prevent two of the dicts in the list from having `lang` python?
1. fix the test from 3.

### DIFF
--- a/conda_build/create_test.py
+++ b/conda_build/create_test.py
@@ -137,8 +137,7 @@ def create_py_files(m):
             # add any imports specifically marked as python
             if (hasattr(import_item, 'keys') and 'lang' in import_item and
                     import_item['lang'] == 'python'):
-                imports = import_item['imports']
-                break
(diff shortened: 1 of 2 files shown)


msarahan:
Excellent, thanks @timsnyder

PR #1969, opened 2017-04-28
Add libgcc to run dependencies.

johanneskoester:
This PR adds libgcc to the runtime dependencies when the recipe needs compilation.


msarahan:
This is fine for 2.1.x.  In 3.0, this will be unnecessary, because the compiler packages there will add runtimes as necessary for you.  See https://github.com/conda/conda-docs/pull/414/files#diff-ccb76d9bfb39ca714ae7f27cdf13bc7cR445 for details.

johanneskoester:
Thanks!

PR #1972, opened 2017-05-01
Do not add python.app to run_reqs multiple times

mingwandroid:
I ran into this problem building spyder 3.1.4


msarahan:
Good catch, thanks.

PR #1982, opened 2017-05-02
post: Fix $ORIGIN replacement

mingwandroid:
Literal RPATHs of '$ORIGIN' were falling through to the fixup code
when they should have been passed unmodified. A trailing slash was
the cause and this broke openjdk.


msarahan:
Dang it.  Thanks for the fix.

PR #1984, opened 2017-05-02
Add basic environment marker support to conda skeleton pypi

### LINKED ISSUE #1974: Conda skeleton does not support environment markers
[PEP 508](https://www.python.org/dev/peps/pep-0508) added support for [environment markers](https://www.python.org/dev/peps/pep-0508/#environment-markers) in PyPI version specifications. Specifically, environment markers allow you to have versions like ([example on PyPI here](https://pypi.python.org/pypi/coconut))
```
pygments<2.3,>=2.2;python_version>='2.7'
prompt-toolkit<1.1,>=1.0;python_version>='2.7'
futures<3.1,>=3.0;python_version<'3'
argparse<1.5,>=1.4;python_version<'2.7'
```
where the package on the left is only a dependency if the current Python version meets the requirement on the right. This is a very useful feature for packages that aim to support multiple Python versions.

However, when `conda skeleton pypi` creates `meta.yaml`, it does not respect the environment markers, leaving any dependencies with environment markers _completely out of the dependency structure_. Thus, when one runs `conda build` on such a skeleton, it fails with (full error at https://github.com/evhub/coconut/issues/246)
```
RuntimeError: Setuptools downloading is disabled in conda build. Be sure to add all dependencies in the meta.yaml
```
since `setuptools` tries to respect the environment markers, but `conda` won't let it.

This is a major problem for packages that use environment markers in their dependencies. It'd be great if this could be fixed, but until then, do you have any ideas for a suitable workaround?

evhub:
Looks for simple environment markers of the form `^:python_version(<|<=|!=|==|>=|>)(.+)$` and adds in their requirements if the current Python version matches the specification. Resolves #1974.

### DIFF
--- a/conda_build/skeletons/pypi.py
+++ b/conda_build/skeletons/pypi.py
@@ -896,6 +896,33 @@ def get_requirements(package, pkginfo, all_extras=True):
             extras_require = [pkginfo['extras_require'][x] for x in extras]
         except KeyError:
             sys.exit("Error: Invalid extra features: [%s]" % ','.join(extras))
+        # match PEP 508 environment markers; currently only matches the
+        #  subset of environment markers that compare to python_version
+        #  using a single basic Python comparison operator
+        version_marker = re.compile(r'^:python_version(<|<=|!=|==|>=|>)(.+)$')
+        for extra in pkginfo['extras_require']:
+            match_ver_mark = version_marker.match(extra)
+            if match_ver_mark:
+                op, ver = match_ver_mark.groups()
+                try:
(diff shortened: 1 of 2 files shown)


msarahan:
Looks good.  Could you please add some tests so that we're actually exercising your addition (mostly to make sure that no one, especially me, breaks your new functionality later).  If you need help on how conda-build's tests work, don't hesitate to ask.

evhub:
@msarahan Test added and passing!

msarahan:
Cool, looks good enough for now.  I may rewrite it later to be more of a unit test than an integration test, but this is certainly good for now.  Thanks!

PR #1988, opened 2017-05-04
WIP: Fix cross compiling

msarahan:
We discovered a number of issues with the cross-compiling support recently.  This PR is still failing with several git-jinja2 related failures.  These have to do with work_dir not matching up and source code not being present.

I'm going to come back to this late Sunday or first thing Monday.

### DIFF
--- a/conda_build/api.py
+++ b/conda_build/api.py
@@ -26,14 +26,17 @@ from conda_build.utils import get_logger as _get_logger
 
 
 def render(recipe_path, config=None, variants=None, permit_unsatisfiable_variants=True,
-           **kwargs):
+           finalize=True, **kwargs):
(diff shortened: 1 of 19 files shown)


kalefranz:
https://en.wikipedia.org/wiki/Unix_time

msarahan:
You're not wrong, but the ship really has sailed here.  Ray built about 4000 packages across the different platforms that all have timestamps that would be 100 times whatever we would correct them to.  Sorry for the confusion.  If you can propose a correction scheme, I'll consider it.  Otherwise, please deal with it.

kalefranz:
Yeah I can correct it on the conda side.  Give me a sec to make a PR against 4.3.

kalefranz:
https://github.com/conda/conda/pull/6001

PR #1989, opened 2017-05-04
variants: Join and split keys + values by '#' not ','

mingwandroid:
It is more likely that someone will use ',' in a value for example:
LDFLAGS=-Wl,-rpath,blah


jjhelmus:
`#` might not be a good choice for specifying variants for command line usage.  Many shells ignore anything after a `#`.  Would `+` work?

msarahan:
@jjhelmus your line of thought is good, but this representation is only passed around within Python.  It never makes it out to the shell.  It is strictly an intermediate representation of keys that get joined so that they don't become extra dimensions (``zip_keys``).

PR #2001, opened 2017-05-08
support optional dependencies spec from conda 4.4

### LINKED ISSUE #1964: Support conda 4.4 optional/constrained dependencies
With https://github.com/conda/conda/pull/4982, conda is ready to support constrained dependencies, and now we need to add the capability into meta.yaml.

Under the requirements section in meta.yaml, we need to add a `run_constrained` key that takes a list of specs, just like `build` and `run`.  The result should be an additional `constrains` key in `index.json`, parallel to `depends` that comes from the `run` key in meta.yaml.

msarahan:
fixes #1964

### DIFF
--- a/conda_build/metadata.py
+++ b/conda_build/metadata.py
@@ -264,6 +264,7 @@ default_structs = {
     'requirements/host': list,
     'requirements/run': list,
     'requirements/conflicts': list,
+    'requirements/run_constrained': list,
     'test/requires': list,
     'test/files': list,
     'test/source_files': list,
@@ -370,7 +371,7 @@ FIELDS = {
               'pin_depends', 'include_recipe',  # pin_depends is experimental still
               'preferred_env', 'preferred_env_executable_paths',
               ],
-    'requirements': ['build', 'host', 'run', 'conflicts'],
+    'requirements': ['build', 'host', 'run', 'conflicts', 'run_constrained'],
     'app': ['entry', 'icon', 'summary', 'type', 'cli_opts',
             'own_environment'],
     'test': ['requires', 'commands', 'files', 'imports', 'source_files'],
@@ -961,6 +962,11 @@ class MetaData(object):
         if preferred_env:
             d['preferred_env'] = preferred_env
 
+        # conda 4.4+ optional dependencies
+        constrains = self.get_value('requirements/run_constrained')
+        if constrains:
+            d['constrains'] = constrains
+
         if self.get_value('build/features'):
             d['features'] = ' '.join(self.get_value('build/features'))
         if self.get_value('build/track_features'):
@@ -1276,10 +1282,12 @@ class MetaData(object):
             if hasattr(output_reqs, 'keys'):
                 build_reqs = output_reqs.get('build', []) + requirements.get('build', [])
                 run_reqs = output_reqs.get('run', [])
+                constrain_reqs = output_reqs.get('run_constrained', [])
             else:
                 output_reqs = ensure_list(output_reqs)
                 build_reqs = output_reqs + requirements.get('build', [])
                 run_reqs = output_reqs
+                constrain_reqs = []
             if 'name' in output:
                 # since we are copying reqs from the top-level package, which
                 #   can depend on subpackages, make sure that we filter out
@@ -1290,6 +1298,8 @@ class MetaData(object):
 
             requirements['build'] = build_reqs
             requirements['run'] = run_reqs
+            if constrain_reqs:
+                requirements['run_constrained'] = constrain_reqs
             output_metadata.meta['requirements'] = requirements
             output_metadata.meta['package']['version'] = output.get('version') or self.version()
             extra = self.meta.get('extra', {})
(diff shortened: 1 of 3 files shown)


msarahan:
@kalefranz please review - especially make sure that the test case adequately captures the functionality that you had in mind.

kalefranz:
codecov has been giving me problems recently on Appveyor too.  I pinned to `pip install codecov==2.0.5` a few days ago.  That seemed to help.

msarahan:
Thanks.  Merging.

183amir:
There is this option in conda build:
```yaml
build:
  pin_depends: strict
```
Can this be used for `run_constrained` too? Or is something similar available?

msarahan:
no, pin_depends is something completely different.  See this for its origin: https://github.com/conda/conda-build/pull/741

The idea for it was to constrain dependencies tightly.  optional dependencies are completely different - they say "if this dependency is installed, make sure it meets these additional constraints"

183amir:
So I have this recipe:
```yaml
package:
  name: bob-devel
  version: {{ os.environ.get('CI_BUILD_TAG', 'vx.y.z')[1:] }}

build:
  number: 0
  pin_depends: strict

requirements:
  build:
    - numpy x.x
  run:
    - python
    - numpy x.x
    # Our dependency list.
    # Everything is pinned to allow better reproducibility.
    # The pinnings below are coming from anaconda 4.4.0
    - curl 7.52.1
    - cython 0.25.2
    - cytoolz 0.8.2
    - dask 0.14.3
    - freetype 2.5.5
    - hdf5 1.8.17
    - imagesize 0.7.1
    - ipython 5.3.0
    - jpeg 9b
    - libpng 1.6.27
    - libtiff 4.0.6
    - matplotlib 2.0.2
    - mkl 2017.0.1
    - nose 1.3.7
    - numba 0.33.0
    - numba 0.33.0
    - numexpr 2.6.2
    - pandas 0.20.1
    - pillow 4.1.1
    - pip 9.0.1
    - scikit-image 0.13.0
    - scikit-learn 0.18.1
    - scipy 0.19.0
    - setuptools 27.2.0
    - six 1.10.0
    - sphinx 1.5.6
    - sqlalchemy 1.1.9
    - toolz 0.8.2
    # these pinnings of indirect dependencies make sure bob-devel is compatible with anaconda 4.4.0, hopefully
    - bokeh 0.12.5
    - nbconvert 5.1.1
    - testpath 0.3
    # the rest of the dependencies
    - boost 1.61.0
    - caffe 1.0.0rc5  # [linux]
    - cmake 3.6.3
    - coverage 4.3.4
    - cyvlfeat 0.4.5
    - dlib 19.0
    - docopt 0.6.2
    - ffmpeg 2.8.10
    - gcc 4.*  # [linux]
    - giflib 5.1.4
    - ipdb 0.10.2
    - kaldi r7271.1a4dbf6  # [linux]
    - keras 2.0.8  # [linux]
    - libblitz 0.10
    - libgcc 4.*  # [linux]
    - libmatio 1.5.6
    - libogg 1.3.2
    - libsvm 3.21
    - madmom 0.15.1
    - menpo 0.8.1
    - menpofit 0.5.0
    - menpowidgets 0.3.0p6
    - mne 0.14.1
    - opencv 3.1.0
    - pkg-config 0.28
    - pkgtools 0.7.3
    - pyedflib 0.1.9
    - schema 0.6.5
    - sox 14.4.2
    - sphinx_rtd_theme 0.2.4
    - tensorflow 1.2.1  # [linux]
    - toolchain 2.3.2
    - virtualenv 15.1.0
    - vlfeat 0.9.20
    - zc.buildout 2.9.2

test:
  imports:
    - tempfile

about:
  home: http://www.idiap.ch/software/bob
  license: BSD 3-Clause
  summary: A meta-package to define an environment for Bob package development

extra:
  recipe-maintainers:
    - anjos
    - amohammadi
```
And I pin the versions that I want in it. But the great thing about this is that it pins the indirect dependencies too. It's like taking a snapshot of the environment while only pinning the relevant ones in the build dependencies.

Now what I really want is actually this frozen environment (with direct and indirect pins) to be in `run_constrained` instead of `run`. This way I don't have to install all the dependencies when I install `bob-devel` but if in future I decide to install one of those dependencies in my environment, I will get a version that is mandated by `bob-devel`.

Does that make sense?

msarahan:
Yes, I can see the draw in that.  I'd gladly consider a PR.  It may also be better to do this with configuration on the conda side somehow - just depends which way is easier to share with other people.

183amir:
Is it possible to do this with conda configurations?
I would also see that making a package like `bob-devel` a build dependency as well would be useful.

msarahan:
I think it's possible, but you should file an issue on the conda tracker and ask for examples and docs.

kalefranz:
If I understand correctly, this is exactly how the `pinned_packages` configuration parameter works in conda 4.4 now.  Note you can have an environment-only `.condarc` file (at the prefix root of the environment).  The `--env` flag on `conda config` also manipulates this file.

183amir:
Hi @kalefranz while having this as the conda configuration is useful. I don't think you can benefit from that during the conda-build process. Right now, with the introduction of the variants, it is very easy to pin direct dependencies during build time. But, it is not possible to pin the indirect recipes during build time. But I am thinking that I can achieve this with the new `run_constrained` feature.

jakirkham:
Also not seeing docs. Is there an issue for this or am I just missing it?

PR #2010, opened 2017-05-10
implement --old-build-string option (no hash in filename)

### LINKED ISSUE #1919: conda-build 3.0.0alpha3 produces conda packages with new build strings
I'm raising some concerns about the new build strings which conda-build 3.0 is creating:
  * the new build strings are different from the existing ones, and this might cause problems for existing users.  The build string does not need to be of a specific format, as far as conda is concerned.  The existing convention, e.g. `np19py27_1` has existed since the beginning of the conda project (this convention actually predates conda-build) and many people are probably very used to it by now.  Moreover, people (such as myself, but likey others as well) have written scripts around conda-build which rely on the existing build string naming convention.
  * the new build strings contain opaque hashes, which makes users wonder what exactly they mean and which information they contain.  In creating the conda ecosystem, one of it's core design principles has always been simplicity and transparency.  The hashes are not transparent and therefore violate this design principle.

msarahan:
closes #1919 

Please review @ilanschnell

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -465,7 +465,8 @@ def create_info_files(m, files, prefix):
         # make sure we use '/' path separators in metadata
         files = [_f.replace('\\', '/') for _f in files]
 
-    write_hash_input(m)
+    if m.config.filename_hashing:
+        write_hash_input(m)
     write_info_json(m)  # actually index.json
     write_about_json(m)
     write_link_json(m)
--- a/conda_build/cli/main_render.py
+++ b/conda_build/cli/main_render.py
@@ -99,6 +99,13 @@ source to try fill in related template variables.",
         help="""Additional variant config files to add.  These yaml files can contain
         keys such as `c_compiler` and `target_platform` to form a build matrix."""
     )
+    p.add_argument(
+        "--old-build-string", dest="filename_hashing", action="store_false",
+        default=True, help=("Disable hash additions to filenames to distinguish package "
+                            "variants from one another.  NOTE: any filename collisions are "
+                            "yours to handle.  Any variants with overlapping names within a "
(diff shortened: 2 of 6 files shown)


teoliphant:
Thanks for working on this.   Wouldn't it be better if we just instead provided a "--build-string-override" option.   Then conda-build wouldn't have to have any of the "old build-string" logic.   But, a package builder could just override if they insist (and are willing to deal with the hassles themselves).    The hashing approach is a much better "standardized approach" for package naming.   It would be great if conda-build just used it, but provided a "escape-hatch" for users of conda-build that for what-ever reason want to use a different package-naming convention in their collections.

ilanschnell:
I would argue that having both the hash and explicit naming is somehow confusing, as the both contain the same information (but in different ways).  The default build string could always be a pure hashsum, but I'm not very concerned about the default.

What I think wold be useful is a `--build-string` option would overwrite the default hash.  The `--build-string` option is also more flexible for possible future conventions which some people might want to implement.

msarahan:
The hash is only an addition to the previous build string behavior.  Hashes do not replace things like py27 or np111, they only supplement them.  Furthermore, the automatic determination logic only takes effect if the ``build/string`` field is not defined in meta.yaml.  @teoliphant I believe that defining the ``build/string`` field in meta.yaml provides the customization that you're after.  Perhaps the flag added by this PR would be better named ``--no-hash-in-build-string``

I don't see good reason to replace the existing customization of the build string via meta.yaml with a CLI flag that would presumably be set to some static string.  Do you?  It would be much less flexible than the existing scheme (no jinja2 templating at all)

msarahan:
Discussed with @ilanschnell during standup this morning.  Option added by this PR is acceptable.

PR #2014, opened 2017-05-10
add sha256 value to repodata.json produced by 'conda index'

kalefranz:
Although no new tests added, this PR should be pretty safe.  The function `hashsum_file` being imported from conda_interface is used by `md5_file` one level down.  So there really shouldn't be any new code paths or a net negative to test coverage.


msarahan:
Looks fine to me.  merging.

PR #2025, opened 2017-05-16
Replace setuptools non-unicode non-space capable cli-??.exe with CPython ones

mingwandroid:
Built with m2w64-toolchain, though we should build these from source instead.


mingwandroid:
Changes I had to make for `mingw-w64`:

```
0313-rdonnelly:setuptools rdonnelly$ diff -urN launcher.c.orig launcher.c
--- launcher.c.orig	2017-05-16 02:33:34.000000000 +0100
+++ launcher.c	2017-05-16 02:32:27.000000000 +0100
@@ -765,6 +765,25 @@
 
 static COMMAND path_command;
 
+#if 1 // !defined(_MSC_VER)
+errno_t _wdupenv_s_emulated(wchar_t **buffer,  
+                            size_t *numberOfElements,  
+                            const wchar_t *varname)
+{
+    size_t szreq;
+
+    errno_t err = _wgetenv_s(&szreq, NULL, 0, varname);
+    if (szreq == 0)
+        return 1;
+    *buffer = (wchar_t*) malloc(sizeof(wchar_t) * szreq);
+    if (!*buffer)
+        return 1;
+    err = _wgetenv_s(&szreq, *buffer, szreq, varname);
+    return err;
+}
+#define _wdupenv_s _wdupenv_s_emulated
+#endif
+
 static COMMAND * find_on_path(wchar_t * name)
 {
     wchar_t * pathext;
@@ -1565,9 +1584,14 @@
 
 #else
 
+#if defined(_MSC_VER)
 int cdecl wmain(int argc, wchar_t ** argv)
+#else
+int cdecl main(int argc, wchar_t ** argv)
+#endif
 {
     return process(argc, argv);
 }
 
 #endif
+
```

.. not much. Will retarget tomorrow, this was only a way of getting exes to you and @msarahan for testing (and also to see what the CI would make of it).

I think I want to build from source here though, and add m2w64-toolchain as a build depend for conda-build. After all, why would be be repackaging binaries when building from source is trivial?

Night!

msarahan:
Seems like it doesn't get PYTHONPATH set right or something.  All the entry point tests are failing their imports:

https://ci.appveyor.com/project/ContinuumAnalyticsFOSS/conda-build/build/1.0.1945/job/vduko7e5la26l91j#L2299

mingwandroid:
I'll look into that.

mingwandroid:
OK, seems it's finding my system Python. Looking into why ..

PR #2031, opened 2017-05-16
Replace setuptools non-unicode non-space capable cli-??.exe with CPyt…

mingwandroid:
…hon ones

Built with m2w64-toolchain, though we should build these from source instead.

Also added gui versions as I think we want to use those in some situations.

And the patch and build script, for safe keeping.


mingwandroid:
OK, let's see how this one performs. Having studied the code more, it allows us to use `wchar_t` in our sys.prefixes/shebangs, which in MS-land are 16-bit. This is not ideal, but better than nothing.

msarahan:
How does the switch to 8-bit unicode in python 3.6 affect this?  Does it matter at all?

mingwandroid:
No effect on this. That's internal processing. Windows still uses UTF-16 as the only way of passing unicode to main() or via env vars. Here all I am doing is avoiding the old cli-XX.exe programs which coerce everything into ASCII (turning nearly all Unicode into ASCII '?' characters).

msarahan:
The appveyor fail looks like a fluke to me.  Merging.

PR #2033, opened 2017-05-17
{cli,gui}-{32,64}.exe now with added manifests

mingwandroid:


mingwandroid:
@msarahan here you go.

msarahan:
Merging.  Ray has done tests locally.  We need to push out this in order to build conda with these changes.

PR #2034, opened 2017-05-17
launcher_sources: Add manifest.xml

mingwandroid:


mingwandroid:
This has no bearing on the build you just made, but would be nice for anyone who tries to build the launchers from source.

msarahan:
Thanks.  Merging.

PR #2037, opened 2017-05-18
Removed old example recipes

mandeep:
Removed the example_packages directory from the repository root as the examples were no longer working with the most recent conda-build versions.


msarahan:
Thanks.

PR #2039, opened 2017-05-18
Fix {cli,gui}-{32,64}.exe for spaces in the final exe path

mingwandroid:
Sorry about this @msarahan.


msarahan:
No problem.  Thanks for fixing.

PR #2042, opened 2017-05-18
allow pin_run_as_build to be a single 'x.x' string

minrk:
[docs](https://github.com/conda/conda-docs/pull/414/files#diff-d1a5dbb2f223de4935d93990676c7074) suggest that pinning should be specified as

```yaml
pin_run_as_build:
  pkg: x.x
```

but the code required a dict:

```yaml
pin_run_as_build
  pkg:
    min_pin: x.x
    max_pin: x.x
```

This transforms `'x.x'` string from the docs example into the same value for min/max kwargs. I'm not sure if this is the right place to do the transform, or if it should be done when the config file is loaded.

### DIFF
--- a/conda_build/render.py
+++ b/conda_build/render.py
@@ -111,8 +111,11 @@ def get_pin_from_build(m, dep, build_dep_versions):
     if (version and dep_name in m.config.variant.get('pin_run_as_build', {}) and
             not (dep_name == 'python' and m.noarch) and
             dep_name in build_dep_versions):
-        pin = utils.apply_pin_expressions(version.split()[0],
-                                            **m.config.variant['pin_run_as_build'][dep_name])
+        pin_cfg = m.config.variant['pin_run_as_build'][dep_name]
+        if isinstance(pin_cfg, str):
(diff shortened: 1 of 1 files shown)


msarahan:
For the sake of usability, do you find it better that this is specifiable as just one thing, as a shortcut to avoid the dict?  I guess the question is: are the docs wrong, and need to show just the dict way of doing things, or do the docs present a use case that is nicer?

I don't really care either way - just wondering what your opinion is.  Thanks for the PR.

msarahan:
Merging.  Looks like you got started when I had an error that slipped in from another PR.

I have thought about this, and I like being able to go either way.  Thanks for the PR, @minrk.

minrk:
> For the sake of usability, do you find it better that this is specifiable as just one thing

Sorry, I haven't been keeping up on my notifications this week due to PyCon. Yes, I think I'll ~always want to pin the packages I've been to `x.x` or `x.x.x` rather than having different values for min and max, so setting with a single string seems most logical to me.

PR #2069, opened 2017-06-05
Allow ~ in source urls and aggregate 'host' and 'build' deps when not cross compiling.

mingwandroid:


mingwandroid:
Urgh, seems we already had a test_source.py. Closing until I fix this.

PR #2072, opened 2017-06-05
add Windows env vars passthrough related to CPU info

### LINKED ISSUE #2064: During testing phase, platform.machine() returns None on Windows (regression)
I am maintaining a conda package for [py-cpuinfo](https://github.com/workhorsy/py-cpuinfo), but with recent conda-build versions (from the 2.1.x branch) it fails during the test phase when it tries to check via `platform.machine()` whether it runs on a supported platform. The reason is that `platform.machine()` does not return `AMD64` as expected, but instead `None` (it works during the build phase).
I bisected the problem to commit 430ce9928, although I can't say I see how the commit would lead to this outcome. Here's a pretty minimal `meta.yaml` file that reproduces the issue for me, it fails with an `AssertionError` with conda-build versions after the above mentioned commit (or from release 2.1.6 on):
```yaml
package:
  name: dummy
  version: 1.0

requirements:
  run:
    - python

test:
  commands:
    - python -c "import platform; assert platform.machine()"
```
The problem seems to be Windows-specific (I only tested 64bit), it works fine for me on Linux and OS X.

msarahan:
fixes #2064

### DIFF
--- a/conda_build/environ.py
+++ b/conda_build/environ.py
@@ -433,6 +433,10 @@ def windows_vars(prefix):
         'USERNAME': os.getenv('USERNAME'),
         'USERPROFILE': os.getenv('USERPROFILE'),
         'windir': os.getenv('windir'),
+        # CPU data, see https://github.com/conda/conda-build/issues/2064
+        'PROCESSOR_ARCHITEW6432': os.getenv('PROCESSOR_ARCHITEW6432'),
+        'PROCESSOR_ARCHITECTURE': os.getenv('PROCESSOR_ARCHITECTURE'),
+        'PROCESSOR_IDENTIFIER': os.getenv('PROCESSOR_IDENTIFIER'),
     }
 
 


AndresGuzman-Ballen:
@msarahan Hi. I noticed that when I upgrade from conda-build 2.1.15 -> 2.1.16, I get a failure with running PyTables' setup.py:

```
  File "setup.py", line 35, in <module>
    import cpuinfo
  File "C:\Users\aguzmanb\Development\SAT.2018.0.0\conda-build\pytables_1500650619595\work\cpuinfo.py", line 1565, in <module>
    _check_arch()
  File "C:\Users\aguzmanb\Development\SAT.2018.0.0\conda-build\pytables_1500650619595\work\cpuinfo.py", line 1522, in _check_arch
    raise Exception("py-cpuinfo currently only works on X86 and some ARM CPUs.")
Exception: py-cpuinfo currently only works on X86 and some ARM CPUs.
```

This is the line in question from https://github.com/PyTables/PyTables/blob/v3.3.0/cpuinfo.py#L1519
```
def _check_arch():
	arch, bits = parse_arch(DataSource.raw_arch_string)
	if not arch in ['X86_32', 'X86_64', 'ARM_7', 'ARM_8']:
		raise Exception("py-cpuinfo currently only works on X86 and some ARM CPUs.")

```

msarahan:
No idea off the top of my head, but you should try to debug what exactly raw_arch_string is coming out as for you, and track backwards from there.  We may need to include more env vars.

PR #2073, opened 2017-06-06
add new build arg to ignore certain run_exports

### LINKED ISSUE #2058: Provide method to override run_exports
Ideally, people never need to do this.  However, it's important to have the ability.

msarahan:
fixes #2058


mingwandroid:
Out of interest, when would you expect this to be used?

msarahan:
If anyone using run_exports was overly zealous in pinning, this may be necessary to override the upstream pin.  I hope it will never be necessary, but I think it might be necessary for temporary fixes.

PR #2076, opened 2017-06-06
Skeleton packages not found during build

waltonseymour:


msarahan:
This is an interesting idea, but I'm afraid I don't trust that our skeletonizer is reliable enough to serve here.  Maybe instead we should search some common locations and recommend places for people to get recipes?  Something like:

Package `abc` is not available in your configured channels, and no recipe could not be found locally.  However, there are packages available at:

(return search results from anaconda client)

Additionally, recipes are available at:

(look for recipe on conda-forge)

Finally, you may be able to generate a recipe for this package using

(conda skeleton command with some detection to try to recommend cran, pypi, cpan, or otherwise)

waltonseymour:
Yeah I'm not sure what the robust solution should be. 
This was something I wrote as an alternative of the `recursive` option which skeletons everything regardless if a package exists in your current channels.
It works pretty well for my needs, but can get bogged down with deep dependency trees.

Figured I'd share to spark a discussion at the least. 😄

PR #2078, opened 2017-06-08
log info/debug to stdout.  Log warn/error to stderr.  dedupe better.

### LINKED ISSUE #1827: Feature: Add logging option for INFO and WARNING
We have a customer request that an logging handlers be added for redirecting INFO and WARNING produced from the logging package to another place. Non critical issues could be sent to a file or suppressed. Possibly add these options for CRITICAL and ERROR. They would like separate handling for a cleaner run log as they sometimes build/deploy to other users systems which create concerns with the output.

msarahan:
fixes #1827 

This also adds configurability to conda-build's logger usage.  Users may configure the logging using the YAML syntax presented at the bottom of the section at https://docs.python.org/3.6/howto/logging.html#configuring-logging

and by putting an entry into their .condarc files:

```
conda-build:
    log_config_file: (path to config file)
```

### DIFF
--- a/ci/travis/run.sh
+++ b/ci/travis/run.sh
@@ -9,7 +9,7 @@ if [[ "$FLAKE8" == "true" ]]; then
     conda build conda.recipe --no-anaconda-upload
     conda create -n _cbtest python=$TRAVIS_PYTHON_VERSION conda-build
     # because this is a file, conda is not going to process any of its dependencies.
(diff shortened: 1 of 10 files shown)


msarahan:
Thanks @mingwandroid - I think that's a good idea.  I'll add it in 3.0.1.

PR #2082, opened 2017-06-09
Add pyldd and use it when ldd/otool fails

mingwandroid:
For now, both ldd/otool and pyldd are used and disparities are warned about.

### DIFF
--- a/conda_build/api.py
+++ b/conda_build/api.py
@@ -297,11 +297,11 @@ def test_installable(channel='defaults'):
 
 
 def inspect_linkages(packages, prefix=_sys.prefix, untracked=False, all_packages=False,
-                     show_files=False, groupby='package'):
+                     show_files=False, groupby='package', sysroot=''):
     from .inspect import inspect_linkages
     packages = _ensure_list(packages)
     return inspect_linkages(packages, prefix=prefix, untracked=untracked, all_packages=all_packages,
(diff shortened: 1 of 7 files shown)


mingwandroid:
I think this should be good to go. The test failures do not look to be related to this code.

I didn't add any new tests though, do you feel the existing tests are sufficient or do we need some more?

`conda inspect linkages` needs more work IMHO. On Linux it's recursive and on macOS it isn't due to the differences between `otool -L` and `ldd`. This is a bit daft. I think we should be recursive on both systems. For now, I emulate the native tool (but this is very easy to change).

I also think we should be using this at the end of every build and warning about both missing and unnecessary dependencies.

msarahan:
Why don't you go ahead and just make this everything you want it to be?  Recursive on Mac, and with the warnings you want to see.  I guess you would need to add tests for the recursive behavior, and for the warnings. TDD?

mingwandroid:
I would kind of like one iteration of a release where it is like for like and compared with the native tool (but now also cross capable) and when I have time, I will think about how to best leverage it to help with package creation. I need to get CDT packages working ASAP and I don't want you to hold up CB3 (I am also happy for this not to go in the first version of CB3 if it reduces the risk).

msarahan:
Yep, failures look unrelated.  Thanks.

PR #2087, opened 2017-06-13
test failure using filename_hashing=False

neok-m4700:
v3.0.0
avoid failing test when using filename_hashing=False (reading non-existent .json file)


msarahan:
Thanks!

PR #2012, opened 2017-05-10
Added semicolons to the printed print statement for the test to preven…

vinjana:
… to prevent syntax errors in Perl (Python and Lua accept semicolons at statement ends).

PR #2089, opened 2017-06-14
Backport of #2012 to 2.1.x branch

nsoranzo:
Added semicolons to the printed print statement for the test to prevent syntax errors in Perl (Python and Lua accept semicolons at statement ends).


msarahan:
I think the CLA is unnecessary here because you're only resubmitting work that is already covered.  Merging.

nsoranzo:
Thanks @msarahan!

PR #2090, opened 2017-06-14
Fixed issue in conda_convert where dependencies weren't being added

### LINKED ISSUE #2085: conda convert ignores dependencies
```
conda convert pakcage-xyz-linux-bla.tar.bz -d 'numpy >=1.8'
```
The dependency 'numpy >=1.8' will not be in the converted package.

How it should be?
The dependency should be included in the new package. Even better, the parsed meta.yaml with conditional depenencies should be used for the converted package.

Example:
meta.yaml
```
requirements:
    build:
        - python
    run:
        - python
        - pywin32 # [win]
```

when first built on linux and then converted to win64 using `conda convert --platform win-64 -o /opt/miniconda3/conda-bld/ /opt/miniconda3/conda-bld/linux-64/package_xy-0.1.31-py34_0.tar.bz2 -d 'pywin32'` the converted package will not have the dependency of pywin32.
The "quick hack" with -d failed as well as mentioned above.

Fix for convert:
See https://github.com/conda/conda-build/blob/master/conda_build/convert.py#L235
There should be something like:
```
# see https://github.com/conda/conda-build/blob/master/conda_build/convert.py#L412
        if dependencies:
            info['depends'].extend(dependencies)
```

mandeep:
fixes #2085

It looks like dependencies weren't being passed to the get_pure_py_file_map function from the conda_convert function. I added logic in conda_convert to deal with dependencies and added an argument to get_pure_py_file_map to accept these dependencies. This fix includes a test that converts a package with two dependencies and checks that these dependencies are listed in index.json.

### DIFF
--- a/conda_build/convert.py
+++ b/conda_build/convert.py
@@ -208,7 +208,7 @@ path_mapping_identity = [
 pyver_re = re.compile(r'python\s+(?:[<>=]*)(\d.\d)')
 
 
-def get_pure_py_file_map(t, platform):
+def get_pure_py_file_map(t, platform, dependencies):
     info = json.loads(t.extractfile('info/index.json').read().decode('utf-8'))
     try:
         paths = json.loads(t.extractfile('info/paths.json').read().decode('utf-8'))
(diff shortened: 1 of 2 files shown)


msarahan:
Good fix.  Thanks.

PR #2094, opened 2017-06-15
Backport #1968

nsoranzo:
Backport of #1968 to fix https://github.com/conda/conda-build/issues/1967 on conda-build 2.1.x.

### DIFF
--- a/conda_build/create_test.py
+++ b/conda_build/create_test.py
@@ -131,8 +131,8 @@ def create_py_files(dir_path, m):
     likely_non_python_pkg = likely_r_pkg or likely_lua_pkg or likely_perl_pkg
 
     if likely_non_python_pkg:
-        imports = ensure_list(m.get_value('test/imports', []))
-        for import_item in imports:
+        imports = []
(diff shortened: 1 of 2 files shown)


nsoranzo:
I guess the test framework is too different between the 2 branches, should I just cherry-pick the changes to `conda_build/create_test.py` ?

msarahan:
I think all you need to change is ```testing_metadata``` -> ```test_metadata``` in the test function definitions.

I was annoyed by the inconsistency I created, and fixed it in 3.x.  The other errors look like sporadic failures that we have yet to nail down.

nsoranzo:
I'm also:
- adding `testing_workdir` as first parameter of `ct.create_py_files()`
- using `testing_workdir` instead of `testing_metadata.config.test_dir` when creating the `test_file` variable

I hope that's correct.

nsoranzo:
The added tests now pass.

msarahan:
Thanks @nsoranzo and thanks again @timsnyder for originally fixing this.

PR #2096, opened 2017-06-16
Removed subdir as cli platform default and now checking for length of platforms

mandeep:
Using the subdir as the default platform on the subcommand parser causes two problems: 

When converting from one platform to the subdir platform, the platforms list will contain a duplicate of the subdir platform. This causes conda convert to convert the package to the subdir platform twice.

The other problem is that when converting from one platform to another platform and both platforms are not the subdir platform,  the package is needlessly converted to the subdir platform. This may be a bug in how conda.base.context handles subdir, but I think this fix is necessary in conda-build anyway.


msarahan:
closing as covered by #2123 and #2116

PR #2098, opened 2017-06-16
Expose DISPLAY variable on Linux (2.1.x)

gqmelo:
This is needed when testing some packages involving GUI.

### DIFF
--- a/conda_build/environ.py
+++ b/conda_build/environ.py
@@ -469,8 +469,9 @@ def linux_vars(compiler_vars, prefix, config):
     if config.arch == 32:
         compiler_vars['CFLAGS'] += ' -m32'
         compiler_vars['CXXFLAGS'] += ' -m32'
-    return {}
-
+    return {
+        'DISPLAY': os.getenv('DISPLAY'),
+    }
 
 def system_vars(env_dict, prefix, config):
     d = dict()


conda_build/environ.py
```diff
@@ -469,8 +469,9 @@ def linux_vars(compiler_vars, prefix, config):
     if config.arch == 32:
         compiler_vars['CFLAGS'] += ' -m32'
         compiler_vars['CXXFLAGS'] += ' -m32'
-    return {}
```
msarahan: This is not your bug, but surely this is not right?  It's always returning an empty dict?  I guess we get away with it because we're modifying compiler_vars in place here, but this code smells.

I'll add a separate issue for this code smell.

gqmelo: I think it's fine. ```linux_vars``` is only used to update the other dict:

```d.update(linux_vars(compiler_vars, prefix, config))```

msarahan: Yeah, it works, but we're both modifying compiler_vars in place, and then also doing the update?  That seems confusing and fragile.  We should do all of one or the other - set CFLAGS/CXXFLAGS in the return dictionary here (no modification in place) or set everything on compiler vars, and remove the update call.

gqmelo: Ah, OK. I agree is confusing. I would rather only return dicts instead of modifying in-place.

msarahan:
Thanks @gqmelo - merging, and will take care of the smell in another PR.

PR #2102, opened 2017-06-16
Added test actions so that conda build errors out earlier if test dependencies not found

### LINKED ISSUE #671: Check that test/requires packages are available before building
I have some packages whose builds fail on certain platforms because packages required for the tests are not available. Currently, the failure is only reported after the package has been completely built. If possible, it would be nice to check if the test dependencies are satisfiable before running the build.

mandeep:
fixes #671

Used get_install_actions from environ.py to check if the test dependencies listed in meta.yaml could be found. This fix also includes a test to make sure that a DependencyNeedsBuildingError is raised when a test dependency can't be found.

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -925,20 +925,27 @@ def build(m, post=None, need_source_download=True, need_reparse_in_env=False, bu
                                                     output_folder=m.config.output_folder,
                                                     channel_urls=tuple(m.config.channel_urls))
 
-        # make sure test deps are available before taking time to create build env
-        environ.get_install_actions(m.config.test_prefix,
-                                    tuple(test_run_ms_deps), 'test',
-                                    subdir=m.config.build_subdir,
-                                    debug=m.config.debug,
-                                    verbose=m.config.verbose,
-                                    locking=m.config.locking,
-                                    bldpkgs_dirs=tuple(m.config.bldpkgs_dirs),
-                                    timeout=m.config.timeout,
-                                    disable_pip=m.config.disable_pip,
-                                    max_env_retry=m.config.max_env_retry,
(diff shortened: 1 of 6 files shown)


conda_build/build.py
```diff
@@ -888,6 +888,8 @@ def build(m, post=None, need_source_download=True, need_reparse_in_env=False, bu
         # installed files at packaging-time.
         host_ms_deps = None
         build_ms_deps = None
+        test_ms_deps = m.get_value('test/requires', [])
```
msarahan: You should add the runtime dependencies to this also.  Test time deps are actually run + test deps.

mandeep:
Added the runtime dependencies to the dependency check and added an exception test.

marscher:
thank you for solving this!

mandeep:
@msarahan It looks like this PR breaks tests that use false package names in runtime dependencies. Travis shows test_subpackages::test_toplevel_entry_points_do_not_apply_to_subpackages breaking, but there may be more. What's the best course of action?

msarahan:
@mandeep it looks like this is affecting recipes that are effective passed "out of order" - out of the order in which they should be built.  It should be possible to rearrange things so that your exceptions here help us determine the correct order.  I'll take a look when I get a chance.

Also, as a note to anyone looking at this: this is going to be a 3.x feature.  It is too involved to port this back to 2.1.x.

msarahan:
Aside from conda issues https://github.com/conda/conda/issues/5614 and https://github.com/conda/conda/issues/5610, this is working well.

Thanks @mandeep - this is a nice addition.  We'll have to see about profiling to speed things up, as we've taken a speed hit here, but it's fine for now.

PR #2105, opened 2017-06-18
Fix variable name in `bldpkgs_dirs` comment

jakirkham:
The variable `cc.subdir` was renamed to the property `subdir` in a previous commit (referenced below). However the comment above was not. This corrects the comment to clarify what this now represents.

ref: https://github.com/conda/conda-build/commit/f99776f9a0f25e6cd2aa363ef036b8b7bbece7a6


jakirkham:
Have skipped CI here as this change is only to the comment (not the code).

msarahan:
Thanks, good catch.

PR #2108, opened 2017-06-19
Added a check to skip conversion if source platform equals target platform

mandeep:
Currently, conda convert will convert a package even if both platforms and architectures are the same.  This PR includes a check that will skip conversion if the source platform and architecture equals the target platform and architecture. 

Also included is a test to make sure that these packages are not built. 

Because this feature broke the current convert test suite, the test suite has been refactored to omit conversions between the same platform and architecture.

### DIFF
--- a/conda_build/convert.py
+++ b/conda_build/convert.py
@@ -172,6 +172,31 @@ def _check_paths_version(paths):
         raise RuntimeError("Cannot handle info/paths.json paths_version other than 1")
 
 
+def _is_same_platform_arch(path, platform):
+    """Compare the source platform and architecture to the target platform and architecture.
+
+    If the source platform and architecture are the same as the target platform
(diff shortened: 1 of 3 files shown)


conda_build/convert.py
```diff
@@ -172,6 +172,31 @@ def _check_paths_version(paths):
...
+    elif info.get('arch') == 'x86':
+        source_architecture = 32
+    else:
+        # if the source arch can't be determined then set it to
+        # an abritrary number 0 so that the package still builds
+        source_architecture = 0
+
+    source_platform_arch = '{}-{}' .format(source_platform, source_architecture)
+
+    return platform == source_platform_arch
```
msarahan: this is probably fine for now, but should account for "aliases" at some point.

msarahan:
thanks @mandeep

PR #2109, opened 2017-06-19
skeleton: Add rpm

mingwandroid:
From my perspective this is ready for review now.

A commandline you could use to try this out is:

`conda skeleton rpm --recursive --distro centos6 --output-dir=~/conda/conda-recipes/cdts/centos6-x86_64 --architecture x86_64 libselinux-devel libX11-devel libselinux-devel java-1.7.0-openjdk java-1.7.0-openjdk-devel libgcj mesa-libGL-devel alsa-lib-devel alsa-lib alsa-utils gtk2-devel gtk+-devel libXrandr-devel ruby gperf xorg-x11-proto-devel`

Points for discussion:

What extra metadata do we want to include with these packages?
Should I allow passing `--architecture` multiple times?

### DIFF
--- a/conda_build/jinja_context.py
+++ b/conda_build/jinja_context.py
@@ -441,7 +441,7 @@ def cdt(package_name, config, permit_undefined_jinja=False):
         "DEBUG_FFLAGS": "-Og -g -Wall -Wextra -fcheck=all -fbacktrace -fimplicit-none -fvar-tracking-assignments",
       }
     }
-    """
+    """  # NOQA
 
     cdt_name = 'cos6'
     arch = config.host_arch or config.arch
@@ -456,6 +456,7 @@ def cdt(package_name, config, permit_undefined_jinja=False):
     else:
         result = (package_name + '-' + cdt_name + '-' + cdt_arch)
     return result
+
 
 def context_processor(initial_metadata, recipe_dir, config, permit_undefined_jinja,
                       allow_no_other_outputs=False, bypass_env_check=False):
--- a/conda_build/jinja_context.py
+++ b/conda_build/jinja_context.py
@@ -391,9 +391,23 @@ def compiler(language, config, permit_undefined_jinja=False):
 
         # support cross compilers.  A cross-compiler package will have a name such as
         #    gcc_target
-        #    gcc_linux-cos5-64
+        #    gcc_linux-cos6-64
         compiler = '_'.join([compiler, config.variant['target_platform']])
     return compiler
+
+
+def cdt(package_name, config, permit_undefined_jinja=False):
+    """Support configuration of Core Dependency Trees."""
+    cdt_name = 'cos6'
+    arch = config.host_arch or config.arch
+    cdt_arch = 'x86_64' if arch == '64' else 'i686'
+    if config.variant:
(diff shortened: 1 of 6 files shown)


conda_build/jinja_context.py
```diff
@@ -391,11 +391,25 @@ def compiler(language, config, permit_undefined_jinja=False):
 
         # support cross compilers.  A cross-compiler package will have a name such as
         #    gcc_target
-        #    gcc_linux-cos5-64
+        #    gcc_linux-cos6-64
         compiler = '_'.join([compiler, config.variant['target_platform']])
     return compiler
 
 
+def cdt(package_name, config, permit_undefined_jinja=False):
```
msarahan: Can you add some tests here?

conda_build/jinja_context.py
```diff
@@ -391,11 +391,25 @@ def compiler(language, config, permit_undefined_jinja=False):
...
+    """Support configuration of Core Dependency Trees."""
+    cdt_name = 'cos6'
+    arch = config.host_arch or config.arch
+    cdt_arch = 'x86_64' if arch == '64' else 'i686'
+    if config.variant:
+        cdt_name = config.variant.get('cdt_name', cdt_name)
+        cdt_arch = config.variant.get('cdt_arch', cdt_arch)
+    if ' ' in package_name:
+        return (package_name.split(' ')[0] + '-' + cdt_name + '-' + cdt_arch + ' ' + ' '.join(package_name.split(' ')[1:]))
```
msarahan: This is more of a code style suggestion than a requirement: please don't ever return from more than one place.  Store a value in one of many ways, and then have one return line that just returns that value.

It's a nitpicky thing (sorry) but I find that it makes code easier to follow.

conda_build/skeletons/rpm.py
```diff
@@ -0,0 +1,668 @@
+import argparse
```
msarahan: Please include at least one test of this.  Skeletons are where we have the worst coverage, and I don't want to keep up that trend.

mingwandroid:
Thanks @msarahan, before I rework this any further I'd prefer to address the discussion points.

For example, if we want/need to add metadata then that will likely change the `cdt` function significantly.

msarahan:
>The LC_ALL / LANG commit isn't really related, I can make a separate PR if you like?

Nah, it's fine here.

> What extra metadata do we want to include with these packages?

I don't really know.  Anything that helps with reproducibility.

> Is passing centos6 multiple times this way a bit ugly? Should I just have a --distro flag instead?

Yes, I think so.  You may consider having a 'distro' key in the user-wide conda_build_config.yaml, and looping over any entries there.

> Should I allow passing --architecture multiple times?

Up to you.  I think it would be better to support only a single one from the CLI, but allow looping with entries in conda_build_config.yaml.

conda_build/jinja_context.py
mingwandroid: I can't add tests to cdt until we've locked down what a cdt is. This is enough to get me past the speed bump of needing to install cdt packages.

msarahan:
Looking good.  Thanks for adding the test.  Merging.

Issue #2101, opened 2017-06-16
work dir is not removed/restored when --no-build-id is used


gqmelo:
When trying to use ```--no-build-id``` on a conda-forge feedstock I've got the following error: https://circleci.com/gh/gqmelo/llvmdev-feedstock/79
conda is always keeping the old work dir

This minimal example can reproduce the problem:

```bash
mkdir -p /tmp/conda_build_id_bug

echo "
package:
  name: foo
  version: 0.1
source:
  url: https://pypi.io/packages/source/e/exec-wrappers/exec-wrappers-1.0.1.tar.gz
" > /tmp/conda_build_id_bug/meta.yaml

echo "
test -f setup.py
rm setup.py
" > /tmp/conda_build_id_bug/build.sh

conda build --no-build-id /tmp/conda_build_id_bug
conda build --no-build-id /tmp/conda_build_id_bug
```

The second ```conda build --no-build-id /tmp/conda_build_id_bug``` fails on ```test -f setup.py``` because the first build removed it.

gqmelo:
However using a local path as source makes the work dir to be overwritten:

```bash
mkdir -p /tmp/conda_build_id_bug/src
touch /tmp/conda_build_id_bug/src/somefile
echo "
package:
  name: foo
  version: 0.1
source:
  path: /tmp/conda_build_id_bug/src
" > /tmp/conda_build_id_bug/meta.yaml
echo "
test -f somefile
rm somefile
" > /tmp/conda_build_id_bug/build.sh
conda build --no-build-id /tmp/conda_build_id_bug
conda build --no-build-id /tmp/conda_build_id_bug
```

This works fine

neok-m4700:
Hi, is it related to https://github.com/conda/conda-build/issues/2104 by any chance, and if yes could you try the proposed patch ?

gqmelo:
Thanks @neok-m4700. It may be related, I'll try your patch after the weekend

gqmelo:
Well, unfortunately your patch is applied to the master branch and I couldn't apply to conda-build 2.1.x. I think this code changed a lot

gqmelo:
I tested and this PR really fixes my issue: https://github.com/conda/conda-build/pull/2100

@neok-m4700 maybe it fixes your too?

msarahan:
Fixed by #2100, tested in #2111 and #2112.  Thanks @isuruf @gqmelo @neok-m4700

PR #2111, opened 2017-06-20
Add test case for  #2100 and #2101 (2.1.x)

gqmelo:
This test make sure the work dir is removed regardless of the --no-build-id option

I made it sure the test fails before #2100


msarahan:
Thanks @gqmelo - good test adaptation there.

PR #2112, opened 2017-06-20
Add test case for  #2100 and #2101

gqmelo:
This test make sure the work dir is removed regardless of the --no-build-id option

This was ported from #2111. Apparently this issue was already fixed on master, but it's good to have a regression test anyway.


msarahan:
Thanks @gqmelo

PR #2113, opened 2017-06-20
update conda_interface

kalefranz:
An update to conda_interface, which should be merged up to the master branch as well to ensure compatibility with conda 4.4.

### DIFF
--- a/conda_build/conda_interface.py
+++ b/conda_build/conda_interface.py
@@ -20,7 +20,11 @@ except ImportError:
     pass
 
 
-from conda.plan import display_actions, execute_actions, execute_plan, install_actions
+if parse_version(CONDA_VERSION) >= parse_version("4.4"):
+    from conda.exports import display_actions, execute_actions, execute_plan, install_actions
+else:
+    from conda.plan import display_actions, execute_actions, execute_plan, install_actions
+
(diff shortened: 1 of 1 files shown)


msarahan:
This breaks conda-build with conda 4.3 - https://travis-ci.org/conda/conda-build/jobs/245067964#L1101

We can't just shift everything over to exports.  Earlier versions of conda 4.3 don't have the appropriate code defined there.  I looked into this a little bit this morning, and the code is just going to be nasty - some kind of dynamic import or a mess of try/except clauses.

kalefranz:
Does the commit I just pushed not cover it?  I might be missing something.

msarahan:
The commit you did does fix it, but not in an ideal way, IMHO.  Yours is a catch-all that gives you no freedom from the exports interface - if any single one of the imports fails, you are forced to keep  the rest of the imports working.  Allowing each one to try in exports first, then fall back to some prior location would be ideal, rather than as a whole group.  Any ideas on how to accomplish that?

msarahan:
That looks like a much better approach.  Thanks.  Waiting on CI, but will merge if it works.

msarahan:
CI looks about as good as it gets right now.  Merging.

PR #2116, opened 2017-06-23
Convert rewrite [WIP]

mandeep:
There are still some tests that I would like to add to this branch; however, I believe everything in convert.py is ready.

### DIFF
--- a/conda_build/convert.py
+++ b/conda_build/convert.py
@@ -257,10 +257,10 @@ def update_lib_contents(lib_directory, temp_dir, target_platform, file_path):
                 new_lib_file = re.sub('Lib', 'lib/{}' .format(python_version), lib_file)
                 os.renames(lib_file, new_lib_file)
 
+            os.rename('{}/Lib' .format(temp_dir), '{}/lib' .format(temp_dir))
+
         except OSError:
             pass
-
-        os.rename('{}/Lib' .format(temp_dir), '{}/lib' .format(temp_dir))
 
 
 def update_executable_path(file_path, target_platform):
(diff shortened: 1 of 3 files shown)


conda_build/convert.py
```diff
@@ -8,517 +8,701 @@
...
-                if 'python' not in firstline:
-                    if show:
-                        print("Non-Python plaintext file %s" % m.path)
-                    matched = True
+    # split dependencies away from their version numbers since we need the names
+    # in order to evaluate duplication
+    dependency_names = set(dependency.split()[0] for dependency in new_dependencies)
+    index_dependency_names = set(index.split()[0] for index in existing_dependencies)
+
+    repeated_packages = index_dependency_names.intersection(dependency_names)
```
msarahan: I believe you could simplify this by converting index_dep_names and dep_names to sets, calling index_dep_names.update(dep_names), and then go back to a list as necessary.

conda_build/convert.py
```diff
@@ -8,517 +8,701 @@
...
+    that the executables are renamed without their file extensions.
+
+    Positional arguments:
+    executable (str) -- the executable to rename including its file extension
+    """
+    return os.path.splitext(os.path.basename(executable))[0]
+
+
+def rename_executable(directory, executable, target_platform):
+    """Rename an executable file when converting between platforms.
```
msarahan: I'm a little wary on this one.  Files in the bin folder can binaries, right?  Do we have a way of distinguishing them from things that are being used as entry point scripts, or just scripts (both in the setuptools sense of the word).  entry_points are things that point at a particular python function, where scripts are files that get run (also by the same exe's on Windows that entry points use).

msarahan:
This is really good, solid code.  Nice work.  Let's talk a bit more about when executables are considered something that should get renamed, but this is pretty close to ready.

msarahan:
test failures all look unrelated to me.  Nice work.  Merging.  I will cherry-pick this onto 2.1.x.

PR #2119, opened 2017-06-24
Add new functionality to --keep-work-dir flag

AndresGuzman-Ballen:
The --keep-old-work flag was deprecated and ended up having the same functionality as the --dirty flag. However, there are times when a developer would want to have the work directory remain intact even after a successful build/test. Rather than create a new flag, the --keep-work-dir flag will be used for this particular case.

@msarahan

### DIFF
--- a/conda_build/cli/main_build.py
+++ b/conda_build/cli/main_build.py
@@ -108,8 +108,9 @@ different sets of packages."""
     p.add_argument(
         '--keep-old-work',
         action='store_true',
-        dest='dirty',
-        help="Deprecated.  Same as --dirty."
+        dest='keep_old_work',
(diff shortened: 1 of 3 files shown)


AndresGuzman-Ballen:
I don't know why this build is failing in CI...I am not seeing this problem appear on my local environment.

msarahan:
I pushed a fix, I think.  The CLI always provided this option, but it would not be there for tests or other usage using the API.

msarahan:
Now it's fine.  The Travis errors right now are unrelated.  Thanks - merging.

AndresGuzman-Ballen:
Thank you @msarahan !

PR #2120, opened 2017-06-25
skeleton generation: replace legacy syntax for noarch pkgs

nehaljwani:
`noarch_python` is legacy syntax

This patch modifies the skeleton generation for pypi and luarocks pkgs.


msarahan:
Thanks @nehaljwani

PR #2134, opened 2017-06-28
Added another conditional to remove pyc files in bin directory

mandeep:


msarahan:
Looks like exactly what we needed.  Thanks.

Issue #2135, opened 2017-06-29
Simple syntax for library loading test


jakirkham:
Would be nice to have some simplified syntax for loading libraries as part of the test phase of a recipe build. An example of how this is being done now can be seen in this [commit]( https://github.com/conda-forge/arpack-feedstock/pull/7/commits/1ce86bf4006f4298c30b74b475cec44c9a4bee48 ). Absorbing the boiler plate into `conda-build` would make it easier for users to try and apply this nice testing technique.

cc @dougalsutherland @jschueller

jschueller:
something like:
````
test:
   libs:
    - libarpack
```
would test existence then load using ctypes.

jakirkham:
Was going to add that printing out linkages could be good as well. Though there are other ways to get at that info.

mingwandroid:
conda-build 3 has a [new implementation of ldd and otool](https://github.com/conda/conda-build/blob/master/conda_build/os_utils/pyldd.py) that could (and should IMHO) be used to record and check all library linkages at the end of each package build.

I want to add symbol parsing to it too and record that stuff.

jakirkham:
That's cool. Is this or will this be accessible from the CLI?

mingwandroid:
I don't know, I mean, `conda inspect linkages` runs it at present, but only in a checking fashion (and also for when the native tool fails as it does in cross-compilation scenarios).

Personally, I would have something similar to `conda inspect linkages` happen at the end of every package build and it would run either in an advisory or a detecting problems / errors mode (here problems would include things like unlinked to library dependencies, so dependencies that are *likely* not needed). Another problem would be linking to a system library outside of the CDT sysroot packages (those are not live yet though).

jakirkham:
As noted above, a lot of the functionality is there now. We could certainly improve on it, but going to go ahead and close this out.

PR #2136, opened 2017-06-29
Add meta.yaml lib test section

jschueller:
This adds a testing section in meta.yaml for library loading test as in #2135:

```
test:
   libs:
    - libmuparser
```

This would test existence and loading of libmuparser shared libray.

### DIFF
--- a/conda_build/create_test.py
+++ b/conda_build/create_test.py
@@ -12,6 +12,7 @@ import sys
 
 from conda_build.utils import copy_into, get_ext_files, on_win, ensure_list, rm_rf
 from conda_build import source
+from conda_build.environ import get_shlib_ext
 
 
 def create_files(m):
@@ -85,6 +86,26 @@ def create_shell_files(m):
                 f.write('\n')
                 if sys.platform == 'win32':
                     f.write("if errorlevel 1 exit 1\n")
+                has_tests = True
+
+    libs = ensure_list(m.get_value('test/load_libs', []))
+    if libs:
+        if sys.platform.startswith('win'):
+            lib_path = os.path.join(m.config.test_prefix, 'Library', 'bin')
+        else:
+            lib_path = os.path.join(m.config.test_prefix, 'lib')
+
+        with open(join(m.config.test_dir, name), 'a') as f:
+            f.write('\n\n')
+            for lib in libs:
+                lib_loc = os.path.join(lib_path, lib + get_shlib_ext())
+                # test existence
+                if sys.platform.startswith('win'):
+                    f.write('if not exist ' + lib_loc + ' exit 1\n')
+                else:
+                    f.write('test -f ' + lib_loc + '\n')
+                # test loading
+                f.write('python -c "import ctypes; ctypes.cdll[\'' + lib_loc + '\']" || exit 1\n')
                 has_tests = True
 
     return has_tests
--- a/conda_build/metadata.py
+++ b/conda_build/metadata.py
@@ -327,6 +327,7 @@ default_structs = {
     'test/source_files': list,
     'test/commands': list,
     'test/imports': list,
+    'test/load_libs' : list,
     'package/version': text_type,
     'build/string': text_type,
     'build/pin_depends': text_type,


jschueller:
cc @jakirkham @dougalsutherland

mingwandroid:
Why not use a compiler instead of Python? `lib` is a poor name imho. Something like `check_shared_libs_load` would be better. Also why not extend it to optionally check that specific symbols are present?

jschueller:
Ok for the name, thanks!
How would you use the compiler, wouldnt it require much more options ? How to add symbol options ?

jakirkham:
We don't what compiler is there. Also would that work on Windows. I thought the compiler was not activate in `test` or has this changed?

Being more explicit with the name is good, but would be good if we could still keep it short as this is something that would ideally be used a lot. What about `load_libs`? Hopefully it is known they are shared only

conda_build/create_test.py
```diff
@@ -87,6 +88,28 @@ def create_shell_files(m):
...
+            f.write('\n\n')
+            for lib in libs:
+                lib_loc = os.path.join(lib_path, lib + get_shlib_ext())
+                # test existence
+                if sys.platform.startswith('win'):
+                    f.write('if not exist ' + lib_loc + ' exit 1\n')
+                else:
+                    f.write('test -f ' + lib_loc + '\n')
+                # test loading
+                f.write('python -c "import ctypes; ctypes.cdll[\'' + lib_loc + '\']"\n')
```
dougalsutherland: Is Python guaranteed to exist here, or does it need to be added as a test requirement somehow?

jschueller: from testing, the same python executable is run whether or not I require 'python' in the test section.

jschueller:
Ok, let's keep it short: load_libs.

conda_build/create_test.py
msarahan: conda-build adds a python test req for you if you have any import tests or a run_test.py file.

jschueller: I have only 'commands' section, the result is unchanged: miniconda/bin/python is used

jakirkham: One thing to think about here, if this is being used to test a library on Windows, we will need to constrain that `python` version to match the build version. I'm not sure if there is a good way to do that. Any suggestions @msarahan?

jakirkham: > miniconda/bin/python is used

👍 Was hoping this would be the case.

msarahan: https://github.com/conda/conda-build/blob/master/conda_build/build.py#L1283-L1325

jakirkham: So it is adding `python`. Hmm...so I guess my Windows question still stands.

mingwandroid: You could end up trying to load VS2008 libraries using VS2015 Python. To make a lib you already needed a compiler. Imagine a conda-build written in something other than Python, let's not assume Python is a given in all situations.

You could write some lib testing programs that are bundled with `conda-build` that take as input the shared libraries to attempt to load.

dougalsutherland: @mingwandroid This is why we need `python` in `test/require` (and to make sure it uses that one), right? Then, assuming the library handles the VS features properly, it'll be loading with the right VS.

Using a compiler would mean installing the compiler in the test env, which then might end up bringing in libraries you accidentally linked against but didn't include in the test env...which would have masked exactly the problem that https://github.com/conda-forge/arpack-feedstock/pull/7 found.

mingwandroid: You can use a compiler or write a very simple program.

mingwandroid: We need to stop being so Python-centric, it's not the right tool for every job.

jakirkham: It takes one line to do this with Python. Seems like a good use of Python to me. 😕

Admittedly there is this issue of loading the library with the wrong runtime on Windows. However wouldn't this be an issue even if Python weren't used? Whatever tool we need to use will still be loading the library as that is the goal. If it has the wrong runtime, won't this still be a problem? How do you envision avoiding this conundrum?

As to the compiler, it is my understanding we can't use the compiler on Windows in the test phase correctly. So this is a different issue. If we could use the compiler on Windows, that would be great and would help solve other problems. Then there is the question of what compiler we are using on the other platforms. TBH this sounds more challenging to me than just getting Python to work here. Though gains here would definitely be valuable in other problems, so not opposed to it.

mingwandroid: I proposed another option if we have compiler troubles, namely building some simple executables and bundling them with conda-build. I think installing python is massive overkill and the wrong tool here.

mingwandroid: I am very concerned that the entire Anaconda Distribution and conda-forge stacks be bootstrappable. This naturally leads to cyclic dependencies not being allowed. The C libraries and tools necessary to bootstrap `Python` (on Linux) are:
`m4 perl autoconf automake pkg-config zlib xz bzip2 tk libffi openssl ncurses readline sqlite yaml`

Making this new feature depend on Python precludes using it in any of those libraries. Python is being used to implement this because it is quick and easy and not because it is the correct way to do it.

We should compile simple test helper executables that takes as arguments e.g.: `libname1,symbol1,symbol2` `libname2,symbol3,symbol4`

mingwandroid: Using the wrong runtime would not be an issue. Conda-build knows what VC runtime was in-play and should run the correct executable.

msarahan:
@mingwandroid could you please clarify how you'd like to see this done?  If compilers are necessary, we need to tweak recipes to use the {{compiler()}} syntax in conda-build 3, but it's not a showstopper.

msarahan:
Marking this for the SciPy sprints next week.  It seems like this would benefit from in-person discussions, and @mingwandroid will be in Austin.  If any of you others will not be at the sprints, we can arrange a video call time.

jschueller:
Why not. I will only attend euroscipy.

Also, I can add that often the name is libxxx.so|dylib and the native windows name is xxx.dll ; without the lib prefix. But that can be false, for example when the lib is compiled from a MinGW toolchain, but exceptions to this exception also exist.

PR #2142, opened 2017-06-30
Add SSL_CERT_FILE and REQUESTS_CA_BUNDLE to build environment

aburgm:
Just like the HTTP proxies, these are part of the system configuration that the build might rely on if it uses requests to fetch an HTTPS resource. An example would be building sphinx-based documentation with the intersphinx extension enabled attempting to download the inventory from `docs.python.org`.

### DIFF
--- a/conda_build/environ.py
+++ b/conda_build/environ.py
@@ -289,6 +289,7 @@ def conda_build_vars(prefix, config):
         'SRC_DIR': src_dir,
         'HTTPS_PROXY': os.getenv('HTTPS_PROXY', ''),
         'HTTP_PROXY': os.getenv('HTTP_PROXY', ''),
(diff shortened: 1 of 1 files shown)


msarahan:
Thanks!  You put this PR up on the master branch, which is the 3.0 series of releases right now.  Do you also need it for the 2.1.x series?

aburgm:
Yes, if you agree this is the right thing to do it would be great if it could be backported to 2.1.x as well!

msarahan:
Sure.  I'll cherry-pick it back there.  I just did a release today, though, so it might be a little while before this makes it into a release.

aburgm:
Thanks! Unfortunate timing on my part, but no problem :)

msarahan:
cherry-picked at https://github.com/conda/conda-build/commit/1ea628795536dce848d26adaff6db02585afd483 - thanks again.

Issue #2140, opened 2017-06-30
Improve reproducibility of artifacts


msarahan:
@bollwyvl raised this privately.  He would like conda packages to be completely verifiable.  There's some good guidelines on this at https://reproducible-builds.org/

@bollwyvl has some preliminary work implemented in a jupyter notebook that I'm happy to share if anyone would like to see.

bollwyvl:
Thanks, @msarahan! Also thanks for not just posting that notebook... it's of course broken 😈 .

In summary:

My early work has a lot of shortcomings:
- only tried with `conda-build` 2.1.6
- only tried with pure python packages: a toy example, and `conda-build` 2.1.16 itself
- only tried on OSX

Observations in no particular order: 
- [x] `info/index.json` wasn't being sorted (has been fixed on master) 🎆 
- [ ] tarballs preserve timestamp and owner
  - easy enough to fix, either by post-processing or something
- [ ] `info/files` is not [sorted](https://github.com/conda/conda-build/blob/56e8ceba1d016a23ba419ad93a6d77b017c6ca14/conda_build/build.py#L324)
- [ ] `info/recipe/meta.yaml` has non-deterministic ordering of sets (mainly package spec listings)
- [ ] `versioneer.py` (and maybe other things that hack how pip/setuptools do their thing) apparently don't respect `PYTHONHASHSEED` and `DETERMINISTIC_BUILD`, so they end up with 1-2 bytes of header differences in `pyc` files
- [ ]  `info/about.json`
  - [ ] there's no easy way to extract`root_pkgs` directly to reproduce the build environment
  - [ ] `env_vars` includes username and other transient path information

In working on this, I have found [`diffoscope`](https://diffoscope.org/) super valuable, and am trying to get it built for conda, but don't have any success there yet!

mingwandroid:
Good to see someone paying attention to this stuff!

bollwyvl:
An update: diffoscope PR on conda-forge:
- https://github.com/conda-forge/staged-recipes/pull/3282
  - currently blocked by `libarchive-c` https://github.com/conda-forge/staged-recipes/pull/3281

bollwyvl:
@mandeep you are a reproducibility-enhancing MACHINE. 🎸 on.

Been a bit heads-down, but just updated conda-forge/staged-recipes#3281 and conda-forge/staged-recipes#3282 with some naming issues, extra deps, and typos on my part. Also started sneaking in:
```bash
[...]

PR #2143, opened 2017-07-01
Sort files before they're written to the files file

mandeep:
In reference sorted files in issue #2140 

For build.py I merely added a sorted() call around the files argument. Even though the if statements are not calling paths in alphabetical order, the sorted files argument seems to be enough.

For convert.py I decided to append all filepaths to a list then write that sorted list into the files file.

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -330,7 +330,7 @@ def write_info_files_file(m, files):
         if m.get_value('build/noarch_python'):
             fo.write('\n')
         elif m.noarch == 'python':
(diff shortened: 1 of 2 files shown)


msarahan:
Thanks - is paths.json sorted?

mandeep:
Yep, for the paths.json the files already have the sorted() call. I checked a package on Windows and OSX to be sure and all looks okay.

msarahan:
Great.  Good work.  Thanks for being proactive on this.

PR #2146, opened 2017-07-02
Use glob2 to resolve recursive pattern /**/

MatthieuDartiailh:
This adds glob2 as a dependency to the project to support patterns such as:
```
**/my_project/**/.txt
```
in all entries using expand_globs. 

Furthermore the output of expand_globs is modified to replace os.path.sep by / as conda manipulate paths containing only / (this is for example the case for ignore_prefix_files, with the modification the filtering fails on Windows). I am not sure this is the proper fix.

If you are favorable to such changes, I will add tests and update the docs on conda-docs. One open question is that glob2 is not available on conda and so it needs to be added as otherwise the conda recipe will not valid.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,7 +56,7 @@ install:
         conda info;
     fi
   - conda install -q anaconda-client requests filelock contextlib2 jinja2 patchelf python=$TRAVIS_PYTHON_VERSION
-  - conda install -q pyflakes conda-verify beautifulsoup4 chardet pycrypto
+  - conda install -q pyflakes conda-verify beautifulsoup4 chardet pycrypto glob2
   - pip install pkginfo
   - if [[ "$FLAKE8" == "true" ]]; then
       conda install -q flake8;
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -43,7 +43,7 @@ install:
   - python -c "import sys; print(sys.executable)"
   - python -c "import sys; print(sys.prefix)"
   - conda update -q --all
-  - conda install -q pip pytest pytest-cov jinja2 patch flake8 mock requests contextlib2 chardet
+  - conda install -q pip pytest pytest-cov jinja2 patch flake8 mock requests contextlib2 chardet glob2
   - conda install -q pyflakes pycrypto posix m2-git anaconda-client numpy conda-verify beautifulsoup4
   - conda install -c conda-forge -q perl
   # this is to ensure dependencies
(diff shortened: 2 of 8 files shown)


msarahan:
Excellent, thanks.  I am definitely in favor of this change - please add tests and docs.  Please add this extra requirement to conda.recipe/meta.yaml, also.

I will take care of getting glob2 available as a conda package.

MatthieuDartiailh:
Thanks will do ! Does the change about converting the os.path.sep to / also make sense to you ?

msarahan:
Yes, if you're conforming to what glob2 does, it makes sense.  Lots of code coerces to unix-style pathsep.  With several of these settings, we even complain if people use windows slashes: https://github.com/conda/conda-build/blob/master/conda_build/metadata.py#L1172

As long as we test it on Unix and Windows, we're good here.

MatthieuDartiailh:
Actually, when matching enaml/core/*/*.py, even plain glob will convert the / close to a \* to \\\\ on Windows  but leave the the / in between enaml and core and as conda reason about prefix files only in terms of / it can break the identification of files to ignore on windows. I will try to get the tests and docs in tonight if possible. Thanks for your responsivity.

msarahan:
This looks ready - thanks.  I will merge it as soon as we have the glob2 package available in defaults.  I have submitted a recipe, and it needs to get reviewed and built.

MatthieuDartiailh:
Thanks let me know if anything goes wrong. By the way do you have any idea when a release including this will be published ?

msarahan:
I will try to tag a release quickly.  I don't know how long it will take us to get the glob2 package out, but that is really the rate limiting step here.

msarahan:
glob2 is now available on defaults.  I have edited the CI config files to install glob2 from conda.  Assuming things pass, I'll merge this.  A release is happening probably today or tomorrow.

MatthieuDartiailh:
I tried to give a look at the failures but they seem quite cryptic to me. If you something that I need to change @msarahan, let me know.

msarahan:
Don't worry about it. We have several other issues that are probably
responsible here.  I'll take one more look later today and most likely
merge and tag.

msarahan:
Thanks @MatthieuDartiailh for helping carry this forward.

MatthieuDartiailh:
You are welcome, the fix was easy and will allow me to package easily enaml cache files, so this is also a big win for me.

MatthieuDartiailh:
Any idea about when the release will appear in anaconda default channel ?

MatthieuDartiailh:
Or will this be backported to the 2.1 series ?

msarahan:
The 3.0 release should appear in defaults today.  Ironing out one last bug now.

PR #2151, opened 2017-07-04
pyldd: Directories are not codefiles.

mingwandroid:
Also use os.path.realpath instead of os.readlink so that we follow
links to links.


mingwandroid:
Closing this. The reversions weren't meant to make it into the PR.

PR #2153, opened 2017-07-05
cran-skeleton: allow certain blank keys

nehaljwani:
The 'Writing R Extensions' manual doesn't mandate that if there are no
suggestions for a package, then the field should not be present. It can
still be present, with a blank value, for example:
https://cran.r-project.org/web/packages/corpcor/DESCRIPTION

This patch adds the functionality to not throw an error during parsing
if the value of certain fields is empty. It should just ignore that
field and continue.

Fixes https://github.com/conda/conda/issues/5624

### DIFF
--- a/conda_build/skeletons/cran.py
+++ b/conda_build/skeletons/cran.py
@@ -150,6 +150,9 @@ CRAN_KEYS = [
     'Maintainer',
 ]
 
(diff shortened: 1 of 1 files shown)


msarahan:
Seems fine to me.  I guess people will just need to add other fields that are allowed to be blank as they bump into them.

PR #2149, opened 2017-07-03
Sort output metadata prior to writing to meta.yaml

mandeep:
In relation to #2140:

The field values in the metadata are sorted alphabetically prior to outputting the meta.yaml file in the info/recipe directory. The build/script and test/commands values are omitted.

PR #2154, opened 2017-07-05
Moved try block in copy_recipe further up in order to catch values that are lists

mandeep:
@mingwandroid brought up the issue that PR #2149 may result in an AttributeError traceback when a section contains a list as its value instead of a dictionary. This PR will fix that issue.

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -214,12 +214,13 @@ def copy_recipe(m):
         # each metadata field has a dict as a value and each value may contain
         # a list of strings that needs to be sorted alphabetically
         for field, value in output_metadata.meta.items():
-            for key in value.keys():
-                if '{}/{}' .format(field, key) not in ('build/script', 'test/commands'):
-                    try:
+            try:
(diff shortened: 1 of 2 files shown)


msarahan:
I'm not sure this is the right fix.  Let's take a step back and figure out exactly what data types we expect.  From there, we can write better code to either check type and dispatch accordingly, or write better comments for what a given try/except block is handling, and also what it is ignoring.

mandeep:
I'm with you. Do you want me to push a commit to this branch that removes the code since its breaking things?

msarahan:
eh, just use this branch to do the right thing.  It's fine that master is a bit broken right now.  Ray has reverted it locally anyway, and hopefully you can figure out the right, clean way to do this quickly.

msarahan:
@mingwandroid any opinions on this?  I think it should work better than the code that broke you, but I'd also appreciate if you could tell us which recipe triggered the bug you saw.

msarahan:
Thanks @mandeep

PR #2156, opened 2017-07-05
Feature/rework cpan

jerowe:
@msarahan , I don't think that this is really ready for prime time, but I wanted to get it out there to motivate me to work on it more.

I have been using it for my own purposes for several months now.


msarahan:
Excellent, thanks!  I'll take a good look soon - tied up with conferences, but at the very latest, I'll review this during the SciPy sprints next week.

jerowe:
I didn't know this was a thing! Bah one of these days I need to come to Texas.

msarahan:
Yes, you should.  The SciPy conference is especially good.  Hope to see you maybe next year.

jerowe:
No rush. Like I said - it could use some cleanup still. 

The only major design change that I have made so far is to change what the definition of a core module is. Technically, anything that is in the perl 'corelist' is a module.

BUT, sometimes core modules are updated. So, in terms of building recipes. our definition should be whether or not it contains a release on cpan.

If you map it to urls core modules with no further development exist in the `fastapi.cpan.api/module/CoreModule`, while anything that has a release on cpan has an entry in `fastapi.cpan.api/release/CoreModuleWithRelease`.

I don't know if I explained that clearly...

jerowe:
My youngest is still kind of young for me to do a lot of traveling. When she is old enough to join my older one in camps and has moved past the 'I am a force of destruction' stage  I have big plans to get out to more events like this. For now I will just bug you through the internet. ;-)

jerowe:
I don't know what kind of black magic you guys did, but the skeleton feature is waaay faster than it used to be!

msarahan:
Glad to hear it. I don't think we've specifically optimized it, but it
might benefit from work we've done elsewhere in conda-build, or it might be
improvements in conda, which has also gotten faster.

jerowe:
Hi @msaharan. The gcc discussion just reminded me I should check in on this. How is it going? How was the conference?

msarahan:
Hi @jerowe - conference was good.  Great conda-forge participation.  Those tags were my rough estimates of what anyone should expect to encounter if they were to work on this PR - how much they would need to work, and how much they would need to know about conda-build (and perl) in order to do any work at all.

As for this PR - where is it in your opinion?  Do you want to do more cleanup on it?  Do you want me to review it as is and perhaps merge it?

jerowe:
I think this could be a good first pass. It could use some more cleaning up, but the overall structure of the code is mostly the same and maybe that would make it easier to review?

Did my definition of what is or isn't a core module make sense? That is the biggest change. If you agree with that I think it is ok to review.

msarahan:
I think your changes to what is core are fine.  Perhaps adding a test of what is core and what is not might be helpful to make your behavior more explicit?  For cleanup, I request removing the stuff you've commented.  Print lines, unused code, etc.  If people need to see how it used to be done, they can look at the git history.  Commented things that aren't explanations make code harder to read, IMHO.

jerowe:
Those should definetely be removed. I like to keep them around until I merge because I am always bouncing between projects, but I will get them removed for testing.

The test for what is core and what is not might be difficult if something suddenly becomes core, but I could have a test for things I know are not core.

msarahan:
Merging, since cpan is broken now without this - tested locally and saw the Travis tests pass OK on cpan.  Thanks @jerowe

jerowe:
I'm glad this helped, and I'm sorry I didn't get to the fixes myself! Thanks for cleaning up. ;-)

conda_build/skeletons/cpan.py
```diff
@@ -571,32 +620,74 @@ def dist_for_module(cpan_url, module, perl_version, config):
...
+        mod_dict['distribution'] = 'perl'
+
+    return mod_dict
+
+
+@memoized
+def metacpan_api_is_core_version(cpan_url, module):
+
+    url = '{0}/release/{1}'.format(cpan_url, module)
+    url = url.replace("::", "-")
```
timsnyder: `metacpan_api_is_core_version()` only works if the module->distribution mapping is a simple `s/::/-/g` but that doesn't hold true for all modules, namely `YAML::XS` which kept it's old distribution name of `YAML-LibYAML` (see https://fastapi.metacpan.org/v1/module/YAML::XS).

I'm going to open an issue and submit a PR for this.

PR #2158, opened 2017-07-05
Improve error message when git describe fails

dhirschfeld:
Clarify suggestion to use annotated tags


msarahan:
Much more clear.  Thanks!

dhirschfeld:
As discussed on gitter *Are you using annotated tags?* is ambiguous and could be interpreted to mean  that the reason git describe failed was *because* you were using annotated tags which is the opposite of the intended meaning.

PR #2160, opened 2017-07-06
fix inspect_linkages call in os_utils.ldd.get_linkages

dougalsutherland:
The call to `conda_build.os_utils.pyldd.inspect_linkages` in `conda_build.os_utils.ldd.get_linkages` passes `sysroot` as a second positional argument, but the second positional argument to that function [is actually `resolve_filenames`](https://github.com/conda/conda-build/blob/667ef45bfd9e77d365403279f416eb5eb0c7fad5/conda_build/os_utils/pyldd.py#L842).


msarahan:
Nice fix.  Keyword args for the win.

PR #2180, opened 2017-07-10
fix priority of setup.cfg over setup.py

### LINKED ISSUE #2164: setup.cfg vs setup.py precedence is backwards in reading setup.py data
Discussion at:
https://github.com/conda/conda-build/commit/be5cc63cfa2f421d4ebf14e3634c21ea7939d486

msarahan:
fixes #2164 - thanks @mbargull


mbargull:
LGTM, I guess. Though `test_load_setup_py_data_from_setup_cfg` is still always skipped due to the older `setuptools` package in `defaults`. Maybe a local test with a more recent version of `setuptools`, e.g., from `conda-forge`, might be a good thing.
A backport to `2.1.x` should be very similar.

PR #2205, opened 2017-07-16
Rewrite skeleton template and match conda-forge standard

mwcraig:
This replaces the string template with what is essentially a nested ordered dict (`ruamel_yaml.comments.CommentedMap`) and writes the recipe out in conda-forge order. 

This needs a bit of cleanup still (feel free to do that, I won't be able to come back around to it until later on Sunday).

A few thoughts that I won't remember in the morning:

+ It would be nice to put the "standard order" variables (`EXPECTED_SECTION_ORDER`, `REQUIREMENTS_ORDER`, maybe `ABOUT_ORDER`) some place "standard" and have `conda skeleton` and `conda-smithy` import from that place rather than trying to keep the values synced across packages.
+ With a little refactoring I think the recipe generation conda build pulled out into a function that can take in either the dictionary `d` generated in `skeletonize` or metadata pulled from an existing recipe. With that change, with same function could be used to "conda forge-ify" existing recipes.

### DIFF
--- a/conda_build/skeletons/pypi.py
+++ b/conda_build/skeletons/pypi.py
@@ -35,6 +35,7 @@ from conda_build.environ import create_env
 from conda_build.config import Config
 from conda_build.metadata import MetaData
 from conda_build.license_family import allowed_license_families, guess_license_family
+from conda_build.render import FIELDS as EXPECTED_SECTION_ORDER
 
 pypi_example = """
 Examples:
@@ -52,11 +53,8 @@ Use the --pypi-url flag to point to a PyPI mirror url:
     conda skeleton pypi --pypi-url <mirror-url> package_name
 """
 
-# Definitions of EXPECTED_SECTION_ORDER and REQUIREMENTS_ORDER below are from
+# Definition of REQUIREMENTS_ORDER below are from
 # https://github.com/conda-forge/conda-smithy/blob/master/conda_smithy/lint_recipe.py#L16
-EXPECTED_SECTION_ORDER = ['package', 'source', 'build', 'requirements',
-                          'test', 'app', 'about', 'extra']
-
 REQUIREMENTS_ORDER = ['build', 'run']
 
 # Definition of ABOUT_ORDER reflects current practice
@@ -64,7 +62,7 @@ ABOUT_ORDER = ['home', 'license', 'license_family', 'license_file', 'summary',
                'description', 'doc_url', 'dev_url']
 
 # This may be overkill, but some day sha256 won't be enough. Might as well be
-# ready...list this in order of decreasing preference.
+# ready...list these in order of decreasing preference.
 POSSIBLE_DIGESTS = ['sha256', 'md5']
 
 PYPI_META_HEADER = """{{% set name = "{packagename}" %}}
@@ -547,7 +545,7 @@ def get_download_data(pypi_data, package, version, is_url, all_urls, noprompt, m
         else:
             # That didn't work, even though as of 7/17/2017 some packages
             # have a 'digests' entry.
-            # As a last-ditch effort, try for this entry
+            # As a last-ditch effort, try for the md5_digest entry
             try:
                 digest = ('md5', url['md5_digest'])
             except KeyError:
--- a/conda_build/skeletons/pypi.py
+++ b/conda_build/skeletons/pypi.py
@@ -19,6 +19,7 @@ from tempfile import mkdtemp
 import pkginfo
 import requests
 from requests.packages.urllib3.util.url import parse_url
+from six.moves.urllib.parse import urljoin
 import yaml
 import ruamel_yaml
 
@@ -176,7 +177,7 @@ def skeletonize(packages, output_dir=".", version=None, recursive=False,
         if is_url:
             package_pypi_url = ''
         else:
-            package_pypi_url = '/'.join([pypi_url, package, 'json'])
+            package_pypi_url = urljoin(pypi_url, '/'.join((package, 'json')))
 
         if not is_url:
             dir_path = join(output_dir, package.lower())
@@ -237,8 +238,8 @@ def skeletonize(packages, output_dir=".", version=None, recursive=False,
         # from PyPI. summary will be pulled from package information in
         # get_package_metadata or a default value set if it turns out that
         # data['summary'] is empty.
-        d['summary'] = data['summary']
-        d['description'] = data['description']
+        d['summary'] = data.get('summary', '')
+        d['description'] = data.get('description', '')
         get_package_metadata(package, d, data, output_dir, python_version,
                              all_extras, recursive, created_recipes, noarch_python,
                              noprompt, packages, extra_specs, config=config,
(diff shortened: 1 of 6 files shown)


msarahan:
Really nice work here, Matt.  Thanks for working on it.  @ocefpaf was asking me about a few of the other aspects, namely sha256.  I looked into it a little bit.  The new pypi.io JSON api will let us do it reasonably easily, I think.

Check this out:

```
import requests
import json
d = json.loads(requests.get('https://pypi.io/pypi/click/json').content)
# show current release version
d['info']['version']
# show files available for given release
d['releases']['6.7']
```

As things stand, we only support sdists.  We could filter the list of files to maintain this behavior, or we could try to support wheels.

I need to get back to other things right now.  Would you like to push further on this, or should we merge as is and leave the PyPI JSON api for another day?

conda_build/skeletons/pypi.py
```diff
@@ -437,18 +418,15 @@ def skeletonize(packages, output_dir=".", version=None, recursive=False,
...
-                                       width=200)
-
-            rendered_recipe += content
+                else:
+                    rendered_recipe += ruamel_yaml.dump({key: ordered_recipe[key]},
+                                                Dumper=ruamel_yaml.RoundTripDumper,
+                                                default_flow_style=False,
+                                                width=200)
+                    rendered_recipe += '\n'
+            # make sure that recipe ends with one newline, by god.
```
mwcraig: 🤣

mwcraig:
Given the loud comment at the top of https://wiki.python.org/moin/PyPIXmlRpc, switching to JSON seems like the way to go. It mostly looks straightforward, just one question.  [The code](https://github.com/conda/conda-build/pull/2205/files#diff-c61365ff02198f8397cfa6e399b6ef8bR302) tries to check for hidden vs non-hidden versions, and uses the latest non-hidden version. The JSON output (and the [docs for PyPA](https://warehouse.readthedocs.io/api-reference/xml-rpc/#changes-to-legacy-api)) don't distinguish between hidden or not.

Was going to ask how to try to handle that, but think I answered the question is asking it. We can't get the hideen status so we'll ignore it.

mwcraig:
I'll do as much as I can before noon Monday, then we'll need to call it done (on my end).

Do you need me to fix up the tests?

mwcraig:
One other minor spacing issue.

As it currently stands in the PR, indentation in the output of skeleton looks like this:

```
requirements:
  build:
  - python
  - setuptools
  - astropy >=1.0
```

so the lists aren't indented. I can add a couple of args to `dump` to get the output to look like this instead (list indented nicely, but build indented by 4 not 2):

```
requirements:
    build:
      - python
      - setuptools
      - astropy >=1.0
```

**Preferences?**

For future me, the extra arguments are `indent=4, block_seq_indent=2`, relevant documentation at http://yaml.readthedocs.io/en/latest/detail.html#indentation-of-block-sequences

msarahan:
Please keep all indents equal. I'm fine with all 2 or all 4, whatever
matches most recipes we already have.

mandeep:
Regarding the PyPI API, it may be simpler to use requests' json() method on the response:

```python
import requests

r = requests.get('https://pypi.io/pypi/click/json').json()

r['info']['version']
r['releases']['6.7']
```

This way json doesn't have to be imported, and the code will work both on Python 2 and 3.

mingwandroid:
I have a strong personal preference for 2 spaces instead of 4, FWIW.

mwcraig:
The issue isn't 4 spaces vs 2...if it were, I'd go with 2 because that is what conda-forge currently does.

The issue is that the indentation produced by `ruamel.yaml` is odd, effectively producing either a mixx of 2 and 0 spaces or a mix of 4 and 2 spaces.  Either of those is 👎 .

The first case below (like above, but spaces more explicitly shown) has `dict` keys (like `build`) are indented 2 spaces, but for list items, like the requirements under `build`, the items are indented two spaces but the dashes are not.

In the second case I add the commands to indent the list dashes two space (good) but then the `dict` keys (like `build`) are indented 4 spaces (bad).

Pointers on where to look for fixes are welcome :). 

First case:

```
requirements:
⊔⊔build:
⊔⊔-⊔python
⊔⊔-⊔setuptools
⊔⊔-⊔astropy⊔>=1.0
```

Second case:

```
requirements:
⊔⊔⊔⊔build:
⊔⊔⊔⊔⊔⊔-⊔python
⊔⊔⊔⊔⊔⊔-⊔setuptools
⊔⊔⊔⊔⊔⊔-⊔astropy⊔>=1.0
```

mwcraig:
about half way through the json refactor, please don't merge until pushed (I've passed my self-imposed noon deadline)

msarahan:
I don't think we use ruamel yaml in conda build. We can, if it makes life
easier. I would really like to eliminate our fork (ruamel_yaml) which was
done to avoid the complexity of dealing with namespace packages.  I think
there are some examples on conda forge of how to properly deal with
namespace packages and perhaps the conda forge ruamel.yaml package is
already ok and we should switch to it for conda itself.

Whatever. All this is to say that aside from our non-ideal fork, Anton, the
author of ruamel.yaml, is very helpful and may have ideas for us here.

mwcraig:
@msarahan -- do you want me to keep the function signatures for `get_download_data` and `get_package_metadata` unchanged?

msarahan:
Those are internals. Change them however you want. Until we have a standard
interface that every skeleton should implement, it's the wild West.

mwcraig:
Added a few things:

+ JSON API to PyPI
+ Try really hard to calculate a `sha256`.
+ Super ugly hack to indent lists by two spaces. Didn't have a chance to check with Anton or post to SO.

I'm about out of time on this, but could change the yaml rendering to use some of the [dumpers in `render.py`](/Users/mcraig/miniconda-build/conda-bld/src_cache) with plain yaml instead of ruamel.yaml.

mwcraig:
One other question...in this PR I add a variable for the order of the top-level keywords in a recipe. At https://github.com/conda/conda-build/blob/master/conda_build/render.py#L587 there is another, which is identical to the one in skeleton except that it does not include 'app'.

Can I add `app` to that and just import from render?

msarahan:
@mwcraig yes, I believe adding "app" and importing from render is the right way.

mwcraig:
@msarahan -- I've done what I can for now. I got close to having nice looking output that preserved order using `render.output_yaml`, but didn't have time to figure out how to make it preserve order of the keys in the `about` section. 

I think preserving that order is important for the human reviewers at staged recipes, so stuck with the hack I have now. 

I can push the changes I tried along those lines if you have time, or am fine with merging as is. Would like to say I'll have some time later this week but doubt I actually will.

Can ifx tests if you want me to.

msarahan:
I'm fixing tests as we speak.  Don't worry about it.  I'll pull in your changes and I think we'll be ready to merge.

msarahan:
Thanks @mwcraig - excellent sprint contribution here.  Thanks for sticking with it.

mwcraig:
Thanks for your help with it! Looking forward to being able to more easily move a few dozen packages over to conda-forge.

PR #2209, opened 2017-07-16
remove existing entry points (esp. links) before writing new ones

msarahan:
For conda especially, what was happening was that writing the new entry points was following the symlink through to the root environment, causing the root env conda to break.


ericdill:
Thanks @msarahan !

PR #2218, opened 2017-07-19
Fixed issue where test dependencies were being checked even when notest was passed

### LINKED ISSUE #2217: Building a package with build and test requirement, the test requirement is used
We have seperate build and test stages, so we build once with `--no-test` and another time with `-t`.

A package that has an asymetric build and test requirement will favor the test requirement in the build stage even when `--no-test` is specified...

Example:

lets suppose cmake is available in version 3.5, so the build requirement is satisfied, but not the test requirement. We are building with `--no-test`.  So in any case I'd expect CB to build,

```
package:
  name: "teskpkg"
  version: "1.0.0"
  
requiremnts:
  build:
    - cmake >= 3.5
     
test:
  requires:
    - cmake >= 3.7
```

When cmake 3.5 is available and we build with `conda build . --no-test` we get an error that the requriement cmake >= 3.7 cannot be not satisfied.

I noticed that in contrast to CB2 the test requirements are evaluated even before build. That might be the reason for htis. I dont specially like this because it breaks isolation between build and test stage, which imho should be isolated to allow for multi-stage building

mandeep:
fixes #2217 

I added the `notest` keyword argument to the `build` function which it receives from `build_tree`. I was able to duplicate the error in #2217 with a test that I wrote and this fix passes that test.


tests/test_api_build.py
```diff
@@ -1094,3 +1094,8 @@ def test_indirect_numpy_dependency(testing_metadata):
     testing_metadata.meta['requirements']['build'] = ['arrow-cpp 0.5.*']
     testing_metadata.config.channel_urls = ['conda-forge']
     api.build(testing_metadata, numpy=1.13)
+
+
+def test_dependencies_with_notest(testing_workdir, testing_config):
+    recipe = os.path.join(metadata_dir, 'test_dependencies')
+    api.build(recipe, config=testing_config, notest=True)
```
msarahan: To make this test a little more rigorous, please add a small testcase like this:

```
with pytest.raises(DependencyNeedsBuildingError):
    api.build(recipe, config=testing_config, notest=False)
```

msarahan:
Nice, thanks.

PR #2221, opened 2017-07-20
try to avoid downloading until it's truly necessary

### LINKED ISSUE #2216: Download of source packages is done before skip-existing
This behaviour changed between CB2 and 3: in CB2 the source download is done after deciding whether we can skip the build while in CB3 not.. That wastes a lot of space and computing time. Could that be changed back?

msarahan:
fixes #2216


msarahan:
another heisenbug here.  The disable_pip test passes fine locally, but dies on CI.  I'm starting to suspect a lack of test isolation on pytest, somehow.  Merging this for now.

PR #2226, opened 2017-07-21
fix test python not using correct metadata config object

msarahan:
@mingwandroid - this should fix you, but I'd also appreciate if you could help me understand exactly what you were doing, so I can write a test.


mingwandroid:
I was doing: `conda-build -t $CONDA_PREFIX/pkgs/python-*.tar.bz2 --debug`

Note, this is in the pkg cache so you may need to copy it elsewhere to prevent `clean_pkg_cache` from removing it (though I need to figure out what to do about that wrt testing..)

PR #2230, opened 2017-07-22
Do not add setuptools to runtime dependencies

mwcraig:
`skeleton pypi` currently adds `setuptools` to both `build` and `run` dependencies if it is not already present in the dependencies extracted from `setup.py`. My understanding is that `setuptools` should not be needed at run time, only at build time. 

This PR drops the addition of `setuptools` to the `run` requirements; if the package explicitly lists `setuptools` as a requirement in `setup.py` then it remains in the list of dependencies.


msarahan:
I stripped off your ProgressiveFetchExtract thing.  It seemed like it was fixed in a different way on master.

Merging, thanks.

PR #2232, opened 2017-07-23
Testing changes

mingwandroid:
Mainly this is about copying the `test/source_files` into `info/test`, but also includes some minor fixes.


msarahan:
Not good to include source_files in info/test, imho. The whole point is
that people most often want to omit tests and test data from packages.
These files should get downloaded at test time if the source folder isn't
there, but they shouldn't be in the package by default. Perhaps info/test
is a good option when people opt to exclude the recipe from the package, as
the recipe is necessary right now to do the source download.

mingwandroid:
There are pros and cons. I *believe* that the old way of testing created a package that, for Anaconda, was, in total 17k. I cannot verify that because the current work flow is to remove those packages after a release. In the Minicondas I've been building I am seeing a 0.5% size increase. Please feel free to edit this comment so it's a live discussion. I'm probably strongly biased for doing it this way (because otherwise I'd not have done it this way).

Cons:
1. Slight size increase.

Pros (mostly this is just repeated from https://github.com/conda/conda-build/pull/2232/commits/0957e761bd5ebfb469f26ecde13a7c19ec913eef):
1. Bundling the test files improves the package as it is atomically testable.
2. Testing works in air-gapped situations.
3. Testing works if the source code vanishes or changes its location.
4. Testing works for recipes that specify their source with `source/path: ..` (or any other relative path) => the issue here is that the location of the sources on the build machine get baked into the rendered recipe's `source/path`.

I can see that bundling tests may lead to a situation where people avoid bundling large test-suites or datasets (similarly to how `conda-forge` currently delete the `Lib/tests` folder) , but for that I would like to expand our testing into two categories, quick inexpensive smoke tests and ones with `test/requires` that can use sub-packages and therefore require internet connectivity, when I get a chance.

At present I just need to be able to get QA to test the new stuff ASAP and this seemed to me to be the right way to do it.

msarahan:
The old way is not a good comparison because it mostly only ran trivial
import tests.  However, I do agree with your many benefits, and I think the
completely correct procedure is probably your addition here plus a split
package for any large test data.  Good thinking and thanks for the
explanation.

mingwandroid:
No problem! I'll keep working on this then. There's a few issues to sort out.

mingwandroid:
This should be good to go now provided the tests pass.

PR #2233, opened 2017-07-23
Fallback to using ruamel.yaml in conda skeleton

jjhelmus:
Use the namespaced ruamel.yaml import in conda skeleton pypi if
importing ruamel_yaml fails.  If both fail to import raise a helpful
error message.

Import logic adapted from common/serialize.py file in conda source code.


mandeep:
Thanks for this. Since ruamel_yaml is included in the skeleton, maybe it should be listed in the conda-build runtime requirements. I also need to discuss namespace packages with Ilan since it may be possible to use ruamel.yaml now instead of the fork.

msarahan:
Thanks @jjhelmus

PR #2234, opened 2017-07-24
Strip timestamp and user information from target archive

### LINKED ISSUE #1905: use root owner and group in the package tar files
The package tar.bz2 files produced by conda-build set owner and group membership to the conda-build user.  This makes package files non-reproducible if prepared by other user.  More importantly, they can lead to confusing results when installed by the root user, because the packager's user and group ownership could be restored if the same user or group existed on the target system.

Please consider resetting group and user name to `"root"` and group and user IDs to `0` when creating a package archive.  I use a similar approach in some of my packages to produce reproducible source distributions - https://github.com/diffpy/diffpy.structure/blob/master/devutils/makesdist#L30-L48

mandeep:
fixes #1905 

In an effort to be able to reproduce builds (issue #2140), this PR strips member information from files as they are added to the target archive. 

cc @msarahan @mingwandroid


conda_build/build.py
```diff
@@ -432,6 +432,10 @@ def write_info_json(m):
             info_index['depends'] = [' '.join(dist.split('::', 1)[-1].rsplit('-', 2))
                                      for dist in dists]
 
+    # change index.json timestamp for reproducibility
+    if 'timestamp' in info_index:
+        info_index['timestamp'] = 86400
```
bollwyvl: Perhaps name it, initialize in one place, and preferentially use `SOURCE_DATE_EPOCH`:
```py
DUMMY_TIMESTAMP = int(os.environ.get('SOURCE_DATE_EPOCH', 86400))
```

mandeep: Great point. I also set the default to 0 instead of 86400 as the next PR will contain a patch that sets pyc file timestamps to 0.

mingwandroid: Is 1 not safer than 0? Do we know if the comparison is `>` or `>=`?

mandeep: Yep, what I ended up doing was adding 1 to the DUMMY_TIMESTAMP for the pyc files.

mandeep:
I added a commit to this branch that modifies the pyc file timestamps to be 1 second newer than their related py files. With this PR and the other reproducibility PRs, I'm able to build the same package repeatedly with the resulting packages showing identical sha256 hashes. 

The only problem is that the patch_pyc_timestamp function in utils.py is not compatible with Python 2. I'm still searching for a solution for this but wanted to present the code in case anyone had questions.

conda_build/utils.py
```diff
@@ -1258,3 +1258,14 @@ def insert_variant_versions(metadata, env):
                 reqs.insert(i, ensure_valid_spec(' '.join((x.group(1),
                                                 metadata.config.variant.get(x.group(1))))))
     metadata.meta['requirements'][env] = reqs
+
+
+def patch_pyc_timestamp(filepath):
+    with open(filepath, 'rb') as pyc_file:
+        pyc_bytes = pyc_file.read()
+
+    new_timestamp = (DUMMY_TIMESTAMP + 1).to_bytes(4, byteorder='little')
```
mingwandroid: Hi @mandeep, please change byteorder='little' to use detection of endianness by checking the first 4 bytes. Even if all of the platforms we currently build for are little-endian, we shouldn't assume that's the case for everyone else.

msarahan:
The correct place for this content is now over at conda/conda-package-handling

Closing - happy to help with any migration questions if you'd like to continue this work.

mandeep:
Happy to help. Should I just piggy back off of https://github.com/conda/conda-package-handling/pull/3 ?

msarahan:
Sure!  Anything you do to improve the situation is much appreciated.  If you have questions, let us know.  I trust your judgement, though.

PR #2235, opened 2017-07-24
Add filename to order function so that filenames with identical filesizes are sorted alphabetically

mandeep:
In relation to reproducibility (issue #2140), this PR adds the filename of the file as a last resort when ordering. Now, two files with identical filesizes will be sorted alphabetically.

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -752,7 +752,7 @@ def bundle_conda(output, metadata, env, **kw):
             fsize = os.stat(join(metadata.config.host_prefix, f)).st_size or 100000
             # info/* records will be False == 0, others will be 1.
             info_order = int(os.path.dirname(f) != 'info')
-            return info_order, fsize
+            return info_order, fsize, f
 
         # add files in order of a) in info directory, b) increasing size so
         # we can access small manifest or json files without decompressing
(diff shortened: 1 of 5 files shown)


msarahan:
Please add a test

msarahan:
Thanks @mandeep.  I'm going to make the test you added a little lighter in terms of dependencies and source needed and clean up the merge conflict.  I'll merge after that.

mandeep:
@msarahan I picked conda-build since I knew it had files with identical sizes. Do you want me to find package that will result in a smaller size? Changing the source will mean that the tests will need to change and I don't want you to have to go through that trouble.

mandeep:
I've rebased the commits on top of the master branch. The merge conflict was the typical test at EOF conflict.

mandeep:
@msarahan I tried some other packages in the conda repository and couldn't find any that contained files of the same size. I also refactored this test so that we test all files with the same filesize.

msarahan:
Thanks for looking @mandeep.  I think you should be able to create files with identical size, right?

```
with open('file_a', 'w) as f:
    f.write('123')

with open('file_b', 'w) as f:
    f.write('123')
```

As long as both files have the same number of characters written to them (and have the same encoding), they should be the same size, right?

mandeep:
We could create files and them add them to a new tarfile and then check the tarfile to see if the members are in order. I used the conda-build recipe because I wanted a test that was a bit more functional. What do you think?

msarahan:
The correct place for this content is now over at https://github.com/conda/conda-package-handling

Closing - happy to help with any migration questions if you'd like to continue this work.

PR #2238, opened 2017-07-26
Only write one build script to build file and warn when multiple scripts found

### LINKED ISSUE #2231: Build fails if both meta.yaml build script and build.sh are specified
This was introduced in conda-build 3.0. If you specify a build script and build.sh exists the build fails. Steps to reproduce:
* Add the following to a meta.yaml file:
```
build:
  script: python setup.py install
```
* `echo "python setup.py install" > build.sh`

Results in (just the useful part of the error):
```
running install_scripts
writing list of installed files to 'record.txt#!/bin/bash'
error: [Errno 2] No such file or directory: 'record.txt#!/bin/bash'
```
If you rename build.sh or remove the script line from meta.yaml the build works fine. This was not the behavior in 2.x. Seems like the current way (you can have one or the other but not both) is probably "right", but maybe the default behavior could be something better than breaking (took me a _while_ to debug this).

mandeep:
fixes #2231

If a build.sh exists and a script value exists in meta.yaml's build field, conda build will build with the script found in meta.yaml. Also a warning will be raised to let the user know that two scripts exist.

I tried adding a test for the warning that used capfd to capture stderr, but there's too much output from the build process for capfd to capture.

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -1043,9 +1043,9 @@ def build(m, post=None, need_source_download=True, need_reparse_in_env=False, bu
             else:
                 build_file = join(m.path, 'build.sh')
                 if isfile(build_file) and script:
-                    raise CondaBuildException("Found a build.sh script and a build script "
+                    raise CondaBuildException("Found a build.sh script and a build/script section"
                                               "inside meta.yaml. Either remove the build.sh "
-                                              "script or remove the build script in meta.yaml.")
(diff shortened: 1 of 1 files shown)


msarahan:
Could we just write the script commands to the end of the existing file? I
don't have strong preference here, but it seems like an error or appending
the script section would be better than the warning. People don't always
see warnings...

PR #2239, opened 2017-07-26
Remove user specific information from about.json

mandeep:
In relation to #2140, this PR removes the writing of user specific information to about.json. I spoke with @kalefranz and @abarto and neither conda nor Anaconda.org uses the metadata from about.json. I wasn't sure if it was safe to remove the about.json file completely so I limited the content to package information.

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -382,30 +382,6 @@ def write_about_json(m):
             if value:
                 d[key] = value
 
-        # for sake of reproducibility, record some conda info
-        d['conda_version'] = conda_version
-        d['conda_build_version'] = conda_build_version
-        # conda env will be in most, but not necessarily all installations.
-        #    Don't die if we don't see it.
-        stripped_channels = []
-        for channel in get_rc_urls() + list(m.config.channel_urls):
-            stripped_channels.append(sanitize_channel(channel))
-        d['channels'] = stripped_channels
-        evars = ['PATH', 'PYTHONPATH', 'PYTHONHOME', 'CONDA_DEFAULT_ENV',
-                 'CIO_TEST', 'CONDA_ENVS_PATH']
-
-        if cc_platform == 'linux':
-            evars.append('LD_LIBRARY_PATH')
-        elif cc_platform == 'osx':
-            evars.append('DYLD_LIBRARY_PATH')
-        d['env_vars'] = {ev: os.getenv(ev, '<not set>') for ev in evars}
-        # this information will only be present in conda 4.2.10+
-        try:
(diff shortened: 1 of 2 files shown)


bollwyvl:
Ooh, don't know about this. I had done this in my hacky approach, but only because I didn't have anywhere canonical to put it. I don't think it's reasonable to suggest that two versions of `conda-build` will produce the same output, so at least that would have to be captured... but at that point, might as well specify everything that was available as who knows (yet) what might be needed.

Anyhow, I had been dreaming of a plugin/feature like:

```bash
conda-build --export-build-env foo-0.0.0.tar.bz2 > environment.yml
```

Which would _indeed_ use the `about.json` specs to make an `environment.yml` that could be used to provision a build environment. Instead of a local file path, accepting a URL would be wonderful.

Better still would be...
```bash
conda-build --reproduce foo-0.0.0_0.tar.bz2
```

Which would do the same extraction as above, but then...
- build a temporary environment with what it found
- use the temporary `conda` in it
  - using the embedded `recipe`, run `conda-build`
- check the hashes
  - if not the same, and available, use `diffoscope` against the generated `tar.bz2` to report back what changed

The challenging thing in this are the channels. Without a `requirements.txt` style representation, one can't necessarily guarantee that conda will resolve what is reported there, so it might be better to reduce it down to a `requirements.txt` or `environment.yaml`-style representation. And indeed, at the end of the day, the goal is for two packages in two channels to be the same!

The other thing to do with this would be the [`.buildinfo`](https://wiki.debian.org/ReproducibleBuilds/BuildinfoFiles) approach, and have a separate file which captured this information.

One could also imagine checking/generating `.sig` files and all kinds of other goodies. For example, if a URL is used, having a well-known location, i.e. `foo-0.0.0_0.sig` could be automatically checked when reproducing. There's really no claim to be satisfied with a separate `.buildinfo`, but yet it would still be worthwhile to sign.

So at the end of the day, one could imagine a single run of `conda-build` generating:
```
foo-0.0.0_0.buildinfo
foo-0.0.0_0.buildinfo.sig
foo-0.0.0_0.tar.bz2.sig
foo-0.0.0_0.tar.bz2
```

But now we start getting to some pretty deep stuff, as _really_ what one would want is the ability for _other_ people to sign packages, not just the author, such that an organization could demand that their internal build system had reproduced something claimed in public.

mandeep:
@bollwyvl You bring up some great points! I think having those two commands in conda-build would be awesome. I'll have to discuss this with @msarahan to see if there's a certain way he would want to approach this.

PR #2242, opened 2017-07-28
Fix "ResourceWarning: unclosed file"

Bezier89:
This is causing our tests to fail unless we ignore warnings ;(


msarahan:
Test failures here look unrelated.  Thanks for the fix!

Issue #2236, opened 2017-07-25
AttributeError when building from multiple sources


jlumpe:
I am attempting to build my recipe from multiple sources, but conda-build is raising an exception on the source step. My desired sources in `meta.yaml` look something like this:

```yaml
  - git_url: https://github.com/jlumpe/lyve-SET.git
    git_rev: v1.1.4-head
    patches:
      - lyve-set.patch

  - git_url: https://github.com/lskatz/Schedule--SGELK.git
    folder: build/Schedule

  - git_url: https://github.com/lskatz/cg-pipeline
    folder: libs/cg-pipeline

  - url: http://phast.wishartlab.com/phage_finder/DB/prophage_virus.db
    folder: build

  - path: templates
    folder: build/templates
```

Running `conda-build -s .` in the directory with my meta.yaml file I get the following error:

```
Traceback (most recent call last):
  File "/home/jared/anaconda3/bin/conda-build", line 6, in <module>
    sys.exit(conda_build.cli.main_build.main())
  File "/home/jared/anaconda3/lib/python3.5/site-packages/conda_build/cli/main_build.py", line 342, in main
    execute(sys.argv[1:])
  File "/home/jared/anaconda3/lib/python3.5/site-packages/conda_build/cli/main_build.py", line 329, in execute
    outputs = [action(recipe, config) for recipe in args.recipe]
  File "/home/jared/anaconda3/lib/python3.5/site-packages/conda_build/cli/main_build.py", line 329, in <listcomp>
    outputs = [action(recipe, config) for recipe in args.recipe]
  File "/home/jared/anaconda3/lib/python3.5/site-packages/conda_build/cli/main_build.py", line 272, in source_action
    source.provide(metadata)
  File "/home/jared/anaconda3/lib/python3.5/site-packages/conda_build/source.py", line 608, in provide
    path = normpath(abspath(join(metadata.path, metadata.get_value('source/path'))))
  File "/home/jared/anaconda3/lib/python3.5/site-packages/conda_build/metadata.py", line 891, in get_value
    value = self.get_section(section).get(key, default)
AttributeError: 'list' object has no attribute 'get'
```

It looks like it is trying to treat the `source` key as a dictionary instead of recognizing it is a list of dictionaries. The strange thing is various subsets of these sources will work or not work.

Here is a very basic version that reproduces the problem:

```yaml
source: 
  - git_url: https://github.com/jlumpe/lyve-SET.git
  - path: templates
```

[...]

PR #1929, opened 2017-04-14
add support and tests for multiple source files in one meta.yaml

msarahan:

Issue #1466, opened 2016-10-16
Multiple sources


jakirkham:
Would be nice if we had support for downloading multiple sources and combining them in some fashion.

minrk:
Since my case the extra packages happened to be Python packages, I was able to do it relatively simply with a requirements.txt with hash checking in build.sh.

https://github.com/conda-forge/fenics-feedstock/pull/1/commits/556756f9a7e265afaad50536f158cff710b090e6

This should have similar caching / hash checking to the source item, so ends up only a minor degradation from having multiple sources in meta.yaml.

mwiebe:
Some more references about this:

https://groups.google.com/a/continuum.io/forum/#!topic/conda/vxK2dnQzNz4
https://github.com/stuarteberg/conda-multisrc-example

The cases I've run into where this would be nice are LLVM + clang + friends (as in the email referenced), cases where there is a cmake build system separate from the main repo like https://github.com/python-cmake-buildsystem/python-cmake-buildsystem, and https://github.com/bkaradzic/bgfx which wants to be built together with https://github.com/bkaradzic/bx right beside it.

jakirkham:
cc @stuarteberg

jakirkham:
CC-ing @kain88-de @isuruf based on similar [discussion]( https://gitter.im/conda-forge/conda-forge.github.io?at=58f33e8408c00c092a862622 ) on Gitter.

msarahan:
There is an implementation of this at #1929 - Mostly working, but I want to add one more test of what happens when two tarballs that each contain a single folder get extracted into the same place.  They should both get "hoisted" but I'm not sure that is done right in the code right now.

I don't currently have plans to backport it to 2.1.x, but if anyone else wants to, go for it.

jakirkham:
Looks like that PR was merged. Does that mean this is solved in `conda-build` 3?

msarahan:
I think so.  Would be nice to have more testing by other people, though.  This should be in the current rc0 release on conda-canary.

jakirkham:
Closing as this has been solved in `conda-build` 3.

PR #2247, opened 2017-08-01
Fix bugs related to multiple sources

stuarteberg:
After #1929, a single recipe may specify multiple source repos (fixing #1466).  In that case, the metadata `source` section is written as a **list** of source dicts instead of a single dict:

```yaml
source:
  - git_url: https://github.com/foo/foo
    git_tag: 1.0

  - git_url: https://github.com/foo/bar
    git_tag: 1.1
    folder: bar
```

But that breaks an assumption in the convenience function `MetaData.get_value()` -- that every section is a `dict`.  Calls like `meta.get_value('source/git_url')` result in errors.  For example:

```pytb
  File "/miniforge/lib/python3.6/site-packages/conda_build/metadata.py", line 888, in get_value
    value = self.get_section(section).get(key, default)
AttributeError: 'list' object has no attribute 'get'
```

~~This PR "fixes" the problem by merely assuming the caller is interested in the FIRST source in the list.~~

If you don't like this proposal, here are some alternative solutions we could consider:
- ~~Add an assertion to forbid using `MetaData.get_value()` for the `source` section altogether, and require callers to check the type of `meta.get_section('source')` themselves.~~
- Extend `MetaData.get_value()` to support indexes, e.g. `meta.get_value('source/0/git_url')`.

cc: @msarahan (Since you authored #1929)

### DIFF
--- a/conda_build/environ.py
+++ b/conda_build/environ.py
@@ -380,7 +380,8 @@ def meta_vars(meta, config):
 
     git_exe = external.find_executable('git', config.build_prefix)
     if git_exe and os.path.exists(git_dir):
-        git_url = meta.get_value('source/git_url')
+        # We set all 'source' metavars using the FIRST source entry in meta.yaml.
+        git_url = meta.get_value('source/0/git_url')
 
         if os.path.exists(git_url):
             if sys.platform == 'win32':
@@ -396,9 +397,9 @@ def meta_vars(meta, config):
                                  git_url,
                                  config.git_commits_since_tag,
                                  config.debug,
-                                 meta.get_value('source/git_rev', 'HEAD'))
+                                 meta.get_value('source/0/git_rev', 'HEAD'))
 
-        if _x or meta.get_value('source/path'):
+        if _x or meta.get_value('source/0/path'):
             d.update(get_git_info(git_exe, git_dir, config.debug))
 
     elif external.find_executable('hg', config.build_prefix) and os.path.exists(hg_dir):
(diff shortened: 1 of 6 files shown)


msarahan:
Hmm, sorry about this.  I have found the get_* methods to be annoyingly constrained, and rarely use them.  It shows, I guess. =(

Are you seeing this because you're manipulating MetaData objects yourself in other code?  It's OK if you are, I just want to understand if there's conda-build bugs here, or if I'm only breaking MetaData's API for outside users.

Given your three options, I think a combination of this PR, plus your last suggestion, is nicest.  When a list is present and no index is present in the path, implicitly return only the first element (but show a warning with the index syntax).

stuarteberg:
>Are you seeing this because you're manipulating MetaData objects yourself in other code?

No, I was merely trying to write a recipe.  The test case I added in this PR is an example that fails on master.

>I think a combination of this PR, plus your last suggestion, is nicest.

OK, sure.  Now I've updated the PR.

stuarteberg:
AFAICT, the Travis failures are unrelated these changes.

For `CONDA_VERSION=4.3.x`, there's just one failure:

```
[gw0] FAILED tests/test_api_build.py::test_disable_pip 
tests/test_api_build.py::test_noarch_none_value 
```

stuarteberg:
Regarding the complaint from codacy: Do I really need to avoid using `assert`?!!??

msarahan:
RE: codacy, I've been ignoring that, but they have a point.  If those checks are ever skipped, our programs will exit in perhaps strange, uninterpretable ways.  I have that on the way-back burner.

The other stuff looks great.  Thanks for humoring me.  I'm going to go ahead and merge this.  Hopefully I can get the tests all green again soon.

mandeep:
Thanks for this. I believe this fixes #2236 as well.

PR #2248, opened 2017-08-01
conda 4.4 compatibility changes

kalefranz:


msarahan:
Looking much, much better.  Any idea on these Dist errors?  https://travis-ci.org/conda/conda-build/jobs/259957668#L2734

kalefranz:
Yeah I spent the last two hours trying to dig in and figure this out.  Frustratingly had *different* dist errors locally than on Travis CI.  Then I did an `rm -rf pkgs/*` to both of my package caches, and the the tests passed.  So we have a state issue with conda-build and the package cache I think.

If this PR doesn't introduce any new errors, my advice would be to go with it what we have here.  It's incrementally a big improvement.  The fixes to these dist errors might be in the conda code.

msarahan:
OK, sounds good to me.  It is definitely already a big improvement.  Merging.

kalefranz:
I think I discovered the problem with those tests (and probably others).  By default, `get_index()` hasn't included packages in the package cache.  It's taken a special ([help suppressed](https://github.com/conda/conda/commit/f761f65a82b739562a0d997a2570e2b8a0bdc783)) command line argument to do that.  My understanding was a little wrong there.  Working on conda code now to bring 4.4 back to the established behavior.

PR #2249, opened 2017-08-01
Add source cache to config and remove conda legacy imports

mandeep:
Add src_cache_root to Config objects and allow setting via --cache command line flag.


msarahan:
Looks great.  I tweaked the CLI argument and fixed a flake8 thing.  Merging.

PR #2250, opened 2017-08-02
do not modify user-specified build/string with hash

### LINKED ISSUE #2245: Problem: Conda build --croot --test failes to to missing builded package
On our Docker CI workflow we work with two seperate stages for build and test a package. The builded conda packages are transfered by gitlab runner artifacts.

The build command is `conda build path/to/recipe --croot $(pwd)/builds --no-test` while test stage is run by `conda build path/to/recipe --croot $(pwd)/builds --test`.
While the build stage runs successfully and provide the compiled packages as artifacts, the test stage will failed with following errors:
```
  File "/opt/miniconda3/lib/python3.5/site-packages/conda_build/environ.py", line 680, in get_install_actions
    actions = install_actions(prefix, index, specs, force=True)
  File "/opt/miniconda3/lib/python3.5/site-packages/conda/plan.py", line 461, in install_actions
    update_deps, pinned)
  File "/opt/miniconda3/lib/python3.5/site-packages/conda/plan.py", line 634, in get_actions_for_dists
    pkgs = r.install(specs, installed, update_deps=update_deps)
  File "/opt/miniconda3/lib/python3.5/site-packages/conda/resolve.py", line 809, in install
    pkgs = self.solve(specs, returnall=returnall)
  File "/opt/miniconda3/lib/python3.5/site-packages/conda/resolve.py", line 849, in solve
    reduced_index = self.get_reduced_index(specs)
  File "/opt/miniconda3/lib/python3.5/site-packages/conda/resolve.py", line 374, in get_reduced_index
    specs, features = self.verify_specs(specs)
  File "/opt/miniconda3/lib/python3.5/site-packages/conda/resolve.py", line 300, in verify_specs
    raise ResolvePackageNotFound(bad_deps)
  - sw-package 1.0 sw_package

  File "/opt/miniconda3/bin/conda-build", line 6, in <module>
  File "/opt/miniconda3/lib/python3.5/site-packages/conda_build/cli/main_build.py", line 342, in main
    execute(sys.argv[1:])
  File "/opt/miniconda3/lib/python3.5/site-packages/conda_build/cli/main_build.py", line 329, in execute
    outputs = [action(recipe, config) for recipe in args.recipe]
  File "/opt/miniconda3/lib/python3.5/site-packages/conda_build/cli/main_build.py", line 329, in <listcomp>
    outputs = [action(recipe, config) for recipe in args.recipe]
  File "/opt/miniconda3/lib/python3.5/site-packages/conda_build/cli/main_build.py", line 277, in test_action
    return api.test(recipe, move_broken=False, config=config)
  File "/opt/miniconda3/lib/python3.5/site-packages/conda_build/api.py", line 205, in test
    move_broken=move_broken)
  File "/opt/miniconda3/lib/python3.5/site-packages/conda_build/build.py", line 1422, in test
    channel_urls=tuple(metadata.config.channel_urls))
  File "/opt/miniconda3/lib/python3.5/site-packages/conda_build/environ.py", line 682, in get_install_actions
    raise DependencyNeedsBuildingError(exc, subdir=subdir)
conda_build.exceptions.DependencyNeedsBuildingError: Unsatisfiable dependencies for platform linux-64: ['sw-package']
```

This behaviour also occours localy, even if you run `conda build --croot --no-test` and `conda build --croot --test` seperately without cleanup the `croot` folder.

We work with Conda 4.3.23, Conda Build 3.0.8 under ubuntu linux 64 and windows 64

msarahan:
fixes #2245 

This also makes the hash available as the PKG_HASH env var, usable in build scripts and in meta.yaml, should the user want to put it in themselves.  It does not include the leading h.

### DIFF
--- a/conda_build/api.py
+++ b/conda_build/api.py
@@ -89,7 +89,7 @@ def get_output_file_paths(recipe_path_or_metadata, no_download_source=False, con
     elif isinstance(recipe_path_or_metadata, string_types):
         # first, render the parent recipe (potentially multiple outputs, depending on variants).
         metadata = render(recipe_path_or_metadata, no_download_source=no_download_source,
-                            variants=variants, config=config, **kwargs)
+                            variants=variants, config=config, finalize=True, **kwargs)
     else:
         assert hasattr(recipe_path_or_metadata, 'config'), ("Expecting metadata object - got {}"
                                                             .format(recipe_path_or_metadata))
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -207,12 +207,18 @@ def copy_recipe(m):
                 utils.copy_into(src_path, dst_path, timeout=output_metadata.config.timeout,
                                 locking=output_metadata.config.locking, clobber=True)
 
+        # hard code the build string, so that tests don't get it mixed up
+        build = output_metadata.meta.get('build', {})
+        build['string'] = output_metadata.build_id()
+        output_metadata.meta['build'] = build
+
         # just for lack of confusion, don't show outputs in final rendered recipes
         if 'outputs' in output_metadata.meta:
             del output_metadata.meta['outputs']
 
         utils.sort_list_in_nested_structure(output_metadata.meta,
                                             ('build/script', 'test/commands'))
+
         rendered = output_yaml(output_metadata)
 
         if not original_recipe or not open(original_recipe).read() == rendered:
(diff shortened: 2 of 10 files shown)


argltuc:
I'm not sure, to not modify the build string, if the user set it manualy is the best way to solve the bug of #2245, because this will break the behaviour of conda build again.
I could imagine, that a solution where conda build just adds it build hash to the user given build string before any repository check is done does note break the current beaviour of CB and could solve the missing packages in a testing scenario

msarahan:
The reason conda-build broke was that conda-build assumed that the build string ended with _(build number).  You broke that assumption, which is fine, but the price of breaking that assumption is that you no longer get the default behavior.  It was a mistake for me to try to append the hash to arbitrary user build strings, because I don't know exactly where they want them.  This PR should give people the flexibility to keep the hash if they want, put it where they want, or leave it out.

Omitting the hash is not a breaking change.  If people are using variants, it could result in clobbering packages, as detailed in the --old-build-string option.  If people are updating to use variants, and they're using custom build strings, I expect that they can/should update their recipe to put the hash where they want it.

PR #2253, opened 2017-08-02
ignore feature records in index for 'conda inspect'

kalefranz:
This problem surfaced via https://github.com/conda/conda/pull/5645.  Conda has always added "virtual" feature records to the `index`.  It was never a problem here because there were so few of them in the past.  Now there are a few more, and specifically feature records for python, which is the source of the particular collision here.

The way feature records for the solver are identified in `index` is that the name field ends with an `@` character.


msarahan:
Is there anything worth explicitly testing here?

kalefranz:
That code path will implicitly be covered once that linked PR is merged for conda 4.4.

PR #2254, opened 2017-08-02
fix eventual consistency of files downloaded feeding package version

### LINKED ISSUE #2243: Using variants in source url fails to regenerate setup.py data in 3.0.8
We're using variants to generate multiple package versions, so meta.yaml has:
 ```
 {% set setup_data = load_setup_py_data(from_recipe_dir=True) %}
  package:
    name: my_package
    version: '{{my_version}}.{{setup_data['version']}}'

  source:
  {% if my_version %}
    url: http://server/ver{{my_version}}.zip
  {% endif %}
```
conda_build_config.yml has:
 ```
 my_version:
    - 470
    - 480
    - 490
```
and our setup.py contains this logic:
```
from setuptools import setup
if 'SRC_DIR' in os.environ:
  setup(version=work_it_out(os.environ['SRC_DIR']))
else:
  setup(version='0')
```
With conda-build 3.0.4, this works fine - setup.py is re-evaluated after the variants are expanded and each source package downloaded, however after upgrading to 3.0.8 it seems that it sticks with the original version number '0' for all the rendered yaml. 

Is there any way to force conda-build to re-render the templates after source downloading?

msarahan:
fixes #2243


msarahan:
@mingwandroid - I will come back later to fix the cross-compiler issue, once we can figure it out.  I want to get this release out so that we can move forward on the build system.  Merging.

PR #2265, opened 2017-08-08
Add newlines

neok-m4700:
Prevents issues when merging meta test commands and test scripts in run_test.{sh,bat}

I've seen some test failures on recipes when newlines are missing.


msarahan:
Cool, thanks!

Issue #2266, opened 2017-08-08
conda render ignores channels specified on command line


chaubold:
I noticed that `conda render` complained about missing package dependencies, but `conda build` did not. Looking at the code it turns out that `conda render` does not pass on any channels specified on the command line, it only considers those from the `.condarc` file.

To be more precise: the `execute` method in `conda render`'s command line interface [cli/main_render.py#L138](https://github.com/conda/conda-build/blob/master/conda_build/cli/main_render.py#L138) should probably also contain those lines from `conda build` [cli/main_build.py#L309-L323](https://github.com/conda/conda-build/blob/master/conda_build/cli/main_build.py#L309-L323).

msarahan:
what version of conda-build is this?

chaubold:
The very latest one (3.0.9), and my links to the code above were all to the current master.

(Give me one sec, I'm preparing a PR with a fix, just waiting for the test suite to run through...)

PR #2267, opened 2017-08-08
Consider channels specified via CLI in conda-render

chaubold:
This aims to fix the problem that `conda-render` ignored channel settings provided on the command line (#2266) by copying over the respective code snippet from `conda_build/cli/main_build.py`.

The PR includes a test for success and failure depending on whether the channel is provided as command line flag or not. The check for success is currently whether the rendered `meta.yaml` contains the correct version number for the required package or not. @msarahan I'm not sure whether that's too much of an assumption that the version number is present. If you have better criteria in mind, I'm happy to change that.

### DIFF
--- a/conda_build/cli/main_render.py
+++ b/conda_build/cli/main_render.py
@@ -8,8 +8,9 @@ from __future__ import absolute_import, division, print_function
 
 import logging
 import sys
+import os
 
-from conda_build.conda_interface import ArgumentParser, add_parser_channels, cc_conda_build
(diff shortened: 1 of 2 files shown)


msarahan:
Nice tests.  version numbers are guaranteed to be present after rendering as long as the recipe is "finalizable" (all deps are installable) and you haven't disabled finalizing.

Failures are just one flake8 and several pypi download errors.  Seems OK to me.  I'll fix the flake8 down the line a bit.  Thanks!

chaubold:
You're welcome! Thanks for providing us with conda ;-)

PR #2268, opened 2017-08-08
metadata: Fix mistake in get_value() (bug from #2247)

stuarteberg:
Well, this is embarrassing.  Last week I submitted a PR (#2247) which had a small bug in it.  If someone uses the new "index" syntax for `get_value()` with a non-zero index (e.g. `meta.get_value('source/2/git_url')`), it won't work.  Oops.

This isn't affecting anyone right now, because at the moment no part of conda-build calls `get_value()` with a non-zero index.  But here's a fix nonetheless.  (With a test case!)

### DIFF
--- a/conda_build/metadata.py
+++ b/conda_build/metadata.py
@@ -924,7 +924,7 @@ class MetaData(object):
             if len(section_data) == 0:
                 section_data = {}
             else:
-                section_data = section_data[0]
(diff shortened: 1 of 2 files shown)


msarahan:
Whoops, sorry I didn't catch that.  Thanks for the fix!

PR #2277, opened 2017-08-13
Pass omit_defaults=m.config.override_channels to get_build_index

mingwandroid:
.. otherwise --override-channels does nothing.


msarahan:
Good catch.  Thanks.

PR #2308, opened 2017-08-23
update index before testing .tar.bz2 and after moving failed

### LINKED ISSUE #2301: When testing a package, the folder containing the package needs be an indexed channel
When testing packages using: `conda-build /data/builds/linux-64/eigen3-3.2.9-he32ef9e_5.tar.bz2 --test` and the artifact is just in a folder without repodata.json, the above command would produce: 

```
  - eigen3 3.2.9 he32ef9e_5
```

after running `conda index` on that folder the above test is run successfully...

Going back to our discussion of #2260, where we concluded that testing recipes is unintuitive: I'd say that its also unintuitive to have to put a package-under-test (that we even specify by full path) into a valid conda channel... 

It should be: *CB, please test that package there!*, not *CB, please test that package in that channel there*.

### LINKED ISSUE #2303: Dirty repodata.json after failed builds
When a package builds successfully but fails the test (overall fails the build) the package is removed from `conda-bld/linux-64/` however repodata.json still lists the package.

cc @msarahan

msarahan:
fixes #2301 
fixes #2303

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -45,7 +45,7 @@ from .conda_interface import dist_str_in_index
 
 from conda_build import __version__
 from conda_build import environ, source, tarcheck, utils
-from conda_build.index import get_build_index
+from conda_build.index import get_build_index, update_index
 from conda_build.render import (output_yaml, bldpkg_path, render_recipe, reparse, finalize_metadata,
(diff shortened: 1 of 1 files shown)


quasiben:
Thanks @msarahan!

Issue #2327, opened 2017-08-29
enhance hash rename in src_cache


neok-m4700:
Hash renaming leaves to many hashes in filename

e.g. for 
```yaml
url: https://ftp.gnu.org/gnu/binutils/binutils-2.28.tar.bz2
 ```
is renamed to `binutils-2.28.tar.bz2 -> binutils-2_6297433ee1.28_6297433ee1.tar.bz2`

Enhancement proposal (**source.py**) l.32:
```python
# ext_re = re.compile(r"(.*?)((?:\.tar)?\.[^.]+?)")
ext_re = re.compile(r"(.*?)(\.(?:tar\.)?[^.]+)$")
```
https://regex101.com/r/SOZuGU/2/

msarahan:
Yes, seems right.  Mind sending in a PR, or would you rather not?

PR #2330, opened 2017-08-29
fix hash regex for downloaded files in src_cache

neok-m4700:
https://github.com/conda/conda-build/issues/2327


msarahan:
Thanks!

PR #2334, opened 2017-08-29
add LD_RUN_PATH back into linux variables for now

msarahan:
cc @neok-m4700


neok-m4700:
Thanks @msarahan , I think that it it safer from a user point of view, when you don't include all the necessary build toolchain, nor set the whole bunch of flags -Wl,-rpath or -rpath-link, but still want to link preferably to the libraries in the prefix.

mingwandroid:
Agreed, definitely safer. Later we will look for a way to allow *us* to avoid needing these variables (I really dislike `LD_*` if you cannot tell!).

Issue #2329, opened 2017-08-29
Are relative paths fully supported?


gabm:
We'd like to use relative paths instead of absolute paths in our scripts on windows and linux:

```
conda-build --croot relative/path/to/something RECIPE_FOLDER
conda-build --test relative/path/to/an/artifact.tar.bz2
...
```

I remember that we had several problems with that in the past so we switched to use absolute paths everywhere. By chance we noticed that something changed recently in path handling in CB3. Seems more robust..

Are the above usages fully supported by CB3 and will they be supported in future?

msarahan:
My intention is that they are supported, but I don't know whether there are explicit tests for relative vs. absolute for every argument.  If you have time, I'd appreciate any contribution you can make towards explicit testing.

gabm:
So it's desired behavior, that relative paths could be used as arguments? in this case I'd provide a PR tomorrow with some testcases to protect relevant usage scenarios for us.

As for the mentioned commands above: croot doesn't seem to work with relative paths .. testing works...

msarahan:
Yes, it's absolutely desired behavior.  Test cases would be much
appreciated.

gabm:
I submitted a PR with the tests and fixes in #2335

msarahan:
fixed by #2335 - closing.

PR #2335, opened 2017-08-30
Problem: relative paths in croot don't work anymore

gabm:
This PR is the result of #2329 

### tests

It adds 3 new testcases:
1. `test_relative_path_croot` which covers:  `conda-build --croot ./relative/path {recipe-folder}`
2. `test_relative_path_test_artifact` which covers: `conda-build --test ./relative/path/to/artifact.tar.bz2`
3. `test_relative_path_test_recipe` which covers `conda-build --test {recipe-folder} --croot ./relative/path/to/croot`

While test 2 and 3 worked out of the box, test 1 fails on linux and most-likely on windows as well. The fix for test 1 is included in this PR in 6720ae1, but I'm not 100% sure, so please @msarahan, review it carefully. 

### fix

I intented to use croot as an absolute path internally by parsing the commandline parameter. It was added back then by #1736. The code to make croot absolute is aparently still there, but it gets overwritten by: 

https://github.com/conda/conda-build/blob/6b9700f2c92db4b2dc822d51829a19be181afd1b/conda_build/config.py#L196

In this part of the code all attributes of the class are set from the kwargs dictionary, although some of them are handled above. My solution is to pop the specially handled args from the kwargs dictionary so they don't get overwritten later on. I suppose it happens to the other args in this function as well, so I popped them as well, but of course I'm unaware of consequences of doing so.

Another option would be to not pop the entries from the dict but instead handle the special cases _after_ setting the defaults...

### DIFF
--- a/conda_build/config.py
+++ b/conda_build/config.py
@@ -147,7 +147,7 @@ class Config(object):
 
     def set_keys(self, **kwargs):
         def env(lang, default):
-            version = kwargs.get(lang)
+            version = kwargs.pop(lang, None)
             if not version:
                 # Hooray for corner cases.
                 if lang == 'python':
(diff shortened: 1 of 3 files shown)


msarahan:
Nice addition, thanks @gabm - CI failures look like nothing new.  Merging.

PR #2341, opened 2017-09-03
Problem: When packages reside in un-indexed channels, the test command fails

### LINKED ISSUE #2301: When testing a package, the folder containing the package needs be an indexed channel

gabm:
This fixes #2301 

@msarahan Please review the fix carefully. There is more logic involved to index the right channels.. Maybe at some point we could teach update_index to do globbing...

There is one point where I'm not sure, and its where I throw the error... I'll mark it in the code

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -1360,8 +1360,7 @@ def _construct_metadata_for_test_from_package(package, config):
             is_channel = True
 
     if not is_channel:
-        log.error("Your package must reside in a channel structure with platform-subfolders!")
-        sys.exit(1)
+        raise ValueError("Your package must reside in a channel structure with platform-subfolders!")
 
     # get channel url
     local_channel = os.path.dirname(local_pkg_location)
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -1360,11 +1360,12 @@ def _construct_metadata_for_test_from_package(package, config):
             is_channel = True
 
     if not is_channel:
-        raise ValueError("Your package must reside in a channel structure with platform-subfolders!")
+        raise ValueError("Your package must reside in a channel structure with "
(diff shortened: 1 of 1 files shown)


conda_build/build.py
```diff
@@ -1347,24 +1347,39 @@ def _construct_metadata_for_test_from_package(package, config):
...
+    # get last part of the path
+    last_element = os.path.basename(local_pkg_location)
+    is_channel = False
+    for platform in ('win-', 'linux-', 'osx-', 'noarch'):
+        if last_element.startswith(platform):
+            is_channel = True
+
+    if not is_channel:
+        log.error("Your package must reside in a channel structure with platform-subfolders!")
+        sys.exit(1)
```
gabm: here I'm not sure how to properly throw an error?!

msarahan: ```raise ValueError("Your package must reside in a channel structure with platform-subfolders!")```

A more specific exception type might be nice, but ValueError is probably fine to start with.

gabm: done...

msarahan:
Thanks!  I'll wait on CI and merge if they come out OK.  Looks fine to me.

msarahan:
Looks great.  Thanks again @gabm.  Merging.

PR #2364, opened 2017-09-13
Set ARCH from host_arch

mingwandroid:


msarahan:
I think the fallback from host_arch to arch is already there.  Was it not working?

Specifically:
https://github.com/conda/conda-build/blob/master/conda_build/config.py#L236

mingwandroid:
I saw that pattern being used in the one place that did use host_arch so I cargo culted it everywhere to be safe.

mingwandroid:
But yes it probably shouldn't be needed. It would be nicer from a DRY perspective if host_arch() called arch() I think, but it's minor. Feel free to change my PR here, I'll be stuck in debugging-land for a while.

msarahan:
superseded by #2368; closing

PR #2367, opened 2017-09-13
BUG: verify source hash

### LINKED ISSUE #2366: source hash not verified
The md5 or sha256 hash of a source tarball is not checked prior to building a package.  For example the following recipe build successfully regardless of which md5 line is un-commented:

``` yaml
package:
  name: imagesize
  version: 0.7.1

source:
  fn: imagesize-0.7.1.tar.gz
  url: https://pypi.io/packages/source/i/imagesize/imagesize-0.7.1.tar.gz
  #md5: 976148283286a6ba5f69b0f81aef8052
  md5: deadbeefdeadbeefdeadbeefdeadbeef

build:
  number: 0
  script: python setup.py install --single-version-externally-managed --record record.txt

requirements:
  build:
    - python
    - setuptools
  run:
    - python
```

A build completes without an indication about an incorrect md5 with conda build `3.0.17`, it fails with conda build 2.1.17 with a notice about an MD5 mismatch.  

Will try other releases in the 3.x line shortly.

jjhelmus:
closes #2366

### DIFF
--- a/conda_build/source.py
+++ b/conda_build/source.py
@@ -90,7 +90,7 @@ def download_to_cache(cache_folder, recipe_path, source_dict):
 
     hashed = None
     for tp in ('md5', 'sha1', 'sha256'):
-        if 'tp' in source_dict:
+        if tp in source_dict:
             expected_hash = source_dict[tp]
             hashed = hashsum_file(path, tp)
             if expected_hash != hashed:


msarahan:
Thanks @jjhelmus.  Merging.

PR #2385, opened 2017-09-18
Fix bug that causes conda-build to ignore filename_hashing setting

Bezier89:


msarahan:
Thanks!

Bezier89:
No problem @msarahan. I should note that this only affected calls through conda_build.api, calls through the CLI worked as expected.

PR #2390, opened 2017-09-19
Remove .a from files not considered for prefix replacement

mingwandroid:
There is no reason on earth for this exclusion.


mingwandroid:
This broke the ability to statically link to openssl.

Issue #1787, opened 2017-03-03
Trying "conda skeleton cran data.tree" gives "Error: Could not parse metadata (Enhances:)"


kalefranz:
_From @JustGitting on March 3, 2017 2:2_

Hi,

I'm trying to install the R package [data.tree](https://cran.r-project.org/web/packages/data.tree/) using the method described [here](https://www.continuum.io/blog/developer-blog/conda-data-science).

I've successfully used it to install a number of R packages not in the conda repository. However, I'm unable to install the data.tree package.

```
C:\Users\user1>conda skeleton cran "data.tree"
Fetching metadata from https://cran.r-project.org/
Error: Could not parse metadata (Enhances:)
```

It would appear to be a parsing problem, because replacing the "." in "data.tree" produces a different error.

```
C:\Users\user1>conda skeleton cran "data-tree"
Fetching metadata from https://cran.r-project.org/
Package data-tree not found
```
or

```
C:\Users\pc83>conda skeleton cran data..tree
Fetching metadata from https://cran.r-project.org/
Package data..tree not found

C:\Users\pc83>conda skeleton cran data\.tree
Fetching metadata from https://cran.r-project.org/
Package data\.tree not found

C:\Users\user1>conda skeleton cran "data/.tree"
Fetching metadata from https://cran.r-project.org/
Package data/.tree not found
```

System info:
```
C:\>conda info
Current conda install:
               platform : win-64
          conda version : 4.3.13
       conda is private : False
      conda-env version : 4.3.13
    conda-build version : 2.1.5
         python version : 3.6.0.final.0
       requests version : 2.13.0
       root environment : C:\Data\AppData\Local\Continuum\Anaconda3  (writable)
    default environment : C:\Data\AppData\Local\Continuum\Anaconda3
       envs directories : C:\Data\AppData\Local\Continuum\Anaconda3\envs
                          C:\Users\user1\AppData\Local\conda\conda\envs
                          C:\Users\user1\.conda\envs
          package cache : C:\Data\AppData\Local\Continuum\Anaconda3\pkgs
                          C:\Users\user1\AppData\Local\conda\conda\pkgs
           channel URLs : https://conda.anaconda.org/bioconda/win-64
                          https://conda.anaconda.org/bioconda/noarch
                          https://repo.continuum.io/pkgs/free/win-64
                          https://repo.continuum.io/pkgs/free/noarch
                          https://repo.continuum.io/pkgs/r/win-64
[...]

PR #2393, opened 2017-09-21
Simplify handling of blank fields in CRAN metadata.

jdblischak:
## Summary

This PR re-implements a past solution for `conda skeleton cran` that handles blank entries for any field in CRAN metadata.

## Background

Sometimes R packages on CRAN have blank metadata fields (e.g. Suggests is blank for [corpcor](https://cran.r-project.org/web/packages/corpcor/DESCRIPTION) and Enhances is blank for [data.tree](https://cran.r-project.org/web/packages/data.tree/DESCRIPTION)). In the past this caused problems when running `conda skeleton cran`, e.g.:

* Issue #1787 from March 2017 for data.tree using conda-build 2.1.5
* Issue https://github.com/conda/conda/issues/5624 from July 2017 for corpcor using conda-build 2.1.16

One solution to this problem was provided on July 5th by @nehaljwani in commit https://github.com/conda/conda-build/pull/2153/commits/4d90305bf675a6ebe49b3ead162820338f74413c in PR https://github.com/conda/conda-build/pull/2153. The implementation was to create a global variable `CRAN_KEYS_CAN_BE_BLANK` that stored metadata fields known to be blank. This fixed the issue with creating a recipe for corpcor.

```
CRAN_KEYS_CAN_BE_BLANK = [
    'Suggests'
]
```

When I saw that Issue #1787 was still open, I figured that I could fix the data.tree recipe by adding `'Enhances'` to `CRAN_KEYS_CAN_BE_BLANK`. However to my surprise, `conda skeleton cran` was already able to create a recipe for data.tree. The reason is because the line that evaluates blank columns always evaluates to `False` because `'Enhances:'.split(':')` returns `['Enhances', '']`:

```
if len(line) < 2 and line[0] in CRAN_KEYS_CAN_BE_BLANK:
    continue
```

In researching this, I discovered that there was a separate solution submitted shortly beforehand. On Jun 14th, @mingwandroid submitted commit https://github.com/conda/conda-build/pull/2088/commits/c1fcde94f0cb0b7d2e7cfe4266e72f2d530776e9 in PR https://github.com/conda/conda-build/pull/2088:

```
if ': ' in line:
    (k, v) = line.split(': ', 1)
else:
    # Sometimes (leaflet) you get lines such as 'Depends:'
    (k, v) = line.split(':', 1)
```

The advantage to this solution is that it works for any blank entry without having to pre-define them. This solution was overwritten in commit https://github.com/conda/conda-build/commit/59e7dbd4a9eb5ca8ab6d0c2a1b24a8cd42430855?diff=split when PR #2153 was merged.

In this PR, I re-implemented the solution from @mingwandroid. I tested that it works for both corpcor and data.tree. It should also work for any other packages that have blank entries in other fields. In my opinion, this solution is easier to maintain, since we don't have to edit the skeleton file each time a new blank field is observed in CRAN metadata. And if a package is missing critical metadata (e.g. License), this solution will fail just the same as one that pre-defines all the fields that are safe to be blank.

### DIFF
--- a/conda_build/skeletons/cran.py
+++ b/conda_build/skeletons/cran.py
@@ -149,10 +149,6 @@ CRAN_KEYS = [
     'Maintainer',
 ]
 
-CRAN_KEYS_CAN_BE_BLANK = [
-    'Suggests'
-]
-
 # The following base/recommended package names are derived from R's source
 # tree (R-3.0.2/share/make/vars.mk).  Hopefully they don't change too much
 # between versions.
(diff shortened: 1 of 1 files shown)


msarahan:
Thanks @jdblischak

PR #2413, opened 2017-10-02
debug appveyor pip issue

msarahan:


msarahan:
Seems like the error is that old versions of miniconda have faulty pip packages that leave the requests folder around after removing pip.  Filed issue for appveyor to update their miniconda:

http://help.appveyor.com/discussions/suggestions/3180-please-update-miniconda-installations-pip-update-issue

PR #2414, opened 2017-10-02
change inspect linkage tests for libpython instead of openssl 

msarahan:
fixes py27 linux tests.

Not sure if this is an actual problem with our py27 packages, though.  Could be just a bug with conda inspect linkages, or could be actually true that python isn't linking against our packaged openssl, but rather the system openssl.  The former would be an issue to fix here, the latter not so much.


mingwandroid:
It's not a problem, it is deliberate (but unfortunate).

mingwandroid:
Note, you should not be seeing libpython *either* here. It also is linked to statically. This is deliberate (and not unfortunate), it's a performance optimization.

PR #2420, opened 2017-10-03
Don’t strip generic files in noarch: python

### LINKED ISSUE #2419: noarch: python packages strip data_files
in setup.py, `data_files` can be used to deliver support files, especially to `$PREFIX/share`. `noarch: python` [doesn't know how to handle these](https://github.com/conda/conda-build/blob/3.0.23/conda_build/noarch_python.py#L90) and ends up omitting them.

It seems like 'unknown' files in the prefix should be included as-is relative to $PREFIX (exactly as in a non-noarch package) rather than excluded.

minrk:
This appears to have been simpler than I thought. Stripping unrecognized files prevents `noarch: python` packages from including data_files, such as those in `$PREFIX/share`. Simply removing the stripping appears to cause the correct behavior of including these files relative to $PREFIX as in a non-noarch package.

closes #2419

### DIFF
--- a/conda_build/noarch_python.py
+++ b/conda_build/noarch_python.py
@@ -85,10 +85,11 @@ def handle_file(f, d, prefix):
     # Include examples in the metadata doc
     elif f.startswith(('Examples/', 'Examples\\')):
         d['Examples'].append(f[9:])
+    # No special treatment for other files
+    # leave them as-is
     else:
         log = logging.getLogger(__name__)
-        log.warn("Don't know how to handle file: %s.  Omitting it from package." % f)
-        os.unlink(path)
+        log.debug("Don't know how to handle file: %s.  Including it as-is." % f)
 
 
 def populate_files(m, files, prefix, entry_point_scripts=None):


jakirkham:
I imagine there is a corresponding change needed for `conda`, but don't really know the details myself. Maybe @kalefranz or @msarahan can point you in the right direction.

minrk:
I also assumed there would be, but I tested with the jupyterhub package I was building (https://github.com/conda-forge/jupyterhub-feedstock/pull/11) and current stable conda appears to do everything right with only this change. I wouldn't be surprised if there are other cases not correctly handled, though.

kalefranz:
For reference, the `os.unlink()` that's being removed was added via #1806 in  d0c007cae588fe5cd883723440b77ebc301bf4bc.  The message for the commit is "don't write shebang in noarch entry points," but I don't see how it's exactly related.

The change in theory seems fine to me right now.  But it's also very possible there are edge cases that aren't coming to mind for me right now.

minrk:
Thanks @kalefranz! Looking at #1806, it looks to have turned a fatal error into a warning and ignored file. But such files were still 'unhandled'.

I 100% believe there are edge cases that I haven't thought of.

kalefranz:
Also, I doubt conda >=4.3 would need any change here for this issue.  It'll just install everything in the paths.json manifest, making the relevant `noarch: python` path translations.

msarahan:
Failures seem unrelated.  Merging.  Thanks @minrk

PR #2427, opened 2017-10-07
Fix pin compatible dependencies by removing tilde from dependency

### LINKED ISSUE #2351: conda skeleton leaves tilde chars from requires.txt
Running conda skeleton pypi fs gives

requirements:
  build:
    - python
    - setuptools
    - appdirs~ 1.4.3
    - pytz
    - setuptools
    - six~ 1.10.0
  run:
    - python
    - setuptools
    - appdirs~ 1.4.3
    - pytz
    - setuptools
    - six~ 1.10.0

mandeep:
fixes #2351 

This PR adds pin compatible versioning functionality to the PyPI skeleton. For example, a version listed as ``package ~= 1.4.3`` will be converted to ``package >=1.4.3,<2``.

The error stems from https://github.com/conda/conda/blob/master/conda/cli/common.py#L106 so I'll be submitting a PR there as well. 

Tests look fine locally.

### DIFF
--- a/conda_build/skeletons/pypi.py
+++ b/conda_build/skeletons/pypi.py
@@ -619,7 +619,7 @@ def convert_version(version):
     version_parts = version.split('.')
     suffixes = ('post', 'pre')
     if any(suffix in version_parts[-1] for suffix in suffixes):
-        version_parts.remove(version_parts[-1])
+        version_parts.pop()
     # the max pin length is n-1, but in terms of index this is n-2
     max_ver_len = len(version_parts) - 2
     version_parts[max_ver_len] = int(version_parts[max_ver_len]) + 1
--- a/conda_build/skeletons/pypi.py
+++ b/conda_build/skeletons/pypi.py
@@ -700,8 +700,12 @@ def get_package_metadata(package, d, data, output_dir, python_version, all_extra
                     spec = spec_from_line(dep)
                     if '~' in spec:
                         version = spec.split()[-1]
+                        version_parts = version.split('.')
                         tilde_version = '~ {}' .format(version)
-                        max_pin = int(version.split('.')[0]) + 1
+                        # the max pin length is n-1, but in terms of index this is n-2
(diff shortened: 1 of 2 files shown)


nehaljwani:
I could be mistaken, but shouldn't `package ~= 1.4.3` mean `package >=1.4.3,<1.4.4`? https://stackoverflow.com/a/39590286/1005215

nehaljwani:
As mentioned in the spec at https://www.python.org/dev/peps/pep-0440/#compatible-release, the example is:
```
~= 1.4.5
>= 1.4.5, == 1.4.*
```

msarahan:
I need to read the pep again. I'll do whatever the pep says. Thanks @mandeep for following up on this.

msarahan:
My reading of the pep is that for a given spec with n places pinned, the upper bound has n-1 places pinned.

1.2.3  has 3 places pinned, so the upper bound should be 1.3

I can work on this later, Mandeep. It would make a lot of sense to use existing logic for this.

mandeep:
I've committed a change to reflect PEP 440. If there's some existing logic that you want me to take a look at just let me know.

msarahan:
This still needs a tiny bit more work.  As the pep mentions, post-versions need to be ignored.  I think the current count will include them.

```
~= 2.2.post3
>= 2.2.post3, == 2.*
```

mandeep:
Do I need to factor in other release types besides pre, post, and developmental?

msarahan:
Only whatever is listed in pep440.  I think in addition to the 3 you listed, there's also alpha, beta, and rc.  I think alpha and beta may be just one letter, too.

mandeep:
@msarahan I added support for pre and post releases per PEP 440. I also included a test for the new function that I created to make sure versions are converted correctly.

msarahan:
Sorry this slipped past me @mandeep.  This is nice work.  Thanks for fixing this up.

Issue #2400, opened 2017-09-26
Cannot set build feature from environment variable


AntoineD:
I tried to set a debug feature as explained [here](https://conda.io/docs/user-guide/tasks/build-packages/environment-variables.html#environment-variables-to-set-build-features):
`FEATURE_DEBUG=1 conda build xyz`
where the recipe of xyz has a another feature already set.
I get the following traceback:
```
Attempting to finalize metadata for xyz INFO:conda_build.metadata:Attempting to finalize metadata for xyz Traceback (most recent call last):
  File "/abc/anaconda2/bin/conda-build", line 11, in <module>
    sys.exit(main())
  File "/abc/anaconda2/lib/python2.7/site-packages/conda_build/cli/main_build.py", line 388, in main
    execute(sys.argv[1:])
  File "/abc/anaconda2/lib/python2.7/site-packages/conda_build/cli/main_build.py", line 379, in execute
    noverify=args.no_verify)
  File "/abc/anaconda2/lib/python2.7/site-packages/conda_build/api.py", line 185, in build
    need_source_download=need_source_download, config=config, variants=variants)
  File "/abc/anaconda2/lib/python2.7/site-packages/conda_build/build.py", line 1771, in build_tree
    notest=notest,
  File "/abc/anaconda2/lib/python2.7/site-packages/conda_build/build.py", line 881, in build
    output_metas = expand_outputs([(m, need_source_download, need_reparse_in_env)])
  File "/abc/anaconda2/lib/python2.7/site-packages/conda_build/render.py", line 543, in expand_outputs
    for (output_dict, m) in _m.get_output_metadata_set(permit_unsatisfiable_variants=False):
  File "/abc/anaconda2/lib/python2.7/site-packages/conda_build/metadata.py", line 1757, in get_output_metadata_set
    permit_unsatisfiable_variants=permit_unsatisfiable_variants)
  File "/abc/anaconda2/lib/python2.7/site-packages/conda_build/metadata.py", line 638, in finalize_outputs_pass
    fm = finalize_metadata(om, permit_unsatisfiable_variants=permit_unsatisfiable_variants)
  File "/abc/anaconda2/lib/python2.7/site-packages/conda_build/render.py", line 289, in finalize_metadata
    permit_unsatisfiable_variants=permit_unsatisfiable_variants)
  File "/abc/anaconda2/lib/python2.7/site-packages/conda_build/render.py", line 112, in get_env_dependencies
[...]

Issue #2122, opened 2017-06-25
bug using features


neok-m4700:
**environ.py** around l. 645:
```python
def get_install_actions(...)
    # ...
    for feature, value in feature_list:
        if value:
            specs.append('%s@' % feature)
```
is throwing a `AttributeError: 'tuple' object has no attribute 'append'` on specs.

a possible correction could be https://github.com/neok-m4700/conda-build/commit/863f3c0bf4379a678bc5a219d71a10a6769a35d2

msarahan:
Seems like the right fix.  Can you propose a test case?  Is there a recipe you'd recommend as a test case?

neok-m4700:
I've played around with the features https://github.com/neok-m4700/test_features.git

It works ie setting the correct env variable picks the correct feature, is it sufficient ?

msarahan:
fixed by #2428, closing.

PR #2428, opened 2017-10-07
Fix FEATURE_* environment variables

lschr:
The `specs` parameter is passed as a tuple to `environ.get_install_actions`
in various places, which makes appending due to FEATURE_* environment
variables (e.g. FEATURE_NOMKL=1) fail. Therefore, explicitly convert
`specs` to a list.


lschr:
I think this fixes #2400 and #2122.

msarahan:
Test failures look unrelated.  Thanks, this is a good fix.  Merging.

PR #2436, opened 2017-10-12
make conda-verify opt-in, not opt-out.  

msarahan:
also fix some fields related to conda-verify tests

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -762,7 +762,7 @@ def bundle_conda(output, metadata, env, **kw):
 
         # we're done building, perform some checks
         tarcheck.check_all(tmp_path, metadata.config)
-        if not getattr(metadata.config, "noverify", False):
+        if getattr(metadata.config, "verify", False):
             verifier = Verify()
             ignore_scripts = metadata.config.ignore_package_verify_scripts if \
                              metadata.config.ignore_package_verify_scripts else None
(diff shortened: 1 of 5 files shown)


mandeep:
I believe the keyword arguments on Line 771 will need to be changed to match conda verify.

https://github.com/conda/conda-build/blob/master/conda_build/build.py#L771

msarahan:
Yes, good catch.  With this in mind, I think I'll probably need to install conda-build from source in the conda-verify Travis setup, so we can get things right there.  Then we'll merge your PR over there, and finally come back here to merge this PR, and update the conda-verify dependency to >=3.0

mandeep:
Sounds good. Let me know if there's anything I can do to help.

msarahan:
Anything you can do to fix the unicode errors on conda-verify will help a lot.  It seems like the latest build passed, so whatever you tried worked OK.  I think I can take it from here, but if there's something else that comes up, I'll let you know.

I don't think I really appreciated how much work you did on conda-verify.  That was a ton of work.  Nice job.  I guess if you can review the changes I made in rebasing your changes, and make sure I didn't screw anything up, that would be helpful.  I did cut several tests because I reworked things to use conda-build's render function if it was available.  When that's used, conda-build errors out on several of the things that you had tests for as far as invalid metadata.

msarahan:
Merging this now, with the understanding that I'm going to come back and change the conda-verify API to match conda-verify 3.0 after releasing that with @mandeep's PR.

PR #2437, opened 2017-10-12
Use proper temp directory

Bezier89:
Windows consoles by default generally start in C:\. It is common to not have write permission in the root C:\ without escalating privilege. It's better to use a proper temp directory, as the current approach fails in this case unless running as administrator, whereas with these changes it runs fine without running as administrator.

### DIFF
--- a/conda_build/post.py
+++ b/conda_build/post.py
@@ -9,6 +9,7 @@ import locale
 import re
 import os
 import stat
+import tempfile
 from subprocess import call, check_output
 import sys
 try:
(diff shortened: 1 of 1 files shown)


msarahan:
Seems fine.  There are occasionally difficulties with long path lengths in windows, and this may run afoul of them.  It's worth a try, though, to see if this works out OK.  Regardless, users should be able to control behavior here with TEMPDIR or TMPDIR.

msarahan:
There's a new test failure that looks caused by this: https://travis-ci.org/conda/conda-build/jobs/287239908#L2510

I can't take a look at this until early next week.  If you have time to, that's great.  No rush, though.

Bezier89:
Thanks for the heads up. I won't be able to look at it today either, but will try to investigate later.

msarahan:
There was a bug where the dest was the folder, not the file.  Needless to say, things weren't working.  Pure luck they were working before - your PR just exposed a latent bug.  Thanks!

Bezier89:
Awesome! Glad I could help ;-)

PR #2439, opened 2017-10-13
Fix bits wrong win cross compile

msarahan:
Thanks @nehaljwani for figuring this out.  I'll come back in a bit to add tests.

### DIFF
--- a/conda_build/windows.py
+++ b/conda_build/windows.py
@@ -11,8 +11,6 @@ import setuptools  # noqa
 from distutils.msvc9compiler import find_vcvarsall as distutils_find_vcvarsall
 from distutils.msvc9compiler import Reg, WINSDK_BASE
 
-from .conda_interface import bits
-
 from conda_build import environ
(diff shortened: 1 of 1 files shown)


conda_build/windows.py
```diff
@@ -55,7 +53,7 @@ def fix_staged_scripts(scripts_dir):
             with open(join(scripts_dir, fn + '-script.py'), 'wb') as fo:
                 fo.write(f.read())
             # now create the .exe file
-            copy_into(join(dirname(__file__), 'cli-%d.exe' % bits),
+            copy_into(join(dirname(__file__), 'cli-%d.exe' % config.host_arch),
```
nehaljwani: `cli-%d.exe` should be `cli-%s.exe`

nehaljwani:
Can confirm that with the above fix, https://github.com/ContinuumIO/anaconda/issues/1114 should be resolved.

PR #2440, opened 2017-10-14
implement requires_features and provides_features

msarahan:
This is for conda 4.4's new key-value pair support

CC @kalefranz

### DIFF
--- a/conda_build/metadata.py
+++ b/conda_build/metadata.py
@@ -330,6 +330,8 @@ default_structs = {
     'build/preferred_env': text_type,
     'build/preferred_env_executable_paths': list,
     'build/ignore_run_exports': list,
+    'build/requires_features': dict,
+    'build/provides_features': dict,
     'requirements/build': list,
     'requirements/host': list,
     'requirements/run': list,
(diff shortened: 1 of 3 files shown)


msarahan:
Note that any provides_features or requires_features will factor into the package's hash.  This is because aside from a few exclusions, the requirements section and the build section are funneled in wholesale.

Code at https://github.com/conda/conda-build/blob/e10432b28916ad45382b6ba9ab571ee693310a45/conda_build/metadata.py#L1028-L1062

kalefranz:
> Note that any provides_features or requires_features will factor into the package's hash.

I think completely appropriate and would not be intuitive for me if it didn't.

kalefranz:
Once conda 4.4 has some time in the wild, we'll probably want to add a type of warning or notice to `features` and `provides_features` fields under the `build` section.

However, it's fine for a recipe to specify both.  If `provides_features` is specified, conda will completely ignore the `track_features` field, and if `requires_features` is specified, conda will completely ignore the `features` field.

If recipes *do* double-specify, and as we start to use features more, I wouldn't be surprised if we find cases where features are just broken even in 4.3.  Given that we've only really had a single often-used feature associated with `defaults`, I don't think it's ever been exercised enough to consider battle-tested.

jakirkham:
Are there corresponding (dev) docs for these?

mbargull:
Hi @jakirkham, likely you are already aware of this, but if not, I'd like to reference this might be subject to change: https://github.com/conda/conda/pull/6347.

PR #2445, opened 2017-10-17
Exactly match package name in build_folder selection

rendinam:
* Corrects behavior in dirty environments where a package name already encountered in a build session may appear as a prefix or suffix, with our without '_' characters, of other package names.
* Meant to be the final incarnation of the previous fixes here #1931, #1953, which were both insufficiently general to cover all combinations of cases.

### DIFF
--- a/conda_build/config.py
+++ b/conda_build/config.py
@@ -380,7 +380,7 @@ class Config(object):
                                                      "to preserve croot during path joins")
             build_folders = sorted([os.path.basename(build_folder)
                             for build_folder in get_build_folders(self.croot)
-                            if os.path.basename(build_folder).startswith(package_name + "_")])
+                            if build_folder[:build_folder.rfind('_')] == package_name])
             if self.dirty and build_folders:
                 # Use the most recent build with matching recipe name
                 self._build_id = build_folders[-1]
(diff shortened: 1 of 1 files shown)


msarahan:
Heh...  May the good luck of finality be with all of us!  Thanks for taking the latest stab.  Waiting on CI, but code looks good to me.

rendinam:
There's a lot going on in that `build_folders` list comprehension, which tripped me up. Perhaps it makes sense to break it down into two steps for clarity.
`tests/test_api_build.py::test_keep_workdir_and_dirty_reuse` is passing locally with the latest change.

msarahan:
yes, breaking it up for readability is fine for me.  You're making one loop into two (n -> 2n), but the loops should be small and not affect performance too much.

rendinam:
Some quick profiling of this section indicates that the submitted approach is ~10% faster than a single equivalent list comprehension alternative that uses repeated calls to `os.path.basename` and `get_build_folders`, . i.e. 

```
build_folders = sorted([os.path.basename(build_folder) for build_folder in get_build_folders(self.croot)
if os.path.basename(build_folder)[:os.path.basename(build_folder).rfind('_')] == package_name])
```
for cases with many (several 10s of) pre-existing build folders.

The absolute execution time on my machine is between 0.0004s & 0.005s for both approaches for scenarios having up to several hundred preexisting build folders.

msarahan:
wow!  Thanks for doing that extra legwork.  I love it when my instincts are wrong.  Makes me learn new stuff!

rendinam:
Indeed! I didn't know what to expect by looking at it, so it seemed like a good opportunity to collect data and build some intuition; ended up being pleasantly surprised. Thanks for the merge.

PR #2455, opened 2017-10-20
fix implict pinning in outputs; fix incorrect matching of subpkgs

msarahan:
before this, error messages like:

```
conda_build.exceptions.RecipeError: Mismatching hashes in recipe. Exact pins in dependencies that contribute to the hash often cause this. Can you change one or more exact pins to version bound constraints?
Involved packages were:
Mismatching package: py-boost; consumer package: boost
```

showed up with non-top-level variants (such as a top-level shared lib with subpackages that consume it)

This PR also fixes some instances where the pins from the variant weren't being inserted early enough into outputs, which led to them not being differentiated well enough, which led to some of them (again, non-top-level ones) being ignored.

This PR needs tests before merging.

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -252,8 +252,10 @@ def copy_license(m):
 def copy_test_source_files(m):
     test_dest = join(m.config.info_dir, 'test')
     create_all_test_files(m, test_dest)
-    with open(os.path.join(test_dest, 'test_time_dependencies.json'), 'w') as f:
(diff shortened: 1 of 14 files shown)


mingwandroid:
Hi @msarahan, what is the status of this PR? Is it redundant now? My current master branch is rebased on top of these two commits and I want to make a PR.

msarahan:
Thanks for the reminder.  I'll try to revisit this one today and get it merged.  I think it still does important things.

PR #2459, opened 2017-10-23
remove leftovers created by utils.get_recipe_abspath

neok-m4700:
Hi,

Temporaries accumulate in `/tmp` and are never cleared up.
The directory is created upon call to `utils.py->get_recipe_abspath()`: `need_cleanup` is set but never used in `build.py`.

On my system, `/tmp` is a tmpfs, and this leads to a fill up of my RAM, when building a lot of recipes.

Ran the tests: some issues, but unrelated to this PR.
If someone could confirm that this PR doesn't break anything, it'll be great.


msarahan:
Thanks.  CI failures look unrelated.  Nice to have this cleanup actually working.

Issue #2015, opened 2017-05-10
Consider standardizing on SPDX license labels


kalefranz:
There's actually a standard for license identifiers (and it's not whatever Ilan thought of at the time).

> The [SPDX License List](https://spdx.org/licenses/) is a list of commonly found licenses and exceptions used for open source and other collaborative software.  The purpose of the SPDX License List is to enable easy and efficient identification of such licenses and exceptions in an SPDX document (or elsewhere). The SPDX License List includes a standardized short identifier, full name, vetted license text, other basic information, and a canonical permanent URL for each license and exception. By providing a short identifier, users can efficiently refer to a license without having to redundantly reproduce the full license.

Consider making these identifiers our standard for conda-build recipes (and I guess by extension conda-verify).

kalefranz:
Link to SPDX specifications: https://spdx.org/specifications

kalefranz:
This is probably a more general "conda community" discussion.  This ticket should probably just be a placeholder for now.  If any discussion is needed, let's consolidate it at https://github.com/conda/conda/issues/5280.

msarahan:
We can add options for this.  Adding to SciPy2017 sprint.

kalefranz:
Commented in https://github.com/conda/conda-build/pull/2463#issuecomment-340510917 regarding this issue.  There is motivation to relax the `license` field to be able to be a list, so that projects licensing under multiple licenses can be better described.  So that we don't break the existing API, and to tie both concepts together, I suggested a new, and in the future preferred, `spdx_licenses` field.

PR #2463, opened 2017-10-27
create channel-level channeldata.json

kalefranz:
This PR creates the equivalent of the current channel-level metadata.json.  The format is new, agreed to after discussion in https://github.com/conda/conda/issues/5340, and hence the new name.  In this case, there is a `channeldata_version` key, which we'll use if/when we need to change format in the future.

This PR also contains the foundation of a test framework for `conda index`.

The `conda index` command has used a caching strategy of keeping an `.index.json` file, so that rebuilding repodata is quicker on an incremental basis.  This PR adds *four* other caching files, that are in line with the typical conda package structure.  Those files are `.about.json`, `.paths.json`, and `.recipe.json`.  Actually, that's three.  The fourth is a directory `.icons/`, which holds extracted icon files for packages that contain them, as identified in the `app` key of a recipe.

All but the `.paths.json` file are necessary and used to implement the `channeldata.json` file.  It's my intention to use `.paths.json` in a near-future PR.  One purpose will be to add the ability to warn when paths overlap across package names.

I don't believe the addition of these four extra caching files will add significant run time to the `conda index` command.  That's because we have to decompress the entire .tar.bz2 file to extract a single element from it, and so we might as well archive all the components we need in one go.

### DIFF
--- a/conda_build/index.py
+++ b/conda_build/index.py
@@ -166,13 +166,10 @@ def _build_channeldata(dir_path, subdir_paths):
             for fn, record in index_data[subdir_path].items():
                 record.update(about_data.get(subdir_path, {}).get(fn, {}))
                 _source_section = recipe_data.get(subdir_path, {}).get(fn, {}).get('source', {})
+                if isinstance(_source_section, (list, tuple)):
+                    _source_section = _source_section[0]
                 for key in ('url', 'git_url', 'git_rev', 'git_tag'):
-                    try:
-                        value = _source_section.get(key)
-                    except AttributeError:
-                        print("WARNING: bad recipe detected for %s/%s" % (subdir_path, fn))
-                        # AttributeError: 'list' object has no attribute 'get'
-                        value = None
+                    value = _source_section.get(key)
                     if value:
                         record['source_%s' % key] = value
                 package_groups[record['name']].append(record)
@@ -203,7 +200,12 @@ def _build_channeldata(dir_path, subdir_paths):
             best_record = sorted(latest_version_records, key=lambda x: x['build_number'])[-1]
             # Only subdirs that contain the latest version number are included here.
             # Build numbers are ignored for reporting which subdirs contain the latest version of the package.
-            subdirs = sorted(set(rec['subdir'] for rec in latest_version_records))
+            try:
+                subdirs = sorted(set(rec['subdir'] for rec in latest_version_records))
+            except Exception as e:
+                import pdb; pdb.set_trace()
+                subdirs = []
+                assert 1
             package_data[name] = {k: v for k, v in best_record.items() if k in FIELDS}
             package_data[name]['subdirs'] = subdirs
 
--- a/conda_build/index.py
+++ b/conda_build/index.py
@@ -309,9 +309,12 @@ def _read_index_tar(tar_path, lock, locking=True, timeout=90):
                 paths_json = {}
 
             try:
-                recipe_json = yaml.load(t.extractfile('info/recipe/meta.yaml').read().decode('utf-8'))
+                recipe_json = yaml.load(
+                    t.extractfile('info/recipe/meta.yaml').read().decode('utf-8')
+                )
             except Exception as e:
-                log.warn('Error extracting info/recipe/meta.yaml in %s: %r', tar_path, e, exc_info=True)
+                log.warn('Error extracting info/recipe/meta.yaml in %s: %r',
+                         tar_path, e, exc_info=True)
                 recipe_json = {}
 
             # If a conda package contains an icon, also extract and cache that in an .icon/
@@ -398,7 +401,9 @@ def _build_channeldata(dir_path, subdir_paths):
                 value = _source_section.get(key)
                 if value:
                     record['source_%s' % key] = value
-            record['_has_icon'] = bool(recipe_data.get(subdir_path, {}).get(fn, {}).get('app', {}).get('icon'))
+            record['_has_icon'] = bool(
+                recipe_data.get(subdir_path, {}).get(fn, {}).get('app', {}).get('icon')
+            )
             package_groups[record['name']].append(record)
 
     FIELDS = (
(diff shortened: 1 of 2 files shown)


kalefranz:
I don't believe this PR is yet ready to review (although it's close).  I'll remove this message and ping reviewers when it is.

kalefranz:
Let me fix these test failures before anyone starts reviewing...

conda_build/index.py
```diff
@@ -5,78 +5,155 @@
...
-                return json.loads(t.extractfile('info/index.json').read().decode('utf-8'))
-            except EOFError:
-                raise RuntimeError("Could not extract %s. File probably corrupt."
-                    % tar_path)
-            except OSError as e:
-                raise RuntimeError("Could not extract %s (%s)" % (tar_path, e))
-            except tarfile.ReadError:
-                raise RuntimeError("Could not extract metadata from %s. "
-                                "File probably corrupt." % tar_path)
+def get_build_index(subdir, bldpkgs_dir, output_folder=None, clear_cache=False,
```
kalefranz: This function is not modified in this PR, just shifted to a new location in the file.  No need to review.

conda_build/index.py
```diff
@@ -5,78 +5,155 @@
...
 
-def _add_extra_path(extra_paths, path):
-    if isfile(path):
-        extra_paths[basename(path)] = {
-            'size': getsize(path),
-            'timestamp': int(getmtime(path)),
-            'md5': md5_file(path),
-        }
+
+def _ensure_valid_channel(local_folder, subdir, verbose=True, locking=True, timeout=90):
```
kalefranz: This function is not modified in this PR, just shifted to a new location in the file.  No need to review.

goanpeca:
This is awesome @kalefranz thanks!

Some questions based on the file I got by email:

**1.) Should this be a string and have major and minor versions etc... or no need? `"1.0"` ?**
```json
{
  "channeldata_version": 1,
```

**2.)   What is the number to the right, a date?***
```json
"icon_hash": "md5:df7feebede9861a203b480c119e38b49:12760",
```

**3.) Does this**
```json
    "icon_url": "icons/notebook.png",
```
Mean that the path to find the icon will be

- `https://repo.continuum.io/pkgs/main/osx-64/icons/notebook.png` or
- `https://conda.anaconda.org/anaconda/notebooks/osx-64/icons/notebook.png`

It seems currently we have files on `https://conda.anaconda.org/anaconda/notebooks/repodata.json`

**4.) The topic on multiple license within a package has come up several times on the conda-forge gitter channel. I think it is time to reconsider having a list instead of a fixed string?**

```json
 "license": "BSD 3-Clause",
```
vs.
```json
 "license": ["BSD 3-Clause"],
``` 

And how are `"license": "MIT license or UIUC License",` this interpreted, or?

?


**5.) what does `reference_package` refer to, or wy it is needed?**

**6.) I was expecting some info on icons for spyder/spyder or glueviz. Was it because the file @kalefranz sent was still not final?

**7.) I know @teoliphant has asked for extra information like tags on packages. Maybe it is something worth exploring?**

```json
    "gobject-introspection": {
...
      "tags": ["list of sentences or", "key", "word"]
    },
```

kalefranz:
> 1.) Should this be a string and have major and minor versions etc... or no need? "1.0" ?

The integer value is correct, representing an API version for the file.  I've not yet seen semver apply to APIs.  The integer follows similar version fields in things like `info/paths.json`.

> 2.) What is the number to the right?

Format is `hash_type:digest_value:size_in_bytes`.  The size is also often called "length," which is the size of what's being hashed.  It's good practice in general to couple hashes with size to make collision attacks more difficult.

> 3.) 

`icon_url` is a relative path from the root of the channel.  So `"icon_url": "icons/notebook.png"` means https://repo.continuum.io/pkgs/main/icons/notebook.png for the https://repo.continuum.io/pkgs/main channel.  The idea of channeldata was that it be an aggregation of information among all the subdirs, so pointing to something within a subdir doesn't seem appropriate.  Still, there's nothing technically barring us from doing so.  In that case, the value would be something more like `"icon_url": "osx-64/icons/notebook.png"` though.

> 4.) The topic on multiple license within a package has come up several times on the conda-forge gitter channel. I think it is time to reconsider having a list instead of a fixed string?

I don't think that's necessarily an issue for this PR.  However, I'd point to https://github.com/conda/conda-build/issues/2015.  If we implement that, we would probably need to make a new field so that we don't break existing code that depends on the old behavior.  The field could probably be named `spdx_licenses` and be a list.  You could probably chime in there with more thoughts.  If an `spdx_licenses` field gets added to meta.yaml in a future PR, it'll be trivial to add it here as well.

> 5.) what does reference_package refer to, or why it is needed?

The `reference_package` field is a pointer to the package that was used to construct the information.  In this `channeldata.json` file, we're aggregating on package names across subdirs, so knowing the actual source of the information that's showing up will probably be helpful at some point.

> 6.) I was expecting some info on icons for spyder/spyder or glueviz. Was it because the file @kalefranz sent was still not final?

Perhaps the latest packages don't have icon information in them?  https://github.com/AnacondaRecipes/glueviz-feedstock/blob/master/recipe/meta.yaml  Doesn't look like that's the case.  I'll dig in on this issue more.

> 7.) I know @teoliphant has asked for extra information like tags on packages. Maybe it is something worth exploring?

Sure, those can be added.  Would probably need to be added in meta.yaml first though.  We can add those to channeldata.json at any time.  We'll also be able to make additions without incrementing `channeldata_version`, as long as they're optional values anyway.  That is, if we add that key to meta.yaml, it wouldn't be a "required" field in the channeldata.json API because plenty of existing packages wouldn't have it.

> 8.) I would also like to request that conda provides some means or commands to download this info and save it a place that can be queried via the conda info or config command?

I definitely think that's part of the plan.

kalefranz:
cd649c1 fixes the issue 6) above about spyder not having icon_url

goanpeca:
Ok @kalefranz thanks, any chance I could get a new version of the channeldata.json :-p ?

kalefranz:
Here's a snippet of what `channeldata.json` looks like:

```
{
  "channeldata_version": 1,
  "packages": {
    "_ipyw_jlab_nb_ext_conf": {
      "home": "https://github.com/ContinuumIO/anaconda-nb-extensions-meta-config",
      "license": "BSD",
      "reference_package": "osx-64/_ipyw_jlab_nb_ext_conf-0.1.0-py27h172cb35_0.tar.bz2",
      "subdirs": [
        "linux-32",
        "linux-64",
        "osx-64",
        "win-32",
        "win-64"
      ],
      "summary": "A configuration metapackage for enabling Anaconda-bundled jupyter extensions",
      "version": "0.1.0"
    },
    "aiobotocore": {
      "description": "Main purpose of this library to support amazon s3 api, but other services should work (may be with minor fixes). For now it has been tested only for upload/download api for s3. It has been reported that SQS and Dynamo services also work.",
      "dev_url": "https://github.com/aio-libs/aiobotocore",
      "doc_url": "http://aiobotocore.readthedocs.io",
      "home": "https://github.com/aio-libs/aiobotocore",
      "license": "Apache 2",
      "reference_package": "osx-64/aiobotocore-0.4.5-py35h035c80b_0.tar.bz2",
      "source_url": "https://pypi.io/packages/source/a/aiobotocore/aiobotocore-0.4.5.tar.gz",
      "subdirs": [
        "linux-32",
        "linux-64",
        "osx-64",
        "win-32",
        "win-64"
      ],
      "summary": "Async client for aws services using botocore and aiohttp",
      "version": "0.4.5"
    },

...

    "zlib": {
      "description": "zlib is designed to be a free, general-purpose, lossless data-compression library for use on virtually any computer hardware and operating system.",
      "dev_url": "https://github.com/madler/zlib",
      "doc_url": "http://zlib.net/manual.html",
      "home": "http://zlib.net/",
      "license": "zlib",
      "reference_package": "osx-64/zlib-1.2.11-h60db283_1.tar.bz2",
      "source_url": "http://zlib.net/zlib-1.2.11.tar.gz",
      "subdirs": [
        "linux-32",
        "linux-64",
        "osx-64",
        "win-32",
        "win-64"
      ],
      "summary": "Massively spiffy yet delicately unobtrusive compression library",
      "version": "1.2.11"
    }
  },
  "subdirs": [
    "linux-32",
    "linux-64",
    "noarch",
    "osx-64",
    "win-32",
    "win-64"
  ]
}
```

kalefranz:
@goanpeca Can you mark approval on this PR if you're good to go now?

@msarahan I believe this PR is now ready for your review.

msarahan:
Looks good to me.  Thanks for the PR @kalefranz and thanks for the thoughtful review @goanpeca.  Merging.

PR #2466, opened 2017-10-31
implement subspace selection

msarahan:
With zip_keys, it was very frustrating to have a central config file that
specified several zip_keys (python, vc, c_compiler, etc.), but to not allow
easy reduction of that build matrix by other config files.  One variable would
get trimmed to the shorter set of values, but then conda-build would fail with
errors about mismatching zip_keys dimensions.

This commit filters out mismatching entries, but does so in a way that allows
the dimensionality of the composite configuration to be maintained.  This also
allows things like

``conda build somerecipe --python=2.7`` to work when larger python build
matrices are defined in conda_build_config.yaml

### DIFF
--- a/conda_build/config.py
+++ b/conda_build/config.py
@@ -14,7 +14,7 @@ import time
 
 from .conda_interface import root_dir, root_writable
 from .conda_interface import binstar_upload
-from .variants import get_default_variants
+from .variants import get_default_variant
 from .conda_interface import cc_platform, cc_conda_build, subdir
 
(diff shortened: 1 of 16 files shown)


msarahan:
fixed win test with python version.  It begs the question of how to treat python on windows without a zip_keys entry in any variant.  Since we currently can't merge zip_keys groups, I've left zip_keys out on this PR, but it means that passing just 'python' in a variant will not select appropriate compiler keys.  Passing python as a function argument works, though.

PR #2474, opened 2017-11-03
Fix typo in convert platforms list

mandeep:
@msarahan sorry for this. cb tests don't behave well locally for me so I missed this on the initial pass.


msarahan:
np.  Sorry I didn't catch it.

PR #2478, opened 2017-11-06
Check architecture in index.json for platforms other than Linux, macOS, and Windows

mandeep:
I realized the last PR added the ability to convert from architectures like arm to linux, but not the other way around. This PR adds that capability.


msarahan:
Thanks!

PR #2487, opened 2017-11-11
Test noarch Python with Python version constrained

jakirkham:
Adds some tests to make sure that noarch Python works when the Python version is constrained to Python 2 only or Python 3 only.

### DIFF
--- a/tests/test-recipes/metadata/_noarch_python2/meta.yaml
+++ b/tests/test-recipes/metadata/_noarch_python2/meta.yaml
@@ -1,0 +1,23 @@
+package:
+  name: noarch_python2_only_test_package
+  version: "1.0"
+
+source:
+  path: ./noarch_python_test_package
+
+build:
+  script: python setup.py install --single-version-externally-managed --record=record.txt
+  noarch: python
+  entry_points:
+    - noarch_python_test_package_script = noarch_python_test_package:main
+
+requirements:
+  build:
+    - python >=2,<3
+    - setuptools
+  run:
+    - python >=2,<3
+  preferred_env: _env_
+  preferred_env_executable_paths:
+    - test/path
+    - test/path2
--- a/tests/test-recipes/metadata/_noarch_python2/noarch_python_test_package/noarch_python_test_package.py
+++ b/tests/test-recipes/metadata/_noarch_python2/noarch_python_test_package/noarch_python_test_package.py
@@ -1,0 +1,11 @@
+""" This functions as a module but also as entry point.
+"""
+
+answer = 142
+
+
+def main():
+    print(answer + 100)
+
+if __name__ == '__main__':
+    main()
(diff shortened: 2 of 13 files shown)


tests/test-recipes/metadata/_noarch_python2/noarch_python_test_package/noarch_python_test_package.py
```diff
@@ -0,0 +1,11 @@
+""" This functions as a module but also as entry point.
+"""
+
+answer = 142
+
+
+def main():
+    print(answer + 100)
```
nsoranzo: You could use some Python2-only syntax here, like:
```
print >> sys.stdout, answer
```

tests/test-recipes/metadata/_noarch_python2/meta.yaml
```diff
@@ -0,0 +1,23 @@
...
+
+build:
+  script: python setup.py install --single-version-externally-managed --record=record.txt
+  noarch: python
+  entry_points:
+    - noarch_python_test_package_script = noarch_python_test_package:main
+
+requirements:
+  build:
+    - python >=2,<3
```
msarahan: This allows python 3 pre-release versions, and might be fragile.  Please prefer

``python >=2,<3.0a0``

tests/test_api_build.py
```diff
@@ -651,6 +651,26 @@ def test_noarch_python_1(testing_config):
...
 
 
+def test_noarch_python2_only(testing_config):
+    output = api.build(os.path.join(metadata_dir, "_noarch_python2"), config=testing_config)[0]
+    assert package_has_file(output, 'info/files') is not ''
+    extra = json.loads(package_has_file(output, 'info/link.json').decode())
+    assert 'noarch' in extra
+    assert 'entry_points' in extra['noarch']
+    assert 'type' in extra['noarch']
+    assert 'package_metadata_version' in extra
```
msarahan: it would be good to add a step to these tests to try creating an env with both matching and mismatching python versions, to ensure that the desired constraint actually works.

jakirkham: How would you propose doing that? FWICT `conda-build` just builds the recipes in this directory and tries to install them in their test environment. Seems like we have to change the testing infrastructure here without breaking the other recipes and how to do that is a bit unclear to me.

tests/test-recipes/metadata/_noarch_python2/meta.yaml
jakirkham: Should we have the `.0` or should it just be `3a0`?

msarahan: I think you need the .0, but I'm not certain.

tests/test_api_build.py
msarahan: use the environ.create_env function: https://github.com/conda/conda-build/blob/master/conda_build/environ.py#L737

Here's an example of it in use in a test: https://github.com/conda/conda-build/blob/e69e2b57cae76bfa4348565a18497c38a9bc0011/tests/test_environ.py#L16

PR #2488, opened 2017-11-11
Write all `about` metadata fields out

jakirkham:
Not sure if there is a reason to write only some of the `about` fields out and not others. If there is, I guess I'll soon find out. 😄 If there isn't, this should make things a little easier to manage and avoid missing a few things.


jakirkham:
Thoughts on this, @msarahan?

msarahan:
Seems fine to me.  Merging.  Thanks.

Issue #2462, opened 2017-10-24
Adding recipe_url to about


jakirkham:
Would be nice to have a `recipe_url` key in the `about` metadata so that users can find out where the recipe that built this package lives.

jakirkham:
Handling in PR ( https://github.com/conda/conda-build/pull/2489 ).

jakirkham:
Also would add this is useful if we want to include this info on a Anaconda.org to reference back to a package's source.

jakirkham:
(bumping)

PR #2489, opened 2017-11-11
Add a `recipe_url` to `about`

jakirkham:
Fixes https://github.com/conda/conda-build/issues/2462

If Anaconda supports this some day, it will make it easier for people to navigate from the package back to the recipe source. Seems pretty handy if one finds the package has an issue and would like to figure out why. Could also be helpful for someone that is curious.

### DIFF
--- a/conda_build/metadata.py
+++ b/conda_build/metadata.py
@@ -446,7 +446,8 @@ FIELDS = {
     'app': {'entry', 'icon', 'summary', 'type', 'cli_opts',
             'own_environment'},
     'test': {'requires', 'commands', 'files', 'imports', 'source_files'},
-    'about': {'home', 'dev_url', 'doc_url', 'doc_source_url', 'license_url',  # these are URLs
+    'about': {'home', 'dev_url', 'doc_url', 'doc_source_url', # these are URLs
+              'license_url', 'recipe_url',  # these are more URLs
               'license', 'summary', 'description', 'license_family',  # text
               'license_file', 'readme',  # paths in source tree
               },


jakirkham:
Any thoughts on this, @msarahan?

msarahan:
This is something that should be dynamically determined, not hard coded.  Conda-build does try to do this somewhat already: https://github.com/conda/conda-build/blob/master/conda_build/build.py#L225-L235

This info isn't really readily accessible, though.  I do think this field is a good thing to have, but I don't think it should be manually set everywhere.

jakirkham:
So my interest in adding `recipe_url` is setting things up so anaconda.org can link back from packages to feedstocks for instance. Not seeing how dynamically determining this will help.

msarahan:
for example with cb3, you could template it with a required entry in conda_build_config.yaml:

```
base_recipe_url:
  - https://github.com/conda-forge
```

and in meta.yaml:

```
about:
  recipe_url: {{base_recipe_url}}/{{PKG_NAME}}-feedstock
```

jakirkham:
That's a good point. Certainly we can template it in conda-forge (and had planned to do so). Just wouldn't want to force that templating on `conda-build`.

msarahan:
OK.  Please propose some docs over in conda-docs that capture the spirit of this discussion with a warning about templating being a best practice.  With a matching docs PR in place, I'll merge this.

jakirkham:
Sure. That sounds reasonable. Will add to my todos.

As far as templating goes, was planning on picking up this info from the CIs. Musing on that idea for a minute we could pick up this info using an environment variable. The template would look something like this.

```yaml
about:
  {% if environ.get("RECIPE_URL", "") != "" %}
  receipe_url: {{ os.environ["RECIPE_URL"] }}
  {% endif %}
```

Thoughts on this solution?

msarahan:
Yes, that seems like a good way to prevent it from being accidentally included.  I like that idea, even if the if block is a little squirrelly to look at.

jakirkham:
Thanks for the feedback. If you have any ideas of how to make it less squirrely, would be interested to hear them.

jakirkham:
SGTM. Should we prefix it with `CONDA_` or similar since it would be set externally?

msarahan:
Sure, that's safe.  I doubt many other people would set such an env var for other purposes, but you never know.

mbargull:
Another valuable information is the exact commit the package was built from. This could either be stored in a separate key or, if you're not comfortable with introducing to many (specific) new keys, just inside the URL itself. E.g., One of
```yaml
recipe_url: https://github.com/conda-forge/conda-build-feedstock/tree/9ff13a5446ab7fadd6d55969608c1c6e33a7d8ff
```
```yaml
recipe_url: https://github.com/conda-forge/conda-build-feedstock
recipe_tag: 9ff13a5446ab7fadd6d55969608c1c6e33a7d8ff
```
```yaml
recipe:
  url: https://github.com/conda-forge/conda-build-feedstock
  tag: 9ff13a5446ab7fadd6d55969608c1c6e33a7d8ff
```

jakirkham:
Good point @mbargull. Have no strong feelings personally.

Is it reasonable to assume if a recipe has a url, it is probably under some form of version control or are there cases that break with this rule of thumb that either of you can think of?

mbargull:
If someone would use `recipe_url` to point to a description rather a repository, e.g.,
```yaml
recipe_url: http://bioconda.github.io/recipes/snakemake/README.html
```
(for the example, let's just imagine there would be no link to the repository itself xD).
Well, I don't have a real world example, but I don't think we should necessarily assume URL <=> version controlled recipe.

So given that argument, I would take back the `recipe_tag` suggestion, and favor either the generic simple one:
```yaml
recipe_url: https://github.com/conda-forge/conda-build-feedstock/tree/9ff13a5446ab7fadd6d55969608c1c6e33a7d8ff
```
or the extended more specific one (+ using the same sub-keys as used for `source`):
```yaml
recipe:
  git_url: https://github.com/conda-forge/conda-build-feedstock
  git_tag: 9ff13a5446ab7fadd6d55969608c1c6e33a7d8ff
```
```yaml
recipe:
  url: https://github.com/conda-forge/conda-build-feedstock/tree/9ff13a5446ab7fadd6d55969608c1c6e33a7d8ff
```

jakirkham:
So we might want to revisit this soonish. Having split packages raises the question of how to get users with an issue involving a specific package trace back to the right recipe/repo. This would be one mechanism (combined of course with relevant Anaconda changes). Advice on how we would like to proceed at this stage would be most welcome.

jakirkham:
@jezdez , this is the PR we discussed in the last community meeting to put one of the pieces in place to add links from Anaconda.org packages to recipes used to build them

jakirkham:
cc @kathatherine

kathatherine:
@jakirkham Thanks for your patience on this, John. I brought it up to the team here at Anaconda, so they are aware. I'll tag @kenodegard in on this, since he can give you a better update.

jakirkham:
Thanks Katherine! 🙏

I'm also happy to refresh this PR

Though at least in the last discussion it sounded like there is a design discussion we want to have. So hearing from stakeholders about design would be helpful as a first step

To that end here are some ideas to bounce around:
* `recipe_url` in `about` like in the current PR
* `conda-build` CLI flag to bake in `recipe_url` info
* Environment variable (`CONDA_RECIPE_URL`?) to configure this metadata
* A `recipe_url` [variant]( https://docs.conda.io/projects/conda-build/en/latest/resources/variants.html#build-variants ) (like in `conda_build_config.yaml` or similar)
* Some mixture of the above ideas?
* Something else?

PR #2529, opened 2017-11-28
Permit recipes to check for jinja variables without triggering an error.

stuarteberg:
Right now, conda-build will (rightfully) complain if `meta.yaml` tries to use a jinja variable that doesn't exist.  Here's a silly example:

```yaml
about:
  summary: {{SUMMARY}}
```

Error:
```
Undefined Jinja2 variables remain (['SUMMARY']).  Please enable source downloading and try again.
```

So far, so good.  However, it would be nice if `meta.yaml` could at least *check* if a variable is defined or not:

```yaml
about:
  {% if SUMMARY is defined %}
    summary: {{SUMMARY}}
  {% endif %}
```

...but currently, this results in the same error as above, even though it isn't ever *used*.

The problem is that we currently record such undefined names (via `UndefinedNeverFail.all_undefined_names.append()`) as soon as the Undefined variable is constructed, rather than waiting to see if the variable is actually *used* somewhere.  So, there's an easy fix: Just move the `all_undefined_names.append()` statement outside of the `UndefinedNeverFail` constructor and into its methods.

Now `meta.yaml` is free to check any variable to see if it's defined, but it's still an error to use undefined variables.

cc @tingzhao


stuarteberg:
BTW: At first there was a CI failure on Python 2, but I fixed that.  Now the new failures are just due to build timeout, so I think this PR is ready for merge.

msarahan:
Thanks @stuarteberg - looks fine to me.  Merging.

PR #2534, opened 2017-11-30
work on test time with conda 4.4

kalefranz:
This PR takes the edge off, by disabling those safety checks.  I don't think it's a full or "real" solution yet.  It *does* install conda 4.4 correctly.  I recommend merging it.

I also think this shows that the safety checks we're disabling aren't the root cause of what's going on.  If that were the case, the test run time would be closer in line to the 4.3.x runs.

For now though, it at least takes off enough time that we should usually stay under the Travis CI cut-off limit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,14 +46,7 @@ install:
   - conda config --set auto_update_conda False
   - conda update -q --all
   - if [ -n "$CONDA_VERSION" ]; then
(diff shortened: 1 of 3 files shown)


msarahan:
closing this, as @kalefranz isolated the speed issue on the conda side, and we are now enjoying much faster builds.

PR #2535, opened 2017-12-01
which_package speedup

mingwandroid:
This is in 2 commits incase memoizing isn't something we want to do with filelists in dists. AFAICT it should be OK.


msarahan:
Thanks @mingwandroid - excellent addition.

PR #2561, opened 2017-12-11
Add test to show that pin_subpackage does not build in correct order

mingwandroid:
test_subpackage_order_bad() is marked as xfail for now, but we need
to remove this marking when this is fixed.


msarahan:
Merging this - let's talk tomorrow about how soon you might need a release.

PR #2570, opened 2017-12-12
Specify 'jinja2' as an install dependency

### LINKED ISSUE #2565: Jinja2 is a direct dependency for CB
https://github.com/conda/conda-build/blob/3.1.2/conda_build/index.py#L21

Although it is present in conda.recipe, shouldn't this also be specified in `install_requires` in setup.py? I was trying to (in the base env):
```
pip install https://github.com/conda/conda-build/archive/master.zip
```
But it didn't pull in jinja2, which is not part of the Miniconda installers.

nehaljwani:
This is helpful for installing this package via setuptools. The conda
recipe already mentions 'jinja2' as a runtime dependency.

Fixes #2565

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -42,7 +42,7 @@ setup(
                             'conda-skeleton = conda_build.cli.main_skeleton:main',
                             ]},
     install_requires=['conda', 'requests', 'filelock', 'pyyaml', 'conda-verify',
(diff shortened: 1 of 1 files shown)


msarahan:
thanks!

PR #2581, opened 2017-12-15
cache used variables

msarahan:
3.1.4 was a dud release because it was much, much slower.  The test suite wasn't terribly affected because it is so simple.  Rendering the conda-build recipe to build the 3.1.4 release was painfully slow.

This was the main hotspot.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,8 +62,7 @@ install:
       conda install -q flake8;
     else
       conda install -c c3i_test -q perl;
-      conda install -q pytest pip pytest-cov numpy mock;
-      $HOME/miniconda/bin/pip install pytest-xdist==1.16.0 pytest-catchlog pytest-mock;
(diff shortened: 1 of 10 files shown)


goanpeca:
>The test suite wasn't terribly affected because it is so simple. Rendering the conda-build recipe to build the 3.1.4 release was painfully slow.

@msarahan would something like running that test 1000 times and have an average be in within some range help to keep track of changes that might disrupt these times? Something along the lines of %timeit (wih ipython/jupyter?)

msarahan:
The best would be to set up air speed velocity on some internal server.  This slowdown _was_ visible before, I just didn't pay enough attention to it.  The magnitude became much more clear when rendering a non-toy example, but it was there to see with other tests.

msarahan:
appveyor fails pass locally.  I think it has to do with the (global) caching.  I'm leaving that for future work.

PR #2618, opened 2018-01-04
fix matching of partial variants when lining up subpackages

msarahan:
Tests on the way.  Putting this up for testing with @jjhelmus and @nehaljwani

### DIFF
--- a/conda_build/api.py
+++ b/conda_build/api.py
@@ -60,7 +60,7 @@ def render(recipe_path, config=None, variants=None, permit_unsatisfiable_variant
                                     raise
                         output_metas[om.dist(), om.config.variant.get('target_platform'),
                                     tuple((var, om.config.variant[var])
-                                        for var in om.get_used_loop_vars())] = \
+                                        for var in om.get_used_vars())] = \
                             ((om, download, render_in_env))
     return list(output_metas.values())
 
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -1009,9 +1009,6 @@ def build(m, post=None, need_source_download=True, need_reparse_in_env=False, bu
                                     " does not yet support Python 3.  Please handle all of "
                                     "your mercurial actions outside of your build script.")
 
-        utils.insert_variant_versions(m.meta.get('requirements', {}), m.config.variant, 'build')
-        utils.insert_variant_versions(m.meta.get('requirements', {}), m.config.variant, 'host')
-
         exclude_pattern = None
         excludes = set(m.config.variant.get('ignore_version', []))
 
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -1023,6 +1023,12 @@ def build(m, post=None, need_source_download=True, need_reparse_in_env=False, bu
         if excludes or output_excludes:
             exclude_pattern = re.compile('|'.join('(?:^{}(?:\s|$|\Z))'.format(exc)
                                             for exc in excludes | output_excludes))
+
+        # this metadata object may not be finalized - the outputs are, but this is the
+        #     top-level metadata.  We need to make sure that the variant pins are added in.
+        if m.meta.get('requirements'):
+            utils.insert_variant_versions(m.meta['requirements'], m.config.variant, 'build')
+            utils.insert_variant_versions(m.meta['requirements'], m.config.variant, 'host')
         add_upstream_pins(m, False, exclude_pattern)
 
         build_ms_deps = m.ms_depends('build')
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -1026,9 +1026,8 @@ def build(m, post=None, need_source_download=True, need_reparse_in_env=False, bu
 
         # this metadata object may not be finalized - the outputs are, but this is the
         #     top-level metadata.  We need to make sure that the variant pins are added in.
-        if m.meta.get('requirements'):
-            utils.insert_variant_versions(m.meta['requirements'], m.config.variant, 'build')
-            utils.insert_variant_versions(m.meta['requirements'], m.config.variant, 'host')
+        utils.insert_variant_versions(m.meta.get('requirements', {}), m.config.variant, 'build')
+        utils.insert_variant_versions(m.meta.get('requirements', {}), m.config.variant, 'host')
         add_upstream_pins(m, False, exclude_pattern)
 
         build_ms_deps = m.ms_depends('build')
--- a/conda_build/jinja_context.py
+++ b/conda_build/jinja_context.py
@@ -317,7 +317,6 @@ def pin_subpackage(metadata, subpackage_name, min_pin='x.x.x.x.x.x', max_pin='x'
                     break
 
         if not key and not allow_no_other_outputs:
-            import ipdb; ipdb.set_trace()
             raise ValueError("didn't find matching subpackage for name {} with variant {} - this "
                              "is probably a conda-build bug.".format(subpackage_name,
                                                                      metadata.config.variant))
--- a/conda_build/jinja_context.py
+++ b/conda_build/jinja_context.py
@@ -300,14 +300,27 @@ def pin_subpackage(metadata, subpackage_name, min_pin='x.x.x.x.x.x', max_pin='x'
             raise ValueError("Bug in conda-build: we need to have info about other outputs in "
                              "order to allow pinning to them.  It's not here.")
     else:
-        key = (subpackage_name, HashableDict(metadata.config.variant))
         # two ways to match:
         #    1. only one other output named the same as the subpackage_name from the key
         #    2. whole key matches (both subpackage name and variant)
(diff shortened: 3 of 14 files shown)


conda_build/build.py
```diff
@@ -1009,9 +1009,6 @@ def build(m, post=None, need_source_download=True, need_reparse_in_env=False, bu
                                     " does not yet support Python 3.  Please handle all of "
                                     "your mercurial actions outside of your build script.")
 
-        utils.insert_variant_versions(m.meta.get('requirements', {}), m.config.variant, 'build')
```
msarahan: This is removed because at this point, the metadata coming in to build should already have its associated pins.

conda_build/api.py
```diff
@@ -60,7 +60,7 @@ def render(recipe_path, config=None, variants=None, permit_unsatisfiable_variant
                                     raise
                         output_metas[om.dist(), om.config.variant.get('target_platform'),
                                     tuple((var, om.config.variant[var])
-                                        for var in om.get_used_loop_vars())] = \
+                                        for var in om.get_used_vars())] = \
```
msarahan: The old way is fine too.  This matches the way it's done elsewhere.

conda_build/jinja_context.py
```diff
@@ -243,18 +243,23 @@ def pin_compatible(m, package_name, lower_bound=None, upper_bound=None, min_pin=
...
     #    2. whole key matches (both subpackage name and variant)
-    keys = list(outputs.keys())
-    matching_package_keys = [k for k in keys if k[0] == subpackage_name]
+
     if len(matching_package_keys) == 1:
         key = matching_package_keys[0]
+    elif len(matching_package_keys) > 1:
+        key = None
+        for pkg_name, variant in matching_package_keys:
+            shared_vars = set(variant.keys()) & set(metadata.config.variant.keys())
```
msarahan: This matches other outputs with any keys that are common to both metadata objects.  For debugging, the keys are always the (package name, used vars+values).  It used to be (package name, variant) - but that was really big and hard to look at.

msarahan: The other important thing here that this PR fixes is that matching ALL of a variant was breaking things like

libxyz (vs2015)
py-xyz (py35, py36)

Only one of those py packages would match up, because the complete variant would only match once.  Now their common subset will be only the c_compiler, and it'll work.

conda_build/metadata.py
```diff
@@ -44,6 +44,9 @@
 ARCH_MAP = {'32': 'x86',
             '64': 'x86_64'}
 
+output_re = re.compile(r"^\s+-\s(?:name|type):.+?(?=^\w|\Z|^\s+-\s(?:name|type))",
```
msarahan: This is used to match all outputs, each as a single match of this re.  We then index by numeric order according to name matching of rendered outputs section.

conda_build/metadata.py
```diff
@@ -609,6 +612,7 @@ def get_output_dicts_from_metadata(metadata):
     for out in outputs:
         if out.get('name') == metadata.name():
             combine_top_level_metadata_with_output(metadata, out)
+    [trim_empty_keys(out) for out in outputs]
```
msarahan: This could use some future work.  Outputs are coming up with None values for some fields.  That trips up later things, and makes checking values more annoying.  This is a band-aid.  The right fix is to fix the creation of those None values.

conda_build/metadata.py
```diff
@@ -628,41 +632,40 @@ def finalize_outputs_pass(base_metadata, render_order, pass_no, outputs=None,
             if base_metadata.config.verbose:
                 log.info("Attempting to finalize metadata for {}".format(metadata.name()))
             om = base_metadata.copy()
+            om.other_outputs = metadata.other_outputs
```
msarahan: other_outputs is the context of what's available for pin_subpackage.  It's stored on the ``metadata`` object here, but not on ``base_metadata``, which ``om`` is a copy of.  Before we do re-rendering of om's metadata, we need to have other_outputs in place, so it can refer to it for any pin_subpackage stuff it has.

conda_build/metadata.py
```diff
@@ -1706,6 +1725,7 @@ def get_output_metadata_set(self, permit_undefined_jinja=False,
             outputs = get_output_dicts_from_metadata(self)[0]
             output_tuples = [(outputs, self)]
         else:
+            all_output_metadata = OrderedDict()
```
msarahan: This is new - we used to only have the whole set of metadata during finalization.  With parsing whole recipes, though, we need to have that information earlier.  This was necessary for the reparsing of the outputs section.

conda_build/metadata.py
```diff
@@ -1789,6 +1818,27 @@ def get_used_loop_vars(self, force_top_level=False):
...
+                                    # second item: the output text for this metadata
+                                    #    object (might be output)
+                                    self.extract_outputs_text())).rstrip()
+
+        outputs = (yaml.safe_load(self._get_contents(permit_undefined_jinja=False,
+                                template_string=template_string)) or {}).get('outputs', [])
+        self.parse_until_resolved()
+        return get_output_dicts_from_metadata(self, outputs=outputs)
+
+    def get_rendered_output(self, name):
```
msarahan: This is for obtaining the rendered, parsed, dictionary-object representation of an output.  It's not useful for saying what variables are used.  You need earlier, more raw versions of the metadata for that.  It is useful, however, for getting updated, re-rendered contents of outputs.

conda_build/metadata.py
```diff
@@ -1818,11 +1868,11 @@ def _get_used_vars_meta_yaml(self, force_top_level=False):
...
         else:
-            recipe_text = self.get_recipe_text(force_top_level=force_top_level).replace(
-                self.extract_outputs_text().strip(), '')
-
-        recipe_text_without_requirements = recipe_text.replace(
-            self.extract_requirements_text(force_top_level=force_top_level).strip(), "")
+            recipe_text = (self.get_recipe_text(force_top_level=force_top_level).replace(
+                                self.extract_outputs_text().strip(), '') +
+                           self.extract_single_output_text(self.name()))
+        reqs_re = re.compile(r"requirements:.+?(?=^\w|\Z|^\s+-\s(?:name|type))", flags=re.M | re.S)
```
msarahan: This is a more general way of clearing requirements whether they are top-level or output section requirements

conda_build/render.py
```diff
@@ -266,14 +266,15 @@ def get_upstream_pins(m, actions, env):
                 with open(downstream_file + '.yaml') as f:
                     specs = yaml.safe_load(f)
         if not specs and os.path.isfile(pkg_loc):
-            legacy_specs = utils.package_has_file(pkg_loc, 'info/run_exports')
```
msarahan: This order swap was done in the hopes of slightly speeding up this process.  Packages with run_exports.yaml will be in the majority, so hits on those should be a bit faster.

msarahan:
@jjhelmus please take a look.  I may revisit this and put my comments from github here into comments in the code.  Otherwise I'm OK with where this is, and it's ready for you to look at it.

msarahan:
It looks like the appveyor error on py27 is real and needs some investigation.  Please go ahead and review if you have time, but I'll need to figure out the issue there.

msarahan:
Remaining appveyor failure is a fluke.  Passes locally.  This is ready to go.

conda_build/build.py
```diff
@@ -1026,6 +1023,12 @@ def build(m, post=None, need_source_download=True, need_reparse_in_env=False, bu
         if excludes or output_excludes:
             exclude_pattern = re.compile('|'.join('(?:^{}(?:\s|$|\Z))'.format(exc)
                                             for exc in excludes | output_excludes))
+
+        # this metadata object may not be finalized - the outputs are, but this is the
+        #     top-level metadata.  We need to make sure that the variant pins are added in.
+        if m.meta.get('requirements'):
```
jjhelmus: I think `utils.insert_variant_versions(m.meta.get('requirements', {}), m.config.variant, 'build')` could be used to avoid the if block

conda_build/metadata.py
```diff
@@ -691,26 +707,21 @@ def read_meta_file(meta_path):
...
 def combine_top_level_metadata_with_output(metadata, output):
     """Merge top-level metadata into output when output is same name as top-level"""
     sections = ('requirements', 'build', 'about')
     for section in sections:
         metadata_section = metadata.meta.get(section, {})
-        output_section = utils.expand_reqs(output.get(section, {}))
+        if section == 'requirements':
+            output_section = utils.expand_reqs(output.get(section, {}))
+        else:
+            output_section = output.get(section, {})
```
jjhelmus: Is this else statement needed?

jjhelmus:
LGTM, to small questions/comments, but I would approve of merging this as it.

PR #2623, opened 2018-01-09
Actually merge the top-level specs when build_is_host is set

mingwandroid:
.. side question, is m.is_cross not redundant now?


msarahan:
is_cross is a bad name.  As we discussed in chat, that really should be something more like ``has_host``.  We're leaving that for a future PR to keep this one smaller.

PR #2650, opened 2018-01-24
correction

neok-m4700:
```python
  File "<...>/conda_build/api.py", line 188, in build
    need_source_download=need_source_download, config=config, variants=variants)
  File "<...>/conda_build/build.py", line 1923, in build_tree
    notest=notest,
  File "<...>/conda_build/build.py", line 1352, in build
    built_package = bundlers[output_d.get('type', 'conda')](output_d, m, env)
  File "<...>/conda_build/build.py", line 809, in bundle_conda
    files = post_process_files(metadata, initial_files)
  File "<...>/conda_build/build.py", line 710, in post_process_files
    post_build(m, new_files, build_python=python)
  File "<...>/conda_build/post.py", line 592, in post_build
    check_overlinking(m, files)
  File "<...>/conda_build/post.py", line 528, in check_overlinking
    _inspect_file_linking(m, path, files, errors, pkg_name, run_reqs, sysroots)
  File "<...>/conda_build/post.py", line 505, in _inspect_file_linking
    msg_prelude, info_prelude)
  File "<...>/conda_build/post.py", line 440, in _find_needed_dso_in_prefix
    if pkg.dist_name not in m.meta.requirements.host:
AttributeError: 'dict' object has no attribute 'requirements'
```

thanks.

### DIFF
--- a/conda_build/post.py
+++ b/conda_build/post.py
@@ -437,7 +437,7 @@ def _find_needed_dso_in_prefix(m, needed_dso, f, files, errors, run_reqs,
                                                                         and_also))
         for pkg in pkgs:
             print_msg(errors, '{}: {}'.format(msg_prelude, pkg))
-            if pkg.dist_name not in m.meta.requirements.host:
+            if pkg.dist_name not in m.meta.get('requirements', {}).get('host', []):
                 print_msg(errors, '{}: .. but {} not in reqs/host (is transitive)'.
                 format(msg_prelude, pkg.dist_name))
     else:


msarahan:
argh.  Sorry we didn't catch that.  Do you have a recipe that shows this problem?  Our test recipe apparently isn't covering enough.

neok-m4700:
Mmmh it seems to be a corner case, give me 5min to find a not-too-overly-complicated recipe to reproduce ...

neok-m4700:
There you go

**meta.yaml**
```
package:
  name: dummy
  version: 1.0

requirements:
  build:
    - gcc
    - libstdcxx-ng  # break on purpose
```

**build.sh**
```
#!/usr/bin/env bash
# borrowed from www.cprogramming.com/tutorial/shared-libraries-linux-gcc.html
cat >foo.h <<'EOF'
#ifndef foo_h__
#define foo_h__
 
extern void foo(void);

#endif  // foo_h__
EOF

cat >foo.c <<'EOF'
#include <stdio.h>
void foo(void)
{
    puts("Hello, I'm a shared library");
}
EOF

cat >main.c <<'EOF'
#include <stdio.h>
#include "foo.h"
 
int main(void)
{
    puts("This is a shared library test...");
    foo();
    return 0;
}
EOF

gcc -c -Wall -Werror -fpic foo.c
gcc -shared -o libfoo.so foo.o
gcc -L. -Wall -o test main.c -lfoo -lstdc++

cp test $PREFIX/bin
cp libfoo.so $PREFIX/lib
```

msarahan:
excellent, thank you.  I'll write this into a test and push a commit to your PR here.

msarahan:
I can't get your test to behave properly because the gcc package won't install for me in my linux docker container.  I'm merging this as-is, and this will also be affected by #2652

Issue #2598, opened 2017-12-22
Path rewriting corrupts files when building on Windows Subsytem for Linux


steamelephant:
The conda/cookiecutter-conda-python project builds a command line script with a hard-coded path. The rewritten version of the file included in the package starts with the intended content but repeats the end of the file. The file is generated properly when installed by setup.py and when installed in Windows directly and in various more normal Linux installations so I'm hypothesizing some odd ineraction between mmap and the abstracted file system.

This is an edge case but it would be very convenient to be able to build in WSL.  

```python
#!/opt/anaconda1anaconda2anaconda3/bin/python`
# EASY-INSTALL-ENTRY-SCRIPT: 'abc==0.1.0','console_scripts','abc'
__requires__ = 'abc==0.1.0'
import re
import sys
from pkg_resources import load_entry_point

if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
    sys.exit(
        load_entry_point('abc==0.1.0', 'console_scripts', 'abc')()
    )
pkg_resources import load_entry_point

if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
    sys.exit(
        load_entry_point('abc==0.1.0', 'console_scripts', 'abc')()
    )
```
Reducing the package name length makes the duplicate tail start earlier:

```python
#!/opt/anaconda1anaconda2anaconda3/bin/python
# EASY-INSTALL-ENTRY-SCRIPT: 'ab==0.1.0','console_scripts','ab'
__requires__ = 'ab==0.1.0'
import re
import sys
from pkg_resources import load_entry_point

if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
    sys.exit(
        load_entry_point('ab==0.1.0', 'console_scripts', 'ab')()
    )
m pkg_resources import load_entry_point

if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
    sys.exit(
        load_entry_point('ab==0.1.0', 'console_scripts', 'ab')()
    )
```

msarahan:
If there's something we can do about it, we'll be happy to, but WSL stuff can get pretty far over my head pretty quickly.  If you can isolate this any further, please let us know.

[...]

PR #2655, opened 2018-01-26
Update build.py to fix rewriting problem when running in WSL

steamelephant:
This change closes the memory map before rewriting it. This prevents problems when running on WSL and does not seem to cause problems in Windows or Linux. This is a fix for issue #2598


msarahan:
Excellent!  Thanks for submitting this, and sorry we didn't act quickly enough on your earlier recommendation of this fix.  Will wait on CI, but merge assuming everything comes out OK.

msarahan:
CI failures look unrelated.  Merging.  Thanks @steamelephant

PR #2659, opened 2018-01-26
fix hash handling of numpy pinning

### LINKED ISSUE #2658: `pin_compatible` has no effect
I created a recipe for the `tifffile` package that uses variants to build multiple versions. Unfortunately, it does not work. `conda build --output -m conda_build_config.yaml tifffile` yields
```
.../conda-bld/linux-64/tifffile-0.13.5-py35_nomklhba1e662_3.tar.bz2
.../conda-bld/linux-64/tifffile-0.13.5-py35hb4372cb_3.tar.bz2
.../conda-bld/linux-64/tifffile-0.13.5-py36_nomklhba1e662_3.tar.bz2
.../conda-bld/linux-64/tifffile-0.13.5-py36hb4372cb_3.tar.bz2
```
There should be 12 outputs (2 Python versions, 2 BLAS implementations, 3 numpy versions). It seems as if `pin_compatible("numpy", min_pin="x.x", max_pin="x.x")` has no effect.

Recipe and build config are attached below.
[meta.yaml](https://github.com/conda/conda-build/files/1668545/meta.yaml.txt)
[conda_build_config.yaml](https://github.com/conda/conda-build/files/1668525/conda_build_config.yaml.txt)

msarahan:
fixes #2658 

The problem here was an unhappy medium between ``numpy x.x`` and having numpy unpinned.  ``numpy x.x`` worked ok, but the equivalent pin_compatible expression wasn't being treated correctly.

There are 3 outcomes with numpy:

1. ``numpy x.x`` or ``{{ pin_compatible('numpy', max_pin='x.x') }}`` - this is what gets you the npXYY in the build string.  You can have min_pin set to anything you want.  The key is having max_pin with **two** x's.
2. usage of ``{{ pin_compatible('numpy') }}`` either without max_pin arguments at all, or with arguments other than ``'x.x'`` - this makes numpy contribute to the hash
3. No usage of pin_compatible - this makes numpy not contribute to the hash

### DIFF
--- a/conda_build/metadata.py
+++ b/conda_build/metadata.py
@@ -52,6 +52,12 @@ ARCH_MAP = {'32': 'x86',
 #    regex, which extract all outputs in order.
 output_re = re.compile(r"^\s+-\s(?:name|type):.+?(?=^\w|\Z|^\s+-\s(?:name|type))",
                        flags=re.M | re.S)
+numpy_xx_re = re.compile(r'(numpy\s*x\.x)|pin_compatible\([\'\"]numpy.*max_pin=[\'\"]x\.x[\'\"]')
+# TODO: there's probably a way to combine these, but I can't figure out how to many the x
+#     capturing group optional.
+numpy_compatible_x_re = re.compile(
+    r'pin_\w+\([\'\"]numpy[\'\"].*((?<=x_pin=[\'\"])[x\.]*(?=[\'\"]))')
+numpy_compatible_re = re.compile(r"pin_\w+\([\'\"]numpy[\'\"]")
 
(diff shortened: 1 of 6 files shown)


mingwandroid:
LGTM; do we have a deprecation plan for `x.x`? Has this been announced if so?

msarahan:
It has been marked as deprecated with a warning showing up since 3.0.  This PR actually gets rid of the warning, since I'm not defining the numpy_xx check to be more broad.  https://github.com/conda/conda-build/pull/2659/files#diff-c86934c27c2f25579540ddb447fa45bdL1526

I could split it out and use two separate regexes.

PR #2661, opened 2018-01-26
implement exclusive config file

msarahan:
This is intended to address @jakirkham comments from https://github.com/conda-forge/conda-smithy/pull/625

This provides a slightly different CLI option and API argument to specify a config file that overrides the usual search in HOME and cwd.  The thought here was to help people avoid "oops" situations where they meant to use some specific config file, but their system-wide defaults crept in.

With an exclusive config file, the hierarchy is:

1. exclusive config file
2. config file in the recipe (alongside meta.yaml)
3. any additional files specified on the CLI with --variant-config-files/-m

### DIFF
--- a/conda_build/cli/main_render.py
+++ b/conda_build/cli/main_render.py
@@ -96,10 +96,18 @@ source to try fill in related template variables.",
     )
     p.add_argument(
         '-m', '--variant-config-files',
-        dest='variant_config_files',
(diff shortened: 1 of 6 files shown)


jakirkham:
Thanks Mike. Think this captures our intended use case well.

Sorry I was a little quiet, was at a conference.

PR #2665, opened 2018-01-28
Error out if version computation fails

nehaljwani:
If git is not in PATH, get_version() returns 'unknown+0' instead of
erroring out, which affects the output for `conda-build -V`. This patch
addresses the issue by erroring out during creation of package if
version cannot be computed.


msarahan:
Good fix.  Thanks.

PR #2685, opened 2018-02-17
Record build stats

msarahan:
The primary purpose of the build stats here is to feed into CI systems that have different classes of workers.  We'll be aggregating the needs of given package builds, and then we'll use that information to dispatch jobs to the smallest worker that is capable (and available).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,7 +59,7 @@ install:
         conda info;
     fi
   - conda install -q anaconda-client requests filelock contextlib2 jinja2 patchelf python=$TRAVIS_PYTHON_VERSION
-  - conda install -q pyflakes conda-verify beautifulsoup4 chardet pycrypto glob2
+  - conda install -q pyflakes conda-verify beautifulsoup4 chardet pycrypto glob2 psutil
   - pip install pkginfo
   - if [[ "$FLAKE8" == "true" ]]; then
       conda install -q flake8;
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,7 +45,7 @@ install:
   - conda config --set always_yes yes
   - conda config --set auto_update_conda False
   - if [[ "$TRAVIS_PYTHON_VERSION" == "2.7" ]]; then
-      conda install -q futures;
+      conda install -q futures scandir;
     fi
   - conda update -q --all
   - if [ -n "$CONDA_VERSION" ]; then
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -44,6 +44,7 @@ install:
   - python -c "import sys; print(sys.prefix)"
   - conda update -q --all
   - conda install -q pip pytest pytest-cov jinja2 patch flake8 mock requests contextlib2 chardet glob2 perl pyflakes pycrypto posix m2-git anaconda-client numpy conda-verify beautifulsoup4 pytest-xdist pytest-mock filelock pkginfo psutil
+  - if "%PYTHON_VERSION%" == "2.7" conda install -q scandir
   # this is to ensure dependencies
   - python --version
   - python -c "import struct; print(struct.calcsize('P') * 8)"
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -43,7 +43,7 @@ install:
   - python -c "import sys; print(sys.executable)"
   - python -c "import sys; print(sys.prefix)"
   - conda update -q --all
-  - conda install -q pip pytest pytest-cov jinja2 patch flake8 mock requests contextlib2 chardet glob2 perl pyflakes pycrypto posix m2-git anaconda-client numpy conda-verify beautifulsoup4 pytest-xdist pytest-mock filelock pkginfo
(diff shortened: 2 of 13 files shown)


mingwandroid:
What about the number of parallel processes used? Can that be collected somehow? It is an important metric for determining needed CPU count.

msarahan:
How so?  It'll either be 1 or whatever $CPU_COUNT is for the machine, right?  I don't think the number of CPUs used matters as much as the simple fact of whether or not a recipe benefits from having more cores.

The one place I'll eat my words, of course, is that the number of CPUs used will also be tied to the memory usage.  Maybe the best way to express memory usage then is in terms of memory used / total core count, in addition to total memory used?

The CPU stats for psutil are all global (https://psutil.readthedocs.io/en/latest/#cpu).  I'm not sure how I'd count this.  Maybe I could count the number of child processes?  That might be off by quite a bit, though.

mingwandroid:
Replying via email since the spot I'm at has blocked github for some reason.

When it is not CPU_COUNT for the bulk of the build process - time at random
intervals and count # of children? perhaps look for fluctuations? - we know
that the build system is not parallelizable. Even though you can pass -jN
to any GNU make build, many projects explicitly disable this feature since
they have race conditions in their build scripts.

msarahan:
I haven't verified this, but I'm pretty sure you'll have total process counts that don't match the CPU_COUNT.  Because we'll have CPU_COUNT *compiler* processes, I anticipate we'll have CPU_COUNT + "some" processes for parallel builds.  "some" might be make itself, or perhaps other process-spawning programs?  I think we only care about leaf nodes in the process tree, but I'm not sure there's any way to get that.

Off-by-one-or-two won't be the end of the world, but I do think we need to keep it in mind.

mingwandroid:
There'll definitely be challenges involved in detecting this. I think a pattern of a high degree of parallelism the falling back to something a lot less is an indicator that it's happened.

An easy alternative is to run -jSOMETHING_REALLY_BIG and see that you reach >=SOMETHING_REALLY_BIG child processes.

I think a number around 40 would be good for SOMETHING_REALLY_BIG. It would be pretty unlikely that any build system that is not parallelized would hit anywhere near 40.

msarahan:
Process counts look strange, as expected:

```
Resource usage statistics from testing conda-build:
   Process count: 4
   CPU time: Sys=0:00:01.5, User=0:00:04.9
   Memory: 67.9M
   Disk usage: 2.5K
   Time elapsed: 0:00:12.3
```

I think the total CPU time might be the most important thing we're after, though.  That'll be roughly invariant to the number of processes, right?  I still don't understand why we care exactly how many processes are used - this is more a product of which machine a build is running on than a product of a recipe's settings (beyond whether parallel compilation is enabled or not).

```
Resource usage summary:

Total time: 0:01:26.0
CPU usage: sys=0:00:01.6, user=0:00:05.2
Maximum memory usage observed: 67.9M
Total disk usage observed (not including envs): 227.7K
```

mingwandroid:
The reason we want to know the process count is to determine if parallelism is worthwhile. Most recipes will not benefit from a worker with more than 1 CPU while some will benefit hugely from a worker with 16 CPUs.

> I still don't understand why we care exactly how many processes are used - this is more a product of which machine a build is running on than a product of a recipe's settings (beyond whether parallel compilation is enabled or not).

No, the number of processes (when *not* capped to some low limit due to, for example, having detected `CPU_COUNT`) will not depend on the number of CPUs on the worker, and this allows us to figure out whether in future this recipe should be built on a worker with a high CPU count, so we need a way to set an artificially high `CPU_COUNT` and then to detect that the concurrent number of child processes reaches this number.

msarahan:
OK

> set an artificially high CPU_COUNT and then to detect that the concurrent number of child processes reaches this number.

This is out of scope for this PR.  It does capture number of child processes, so it should enable you to do what you propose here.  The actual procedure of doing that, though, is something I see as an optimization step outside of the normal build process.  I think you would want to do that with some loop over all (or many) recipes, and figure out which ones to mark as "wants parallel" or whatever, then go back to normal operation.  Is that accurate?

jjhelmus:
Comparing the wall time vs CPU time of the build would give you a rough estimate of how much the build benefited from parallelism.  There are a items that could effect this number but if CPU time > wall time you know that parallelism helped some.

mbargull:
Ah, @jjhelmus, just beat me to it! I was also going to suggest to do a very abstract `real`/`cpu` time comparison. This isn't most accurate, but might give you a reasonable estimate.

mingwandroid:
> This is out of scope for this PR. It does capture number of child processes, so it should enable you to do what you propose here. The actual procedure of doing that, though, is something I see as an optimization step outside of the normal build process. I think you would want to do that with some loop over all (or many) recipes, and figure out which ones to mark as "wants parallel" or whatever, then go back to normal operation. Is that accurate?

Yes that's right, but I think I need a way to set `CPU_COUNT` to my large number (I do not like the cpu time vs wall time idea, it's too sensitive to CPU count and how the build system in question operates wrt sleeping/polling).

msarahan:
You would set CPU_COUNT as an environment variable on the build worker.  You could do this as a prefix_command in c3i.  It will get passed through (conda-build won't override it with its auto-detection).  I think the PR here gives everyone the tools they need, and you all can crunch numbers and play with variables.

PR #2686, opened 2018-02-20
reverse build/host activation order, to give build exe's higher priority

msarahan:
@mingwandroid reported problems building R recipes where he needed the compilers to be in the host section for some reason.  This PR is a WIP as we explore this idea - it might break stuff pretty badly.


mingwandroid:
> he needed the compilers to be in the host section for some reason

This is because R users expect to be able to call install.packages() from within the REPL and for things involving compilation that requires compilers (much the same as on Python with `sysconfig`).

I want our R infrastructure (and MRO) on Linux to use our compilers rather than reporting an error or building broken software.

PR #2687, opened 2018-02-20
save parent recipe in subfolder of outputs for info/recipe

msarahan:
@mingwandroid commented that he had a hard time finding a recipe for curl, which is a split package where neither output matches the top-level name.  We were not copying the original top-level recipe into packages - only the rendered metadata.

This PR puts the top-level recipe into all output packages, in the "parent" subfolder, and maintains the previous behavior otherwise.


mingwandroid:
Great, thanks.

PR #2694, opened 2018-02-21
reduce final specs to tightest possible single spec for each pkg name

msarahan:
meant to address https://github.com/ContinuumIO/anaconda/issues/1283


conda_build/render.py
```diff
@@ -329,6 +330,70 @@ def add_upstream_pins(m, permit_unsatisfiable_variants, exclude_pattern):
...
+            elif len(spec_parts) == 3:
+                constraint = _Constraint(None, None, " ".join(spec_parts[1:]))
+            else:
+                lower_bound = re.search(r'>=?(.*?(?=,|\Z))', spec_parts[1])
+                if lower_bound:
+                    lower_bound = lower_bound.group(1)
+                upper_bound = re.search(r'<=?(.*?(?=,|\Z))', spec_parts[1])
+                if upper_bound:
+                    upper_bound = upper_bound.group(1)
+                constraint = _Constraint(lower_bound, upper_bound, None)
```
jjhelmus: If the regexes fail to match then no constraints are enforced.  Will this break something like `numpy ==1.11`?

msarahan: yeah, I have probably missed several cases.  ``1.21.*`` specs, too.  Please post them here or add them to the tests - I think you can just push to my branch if you want.

PR #2700, opened 2018-02-22
Add relative path support for load_setup_py_data

mariusvniekerk:
When doing something like
```python
load_setup_py_data(setup_file='../setup.py', from_recipe_dir=True)
```

load_setup.py data did not add the correct path to sys.path.
This fix allows this to work cleanly


msarahan:
Thanks.  Merging.

PR #2702, opened 2018-02-22
[discuss] add option --debug_build

looooo:
as we need a way to create debug builds, it would be nice to set this with an option. I guess this is not the right way to do it, but maybe we can start a discussion on how to implement something which adds the ability to build recipes in debug-mode.

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -1081,6 +1081,9 @@ def build(m, stats, post=None, need_source_download=True, need_reparse_in_env=Fa
     if env_path_backup_var_exists:
         env["CONDA_PATH_BACKUP"] = os.environ["CONDA_PATH_BACKUP"]
 
+    if config.debug_build:
+        env["DEBUG_BUILD"] = "True"
+
     # this should be a no-op if source is already here
     if m.needs_source_for_render:
         try_download(m, no_download_source=False)
--- a/conda_build/cli/main_build.py
+++ b/conda_build/cli/main_build.py
@@ -301,6 +301,9 @@ different sets of packages."""
     p.add_argument('--stats-file', help=('File path to save build statistics to.  Stats are '
                                          'in JSON format'), )
 
+    p.add_argument('--debug-build', help=('Sets a environment-variable DEBUG_BUILD, which can be unresolved ',
+                                          'in the meta_yaml and build files.'), )
+
     add_parser_channels(p)
 
     args = p.parse_args(args)


msarahan:
I think this is fine.  The capability is definitely needed.  Most of the work, though, is in figuring out how to get the subprocess shells to be interactive when you want to be debugging them, but not interactive otherwise.  Since we only support bash and batch, it shouldn't be too much work, but it's more than I can do right now.

Thanks for opening this PR.  Maybe you and others can build here.

looooo:
This is more about simple compiling packages with eg.: `-D CMAKE_BUILD_TYPE=Debug`.

Using conda-packages with vs2015/ kdevelop is working with some hacks. But we need packages to be in debug-mode to debug them with the debugging tools from the ide. Especially this is true for windows.

I do the following to interactively work on a c++ code with conda-dependencies:
- run conda build on a recipe where the path to the source is specified directly in the build.sh / build.bat
- abort after the setup-step is done (eg. cmake)
- link the created _b_env into the envs directory.
- open the ide with a script where the linked env is activated first, and some other stuff (windows, UCRT) is called (similar things are set during conda-build)
- sometimes some additional variables have to be set in the ide's

mingwandroid:
IMHO `conda-build` should play no part in deciding that a build is 'debug' or 'release' or 'super-dooper-whizz-bang-mode'. 

Instead this determination should be in the compiler activation scripts and to switch to debug mode, you activate a set of debug compilers. Then, to decide whether to build a package in debug mode or release mode you use conda-build's -m flag which references an appropriate build matrix file. This way you can have debug and release packages built at the same time if you wish.

To see how this works (you can already do it on Linux and macOS):

cat ~/conda/aggregate/conda_build_config.yaml

```
c_compiler:
  - gcc  # [linux]
  - clang  # [osx] 
cxx_compiler:
  - gxx  # [linux]
  - clangxx  # [osx] 
```

cat ~/conda/aggregate/conda_build_config-dbg.yaml

```
c_compiler:
  - gcc-dbg  # [linux]
  - clang-dbg  # [osx] 
cxx_compiler:
  - gxx-dbg  # [linux]
  - clangxx-dbg  # [osx] 
```

To build in non-debug mode:

```
conda build $RECIPE_DIR -m conda_build_config.yaml
```

To build in debug mode:

```
conda build $RECIPE_DIR -m conda_build_config-dbg.yaml
```

This ties in with conda-build 3's `compiler()` jinja2 stuff and these compiler activation packages already exist on the Anaconda Defaults channel. Nothing's been done for Windows yet though.

We do need a way to augment the build string of the packages though so that different variants do not clobber each other's output files and, for python-debug (where the packages are not compatible with python-non-debug) and similarly for Windows with it's _ITERATOR_DEBUG_LEVEL and the different runtime DLLs we need a way to describe and delineate this package (in)compatibility.

msarahan:
Ray said it better than I could have.  I'd only add that if debug is a variant in conda_build_config.yaml, it won't need to be added to the filenames because it'll be part of the hash.  It may still be nice to have it in the filename, but it won't be strictly necessary.  Then there's the question we haven't quite answered yet about how one should tell conda "I want debug variants of packages."

looooo:
>Instead this determination should be in the compiler activation scripts and to switch to debug mode, you activate a set of debug compilers. Then, to decide whether to build a package in debug mode or release mode you use conda-build's -m flag which references an appropriate build matrix file. This way you can have debug and release packages built at the same time if you wish.  

Sounds good to me. So if I want to have debug-packages of dependencies, I can do the same as anaconda does with creating a repo which have all the dependencies as submodules. Afterwards I build all of them with the  `-m build_matrix.yaml` which includes the debug compilers.

But how does the compiler-specification add the `-DBUILD_TYPE="Debug" in th bld.bat / build.sh. Or is the proposed solution a more general way to select the type of the build. If it is really possible to set the build type without touching the build.sh/bld.bat this would be fantastic.

mingwandroid:
The debug compilers set an env. var called `[OPT_CFLAGS](https://github.com/AnacondaRecipes/aggregate/blob/master/ctng-dbg-compilers-activation-feedstock/recipe/activate-gcc.sh#L138-L139)` (which is to be used in your script when you want to avoid the normal `CFLAGS` which are the debug ones in this case), so you should adapt build.sh to check if `OPT_CFLAGS` exists and pass a different value for [this flag](https://github.com/AnacondaRecipes/boost-feedstock/blob/master/recipe/build.sh#L38).

mingwandroid:
> As I understand it this makes only sense for packages which depend on libgcc. Wouldn't it be possible to handle this the same way as python-versions are handled? There also no package with a different python version is included...
> Another way would be using a `feature` (this should be sufficient for linux).

It's got *nothing* to do with libgcc, and no, `features` are terrible things we try to avoid (binary and essentially hacks to the solver). I'd rather use a completely different channel for debug builds than any other option I've seen so far.

PR #2708, opened 2018-02-26
Use my latest version of check_overlinking

mingwandroid:
It diverged a lot from the upstream one, and the split between in_prefix and in_system is less clear-cut
now that whitelists are supported.

Should whitelists be in conda_build_config.yaml and not hardcoded at all? I am leaning that way personally.


mingwandroid:
Hey Mike, this is the most important change to get into the next conda-build if possible.

mingwandroid:
I would like conda-forge to enable these things as errors so they get their recipes fixed.

PR #2710, opened 2018-02-26
skeleton (cran): Add support for building MRO and various other impro…

mingwandroid:
…vements.

* Add support for using `noarch: generic` for packages that do not need compilation.
  To enable this, pass `--use-noarch-generic`.

* Add support for using Rtools on Windows when building a package from source. To
  enable this, pass `--use-rools-win`.

* Add support for binary repackaging (CRAN/MRAN has binaries for Windows and macOS)
  To enable this, pass `--use-binaries-ver=x.x`. If binaries are not found for any
  platform(s) then it will be built from source on those platforms (this is always
  the case on Linux at present).

* Add support for creating a recipe from a source tarball (identified via file://)
  This is needed in-case someone passes you a pre-release source tarball or if you
  are packaging your own local-only/proprietary R package.

* Use jinja2 for `cran_mirror`. This is so recipes can be shared more easily between
  MRO and non-MRO R. It is expected that you will set `cran_mirror` in
  `conda_build_config.yaml`.

### DIFF
--- a/conda_build/skeletons/cran.py
+++ b/conda_build/skeletons/cran.py
@@ -31,9 +31,7 @@ from conda_build.license_family import allowed_license_families, guess_license_f
 from conda_build.utils import rm_rf
 
 SOURCE_META = """\
-  {fn_key} {filename}{sel}
-  {url_key}{sel} {cranurl}
-  {hash_entry}{sel}
+  {archive_keys}
   {git_url_key} {git_url}
   {git_tag_key} {git_tag}
   {patches}
@@ -471,10 +469,16 @@ def yaml_quote_string(string):
     return yaml.dump(string, Dumper=SafeDumper).replace('\n...\n', '').replace('\n', '\n  ')
 
 
-def clear_trailing_whitespace(string):
+# Due to how we render the metadata there can be significant areas of repeated newlines.
+# This collapses them and also strips any trailing spaces.
+def clear_whitespace(string):
     lines = []
+    last_line = ''
     for line in string.splitlines():
-        lines.append(line.rstrip())
+        line = line.rstrip()
+        if not (line == '' and last_line == ''):
+            lines.append(line)
+        last_line = line
     return '\n'.join(lines)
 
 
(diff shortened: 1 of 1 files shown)


msarahan:
Can we set a default hardcoded cran_mirror, so we don't require someone to use conda_build_config.yaml out of the box?

Is it worthwhile to make noarch: generic packages default?  Have you used them enough to have a feeling for whether you trust them?

mingwandroid:
> Can we set a default hardcoded cran_mirror, so we don't require someone to use conda_build_config.yaml out of the box

I don't really know how that would work, if you emit:

```
{% cran_mirror = 'https://cran.blah.org' %}
```

.. wouldn't that prevent the one in `conda_build_config.yaml` being used?

> Is it worthwhile to make noarch: generic packages default? Have you used them enough to have a feeling for whether you trust them?

I've tested them but they require a reworking of the file layout of R on Windows so that it follows the layout on Unix. I've done this for our R and for MRO but conda-forge have not yet done it AFAIK so they wouildn't want this to be the default. Also I cannot make any use of it unfortunately for other reasons.

msarahan:
In variants.py, up near the top:

```
DEFAULT_VARIANTS = {
    'python': '{0}.{1}'.format(sys.version_info.major, sys.version_info.minor),
    'numpy': '1.11',
    # this one actually needs to be pretty specific.  The reason is that cpan skeleton uses the
    #    version to say what's in their standard library.
    'perl': '5.26.0',
    'lua': '5',
    'r_base': '3.4',
    'cpu_optimization_target': 'nocona',
    'pin_run_as_build': {'python': {'min_pin': 'x.x', 'max_pin': 'x.x'},
                         'r-base': {'min_pin': 'x.x.x', 'max_pin': 'x.x.x'}},
    'ignore_version': [],
    'ignore_build_only_deps': ['python'],
    'extend_keys': ['pin_run_as_build', 'ignore_version', 'ignore_build_only_deps'],
}
```

Add an entry in there for cran_mirror, so that a value is not strictly required in every conda_build_config.yaml file.

mingwandroid:
Ah ok I will do this later.

mingwandroid:
Done.

msarahan:
It would be a good idea to rebase, before you push any more changes.  I have sped things up quite a bit and fixed up the flake8 stuff.

mingwandroid:
Did I force push? Sorry if so.

msarahan:
no, I don't think you did.  I just don't want to - I'm recommending that you do it before your next update, because I sped up the testing by about 10 minutes per run.

msarahan:
Windows is my fault.  All looks fine here.

PR #2712, opened 2018-02-26
Allow per-output {pre,post}-{un,}link scripts

mingwandroid:


conda_build/build.py
```diff
@@ -115,18 +115,26 @@ def create_post_scripts(m):
...
     '''
-    recipe_dir = m.path
+    recipe_dir = (m.path or
+                  m.meta.get('extra', {}).get('parent_recipe', {}).get('path'))
     ext = '.bat' if utils.on_win else '.sh'
     for tp in 'pre-link', 'post-link', 'pre-unlink':
-        src = join(recipe_dir, tp + ext)
+        # To have per-output link scripts they must be prefixed by the output name.
+        tp += ext
+        if hasattr(m, 'other_outputs'):
```
msarahan: other_outputs might work, but it's a little vestigial.  It was what I originally based the "used variable" stuff on - tried to keep a relatively raw metadata object, and then compared it to a more fully-rendered one down the road.

Perhaps we should have post/pre-link scripts be manually specified somehow in meta.yaml for each output rather than expecting a particular name?  Ultimately, the filename in the package will be the standard name, since that's what conda is expecting.  This is how the test scripts work for outputs - you specify them by name, and then they are renamed when packing things up.

PR #2715, opened 2018-02-27
Adds a new jinja function: resolved_packages

### LINKED ISSUE #2404: Add an option to export build time dependencies as run_constrained keys and pin strictly
Following the discussion here: https://github.com/conda/conda-build/pull/2001#issuecomment-332327309, I am proposing to add the following feature to conda build.
Currently, there is an option in conda build recipes which allows you to pin **all** the packages that were used during build time as runtime dependencies. See original pull request here: https://github.com/conda/conda-build/pull/741

For example a recipe like:
```
package:
  name: bob-devel

build:
  pin_depends: strict

requirements:
  run:
    - curl 7.52.1
```
will have the following `index.json`
```json
{
  "depends": [
    "curl 7.52.1 0",
    "openssl 1.0.2l 0",
    "zlib 1.2.8 3",
  ],
  "name": "bob-devel",
}
```
where `openssl` and `zlib` are dependencies of `curl`. Note that everything is pinned strictly and normally dependencies of `zlib` and `openssl` get pinned strictly too but I did not put it in the example above.

I propose a very similar functionality to be added for `run_constrained` keys as well.
For example for the following recipe:
```
package:
  name: bob-devel

build:
  pin_constrains: strict

run_constrained:
  - curl 7.52.1
```
we get this:
```json
{
  "constrains": [
    "curl 7.52.1 0",
    "openssl 1.0.2l 0",
    "zlib 1.2.8 3",
  ],
  "name": "bob-devel",
}
```

183amir:
Fixes #2404

### DIFF
--- a/conda_build/jinja_context.py
+++ b/conda_build/jinja_context.py
@@ -429,6 +429,49 @@ def cdt(package_name, config, permit_undefined_jinja=False):
     return result
 
 
+def all_packages(m, env, permit_undefined_jinja=False, bypass_env_check=False):
+    """Returns the final list of packages that are listed in host or build.
+    This include all packages that will be installed in the host or build
+    environment. An example usage of this jinja function can be::
+
+        requirements:
+          host:
+            - curl 7.55.1
+          run_constrained:
+          {% for package in all_packages('host') %}
+            - {{ pin_compatible(package, exact=True) }}
+          {% endfor %}
+
+    which will render to::
+
+        requirements:
+            host:
+                - ca-certificates 2017.08.26 h1d4fec5_0
+                - curl 7.55.1 h78862de_4
+                - libgcc-ng 7.2.0 h7cc24e2_2
+                - libssh2 1.8.0 h9cfc8f7_4
+                - openssl 1.0.2n hb7f436b_0
+                - zlib 1.2.11 ha838bed_2
+            run_constrained:
+                - ca-certificates 2017.08.26 h1d4fec5_0
+                - curl 7.55.1 h78862de_4
+                - libgcc-ng 7.2.0 h7cc24e2_2
+                - libssh2 1.8.0 h9cfc8f7_4
+                - openssl 1.0.2n hb7f436b_0
+                - zlib 1.2.11 ha838bed_2
+    """
+    assert env in ('host', 'build'), 'Currently only host and build dependencies are supported.'
+    package_names = []
+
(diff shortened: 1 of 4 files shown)


183amir:
@msarahan here is the pull request.

- I don't know what would be a good name for the jinja function.
- When requirements has both `build` and `host`, it does not work for `build`.
- I need help to implement the tests properly.
- Should this function return the name of packages in `requirements:run` or `requirements:run_constrained` or `test:requires`? What good is it there even if it is possible.

msarahan:
Good start.

1. I think what you have is fine.  Exact name doesn't matter; having docs for the feature will matter much more.
2. That's odd.  I think it should.  Write a test!  I can help debug some, but having a test will help us figure it out faster.
3. please be more detailed.  I think you need tests that basically follow the pattern of https://github.com/conda/conda-build/blob/master/tests/test_api_render.py#L103-L106 - that is, make a recipe, render it, then assert that its requirements section has the values you expect.  Probably set comparison.
4. No, I think only build and host are relevant.  Limiting it to those as you have done already makes a lot of sense to me.

msarahan:
Looking good.  Please ping me when you're ready for me to do a more detailed review.

msarahan:
Yep.  I'll wait on CI and review later tonight or tomorrow.

conda_build/jinja_context.py
```diff
@@ -429,6 +429,53 @@ def cdt(package_name, config, permit_undefined_jinja=False):
...
+    if env not in ('host', 'build'):
+        raise ValueError('Only host and build dependencies are supported.')
+
+    package_names = []
+
+    # optimization: this is slow (requires solver), so better to bypass it
+    # until the finalization stage as done similarly in pin_compatible.
+    if not bypass_env_check and not permit_undefined_jinja:
+        pins, _, _ = get_env_dependencies(m, env, m.config.variant)
+        package_names = [p.split(' ')[0] for p in pins]
```
msarahan: Why are you keeping only the package names?  Are the pins also valuable?  It seems like you are tossing more info than you maybe want to.  Your example here shows the contents of pins, but we only return package_names.

183amir: Good idea. I do want the pin information.

183amir:
Thank you @msarahan for pushing your changes too. I agree with your changes.

msarahan:
Thanks @183amir - merging.

PR #2724, opened 2018-03-02
add optional argument for providing extra dep specs at test time

msarahan:
Numba team splits their builds between build and test.  With recent conda-build versions, we're no longer re-rendering the test-time recipe - it's baked in.  This is to support testing when the recipe isn't included.

So, to support Numba being able to dynamically specify dependencies at test time, this is a new feature that allows extra specs at the CLI.  This argument is only respected for testing, not for building or rendering.

CC @stuartarchibald

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -1791,6 +1791,8 @@ def test(recipedir_or_package_or_metadata, config, stats, move_broken=True):
         # not sure how this shakes out
         specs += ['r-base']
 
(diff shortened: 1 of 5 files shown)


stuartarchibald:
Thanks writing this, it looks like it should work nicely. It also has the side effect of being able to test combinations of optional packages against a single built package quickly and dynamically, which I think will be really useful.

msarahan:
build failures are due to the cran removal of the NMF package.  This PR is fine.  Merging.

PR #2732, opened 2018-03-05
Use globs for whitelists instead.

mingwandroid:


msarahan:
Cool, lgtm.  This should make it more flexible.

mingwandroid:
I added another bit.

PR #2741, opened 2018-03-07
FIX: source, missing archive file when dirname == filename

neok-m4700:
This fixes fixes this case:

`archive/<single-folder>/<filename>`, where `filename == single-folder`, `hoist_single_extracted_folder` is trying to move filename up 1 level, thus landing again in `single-folder`, and being deleted afterwards.

Well, this a confusing explanation: the added test recipe is more explicit.


msarahan:
Thanks!  Which version of conda-build did you see this problem with?  We hit pretty much this same issue before, and I'm confused on what I messed up with on fixing it.

neok-m4700:
`3.7.0` (latest). This is a dirty fix, there might be a more elegant way to solve this issue. Also I haven't tested it on windows nor macos, only on linux.

msarahan:
great, thanks.  Your test is good - that'll help me think about whether there's a cleaner way.  I'll either merge this or push up a change in a little while.

msarahan:
I implemented your idea a tiny bit differently (context manager), and cleaned up the rest of my junk that was no longer necessary.  Much cleaner as a whole now.  Thanks for suggesting the better way!

PR #2743, opened 2018-03-07
Add noarch builds to bdist_conda

bp-kelley:
This small patch allows ```python setup.py bdist_conda``` to build noarch python builds.  This is useful in our workflow as we won't need to run "conda convert" anymore.
 
usage:

```
setup(
    name="foo",
    version="1.0",
    distclass=distutils.command.bdist_conda.CondaDistribution,
    conda_buildnum=1,
    conda_features=['mkl'],
   conda_build_noarch = True
)
```

### DIFF
--- a/bdist_conda.py
+++ b/bdist_conda.py
@@ -73,6 +73,9 @@ class CondaDistribution(Distribution):
       (enable when installed).  See the features section of the conda build
       documentation for more information about features in conda.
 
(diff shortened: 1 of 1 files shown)


msarahan:
@bp-kelley this would be fine by me if you feel like reopening or resubmitting.  The 2.1.x branch isn't actively developed anymore - you might want to submit against master.

PR #2740, opened 2018-03-07
catch yaml ScannerError for legacy conda packages

msarahan:
fixes #2737

PR #2757, opened 2018-03-14
fix #2756 ignore non-rendered jinja2 in info/recipe/meta.yaml during indexing

### LINKED ISSUE #2756: 'conda index' is broken when info/recipe/meta.yaml contains non-rendered jinja2
Apparently, some older packages can contain non-rendered jinja2 in info/recipe/meta.yaml.

kalefranz:
fix #2756

### DIFF
--- a/conda_build/index.py
+++ b/conda_build/index.py
@@ -311,6 +311,9 @@ def _read_index_tar(tar_path, lock, locking=True, timeout=90):
                 recipe_json = yaml.load(recipe_text)
             except KeyError:
                 recipe_json = {}
+            except Exception as e:
+                log.debug("'info/recipe/meta.yaml' could not be read for '%s': %r", tar_path, e)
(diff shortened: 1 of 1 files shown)


kalefranz:
CC @ilanschnell

ilanschnell:
Thanks!  I will verify that this fix solves the original problem some of our customers have been seeing.

msarahan:
I believe that this does not conflict with #2740 because it is its own section.  I'm OK with this being there as a catch-all gutter that may elicit more bug reports, and thus more refinement of the proper exception handling.  Merging - although version 3.7.1, which included #2740, should already fix @ilanschnell's problem.

ilanschnell:
Thanks for merging.  The original fix #2740 already fixed my problem, and I just verified that this PR also still fixes the problem.

PR #2768, opened 2018-03-20
For discussion only at present.

mingwandroid:
Here I raise an error or warning if any DT_RUNPATH entries are found.

Some people (e.g. glibc developers) consider DT_RUNPATH the right thing to do. It is non-transitive and allows LD_LIBRARY_PATH to take precedence over it. It is enabled by passing --enable-new-dtags to ld, and by default when using gold where it can be disabled with --disable-new-dtags.

I believe the Anaconda Distribution never wants this however. We've wasted a lot of time tracking down non-bugs because someone has set LD_LIBRARY_PATH by mistake and I see no real use for it outside of a software developer testing their own fixes to a system library, personally.

I would welcome other people's opinions and some discussion here though: @msarahan, @nehaljwani, @jjhelmus, @kalefranz, @jakirkham, @mbargull, @ocefpaf, @isuruf.

### DIFF
--- a/conda_build/os_utils/pyldd.py
+++ b/conda_build/os_utils/pyldd.py
@@ -239,6 +239,9 @@ class UnixExecutable(object):
     def is_executable(self):
         return True
 
+    def get_runpaths(self):
+        return self.dt_runpath
+
 
 def read_data(file, endian, num=1):
     """
@@ -474,6 +477,7 @@ class machofile(UnixExecutable):
     def __init__(self, file, arch, initial_rpaths_transitive=[]):
         self.filename = file.name
         self.shared_libraries = []
+        self.dt_runpath = []
         # Not actually used ..
         self.selfdir = os.path.dirname(file.name)
         results = mach_o_find_dylibs(file, arch)
@@ -928,6 +932,9 @@ class inscrutablefile(UnixExecutable):
     def get_resolved_shared_libraries(self, *args, **kw):
         return []
 
+    def get_runpaths(self):
+        return []
+
     def selfdir(self):
         return None
 
@@ -1043,6 +1050,16 @@ def inspect_rpaths(filename, resolve_dirnames=True, use_os_varnames=True,
                 return [cf.to_os_varnames(rpath) for rpath in cf.rpaths_nontransitive]
             else:
                 return cf.rpaths_nontransitive
+
+
+def get_runpaths(filename, arch='native'):
+    if not os.path.exists(filename):
+        return []
+    sysroot = _trim_sysroot(sysroot)
+    arch = _get_arch_if_native(arch)
+    with open(filename, 'rb') as f:
+        cf = codefile(f, arch, ['/lib', '/usr/lib'])
+        return cf.get_runpaths()
 
 
 # TODO :: Consider returning a tree structure or a dict when recurse is True?
--- a/conda_build/os_utils/pyldd.py
+++ b/conda_build/os_utils/pyldd.py
@@ -983,6 +983,14 @@ def is_codefile(filename, skip_symlinks=True):
     return True
 
 
+def codefile_type(filename, skip_symlinks=True):
+    "Returns None, 'machofile' or 'elffile'"
+    klass = codefile_class(filename, skip_symlinks=skip_symlinks)
+    if not klass:
+        return None
+    return klass.__name__
+
+
 def _trim_sysroot(sysroot):
     while sysroot.endswith('/') or sysroot.endswith('\\'):
         sysroot = sysroot[:-1]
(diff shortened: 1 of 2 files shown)


conda_build/os_utils/pyldd.py
```diff
@@ -1045,6 +1052,15 @@ def inspect_rpaths(filename, resolve_dirnames=True, use_os_varnames=True,
                 return cf.rpaths_nontransitive
 
 
+def get_runpaths(filename, arch='native'):
+    if not os.path.exists(filename):
+        return []
+    arch = _get_arch_if_native(arch)
+    with open(filename, 'rb') as f:
+        cf = codefile(f, arch, ['/lib', '/usr/lib'])
```
nehaljwani: What about /lib64 , /usr/lib64 ?

mingwandroid: That may need a bit of a cleanup after I remind myself why I didn't do that, but the same thing happens in the existing, well-trodden codepaths and it's harmless there at least (even if I cannot completely explain it at present!)

All we actually care about for now is whether any DT_RUNPATHs are present and not their resolved (or unresolved) location.

jjhelmus:
I'm in favor of this change.  Using DT_RPATH rather than DT_RUNPATH will result in fewer issues for users who set LD_LIBRARY_PATH either intentionally or by mistake.  Software in the Anaconda Distribution should be as isolated from the system as possible, use of DT_RUNPATH erodes that isolation unnecessarily.

If there is a compelling need to override the rpath objects, I believe LD_PRELOAD can be used for this purpose when DT_RPATH or DT_RUNPATH is used.  I'm not recommending using of this outside of testing but it is an option.  

One concern is that the [ld.so man page](http://man7.org/linux/man-pages/man8/ld.so.8.html#DESCRIPTION) indicates that the "Use of DT_RPATH is deprecated". Removal of this feature from ld.so would break lots of software and will likely never happen.

kalefranz:
> and will likely never happen.

Well that's a little scary though, no?  I don't know the history here, where maybe one of you does.  Why was `DT_RPATH` deprecated to begin with? Is it something we can convince the GNU linker folks to either (1) back off of, given our legitimate use case, or (2) keep in a state of indefinite deprecation without actually ever removing it (i.e. an acknowledgement from them that there actually is no intent to remove it)?

msarahan:
I have nothing to add to this discussion.  I am in agreement with the sentiment of this PR and Jonathan's support here.  I also agree with Kale that some statement from the GNU folks would give us some nice peace-of-mind.  I don't think we necessarily need to require that statement, though.

Ray, we can go ahead with this after making some attempt to contact the GNU people, or with some link to an existing statement from them that "deprecation here does not indicate intent to remove."

mingwandroid:
I can ask on the glibc mailing list. To be clear if they were to remove `DT_RPATH`, then nearly every linux executable in the world would no longer work, and that simply will not happen, Linux's mantra (at the kernel level at least) is "never break userspace" and I hope at least some of that has filtered down to glibc!

mingwandroid:
glibc developers discussing this ~6 years ago:

https://sourceware.org/ml/libc-help/2012-05/msg00011.html

I believe this is sufficient?

msarahan:
Doesn't have a direct reply from the glibc team, but probably close enough.  Is this ready to merge, @mingwandroid ?

mingwandroid:
Jonathan Wakely and Mike Frysinger are core glibc developers (or at least very frequent contributors).

Let's do this.

kalefranz:
> https://sourceware.org/ml/libc-help/2012-05/msg00011.html

I’m convinced.

jakirkham:
Thanks Ray, et al. Sorry to get here late. Generally agree with all the comments that have come up.

Agree that `DT_RPATH` should be preferred. Have definitely seen issues with users setting `LD_LIBRARY_PATH`, which using `DT_RPATH` would fix. Case and point is issue ( https://github.com/conda-forge/curl-feedstock/issues/24 ). Am suspicious that other issues like this are lurking, which we haven't yet caught. While we discourage it every time, users still seem to like to set `LD_LIBRARY_PATH` in their every day environment. Overriding this seems like a good idea.

Haven't looked closely at the specific code here, but trust that has already been done by now.

mingwandroid:
Thanks @jakirkham, to be clear, pyldd.py is read-only so doesn't modify the DSOs it finds with this issue, it only warns (or errors out, please enable that catch though to make sure all these DSO issues become fatal when you adopt cb3).

Then you need to patch the build system of the package to ensure ld does not get passed --enable-new-dtags (I'm happy to add -Wl,--disable-new-dtags to the compiler activation scripts' LDFLAGS env var too, but that'll only make a difference if gold gets picked up and used where new dtags is the default, can't hurt to make it explicit though).

jakirkham:
Thanks for the clarification. Didn't have the impression this was changing the compilation behavior or modifying DSOs.

Agree this would be a nice thing to have explicit in the compilers (especially if we are encouraging this behavior).

PR #2773, opened 2018-03-21
lgtm.com config

kalefranz:
The product team has been known to use lgtm.com.  Conda and conda-build both get (well, fixed in conda now) bad scores primarily because of this one 'redundant-assignment' error.  Our use is perfectly legitimate though.  This is the config pretty much just copied over from conda (https://github.com/conda/conda/blob/master/.lgtm.yml) that wipes that one out for us.

To clarify, the redundant assignment error happens with we do

```python
from conda.exports import foo
foo = foo
```

but, because their inspection rule set is still probably young

```python
from conda.exports import foo, bar
foo, bar = foo, bar
```

doesn't trigger the error indicator.

Regardless, given how flake8 and pycharm interact, our use of the assignment here is the best technical solution.

### DIFF
--- a/.lgtm.yml
+++ b/.lgtm.yml
@@ -1,0 +1,12 @@
+queries:
+  - exclude: py/redundant-assignment
+
+extraction:
+  python:
+    python_setup:
+      version: 3
+    index:
+      exclude:
+        - .git
+        - conda/_vendor
+        - tests
--- a/.lgtm.yml
+++ b/.lgtm.yml
@@ -8,5 +8,4 @@ extraction:
     index:
       exclude:
         - .git
-        - conda/_vendor
         - tests


markshannon:
We (lgtm.com) don't flag `a, b = a, b` simply because it didn't occur to us that it would be worth looking for.
Typically, redundant assignments are typos or forgetting `self.`. The multiple assignment `a, b = a, b` doesn't fit into that pattern.

OOI, why can't you remove the redundant assignments?

kalefranz:
Hi!

The general need to have "redundant assignments" stems from the need to maintain interfaces for downstream code.  That is, conda-build imports code from conda, but through the normal course of conda releases we move or refactor something away.  On the conda side, we have to know what conda-build is using, and maintain what flake8 and pycharm see as "unused imports" to maintain compatibility.  Yes, the `#NOQA` tag would work to calm flake8 down, but pycharm's "Optimize Imports" will automatically remove unused imports (as it should) and break the downstream code unknowingly.

@msarahan and I started working on the conda-build and conda code bases at about the same time, both code bases being already somewhat mature projects, and both of us being new to the code bases.  Conda-build at the time basically treated all of conda's code as "public."  It was a total minefield to advance the state of code on the conda side, not knowing what innocent refactor was going to blow up on the conda-build side. We arrived at a solution where all of conda-build's imports of conda are isolated to a single module (`conda_build/conda_interface.py`) so that they're easier to track and manage and adapt on the conda-build side.  And on the conda side, we built out `conda/conda_exports.py` initially from the original `conda_build/conda_interface.py`.  Both of these give us layers of indirection to aid in natural code evolution and refactors.  And on the conda side, we have something to test against also.

Besides the quieting of "unused import" warnings by flake8, and "anchoring" of those imports for pycharm, another legitimate reason to have the redundant assignment is to maintain the symbol name in `conda/exports.py` through *automatic* IDE-assisted symbol renames.  Let's say I have in `conda/exports.py`

    from conda.somewhere.private import FooBar
    FooBar = FooBar

and we need to rename `FooBar` to `FooBaz`.  In most IDE's, in the `conda/somewhere/private.py` module, I just click on the `FooBar` symbol, choose the "Rename" refactor option, and `FooBar` can easily be everywhere renamed `FooBaz`.  Now our `conda/exports.py` file would look like

    from conda.somewhere.private import FooBaz
    FooBar = FooBaz

and we don't innocently and unknowingly break anything downstream.  Without that initial redundant assignment, we'd just end up with 

    from conda.somewhere.private import FooBaz

in `conda/exports.py`, and existing code that does

    from conda.exports import FooBar

would blow up.

TL;DR these "redundant assignments" are both intentional and necessary to support long-term compatibility across code bases.

PR #2797, opened 2018-04-02
WIP: Specify branch when cloning

jakirkham:
This is needed to make sure shallow clones on tags reference the number of commits relative to the tag location instead of the branch location. Currently this is still WIP and needs tests.

Note: Already know this doesn't work for `sha1`s, but need a mechanism to check that. Not sure if `conda-build` has one.

Note 2: Some versions of git suffer from a bug where combining this with mirror doesn't behave well. ( https://public-inbox.org/git/20180122110357.683b21a7@novascotia/t/ )

cc @jlstevens

xref: https://github.com/conda-forge/holoviews-feedstock/pull/28

### DIFF
--- a/conda_build/source.py
+++ b/conda_build/source.py
@@ -217,6 +217,8 @@ def git_mirror_checkout_recursive(git, mirror_dir, checkout_dir, git_url, git_ca
                        cwd=mirror_dir, stdout=stdout, stderr=stderr)
     else:
         args = [git, 'clone', '--mirror']
+        if git_ref != 'HEAD':
+            args += ['--branch', git_ref]
         if git_depth > 0:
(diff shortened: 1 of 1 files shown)


jakirkham:
Unfortunately this may be a bit trickier than originally imagined, this doesn't seem to work well with `--mirror`, which I don't understand why yet.

Also having to proof against a `sha1` in `git_ref` is tricky without having some way to detect them. We could try this and fallback.

Anyways at this stage this is more of a polite suggestion to make `git_depth` to be a bit more user friendly. It may be this will end up relegated to the issue backlog unfortunately. Though would add that a reasonable `git_depth` feature would help us move away from using GitHub archives, which seems desirable by many.

PR #2801, opened 2018-04-03
Minor update to config log warning

bjlittle:
Minor correction to `config.py` `log.warn` when setting the `arch` or the `platform`.


msarahan:
Thanks.  I don't know exactly what the problem is on appveyor, but it isn't your fault.  Needs some more optimization love, seems like - that test is part of the failure, but the real issue is that the tests didn't even finish in the allotted time.

PR #2819, opened 2018-04-18
Added "Error:" to several exception messages

SchwarzMarek:
Motivation: It was unclear that conda build failed to me (why), because it was not obvious that "Error: Received dictionary as spec.  Note that pip requirements are not supported in conda-build meta.yaml." is an error message.
Fix: added "Error: " string to several error msgs here which were not stating that.

### DIFF
--- a/conda_build/metadata.py
+++ b/conda_build/metadata.py
@@ -216,7 +216,7 @@ def yamlize(data):
 def ensure_valid_fields(meta):
     pin_depends = meta.get('build', {}).get('pin_depends', '')
     if pin_depends and pin_depends not in ('', 'record', 'strict'):
-        raise RuntimeError("build/pin_depends must be 'record' or 'strict' - "
+        raise RuntimeError("Error: build/pin_depends must be 'record' or 'strict' - "
                            "not '%s'" % pin_depends)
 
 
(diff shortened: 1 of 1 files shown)


msarahan:
The exceptions generally have Error in their name, and they show up in the strings, I think.  With your additions here, the strings will be something like:

ValueError: Error: some error message

That's why none of the exception strings include this info.  What do the tracebacks look like on your computer?  Are things not happening the way I think they are?

SchwarzMarek:
Hello,
whole output of conda build of my botched recipe looks like:

```
test@test:~/Downloads/foldx$ conda build .
INFO:conda_build.variants:Adding in variants from internal_defaults
Adding in variants from /home/test/Downloads/foldx/conda_build_config.yaml
INFO:conda_build.variants:Adding in variants from /home/test/Downloads/foldx/conda_build_config.yaml
Received dictionary as spec.  Note that pip requirements are not supported in conda-build meta.yaml.
```
conda version is: 4.5.1
conda build version: 3.8.1

So I wanted to include "Error" directly to error string to make it more obvious.

Best regards
Marek Schwarz

SchwarzMarek:
My test system is pure Ubuntu 14.04 with updates and Miniconda3 (all in virtualbox).
I understand that python raised error should be visible, however it isn't (at least in my test system) and it took me a while to realize that it actually was an error rather then some info message. I thing that redundancy here isn't that bad (I don't know how common is this).

The provided output was directly copied (not redirected) from terminal.

PR #2823, opened 2018-04-19
PoC: downstream testing

msarahan:
Implements new behavior where a package may specify downstream packages from itself that should be tested when a new build of a given upstream is built.  For example,

```
package:
  name: test_upstream_with_downstream_tests
  version: 1.0

test:
  downstreams:
    - test_downstream_with_test_fail
```

```
package:
  name: test_downstream_with_test_fail
  version: 1.0

test:
  commands:
    - exit 1

```

See the test case in this PR: if you build ``test_downstream_with_test_fail`` (disabling testing, of course), then build ``test_upstream_with_downstream_tests``, the test phase after building the latter package will fail because of the forced fail in the downstream package.

This PR is kind of hairy.  It could use some rework to make the implementation simpler.  Please take a look at the general idea here, and let me know what you think.

CC @mingwandroid @jjhelmus @nehaljwani @CJ-Wright

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -1647,9 +1647,20 @@ def _construct_metadata_for_test_from_package(package, config):
             is_channel = True
 
     if not is_channel:
-        raise ValueError("Your package must reside in a channel structure with "
-                         "platform-subfolders.  See more info at what a valid channel is "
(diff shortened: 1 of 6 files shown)


jjhelmus:
LGTM.

Issue #2821, opened 2018-04-18
Test commands executed in unexpected order with recipe_append.yaml


ceball:
Using almost the same package as attached to #2820, but with the addition of a recipe_append.yaml file, I get test commands being executed in an unexpected order. Or maybe I'm getting an unexpected/mixed up set of commands.

I'm not 100% sure if the problem is only with recipe_append.yaml, but it seems that way.

### Actual Behavior

```
$ cat conda.recipe/meta.yaml
package:
  name: test
  version: 0.1

source:
  path: ..

build:
  noarch: python
  script: python setup.py install --single-version-externally-managed --record=record.txt

requirements:
  build:
    - python
    - setuptools
  run:
    - python
test:
  imports:
    - test
  commands:
    - echo 1
    - echo 2
    - echo 3

$ cat conda.recipe/recipe_append.yaml
test:
  commands:
    - echo 20
    - echo 10
    - false

$ conda build conda.recipe --append-file conda.recipe/recipe_append.yaml
[...]
TEST START: /Users/[...]/conda-bld/noarch/test-0.1-py_0.tar.bz2
[...]
Executing transaction: ...working... done
import: 'test'
import: 'test'
import: 'test'
+ echo 20
20
+ echo 2
2
+ false
Tests failed for test-0.1-py_0.tar.bz2 - moving package to /Users/[...]/conda-bld/broken
WARNING:conda_build.build:Tests failed for test-0.1-py_0.tar.bz2 - moving package to /Users/[...]/conda-bld/broken
removing: test-0.1-py_0.tar.bz2
TESTS FAILED: test-0.1-py_0.tar.bz2

$ cat [...]/conda-bld/test_1524051625768/test_tmp/run_test.sh


set -ex



echo 20
echo 2
false
echo 10
echo 1
echo 3
exit 0


set -ex



echo 20
echo 2
false
echo 10
echo 1
echo 3
exit 0


set -ex



echo 20
echo 2
false
echo 10
echo 1
echo 3
exit 0

```

### Expected Behavior

https://conda.io/docs/user-guide/tasks/build-packages/variants.html#appending-to-recipes:

> Any contents in recipe_append.yaml will add to the contents of meta.yaml. List values will be extended [...]

So I was expecting to see:
```
[...]
+ echo 1
1
+ echo 2
2
+ echo 3
3
+ echo 20
20
+ echo 10
10
+ false
[...]

PR #2825, opened 2018-04-19
Recipe appending: avoid losing ordering of both original and appended commands.

ceball:
This PR is to address #2821, for appending to recipes (https://conda.io/docs/user-guide/tasks/build-packages/variants.html#appending-to-recipes):

> Any contents in recipe_append.yaml will add to the contents of meta.yaml. List values will be extended [...]

However, I'm not really clear on the intent of the function I'm modifying here, so I could do with some guidance. For instance, although line 1401 doesn't affect me right now, I also don't see its purpose (I would expect just line 1402).

### DIFF
--- a/conda_build/utils.py
+++ b/conda_build/utils.py
@@ -1401,7 +1401,7 @@ def merge_or_update_dict(base, new, path, merge, raise_on_clobber=False):
                 if base_value and base_value != value:
                     base_value.extend(value)
                 try:
-                    base[key] = list(set(base_value))
+                    base[key] = list(base_value)
                 except TypeError:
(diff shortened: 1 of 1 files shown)


ceball:
> The purpose of the set is deduplication

I'm not clear on why duplicate commands should be removed in the first place. Shouldn't the user of conda build ensure the commands passed in are what they want to run, and conda build should not modify them? (E.g. I might want to test a command runs ok when repeated multiple times, or something like that...)

msarahan:
Yeah, good point.  That's really what the clobber stuff is for.  Great, I'll go ahead and merge this.

PR #2836, opened 2018-04-24
Sync default cran mran setting

msarahan:
Supersedes #2833 

For Anaconda, we'll need to use -m and specify our config file.  I can also put in a condarc setting to help us avoid forgetting to pass that.

CC @mingwandroid

### DIFF
--- a/conda_build/cli/main_skeleton.py
+++ b/conda_build/cli/main_skeleton.py
@@ -57,7 +57,6 @@ def execute(args):
         parser.print_help()
         sys.exit()
 
-    # for package in args.packages:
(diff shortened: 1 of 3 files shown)


mingwandroid:
Doesn't this mean an attempt to build without any -m will fail due to cran_mirror not being set, both for conda-forge and for us? We still need a default and I still prefer MRAN.

mingwandroid:
OK, I missed that line. LGTM then.

msarahan:
default added in https://github.com/conda/conda-build/pull/2868; will be in next release.

PR #2837, opened 2018-04-25
Include `folder` as a field in `source`

jakirkham:
Closes https://github.com/conda-forge/conda-smithy/issues/756

cc @isuruf


msarahan:
I think this is totally a bug, but I'll wait on CI just in case.  It wasn't something I thought of, because with a single source, it doesn't make much sense, since CB will try to CD into the root of the extracted source.  Still, I think it needs to be where you have put it.

jakirkham:
Thanks Mike. Was wondering whether or not this should live here. Well if you and the CIs agree it is ok, that's good enough for me.

msarahan:
Thanks @jakirkham

PR #2839, opened 2018-04-26
fix handling of <= on upper bound constraints being forced to <

### LINKED ISSUE #2806: empty host loses imposed deps from build run_exports
These libraries are expected to be present when compiling, but if there are no other host dependencies, conda-build discards the host section completely and creates no environment for it.  Conda-build needs to resolve all run_exports prior to env creation, and create the host env even when the only deps are those that are imposed by run_exports in the build section (i.e. "strong" run_exports)

msarahan:
Fixes issue raised by @tkelman in https://github.com/conda-forge/conda-build-feedstock/pull/41#issuecomment-384470429

wherein he had an upper bound with <=, and it just so happened that conda-forge's only available packages were exactly that upper bound.  The old behavior of ignoring the operator and always forcing < was broken.  This fixes it and does the comparison a little more straightforwardly.

fixes #2806

### DIFF
--- a/conda_build/render.py
+++ b/conda_build/render.py
@@ -28,7 +28,6 @@ from .conda_interface import pkgs_dirs
 from .conda_interface import conda_43
 from .conda_interface import specs_from_url
 from .conda_interface import memoized
-from .conda_interface import VersionOrder
 
 from conda_build import exceptions, utils, environ
 from conda_build.metadata import MetaData, combine_top_level_metadata_with_output
@@ -359,6 +358,8 @@ def add_upstream_pins(m, permit_unsatisfiable_variants, exclude_pattern):
         extra_run_specs = set(extra_run_specs_from_build.get('strong', []))
         if not m.uses_new_style_compiler_activation:
             extra_run_specs.update(extra_run_specs_from_build.get('weak', []))
+        else:
+            host_deps = set(extra_run_specs_from_build.get('strong', []))
 
     run_deps = extra_run_specs | set(utils.ensure_list(requirements.get('run')))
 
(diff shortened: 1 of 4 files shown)


tkelman:
Could you elaborate on why the transformation from https://github.com/conda/conda-build/pull/2694 is necessary and what it's fixing? The issue it references is private, could the problem be worked-around elsewhere if it only ever shows up on internal infrastructure?

msarahan:
The issue was that conda-build was putting more than one spec into the final dependencies, so you could end up with things like:

mydep >2.0
mydep >2.1,<3

which you might think is fine, but conda 4.4+ actually only respects the last of the specs.  This was especially a problem for our Anaconda metapackage, which ended up with exact pins (which is all it should have) and range constraints (from several run_exports), and it was then ambiguous which behavior would happen.  This meant that the metapackage, which is supposed to represent a frozen collection of packages, was not actually frozen.

This change in conda-build is compensating for the Conda behavior.  We are also fixing this on the conda side, because really it makes much more sense for the solver to be doing this sort of work implicitly.

CC @kalefranz for any status updates on how conda might change handling this.

kalefranz:
> but conda 4.4+ actually only respects the last of the specs

actually, it's `conda >=4.4,<4.5` :)

kalefranz:
https://github.com/conda/conda/pull/6956 introduced merging of MatchSpecs, which fixed this problem on the conda side.  It was released in conda 4.5.0.

msarahan:
Thanks for the clarification, @kalefranz.  With that knowledge, I think I'll probably simplify this to only clobbering any range bounds with exact pins, and not try to merge multiple range bounds.  That will fix the original metapackage issue, and conda will take care of the rest.

tkelman:
Thanks for the fix. Do you intend to tag a 3.9.2 fairly soon?

msarahan:
Yes. Sometime today.

PR #2840, opened 2018-04-26
Ensure u+w permissions before calling install_name_tool

mingwandroid:


conda_build/os_utils/macho.py
```diff
@@ -192,14 +195,35 @@ def get_rpaths(path):
...
+   except (OSError, utils.PermissionError) as e:
+       log = utils.get_logger(__name__)
+       log.warn(str(e))
+
+
+def install_name_tool(args, verbose=False):
+    args_full = ['install_name_tool'] + args
+    if verbose:
+        print(' '.join(args_full))
+    old_mode = stat.S_IMODE(os.stat(args[-1]).st_mode)
```
msarahan: Do you want os.stat or stat.something?  If os.stat, you need to import os.

PR #2844, opened 2018-04-26
add conda 4.6 compat with activate --stack flag

kalefranz:
Per fairly extensive discussion with @mingwandroid, we took out the default environment stacking from conda 4.4.  In conda 4.6, to get environment stacking, you now need to add a `--stack` flag to activate (from https://github.com/conda/conda/pull/7195).  This PR should maintain compatibility here.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ venv
 junit.xml
 .cache
 .DS_Store
+.pytest_cache/
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,7 +48,7 @@ install:
       conda install -q futures scandir;
     fi
   - conda update -q --all
-  - if [ -n "$CONDA_VERSION" ]; then
+  - if [ "$CONDA_VERSION" = "4.3.x" ]; then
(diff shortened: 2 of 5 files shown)


msarahan:
Thanks for putting this up.  I think it's still not working quite right.  Both of my builds against conda master are failing: https://travis-ci.org/conda/conda-build/jobs/371655560#L2306

These are failing in a way that suggests that the host env is not on PATH (and thus activation isn't doing what we expect).  I'm happy to look at it tomorrow with you.

kalefranz:
Yeah let's work through it tomorrow.  I was going to cut a 4.6.0a2 release tonight, but I'll hold off until we get this resolved in case we need to make further changes on the conda side.

msarahan:
Same issue with version ordering that Ray was encountering with the python prerelease:

```
In [1]: from conda_build.conda_interface import conda_46

In [2]: conda_46
Out[2]: False

In [3]: from conda import __version__ as CONDA_VERSION
   ...:

In [4]: CONDA_VERSION
Out[4]: '4.6.0a1'
```

msarahan:
Thanks @kalefranz

PR #2859, opened 2018-05-01
add simple ASV benchmarking setup

msarahan:
docs on running this at https://asv.readthedocs.io/en/latest/using.html

basically,

```
asv run
```

or

```
asv run (start tag/commit)..(end tag/commit)
```

then

```
asv publish
asv preview
```


jakirkham:
Are these results getting published somewhere?

PR #2877, opened 2018-05-10
add support for .tar.lzma

abloemert:


mingwandroid:
IMHO, if we are going to support this it should be done via `libarchive` and `python-libarchive` rather than an external executable.

abloemert:
> IMHO, if we are going to support this it should be done via libarchive and python-libarchive rather than an external executable.

@mingwandroid: I added the external executable for backwards compatibility with Python 2. Python 3 has built-in support for lzma.

mingwandroid:
Sure, but we have python-libarchive with lzma support for Python 2 on all platforms.

Also python-libarchive is way faster than tarfile.

abloemert:
If python-libarchive is really faster, we should use it for all formats indeed. However this will be much more work and could have a big impact. This simple change however will at least add support for .tar.lzma files, which is currently missing in conda-build.

mingwandroid:
But you introduce a feature that works with some pretty bad caveats, that is IMHO worse than not having it.

mingwandroid:
> However this will be much more work and could have a big impact

About 15 minutes work in my estimation.

mingwandroid:
Here you go: https://github.com/conda/conda-build/commit/cfaada3246ab3d773ae221a2a7b64e17acb2eb20

mingwandroid:
> However this will be much more work

In any PR, this is a terrible argument BTW, it doesn't give me confidence that you will maintain the code you propose.

msarahan:
superseded by https://github.com/conda/conda-build/pull/2878

Thanks @abloemert for getting the ball rolling here.

PR #2878, opened 2018-05-10
For discussion: Use libarchive for both source unpack and final tar.b…

mingwandroid:
…z2 creation

This cleans up the code in conda build significantly at the cost of adding `libarchive` and `python-libarchive-c` dependencies.

I've built those for all Pythons though with comprehensive codec support (7z, bzip2, zip, lzma, lz4 etc), even on Windows so I don't see that as a blocker.

Interestingly, if I change:
```
with libarchive.file_writer(tmp_path, 'ustar', 'bzip2') as archive:
```
to
```
with libarchive.file_writer(tmp_path, 'ustar', 'xz') as archive:
```

.. we get a much smaller package with a `.tar.bz2` file extension, however `conda` is still able to install the package (on macOS at least) *NOTE* anyone who sees this and thinks this is a good idea, it is not, just an interesting anecdote. Creating `.tar.xz` files masquerading as `.tar.bz2` files would be a very bad thing to do.

Ping @kalefranz since we sometimes discuss compression things.


msarahan:
This looks great to me.  I especially like the wheel handling.  Thanks for your care there.  I'll wait on CI, and merge or push changes to fix small things.

mingwandroid:
Before merging this needs tests and checking / adding support for `.tar.z`.

mingwandroid:
We should make a recipe with multiple sources, one in each compression type I think?

conda_build/utils.py
```diff
@@ -606,60 +602,12 @@ def relative(f, d='lib'):
...
-    members = t.getmembers()
-    for i, member in enumerate(members, 0):
-        if os.path.isabs(member.name):
-            member.name = os.path.relpath(member.name, '/')
-        if not os.path.realpath(member.name).startswith(os.getcwd()):
-            member.name = member.name.replace("../", "")
-        if not os.path.realpath(member.name).startswith(os.getcwd()):
-            sys.exit("tarball contains unsafe path: " + member.name)
-        members[i] = member
+decompressible_exts = (('.7z', '.tar', '.tar.bz2', '.tar.gz', '.tar.lzma', '.tar.xz', '.tgz', '.whl', '.zip'))
```
mingwandroid: Doubled up brackets here, will fix.

mingwandroid: unLGTMing until I add tests

mingwandroid:
@msarahan, not sure I've done enough / anything to enable my tests. I must confess I usually lean on CI as my test runner because I find that whole side of things a bit of a pain.

mingwandroid:
@jakirkham, @ocefpaf, @isuruf (please @ other CF people who need to know).

This PR adds support for a load of new source archive types (see the test). This means `conda-build` will start to unpack them rather than copying them which means recipes will need to be updated (simplified though, removing unpacker dependencies and hand-unpacking).

Also, for `python 2` on Windows you will need to get my updates to `c99-to-c89` and also the `libarchive` dependency stack.

jakirkham:
Personally won't have much time to do anything with this over the next month. So will defer to others. That said, have a few thoughts.

First will there be an option to allow recipes to opt-out of this behavior if they aren't ready to be switched over?

Second what do we mean by decompressing `.whl` files? These don't really have `setup.py` or anything in them to aid install. Are we planning to copy files over? How will this work?

Third adding 7z support is very welcome.

Fourth AFAIK nothing uses `c99-to-c89` in conda-forge. Though intent was to use it. Could be useful for VS 2008 stuff. What sort of changes have you made? Does this plugin to the VS compiler architecture somehow or do recipe changes need to occur to use this?

Fifth what changes are you referring to in the `libarchive` stack?

That's all from me for now. :)

mingwandroid:
> First will there be an option to allow recipes to opt-out of this behavior if they aren't ready to be switched over?

No, I don't think that's necessary because the changes are simple and obvious.

> Second what do we mean by decompressing .whl files? These don't really have setup.py or anything in them to aid install. Are we planning to copy files over? How will this work?

`conda-build` uncompresses `whl` files for some reason outside of getting source code. I removed the old code to do this (used unzip or whatever) replacing everything with a call to `untar_xz` instead which calls `libpython` via `python-libpython-c`.

> Third adding 7z support is very welcome

Adding `xz` and `lzma` (cleanly across all platforms) is useful too, but the real important thing is to support future compression schemes added to `libarchive` with very little effort (add them to `decompressible_exts`, nothing more).

> Fourth AFAIK nothing uses c99-to-c89 in conda-forge. Though intent was to use it

I wasn't aware of that, we only started using it since you had it. It was pretty buggy (not fit for purpose if I'm being honest) and is still not perfect, but with [my changes](https://github.com/libav/c99-to-c89/compare/master...AnacondaRecipes:master) it is good enough to handle significant codebases now. We use it for both `libarchive` and also `xz`.

> Fifth what changes are you referring to in the libarchive stack?

The ones I mentioned to `libarchive` and `xz` so they can be compiled with VS2008 mainly. Hopefully those will come when we make PRs to `conda-forge`.

mingwandroid:
Since it could get missed, from the comment in the first commit:

@msarahan, before merging we need to assess whether we need to set some
security flags.

Ref: https://github.com/libarchive/libarchive/blob/master/libarchive/archive.h

```
/* Default: Do not try to guard against extracts redirected by symlinks. */
/* Note: With ARCHIVE_EXTRACT_UNLINK, will remove any intermediate symlink. */
..
/* Default: Do not reject entries with '..' as path elements. */
..
/* Default: Do not reject entries with absolute paths */
```

.. we should also consider what is appropriate for the following flags:

```
/* Default: Do obey umask, do not restore SUID/SGID/SVTX bits. */
/* Default: Do not restore mtime/atime. */
```

mingwandroid:
Also I will create another test of archives that try to escape the `SRC_DIR` via relative paths (`../`) and absolute ones (`/tmp/`) and ensure that we do not end up allowing that.

msarahan:
superseded by #3163 - closing

PR #2881, opened 2018-05-14
create_test: Use exit \b N instead of exit N

mingwandroid:
.. this is so that you can 'call run_test.bat' without having it close your cmd.exe prompt
when you are trying to investigate why tests failed.


msarahan:
good one.  Thanks.

PR #2882, opened 2018-05-15
ensure that merge_build_host is updated correctly before build and be…

msarahan:
…fore each output

The state for merging was being considered incorrectly especially with outputs.  It was behaving properly for the top-level, but not being recomputed for each output.


msarahan:
probably best to wait for tests, in case I did something dumb.  Highly likely in current snotty state.

mingwandroid:
OK

PR #2893, opened 2018-05-18
Add creative commons as valid license family

scopatz:
Pretty minor change for an  important class of licenses.


msarahan:
Yep, thanks.

Issue #2900, opened 2018-05-21
`AttributeError: 'NoneType' object has no attribute 'split'`


rainwoodman:
conda-build 3.10.4 fails on previously working script with error:

```
(base) [yfeng1@waterfall conda-channel-bccp]$ conda build -m variants/pc/python-3.6.yaml -c bccp platform/cfastpm
Adding in variants from internal_defaults
INFO:conda_build.variants:Adding in variants from internal_defaults
Adding in variants from variants/pc/python-3.6.yaml
INFO:conda_build.variants:Adding in variants from variants/pc/python-3.6.yaml
Traceback (most recent call last):
  File "/home/yfeng1/anaconda3/install/bin/conda-build", line 11, in <module>
    sys.exit(main())
  File "/home/yfeng1/anaconda3/install/lib/python3.6/site-packages/conda_build/cli/main_build.py", line 420, in main
    execute(sys.argv[1:])
  File "/home/yfeng1/anaconda3/install/lib/python3.6/site-packages/conda_build/cli/main_build.py", line 411, in execute
    verify=args.verify)
  File "/home/yfeng1/anaconda3/install/lib/python3.6/site-packages/conda_build/api.py", line 200, in build
    notest=notest, need_source_download=need_source_download, variants=variants)
  File "/home/yfeng1/anaconda3/install/lib/python3.6/site-packages/conda_build/build.py", line 2146, in build_tree
    bypass_env_check=True)
  File "/home/yfeng1/anaconda3/install/lib/python3.6/site-packages/conda_build/render.py", line 740, in render_recipe
    allow_no_other_outputs=True, bypass_env_check=bypass_env_check)
  File "/home/yfeng1/anaconda3/install/lib/python3.6/site-packages/conda_build/render.py", line 646, in distribute_variants
    bypass_env_check=bypass_env_check)
  File "/home/yfeng1/anaconda3/install/lib/python3.6/site-packages/conda_build/metadata.py", line 946, in parse_until_resolved
    bypass_env_check=bypass_env_check)
  File "/home/yfeng1/anaconda3/install/lib/python3.6/site-packages/conda_build/metadata.py", line 885, in parse_again
    bypass_env_check=bypass_env_check),
  File "/home/yfeng1/anaconda3/install/lib/python3.6/site-packages/conda_build/metadata.py", line 1421, in _get_contents
    env.globals.update(environ.get_dict(m=self))
[...]

PR #2914, opened 2018-05-23
Add a test case for #2912

rainwoodman:
This is intended to be a test case for the fix in #2912.

I followed the conventions in the test-recipes directory, but I am not sure how to trigger these tests.

When triggered manually with conda-build 3.10.3's conda-render, no errors are raised.
When triggered manually with 3.10.4, the error as described in #2900 is reproduced.

### DIFF
--- a/tests/test-recipes/variants/27_requirements_host/conda_build_config.yaml
+++ b/tests/test-recipes/variants/27_requirements_host/conda_build_config.yaml
@@ -1,0 +1,2 @@
+mpi:
+  - openmpi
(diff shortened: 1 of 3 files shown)


msarahan:
Thanks!  I'll take care of getting it worked into the whole suite.  In the meantime 3.10.5 has been released and should fix your issues.  This will be a good regression test to have.

rainwoodman:
Yes 3.10.5 has fixed the issues.

rainwoodman:
I see -- this is how to test against external recipe files.

msarahan:
There's a few different ways.  It depends on the folder.  Any folder in test/test-recipes/metadata that does not begin with an _ is automatically picked up and built.  Those are largely integration tests of the whole build functionality, and they have their tests as assertions in the recipe build.sh/bld.bat or run_test.*.

For variants stuff, there's usually more to test, since you're making assertions above the level of a single build.  That's why they require a little more effort.  I added the mpich thing to your recipe so that it would also be testing proper looping over that variable.

PR #2920, opened 2018-05-25
Fix raise RuntimeError

nicoddemus:
The code was raising `RuntimeError(msg, pattern)` (a tuple) instead of the intended `RuntimeError(msg % pattern)`.


msarahan:
Thanks!

Issue #2936, opened 2018-06-01
invariant exclusive_config_file redundantly parsed


mbargull:
### Actual Behavior

During batch processing of recipes with the same `conda_build.api.Config` object (or copies thereof), low level, invariant configuration files are parsed for every recipe instead of their contents being cached.
For larger configuration files, like the one from `conda-forge-pinning`, this has a large performance impact during (non-finalized) rendering.

<details><summary>ref. Gitter conversation:</summary>

https://gitter.im/conda/conda-build?at=5b0dc3d299fa7f4c062c0217:
>> I noticed that exclusive_config_file and conda_build: config_file/~/conda_build_config.yaml are parsed over and over again in batch-processing recipes (for every call of variants.get_package_variants). Would it be possible to cache their parsed contents somewhere? I think caching that should be fine since it's the "bottom configuration" and doesn't depend on anything in the recipe, right?

> What needs to be clarified here is what the scope of selectors is for any parsed config file. If you can use variables from the file itself as selectors, it needs to be iterative. If you can only use the small set of selectors that are hard-coded, then you can cache. Maybe I'm misunderstanding your question, though. I think if the assumption is "I'm doing batch processing and these files are always the same," then yes, I'd recommend caching it, perhaps by creating the Config instance and passing that around (copying it if you find it getting mutated undesirably)

https://gitter.im/conda/conda-build?at=5b0dc77799fa7f4c062c1006:
>> I think if the assumption is "I'm doing batch processing and these files are always the same," 

> Correct, that's the case.

>> I'd recommend caching it, perhaps by creating the Config instance and passing that around (copying it if you find it getting mutated undesirably)

> That's what I assumed previously to work, but conda_build.variants.get_package_variants calls conda_build.variants.find_config_files and for each of those files (which includes the exclusive_config_file) conda_build.variants.parse_config_file is called. Hence for different recipes get_package_variants does not load the parsed contents of exclusive_config_file from any cache.
</details>

### Expected Behavior

Low level invariant configuration file (`--exclusive-config-file`, `~/conda_build_config.yaml`, `.condarc`:`conda_build/config_file`) that are not recipe-dependent should be parsed once per `Config` object.

### Steps to Reproduce
<details open>

```python
from collections import defaultdict
from os import makedirs
from os.path import abspath, join, normcase
from shutil import copyfileobj
import sys
from tempfile import TemporaryDirectory
from time import time
from urllib.request import urlopen

from conda_build import api, utils, variants

[...]

PR #2937, opened 2018-06-01
utils: replace glob2 by glob for PY3

mbargull:
`glob2` is considerably slower than `glob` from the Python 3's standard library acc. to the measurements from gh-2936:
```
# conda-build 3.10.6
7.650 glob
3.055 parse_config_file
18.984 render

# conda-build 3.10.6 + monkeypatched glob
2.514 glob
3.124 parse_config_file
14.154 render
```
If I didn't miss anything else, the differences between `glob2.glob` and `glob.glob` are that `glob2.glob` offers more fine grained behavior via additional parameters, i.e., `with_matches, include_hidden, norm_paths, case_sensitive, sep`, uses `recursive=True` instead `False` on default and also calls `normcase` on each returned path.

BEWARE: I did not check whether `conda-build` uses any of the additional parameters of `glob2.glob` or if there are other possible differences/cornercases between the two implementations.

But since `glob.glob` (`glob.iglob`) from Python 3 also handles `**` patterns with `recursive=True`, using it could offer a considerable speedup.

### DIFF
--- a/conda_build/utils.py
+++ b/conda_build/utils.py
@@ -4,7 +4,6 @@ import base64
 from collections import defaultdict
 import contextlib
 import fnmatch
-from glob2 import glob
 import json
 from locale import getpreferredencoding
 import logging
@@ -13,7 +12,7 @@ import mmap
 import operator
 import os
 from os.path import (dirname, getmtime, getsize, isdir, join, isfile, abspath, islink,
-                     expanduser, expandvars)
+                     expanduser, expandvars, normcase)
 import re
 import stat
 import subprocess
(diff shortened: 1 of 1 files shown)


msarahan:
The only reason for glob2 is supporting expressions with ``**`` patterns on python 2.  Your change is fine.  Thanks for digging into this!

mbargull:
> The only reason for glob2 is supporting expressions with `**` patterns on python 2.

That's what I hoped for :smile:.

A related test failed on AppVeyor, though:
https://ci.appveyor.com/project/ContinuumAnalyticsFOSS/conda-build/build/1.0.3599/job/ctagelwg04or2gp5#L2427
```python
______________________________ test_expand_globs ______________________________
[gw0] win32 -- Python 3.6.5 C:\Miniconda36-x64\python.exe
  File "C:\projects\conda-build\tests\test_utils.py", line 207, in test_expand_globs
    assert exp == sorted(['abc', 'acb', os.path.sep.join(('sub1', 'def')),
AssertionError: assert ['c:\\cbtmp\\...\\ssub1\\ghi'] == ['abc', 'acb',...\\ssub1\\ghi']
  At index 0 diff: 'c:\\cbtmp\\popen-gw0\\test_expand_globs0\\abc' != 'abc'
  Full diff:
  + ['abc', 'acb', 'sub1\\def', 'sub1\\ssub1\\ghi']
  - ['c:\\cbtmp\\popen-gw0\\test_expand_globs0\\abc',
  -  'c:\\cbtmp\\popen-gw0\\test_expand_globs0\\acb',
  -  'c:\\cbtmp\\popen-gw0\\test_expand_globs0\\sub1\\def',
  -  'c:\\cbtmp\\popen-gw0\\test_expand_globs0\\sub1\\ssub1\\ghi']
```
So, it looks like we now get absolute paths from `utils.expand_globs` on Windows... Not sure if this causes any trouble, though.

msarahan:
Yeah, that's an annoying inconsistency.  I'll work on that.  I don't know if it would cause trouble, but better to just make it consistent.

PR #2939, opened 2018-06-01
utils: apply abspath for consistency between glob and glob2

mbargull:
ref gh-2937

> So, it looks like we now get absolute paths from utils.expand_globs on Windows... Not sure if this causes any trouble, though.

> Yeah, that's an annoying inconsistency. I'll work on that. I don't know if it would cause trouble, but better to just make it consistent.

Not the nicest solution, but I'm just curious if the CI runs fine if we just `map(abspath, ...)` everything...

### DIFF
--- a/conda_build/utils.py
+++ b/conda_build/utils.py
@@ -45,7 +45,7 @@ if PY3:
 
     # stdlib glob is less feature-rich but considerably faster than glob2
     def glob(pathname, recursive=True):
-        return list(map(normcase, iglob(pathname, recursive=recursive)))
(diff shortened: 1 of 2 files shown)


mbargull:
Those two commits are complete nonsense as I neither actually looked at `utils.expand_globs`' source before and also didn't inspect the `glob` output :roll_eyes:. Sorry for the noise, let's better just close this PR :wink:

msarahan:
Okie dokie - well, thanks for looking at it, anyway.

PR #2952, opened 2018-06-06
utils.glob: remove unnecessary normcase, fix test_expand_globs

mbargull:
Followup to https://github.com/conda/conda-build/pull/2937#issuecomment-393959530.
I simply misread `glob2`'s source code. The `normcase` is only used internally but does not affect the output. Hence, the `map(normcase, ...)` isn't needed and should fix the inconsistent output (and therefore the failing `tests.test_utils.test_expand_globs` on Windows + PY3).


msarahan:
Thanks @mbargull

PR #2970, opened 2018-06-15
variants: fix re.escape usage for Python <3.3

mbargull:
Second and hopefully final fix after gh-2949 for issues introduced in gh-2925.


msarahan:
Thanks @mbargull.  I'll get a release with this out quickly so we can unblock conda-forge.

mbargull:
Thanks a bunch! (and sorry for the trouble...)

PR #2972, opened 2018-06-15
Make recipes beautiful again

### LINKED ISSUE #3000: PyPI skeleton not coping well with environment markers?
hdf5storage is failing in our test suite right now, with errors like

```
  File "/home/travis/build/conda/conda-build/tests/test_api_skeleton.py", line 150, in test_setuptools_test_requirements
    api.skeletonize(packages='hdf5storage', repo='pypi')
  File "/home/travis/build/conda/conda-build/conda_build/api.py", line 281, in skeletonize
    recursive=recursive, config=config, **kwargs)
  File "/home/travis/build/conda/conda-build/conda_build/skeletons/pypi.py", line 274, in skeletonize
    setup_options=setup_options)
  File "/home/travis/build/conda/conda-build/conda_build/skeletons/pypi.py", line 737, in get_package_metadata
    sys.exit("Error: Could not parse: %s" % dep)
SystemExit: Error: Could not parse: numpy ; python_version == '2.7'
```

The requirements coming out look like:

```
"numpy < 1.12.0 ; python_version == '2.6'", "numpy ; python_version == '2.7'", "numpy<1.6.0 ; python_version == '3.0'", "numpy<1.8.0 ; python_version == '3.1'", "numpy<1.12.0 ; python_version == '3.2'", "numpy<1.12.0 ; python_version == '3.3'", "numpy ; python_version >= '3.4'", "h5py>=2.1 ; python_version == '2.6'", "h5py>=2.1 ; python_version == '2.7'", "h5py>=2.1,<2.4 ; python_version == '3.0'", "h5py>=2.1,<2.7 ; python_version == '3.1'", "h5py>=2.1,<2.7 ; python_version == '3.2'", "h5py>=2.1 ; python_version >= '3.3'"
```

these need to be translated into deps that conda-build understands.

ocefpaf:
@msarahan hopefully I did not break anything. The goal is to:

- enforce sha256;
- change the `script` section to use `pip`;
- simplify the recipe by reducing the excess of jinja variables.

I'm sure there are more places we can optimize and simplify the code but I'm not too familiar with conda-skeleton to go any further.

Also, I'd like to remove the single quotes from the name and version values in the package section, to keep a consistent style with the rest of the recipe, but I'm not sure how to do that in an `ruamel_yaml.comments.CommentedMap` object (and that is not really that important anyway.)

fixes #3000

### DIFF
--- a/conda_build/skeletons/pypi.py
+++ b/conda_build/skeletons/pypi.py
@@ -69,17 +69,8 @@ REQUIREMENTS_ORDER = ['host', 'run']
 ABOUT_ORDER = ['home', 'license', 'license_family', 'license_file', 'summary',
                'description', 'doc_url', 'dev_url']
 
-# This may be overkill, but some day sha256 won't be enough. Might as well be
-# ready...list these in order of decreasing preference.
-POSSIBLE_DIGESTS = ['sha256', 'md5']
-
-POSSIBLE_FILE_EXTENSIONS = ['tar.gz', 'tar.bz2', 'zip', 'tar', 'gz']
-
 PYPI_META_HEADER = """{{% set name = "{packagename}" %}}
 {{% set version = "{version}" %}}
-{{% set file_ext = "{file_ext}" %}}
-{{% set hash_type = "{hash_type}" %}}
-{{% set hash_value = "{hash_value}" %}}
 
 """
 
@@ -93,8 +84,7 @@ PYPI_META_STATIC = {
         ('version', '{{ version }}'),
     ]),
     'source': ruamel_yaml.comments.CommentedMap([
-        ('url', '/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.{{ file_ext }}'),  # NOQA
-        ('{{ hash_type }}', '{{ hash_value }}'),
+        ('url', '/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz'),  # NOQA
     ]),
     'build': ruamel_yaml.comments.CommentedMap([
         ('number', 0),
@@ -312,6 +302,7 @@ def skeletonize(packages, output_dir=".", version=None, recursive=False,
             base_url = urlsplit(pypi_url)
             base_url = "://".join((base_url.scheme, base_url.netloc))
             ordered_recipe['source']['url'] = urljoin(base_url, ordered_recipe['source']['url'])
+            ordered_recipe['source']['sha256'] = d['hash_value']
 
             if d['entry_points']:
                 ordered_recipe['build']['entry_points'] = d['entry_points']
@@ -319,10 +310,7 @@ def skeletonize(packages, output_dir=".", version=None, recursive=False,
             if noarch_python:
                 ordered_recipe['build']['noarch'] = 'python'
 
-            ordered_recipe['build']['script'] = 'python setup.py install ' + ' '.join(setup_options)
-            if any(re.match(r'^setuptools(?:\s|$)', req) for req in d['build_depends']):
-                ordered_recipe['build']['script'] += ('--single-version-externally-managed '
-                                                      '--record=record.txt')
+            ordered_recipe['build']['script'] = 'python -m pip install . --no-deps --ignore-installed'
 
             # Always require python as a dependency
             ordered_recipe['requirements'] = ruamel_yaml.comments.CommentedMap()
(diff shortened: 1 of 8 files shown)


conda_build/skeletons/pypi.py
```diff
@@ -313,17 +302,15 @@ def skeletonize(packages, output_dir=".", version=None, recursive=False,
...
                 ordered_recipe['build']['entry_points'] = d['entry_points']
 
             if noarch_python:
                 ordered_recipe['build']['noarch'] = 'python'
 
-            ordered_recipe['build']['script'] = 'python setup.py install ' + ' '.join(setup_options)
-            if any(re.match(r'^setuptools(?:\s|$)', req) for req in d['build_depends']):
-                ordered_recipe['build']['script'] += ('--single-version-externally-managed '
-                                                      '--record=record.txt')
+            ordered_recipe['build']['script'] = 'python -m pip install . --no-deps --ignore-installed'
```
msarahan: If we are forcing pip by default, we also need to disable the hermetic build stuff, since that duplicates conda-build's intent and can also break what conda-build does.

https://github.com/pypa/pip/pull/5053/files#diff-1457c307ea972916e3ade69c27f0c41aR456

Now the thing that really stinks is that that flag is only available on new pip, so it'll break old builds, or ecosystems that don't have pip 10 yet.

I'm OK with the rest of the changes in this PR, but I don't think we're ready for this one yet.

jjhelmus: I think there is a environment variable we could set which would turn off build isolation and still work with pip 9

msarahan: great, that would be better.  I'll see if we can find it.

nehaljwani: Can we add `--no-cache-dir` to the pip arguments as well (to force it to re-create/re-download the wheel and not rely on local cache, if present)? I was bit by this quite a few times.

jjhelmus: `PIP_NO_BUILD_ISOLATION=False` disables build isolation.  My expectation was that this should be set to True but pypa/pip#5229 and my own testing confirms that False is the correct setting.  This variable has no effect on older pip versions.

I would also suggest adding `-vvv` to increase verbosity which makes debugging failed builds much easier.

msarahan:
I rebased onto master and added a bit.  I hope that's OK.

conda_build/skeletons/pypi.py
```diff
@@ -313,17 +302,16 @@ def skeletonize(packages, output_dir=".", version=None, recursive=False,
...
                 ordered_recipe['build']['entry_points'] = d['entry_points']
 
             if noarch_python:
                 ordered_recipe['build']['noarch'] = 'python'
 
-            ordered_recipe['build']['script'] = 'python setup.py install ' + ' '.join(setup_options)
-            if any(re.match(r'^setuptools(?:\s|$)', req) for req in d['build_depends']):
-                ordered_recipe['build']['script'] += ('--single-version-externally-managed '
-                                                      '--record=record.txt')
+            ordered_recipe['build']['script'] = ('python -m pip install . --no-deps '
```
nehaljwani: Should this be `$PYTHON` instead or is that not required?

msarahan: not strictly.  python should always be on PATH always in the env.  If you make it $PYTHON, that's more definite, but it's not strictly necessary.  I'll go ahead and change it, though.

mbargull: Hm.. although I'm usually in favor of generalizing and assuring correctness, I'd rather have just `python` instead of `"${PYTHON}"`/`"%PYTHON%"` here. See comments from https://github.com/conda-forge/conda-forge.github.io/pull/551#issuecomment-378331154 onward.
Just `python` is also the prevalent thing on `aggregate` and `feedstocks`.

msarahan: There was some reason why Intel had this pattern.  I can't remember.  It did have some advantage over depending on PATH.

The current code uses jinja2 instead of env vars, and should work.  Errr... it will, when I can get to a better internet connection that doesn't give me git errors.

ocefpaf:
Thanks for the review here. I'll try to address them ASAP, just got caught on a storm here at the day job.

msarahan:
Don't worry about it.  I'll fix the CI crap and it'll be ready to go.  Appveyor is really misbehaving, and I don't understand why yet.

msarahan:
finally fixed.  Having PYTHON set in the Appveyor matrix seemed to be overriding conda-build, which meant that the ``{{ PYTHON }}`` was something dumb, like ``C:\Miniconda2-x64``

PR #2984, opened 2018-06-23
Add QuickTime Kit Framework to default missing_dso_whitelist

nehaljwani:


mingwandroid:
The default whitelist is something I'm not 100% on (should anything exist, should we put this in a file instead?)..

Also not sure this is 'core' enough as not many things will use Quicktime these days and I expect it's deprecated.

Sorry to be so indecisive, passing the buck right along to @msarahan

nehaljwani:
I added it here since wxpython needed it. If it's deprecated, then oh, well.

mingwandroid:
Your can add it to the recipe? builds/missing_dso_whitelist

msarahan:
I'd prefer to add it to wxpython, and if we encounter it again, then we should add it here.

msarahan:
@nehaljwani I'm going to close this for now, but I'm open to re-opening it if you want to discuss this more.

Issue #2500, opened 2017-11-15
Add support for SSL_NO_VERIFY in conda skeleton pypi


bheklilr:
My company uses a very annoying proxy server that acts as a MITM attack on all SSL encrypted connections to hosts outside of our network.  This makes it very difficult to use conda skeleton to create conda packages from pypi packages.  Since it is very difficult and confusing to figure out how to extract the appropriate certificates from the IT approved browser, then set things up so that the `requests` package can access them, I have figured out how to work around this by modifying the code for `conda_build/skeletons/pypi.py`.  However, I would really like this feature to be integrated into conda skeleton so that I don't have to make this change every time I update, and so others can use it as well.  It's a very simple change, only 2 lines modified and 2 lines added in conda-build 3.0.19.

In the function `package_exists`, doing

```python
no_verify = os.environ.get('SSL_NO_VERIFY', '').strip().lower() in ('1', 'true')
r = requests.get(pypi_url + '/' + package_name, verify=not no_verify)
```

And very similarly in `skeletonize`

```python
no_verify = os.environ.get('SSL_NO_VERIFY', '').strip().lower() in ('1', 'true')
pypi_data = requests.get(package_pypi_url, verify=not no_verify).json()
```

With this change, users can simply set the `SSL_NO_VERIFY` environment variable to 1 or any variant of True to disable SSL verification.

(Before you ask: yes, this looks like a massive security hole.  However, my IT department assures me that our proxy server checks the certificates for me and it should be safe.)

msarahan:
Seems fine to me since is isn't default behavior.  Can you please submit a PR?

bheklilr:
Sure, I think I can get that submitted by the end of the day.

Would you also like me to look at adding it for the other types of repos? I shouldn't think it too difficult. Also, I think this would be a good addition to the docs. Would you mind if I added a little blurb about it where applicable?

msarahan:
Sure, go for it.  Whatever you have time for will be great.

bheklilr:
Could you link me to where the documentation source is for conda skeleton?  It doesn't appear to be in this repository.

PR #2991, opened 2018-06-29
Adding fix for bypassing MITM proxies to skeleton cran

j-hartshorn:
Replicated the changes in PR #2500 into `skeleton cran` which add a small function for getting the truthiness of the SSL_NO_VERIFY environment variable and passes that to requests.session to disable SSL verification.

### DIFF
--- a/conda_build/skeletons/cran.py
+++ b/conda_build/skeletons/cran.py
@@ -519,9 +519,19 @@ def get_latest_git_tag(config):
     print("Using tag %s" % tags[-1])
     return tags[-1]
 
+    
+def _ssl_no_verify():
+    """Gets whether the SSL_NO_VERIFY environment variable is set to 1 or True.
+
+    This provides a workaround for users in some corporate environments where
+    MITM style proxies make it difficult to fetch data over HTTPS.
+    """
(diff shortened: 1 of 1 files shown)


nehaljwani:
@j-hartshorn I think a cleaner approach would be for you to use:
```
export REQUEST_CA_BUNDLE=/path/to/your/organizations/internal/CA.pem
```

conda_build/skeletons/cran.py
```diff
@@ -520,8 +520,18 @@ def get_latest_git_tag(config):
...
+
+    This provides a workaround for users in some corporate environments where
+    MITM style proxies make it difficult to fetch data over HTTPS.
+    """
+    return environ.get('SSL_NO_VERIFY', '').strip().lower() in ('1', 'true')
+
+
 def get_session(output_dir, verbose=True):
     session = requests.Session()
+    session.verify = _ssl_no_verify()
```
j-hartshorn: I've just noticed that I made an error here, there should be a `not` before the function call. Apologies. How should I go about rectifying this, another pull request?

msarahan: Yeah, just another PR.  If you can, please add a test to make sure that bugs like this logic error don't creep in.

Issue #2953, opened 2018-06-06
conda convert throwing error with specific conda-build versions


saraneem:
### Actual Behavior
Getting the following error while using the conda convert command with osx package.
But able to convert the package with conda-build v3.4.1.

```
conda convert --platform all gsas2pkg-1.0.4-0.tar.bz2 -o outputdir/

Source platform 'osx-64' and target platform 'osx-64' are identical. Skipping conversion.
Converting gsas2pkg-1.0.4-0.tar.bz2 from osx-64 to linux-32
Converting gsas2pkg-1.0.4-0.tar.bz2 from osx-64 to linux-64
Converting gsas2pkg-1.0.4-0.tar.bz2 from osx-64 to linux-ppc64le
Converting gsas2pkg-1.0.4-0.tar.bz2 from osx-64 to linux-armv6l
Converting gsas2pkg-1.0.4-0.tar.bz2 from osx-64 to linux-armv7l
Converting gsas2pkg-1.0.4-0.tar.bz2 from osx-64 to linux-aarch64
Converting gsas2pkg-1.0.4-0.tar.bz2 from osx-64 to win-32
Traceback (most recent call last):
  File "/Users/smahalingam/anaconda3/bin/conda-convert", line 11, in <module>
    sys.exit(main())
  File "/Users/smahalingam/anaconda3/lib/python3.6/site-packages/conda_build/cli/main_convert.py", line 130, in main
    return execute(sys.argv[1:])
  File "/Users/smahalingam/anaconda3/lib/python3.6/site-packages/conda_build/cli/main_convert.py", line 126, in execute
    api.convert(f, **args.__dict__)
  File "/Users/smahalingam/anaconda3/lib/python3.6/site-packages/conda_build/api.py", line 304, in convert
    dry_run=dry_run, dependencies=dependencies)
  File "/Users/smahalingam/anaconda3/lib/python3.6/site-packages/conda_build/convert.py", line 783, in conda_convert
    dependencies, verbose)
  File "/Users/smahalingam/anaconda3/lib/python3.6/site-packages/conda_build/convert.py", line 675, in convert_from_unix_to_windows
    update_paths_file(temp_dir, target_platform='win')
  File "/Users/smahalingam/anaconda3/lib/python3.6/site-packages/conda_build/convert.py", line 380, in update_paths_file
    path['sha256'] = update_executable_sha(temp_dir, path['_path'])
[...]

PR #3014, opened 2018-07-14
Fix the file renaming during win to linux conversion

mandeep:
When renaming files that end with -script.py, the update_executable_path
function looks for files that end with -script.py and should remove
the -script.py suffix from the file. Currently, only the .py suffix
is being removed. This commit fixes this issue by removing the entire
-script.py suffix.


msarahan:
Could we please add a test here?  #2953 should be a reproducer.

mandeep:
I've added a test that fails on master but passes on this branch.

PR #3022, opened 2018-07-15
support multiple exclusive_config_files

mbargull:
@msarahan in https://gitter.im/conda/conda-build?at=5b0dc3d299fa7f4c062c0217:
>> Would you accept a PR to extend either the --exclusive-config-file CLI parameter or the conda_build: config_file .condarc parameter to accept multiple paths?
    The use case is Bioconda which is essentially a conda-forge downstream repository now. As such we want to use conda-forge-pinning's conda_build_config.yaml but also add our own on top of that. Using --variant-files won't work, due to the config file processing order, as soon as we would start to add per-recipe configuration files.
>
> Sure, I think that's fine. Just be really clear what the priority is. I think it would be most consistent to have later files have higher priority (this is backwards compared to channels in conda, but more consistent with the CLI behavior, and I hope consistent with the variant config files entry in condarc).

@mbargull in https://gitter.im/conda/conda-build?at=5b0dc77799fa7f4c062c1006:
>> I think it would be most consistent to have later files have higher priority
>
> Indeed, it would be `conda-build -e lower-prio-config.yaml -e higher-prio-config.yaml`

**EDIT: NB: not tested yet....**

### DIFF
--- a/conda_build/config.py
+++ b/conda_build/config.py
@@ -119,9 +119,9 @@ DEFAULTS = [Setting('activate', True),
 
             # variants
             Setting('variant_config_files', []),
-            # this file precludes usage of any system-wide or cwd config files.  Config files in
-            #    recipes are still respected, and they override this file.
-            Setting('exclusive_config_file', None),
+            # these files preclude usage of any system-wide or cwd config files.
+            #    Config files in recipes are still respected, and they override this file.
+            Setting('exclusive_config_files', []),
             Setting('ignore_system_variants', False),
             Setting('hash_length', 7),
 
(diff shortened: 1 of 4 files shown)


conda_build/config.py
```diff
@@ -342,6 +342,23 @@ def target_subdir(self):
...
 
+    @property
+    def exclusive_config_file(self):
+        if self.exclusive_config_files:
+            return self.exclusive_config_files[0]
+        return None
+
+    @exclusive_config_file.setter
+    def exclusive_config_file(self, value):
+        if len(self.exclusive_config_files) > 1:
```
mbargull: Is it safe to assume `self.exclusive_config_files` is always a `list` or should we also support the case where someone sets it to `None`?

msarahan: Yes, I think this is a decent assumption.  We can wrap it with ensure_list if necessary.

msarahan:
The codacy error is wrong.  It can't seem to follow the indirect logic.  Not surprising, I guess - it is not entirely straightforward.

You have one build failure in your new tests: https://travis-ci.org/conda/conda-build/jobs/404215494#L2174

I'll fix it and push here.

mbargull:
> The codacy error is wrong. It can't seem to follow the indirect logic. Not surprising, I guess - it is not entirely straightforward.

My thoughts exactly; though I'm surprised it didn't encounter that before. Does Codacy only inspect code affected by changes? If not, we may want to try to convince it it's fine (would it respect `# noqa` or something?) so this doesn't later pop up all the time.

> You have one build failure in your new tests: https://travis-ci.org/conda/conda-build/jobs/404215494#L2174
I'll fix it and push here.

Duh! Thanks!

msarahan:
Test fails unrelated.  Thanks @mbargull - merging.

PR #3027, opened 2018-07-19
fix test/requires being ignored when --no-copy-test-source-files

msarahan:
addresses https://github.com/vmtk/vmtk/pull/313#issuecomment-406146264

CC @jcfr

### DIFF
--- a/tests/test_api_build.py
+++ b/tests/test_api_build.py
@@ -1185,6 +1185,14 @@ def test_copy_test_source_files(testing_config):
             assert not copy, "'info/test/test_files_folder/text.txt' not found in tar.bz2 but copying test source files. File list: %r" % files
 
 
+def test_copy_test_source_files_deps(testing_config):
+    recipe = os.path.join(metadata_dir, '_test_test_source_files')
+    for copy in (False, True):
+        testing_config.copy_test_source_files = copy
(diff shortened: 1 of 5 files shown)


tests/test_api_build.py
```diff
@@ -1185,6 +1185,14 @@ def test_copy_test_source_files(testing_config):
             assert not copy, "'info/test/test_files_folder/text.txt' not found in tar.bz2 but copying test source files. File list: %r" % files
 
 
+def test_copy_test_source_files_deps(testing_config):
+    recipe = os.path.join(metadata_dir, '_test_test_source_files')
+    for copy in (False, True):
+        testing_config.copy_test_source_files = copy
+        # test is that pytest is a dep either way.  Builds will fail if it's not.
```
jcfr: `test is that pytest` -> `test that pytest`

msarahan: Thanks.  I think this is correct wording - I'm saying what the test is.  Anyway, I'm both too lazy and not willing to do another CI cycle to change this.

msarahan:
Test fails are unrelated.  Merging.

jcfr:
Thanks for addressing the issue so quickly :+1:

PR #3029, opened 2018-07-19
Fix documentation of search order for conda_build_config.yaml

tjd2002:
These corrections are based on my testing and reading of the 'find_config_files' function in 'variants.py' in the conda-build 3 source code ( https://github.com/conda/conda-build/blob/e2a4a8fd1d29347f6c12f8464d9537486b4be983/conda_build/variants.py#L140-L170 )

'cwd' was added to the search path for conda-build 3.0.0 in commit: conda/conda-build@36c2725

The behavior where the conda_build/config_file entry overrides the default location in home dir appears to be even older.

[PR moved from conda/conda-docs#595 per request from @msarahan]

### DIFF
--- a/docs/source/variants.rst
+++ b/docs/source/variants.rst
@@ -293,9 +293,10 @@ Creating conda-build variant config files
 
 Variant input files are yaml files.  Search order for these files is the following:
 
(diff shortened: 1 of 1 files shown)


msarahan:
Thanks!  I'll cancel CI on this since it is just docs.

PR #3038, opened 2018-07-20
put local channel behind user-specified channels, for remote priority

### LINKED ISSUE #3036: Test environment incorrectly prioritizes local channel
Scenario:

- Developer is working on two conda packages, "pkgA" and "pkgB".
- Developer builds pkgA at version 1.0
- Developer uploads pkgA=1.0 to conda channel
- Second developer (or the same developer on a different machine) works on pkgA at version 2.0.
- Second developer uploads pkgA=2.0 to conda channel
- Developer starts work on pkgB, which has a dependency on pkgA, and requires functionality only present in pkgA >= 2.0.
- Developer attempts to build pkgB but unit tests fail.

Actual behavior:

The test environment is pulling pkgA=1.0 from the local conda-bld folder, instead of the newer pkgA=2.0 from their conda channel. This causes tests to fail and the build is broken.

Expected behavior:

The test environment pulls from the user's defined set of channels before it falls back to the local channel. This will cause pkgA=2.0 to be pulled as expected and unit tests to pass.

msarahan:
fixes #3036 

Channels specified via --channel, -c on the CLI, or channel_urls (in the API) will take priority over locally built packages.  Locally built packages still take priority over anything configured in condarc.

### DIFF
--- a/conda_build/index.py
+++ b/conda_build/index.py
@@ -71,14 +71,18 @@ def get_build_index(subdir, bldpkgs_dir, output_folder=None, clear_cache=False,
             log_context = partial(utils.LoggingContext, logging.CRITICAL + 1)
             capture = utils.capture
 
-        # priority: channels passed as args, local (can vary with either croot or output_folder),
+        # priority: (local as either croot or output_folder IF NOT EXPLICITLY IN CHANNEL ARGS),
+        #     then channels passed as args (if local in this, it remains in same order),
         #     then channels from condarc.
         urls = list(channel_urls)
 
         # this is where we add the "local" channel.  It's a little smarter than conda, because
         #     conda does not know about our output_folder when it is not the default setting.
         if os.path.isdir(output_folder):
-            urls.append(url_path(output_folder))
+            if 'local' not in urls:
+                urls.insert(0, url_path(output_folder))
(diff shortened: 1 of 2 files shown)


mingwandroid:
@msarahan, local priority should be the default here if people do not specify it, for those wanting a different order they should specify that explicitly I think.

mingwandroid:
Doesn't this change mean it's no longer the case that people who use `-c` (like I do, for every build) will have `local` appear first in the list? It *requires* them to add it themselves which will break peoples' workflows.

msarahan:
If you put it for every build, why don't you put it in condarc instead?  There has to be a way for people to put channels ahead of local.  You know I never want to break your workflow, but this is crippled right now.  I can put this behind a flag, I guess, but having the split between CLI-specified things and condarc things makes sense to me.

msarahan:
Anyway, not going to rush this into the next release, since it is contentious.

msarahan:
@Bezier89 @mingwandroid I think this is better now.  This will preserve existing behavior when local is not one of the manually specified channels.  However, it should now be possible to put some other channel ahead of local by specifying both the other channel and local as arguments:

``conda build -c somechannel -c local somerecipe``

will put somechannel ahead of local, while

``conda build -c somechannel somerecipe``

will not.

Issue #3046, opened 2018-07-26
Additional information in package's info/about.json


mikecormier:
### Actual Behavior

The info/about.json excludes many of the keys in the package's meta.yaml. 
### Expected Behavior

That the  info/about.json includes the information of the "extra" key in the package's meta.yaml file.
Both the bioconda group and our group, Got Get Data (GGD), would like the extra information added to the about.json file in order to utilize this information for registry querying and tag based searching. Conda is also looking into adding an additional metadata channel json file, which would include a combination of the info/index.json and info/about.json with the "extra" key info included. (https://github.com/conda/conda/issues/7583)

Could you add the information from the "extra" key in the meta.yaml file to the info/about.json file?

msarahan:
Seems reasonable.  Please consider putting up a PR.  We're too busy to work on this ourselves right now.

mikecormier:
Hi @msarahan,

I have create a PR that includes the extra field of the meta.yaml file in the about.json file. 
https://github.com/conda/conda-build/pull/3048

PR #3048, opened 2018-07-26
Include "extra" field of meta.yaml in about.json

mikecormier:
https://github.com/conda/conda-build/issues/3046

@msarahan

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -540,6 +540,8 @@ def write_about_json(m):
             pass
         env = environ.Environment(root_dir)
         d['root_pkgs'] = env.package_specs()
+        # Include the extra section of the metadata in the about.json
+        d['extra'] = m.get_section('extra')  
         json.dump(d, fo, indent=2, sort_keys=True)
 
 
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -541,7 +541,7 @@ def write_about_json(m):
         env = environ.Environment(root_dir)
         d['root_pkgs'] = env.package_specs()
         # Include the extra section of the metadata in the about.json
-        d['extra'] = m.get_section('extra')  
+        d['extra'] = m.get_section('extra')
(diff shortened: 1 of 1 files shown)


mikecormier:
@msarahan,

Looking over the failed test from Travis CI and Appveyor I cannot see any reason that the changes I made would cause those errors. Were these errors present prior to my PR? Do you see any reason why the changes I have made would cause these errors? 

Failed Test:
```
tests/test_api_build.py::test_noarch_python_with_tests
tests/test_api_skeleton.py::test_repo[-rpm-libX11-devel-]
```

The only error that can be traced back to the change/addition I made is in the "Allowed Failure" test in Travis CI run. It errors because of an "E501 line to long error", which is caused by the addition of the "extra" field into the about.json file. 
```
./conda_build/build.py:551:44: E261 at least two spaces before inline comment
./conda_build/build.py:551:101: E501 line too long (105 > 100 characters)
```
The "E261 at least two spaces before inline comment" error can be easily fixed. (I just need to add an additional space to the comment I added)

Additionally, the python 2.7 run on Travis CI did not finish running due to the job timing out. We have had similar problems result from Travis CI building on a public repo for a max of 50 minutes: https://docs.travis-ci.com/user/customizing-the-build#build-timeouts. The script timed out at 49 minutes and 58 seconds.

msarahan:
The CI fails are fine.  Please do fix the flake8 stuff, though.

mikecormier:
@msarahan,

The flake8 errors are fixed, and the test is passing. The CI jobs are still running, but I don't expect anything to have changes for them to pass.

mikecormier:
@msarahan,

I have rebased this PR with the most current commit. The Travis CI runs all pass, except for the python 2.7 run which times out. The Appveyor CI runs still fail. These results are consistent with the latest commit.

msarahan:
Thanks, merging

Issue #2748, opened 2018-03-09
Configure `pip` during build to avoid dependencies


jakirkham:
Currently we patch `setuptools` to know when `conda-build` is running and disable pulling in dependencies. This is a good thing and avoids a fair number of issues. However am unaware of us doing anything on this front with `pip`, which could be problematic if users start pulling in dependencies that way by accident. As this has become a recommendation for Python packages in conda-forge ( https://github.com/conda-forge/conda-forge.github.io/pull/518 ), this deserves more consideration.

One option would be to address this in `conda-build`, this could amount to setting a `pip.conf` or `pip.in` for `pip` to use based on one internally included and maintained in `conda-build`. Another option would be to set a number of environment variables for the active build environment to get the same effect. In either case this should allow us to block downloads, avoid build isolation, disable caching, and any other behavior that might be confusing or problematic for end users building `conda` packages with `pip`.

msarahan:
Conda-build is already setting env vars for when you're in a build.  We'd need to patch pip to find and act on those env vars, but I really don't think this is a conda-build issue.  Instead, it is one for pip-feedstock.

jakirkham:
We could patch the `pip` feedstock. It would probably need to be an activation script.

jakirkham:
Alternatively we could have a post-link script that writes to the `pip.conf`/`pip.ini` in whatever standard location we decide it should live in only if `conda-build` is running. ( https://github.com/pypa/pip/issues/5060 )

msarahan:
Why does it have to be a pip.conf file?  I mean, we can certainly do that, if it's the easiest way.  If that's what we do, then doing it in conda-build is going to be much nicer than doing it in a post-link or activate script.

If we have to hack pip at all to make it work, why not patch it like setuptools?

jakirkham:
We could also teach `pip` where to look for this config file ( https://github.com/pypa/pip/issues/5060 ), which could easily be sent upstream.

jakirkham:
Noticed there were a couple things missing. Added PR ( https://github.com/conda/conda-build/pull/3053 ) to set them as well.

PR #3053, opened 2018-07-27
Configure `pip` to disable pulling in dependencies

jakirkham:
Fixes https://github.com/conda/conda-build/issues/2748

We never want `pip` to install dependencies in the build as we will supply them ourselves.

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -1399,6 +1399,10 @@ def build(m, stats, post=None, need_source_download=True, need_reparse_in_env=Fa
                     # hard-code this because we never want pip's build isolation
                     #    https://github.com/conda/conda-build/pull/2972#discussion_r198290241
                     env["PIP_NO_BUILD_ISOLATION"] = False
+                    # some other env vars to have pip ignore dependencies.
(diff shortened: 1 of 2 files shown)


jjhelmus:
Does it make sense to add `env["PIP_NO_BUILD_ISOLATION"] = False` as well to disable build isolation?

jakirkham:
Yeah was just trying to figure out if we were doing that or not earlier as well. Seems like we are, which is good.

jakirkham:
Looks like there are a couple CI failures related to some `rpm` test where the HTTP request fails. Seems unrelated. Also one Travis CI test timed out.

msarahan:
Added some small comments about the counterintuitive logic of the "NO" env vars.  Thanks @jakirkham

jakirkham:
Thanks @msarahan.

Yeah the "NO" environment variables are still very weird. In addition to the counterintuitive logic, some of them will crash if set incorrectly (as they expect more arguments).

xref: https://github.com/pypa/pip/issues/2897

msarahan:
Frustrating, but at least they are consistent.

PR #3054, opened 2018-07-28
skeleton (RPM): Update to CentOS 6.10

mingwandroid:


mingwandroid:
Fixes https://travis-ci.org/conda/conda-build/jobs/409028658#L1659

A *slight* [warning](https://wiki.centos.org/Manuals/ReleaseNotes/CentOS6.10):

```
NOTE: There is NO mechanism to pick only partial upgrades of packages to CentOS-6.10. All packages and updates to the 6.10 tree are built against the 6.10 tree and may not work correctly with older 6.x packages. If you want some packages in 6.10, please upgrade all packages. You will have issues if you perform only partial updates.
```

I think we'll be ok here, probably, buy would welcome thoughts on whether we should stick to 6.9? Chances are we'll not make many more CDT packages anyway and since we don't distribute the packages and the end-user runs their own distro version the changes of a conflict are more slim.

I can probably find the archive for 6.9 if people think that safe? @msarahan, @nehaljwani, @jjhelmus

msarahan:
Seems safe enough to me. Surely enterprise users keep up with minor releases.

PR #3068, opened 2018-08-08
Master small fixes

mingwandroid:
Eh, the cran change isn't that small thinking about it, the other one should catch the problem (or similar ones) that we had with setuptools_scm.

### DIFF
--- a/conda_build/skeletons/cran.py
+++ b/conda_build/skeletons/cran.py
@@ -953,11 +953,10 @@ def skeletonize(in_packages, output_dir=".", output_suffix="", add_maintainer=No
                         print("ERROR: No binary nor old binary found")
                         sys.exit(1)
                     # Version needs to be stored in archive_details.
+                    # TODO: Sort the versions (though I've never seen anything other than 1 element)
                     archive_details['conda_version'] = archive_details['binaries'][package][-1]
                     archive_details['cran_version'] = archive_details['conda_version'].replace('_', '-')
                     avaliable_artefact = True
-            # TODO :: Implement an option to use the most recent binary instead. We'd need to split the versions out.
-            #         For now it will fallback to compiling from source.
             if archive_details['use_this'] and avaliable_artefact:
                 if is_tarfile:
                     filename = basename(location)
--- a/conda_build/skeletons/cran.py
+++ b/conda_build/skeletons/cran.py
@@ -7,8 +7,8 @@ from __future__ import absolute_import, division, print_function
 import argparse
 from itertools import chain
 from os import makedirs, listdir, sep, environ
-from os.path import (basename, commonprefix, exists, isabs, isdir,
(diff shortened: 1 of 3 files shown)


mcg1969:
Can you briefly summarize what the functional changes are to the CRAN skeleton? They look potentially very useful but some I don't quite understand.

jjhelmus:
Would conda-verify be a better location for the dist-info check?  It already has some similar checks for egg, and pth files.

mingwandroid:
I dislike putting simple checks like this into conda-verify because (e.g.) we didn't run it for years and running this check should not be optional.

mingwandroid:
The binary releases for R seem a bit haphazard, i.e. you'll find packages where the latest (at the snapshot date) source bundle is 1.2 but the latest macOS binary is 1.1 while windows has 1.2, so I add an option for what to do in the case when you wanted binary repackaging but the correct version does not exist. The options are to use the older one (may not work, depending on whether it was actually rebuilt or not) or to compile from source in the normal way.

I also do lots of dylib fixups so that our compiler runtimes are used instead of those provided by MRO (because once we load those old ones we cannot offer any modern language features).

conda_build/skeletons/cran.py
```diff
@@ -143,6 +148,33 @@
...
+  if [[ $target_platform == osx-64 ]]; then
+    pushd $PREFIX
+      for libdir in lib/R/lib lib/R/modules lib/R/library lib/R/bin/exec sysroot/usr/lib; do
+        pushd $libdir || exit 1
+          for SHARED_LIB in $(find . -type f -iname "*.dylib" -or -iname "*.so" -or -iname "R"); do
+            echo "fixing SHARED_LIB $SHARED_LIB"
+            install_name_tool -change /Library/Frameworks/R.framework/Versions/3.5.0-MRO/Resources/lib/libR.dylib "$PREFIX"/lib/R/lib/libR.dylib $SHARED_LIB || true
+            install_name_tool -change /Library/Frameworks/R.framework/Versions/3.5/Resources/lib/libR.dylib "$PREFIX"/lib/R/lib/libR.dylib $SHARED_LIB || true
+            install_name_tool -change /usr/local/clang4/lib/libomp.dylib "$PREFIX"/lib/libomp.dylib $SHARED_LIB || true
+            install_name_tool -change /usr/local/gfortran/lib/libgfortran.3.dylib "$PREFIX"/lib/libgfortran.3.dylib $SHARED_LIB || true
```
msarahan: the libgfortran so version changes in recent GCC (it's on 5 in gcc 8.2).  We're still on gcc 4.8 for mac, so it doesn't matter here yet, but should we constrain libgfortran somehow in light of this?

mingwandroid: If I need to pin libgfortran I'll do it in r-base only I think.

Issue #2716, opened 2018-02-28
variant variables are not looped over in source section


teake:
The [Git environment variables](https://conda.io/docs/user-guide/tasks/build-packages/environment-variables.html#git-environment-variables) seem to be stuck when using them versioning variants.

### Actual Behavior

With a `meta.yaml` that contains:
```
package:
  name: my_package
  version: {{ GIT_DESCRIBE_HASH }}

source:
  - git_rev: {{ git_branch }}
    git_url: my_git_url
```
and a `conda_build_config.yaml` that contains:
```
git_branch:
  - master
  - develop
```
`conda build` outputs two packages, but they're identically versioned:
```
my_package-g3b438ce6-pyh6c1e8c2_0.tar.bz2
my_package-g3b438ce6-pyh94fc622_0.tar.bz2
```
(Note that the contents of the packages is different and does correspond to their respective branches, unlike #2450).

### Expected Behavior

I expect to see two differently versioned packages being output (provided that the specified branches point to different commits).

##### `Output of conda info`

```
     active environment : None
          conda version : 4.4.11
    conda-build version : 3.5.1
         python version : 3.6.2.final.0
               platform : osx-64
             user-agent : conda/4.4.11 requests/2.18.4 CPython/3.6.2 Darwin/17.4.0 OSX/10.13.3
                UID:GID : 502:20
             netrc file : None
           offline mode : False
```

msarahan:
Changed the title to something I think is more accurate.  It should be possible to fix this, but it may take some reordering/refactoring, as I'm not sure off the top of my head exactly how/when the source download factors into the iterative rendering.  I don't have a timeline for you, though.  I'll try to get it in soon, but I'm behind on other "must do" things.  PRs welcome.

teake:
I think the title is less accurate now. Variant variables (`git_branch` in my example) do apply to the source section, as the different branches get checked out correctly. But the Git variables (`GIT_DESCRIBE_HASH`) do not get re-derived for each variant (or so it seems). Or I have the terminology wrong, in which case never mind.

Nice to see this being placed on the roadmap (though I'm afraid I won't have time to put in a PR).

msarahan:
I think you're right.  One more try.

PR #3082, opened 2018-08-16
Fix cached git info for variants

### LINKED ISSUE #2450: Build variants does not work with source repo revision variations
With a `meta.yaml` that contains:
```
source:
  svn_url: an_url
  svn_rev: {{ version }}
```
and a `conda_build_config.yaml` with:
```
version:
  - 1
  - 2
```
Both packages are built from the same sources of the first version, the repo is only checked out once.

Am I doing it properly or is it a bug?

teake:
~~Closes #2450.~~

Whoops that would be #2716.


msarahan:
Thanks for finding the time.  Test fails look unrelated.  Merging.

PR #3085, opened 2018-08-20
py37 surprise CI

kalefranz:


kalefranz:
Appears that this just affects master and not 4.5.x.  https://travis-ci.org/conda/conda-build/builds/418371219#L712

PR #3086, opened 2018-08-20
py37 surprise

kalefranz:


msarahan:
The core issue here is reported in https://github.com/conda/conda/issues/8076 and fixed by https://github.com/conda/conda/pull/8083

PR #3102, opened 2018-08-30
filter .la files

ocefpaf:
@msarahan this would be a big help to sort out https://github.com/conda-forge/conda-forge.github.io/issues/621 in `conda-forge`.

### DIFF
--- a/conda_build/utils.py
+++ b/conda_build/utils.py
@@ -1234,6 +1234,7 @@ def apply_pin_expressions(version, min_pin='x.x.x.x.x.x.x', max_pin='x'):
 def filter_files(files_list, prefix, filter_patterns=('(.*[\\\\/])?\.git[\\\\/].*',
                                                       '(.*[\\\\/])?\.git$',
                                                       '(.*)?\.DS_Store.*',
+                                                      '.*\.la',
                                                       'conda-meta.*')):
     """Remove things like the .git directory from the list of files to be copied"""
     for pattern in filter_patterns:
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -218,7 +218,7 @@ def test_filter_files():
     #    lib/python3.4/site-packages/craftr/stl/craftr.utils.git/Craftrfile
     files_list = ['a', 'x.git/a', 'something/x.git/a',
                   'x.git\\a', 'something\\x.git\\a', 'something/.gitmodules',
-                  'some/template/directory/.gitignore']
+                  'some/template/directory/.gitignore', 'file.la', 'something/file.la']
     assert len(utils.filter_files(files_list, '')) == len(files_list)
 
 


conda_build/utils.py
```diff
@@ -1234,6 +1234,7 @@ def apply_pin_expressions(version, min_pin='x.x.x.x.x.x.x', max_pin='x'):
 def filter_files(files_list, prefix, filter_patterns=('(.*[\\\\/])?\.git[\\\\/].*',
                                                       '(.*[\\\\/])?\.git$',
                                                       '(.*)?\.DS_Store.*',
+                                                      '.*\.la',
```
ocefpaf: PS: last time I wrote a regular expression was many years ago! So I'm not certain this is bullet proof. (It work on a few feedstocks I tested.)

jjhelmus: I think this regex needs a trailing $ so it doesn't match files like `example.lab`

msarahan:
makes sense to me.  Thanks for the fix on the tests.  Will wait for CI and merge.

ocefpaf:
> makes sense to me. Thanks for the fix on the tests. Will wait for CI and merge.

Sorry it took me a while the read the tests code and comments :smile:

mbargull:
In comparison to `.git/`, `.DS_Store`, `conda-meta`, that `.la` is slightly more generic. So maybe there is a small probability that some french developer (*Cliché! Cliché! Cliché!* -- *\*slapping myself for it\**) creates some `oh.la.la` file that is no Libtool archive file.
Maybe this could/should also inspect the file's header to make sure it's a proper `.la` file or, alternatively, the user would need a way to negate that filter, e.g., a 
```
build:
  no-no-filter-files-please:  # intentional stupid name: I'm not proposing anything concrete here ;)
    - '*.la'
```
entry in `meta.yaml`.
NB: Talking about stupid cases with minuscule probabilities here, but just wanted to make sure this has been brought up somewhere.

msarahan:
``oh.la.la`` - you crack me up.  It's a good point, and we should revisit it if anyone reports problems.

ocefpaf:
Most linux distros do a simple `rm -f %{buildroot}%{p_libdir}/*.la`.

I guess we can do better by limiting the removal to the `/lib` directory. Still that is not 100% bullet proof.

epruesse:
It might be useful to have `conda-build` at least comment on the files it's ignoring. Something like `Filtered 999 files from package. See https://docs-page`.

I remember spending an entire day trying to figure out why a build system had seized to work, eventually tracing it to Apple's new SIP deleting `DYLD_*` from the environment IFF the binary (in this case `/bin/sh` called by `/usr/local/bin/make`) was a system binary. Oh how I cursed them for that grandiose, at the time completely undocumented "feature".

PR #3104, opened 2018-09-01
Do not disable pip's cache, rather point in to a location inside the work dir

### LINKED ISSUE #3094: Disabling pip's cache directory results in .egg-info directories
#3053 disabled pip's cache directory by default in the build environment.  

An unfortunate side effect of disabling pip's cache directory is that an `.egg-info` directory is produced rather than a `.dist-info` directory.  This is a result a pip not generating wheels during the install process.

For example, take the following recipe:

``` yaml
package:
  name: imagesize
  version: 1.0.0

source:
  url: https://pypi.io/packages/source/i/imagesize/imagesize-1.0.0.tar.gz
  sha256: 5b326e4678b6925158ccc66a9fa3122b6106d7c876ee32d7de6ce59385b96315

build:
  #script: python -m pip install .
  script: unset PIP_NO_CACHE_DIR; python -m pip install .

requirements:
  host:
    - python
    - pip
  run:
    - python
```

As written, pip's cache directory is disabled in conda-build >= 3.12.1.  The resulting package will contain an `.egg-info` directory:

```
lib/python3.6/site-packages/imagesize-1.0.0-py3.6.egg-info/dependency_links.txt
lib/python3.6/site-packages/imagesize-1.0.0-py3.6.egg-info/top_level.txt
lib/python3.6/site-packages/imagesize-1.0.0-py3.6.egg-info/installed-files.txt
lib/python3.6/site-packages/imagesize-1.0.0-py3.6.egg-info/SOURCES.txt
```

Using the commented script, re-enables the cache directory, resulting in a wheel being generated as an intermediate.  The conda package then contains a `.dist-info` directory.

``` 
lib/python3.6/site-packages/imagesize-1.0.0.dist-info/INSTALLER
lib/python3.6/site-packages/imagesize-1.0.0.dist-info/top_level.txt
lib/python3.6/site-packages/imagesize-1.0.0.dist-info/WHEEL
lib/python3.6/site-packages/imagesize-1.0.0.dist-info/RECORD
lib/python3.6/site-packages/imagesize-1.0.0.dist-info/LICENSE.txt
lib/python3.6/site-packages/imagesize-1.0.0.dist-info/METADATA
```

I believe that `dist-info` directories should be preferred and these are currently not the default. 

It is possible that the python version entry in the `egg-info` directories (**py3.6**.egg-info) could be problematic for `noarch: python` packages although I have no evidence of this.

jjhelmus:
closes #3094

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -18,6 +18,7 @@ import string
 import subprocess
 import sys
 import tarfile
+import tempfile
 import hashlib
 import logging
 import time
@@ -1424,6 +1425,9 @@ def build(m, stats, post=None, need_source_download=True, need_reparse_in_env=Fa
                     # pip's cache directory (PIP_NO_CACHE_DIR) should not be
                     # disabled as this results in .egg-info rather than
                     # .dist-info directories being created, see gh-3094
+
+                    # set PIP_CACHE_DIR to a path in the work dir that does not exist.
+                    env['PIP_CACHE_DIR'] = tempfile.mktemp(dir=m.config.work_dir)
 
                     work_file = join(m.config.work_dir, 'conda_build.sh')
                     with open(work_file, 'w') as bf:
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -1421,9 +1421,9 @@ def build(m, stats, post=None, need_source_download=True, need_reparse_in_env=Fa
                     #    See note above about inverted logic on "NO" variables
                     env["PIP_NO_DEPENDENCIES"] = False
                     env["PIP_IGNORE_INSTALLED"] = True
-                    # disable use of pip's cache directory.
-                    #    See note above about inverted logic on "NO" variables
-                    env["PIP_NO_CACHE_DIR"] = False
+                    # pip's cache directory (PIP_NO_CACHE_DIR) should not be
+                    # disabled as this results in .egg-info rather than
+                    # .dist-info directories being created, see gh-3094
 
                     work_file = join(m.config.work_dir, 'conda_build.sh')
                     with open(work_file, 'w') as bf:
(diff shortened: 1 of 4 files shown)


jjhelmus:
Disabling pip's cache results in [pip not building wheels](https://github.com/pypa/pip/blob/10.0.1/src/pip/_internal/commands/install.py#L293-L295) which leads to the conda package containing Python versioned `.egg-info` directories rather than modern `.dist-info` directories.  This change leaves the choice of disabling the cache up to the user (pip by default uses the cache directory) but when the cache is enabled it points to a location inside the work directory.  This avoid situations where the wheels from the default pip cache are installed into the conda environment.

jjhelmus:
My choice of `mktemp` was intentional despite the function being deprecated and unsafe.  Other options from the tempfile module would create a folder inside the work directory which is not ideal.  If there is better method to generate a unique directory name but not actually create the directory I'm all ears.

nehaljwani:
Wouldn't this hurt the builder in the scenario where `merge_build_host: True` ? Wouldn't the cache directory would then become part of the package itself?

How about instead of putting it in the `_build_env` directory, we shove it one level up and call it something like `__pycache_<(not-necessarily!?-)long-path-here>` ?

nehaljwani:
Sorry, I misread. You are actually using the work directory. Nevertheless, I would still like to know your thoughts about a top level __pipcache diretory.

nehaljwani:
Or we could still put it in the work directory, but prefix it with `__pipcache`. I want an identifier, so that I wouldn't have to go inside the work directory and hunt for the cache directory whereas I could simply `/path/to/worl-or-top-level/__pipcache<hit-tab>`

msarahan:
There's one more of these vars in windows.py.

If it helps @nehaljwani, I agree that it would be helpful to have a folder for these files, though I personally don't know that I would use it.

jakirkham:
Thanks all. This seems like a nice, simple fix to the issue.

PR #3108, opened 2018-09-04
Memoize read_meta_file

### LINKED ISSUE #3107: Conda multi output recipe takes 9+ minutes to render
I am using conda build 3.13.0, but the behaviour is similar in the recent versions.
```
/t/llvm-compilers-linux ❯❯❯ /usr/bin/time ~/m3/bin/conda render . > /dev/null                                                                                                                               ✘ 127 
645.78user 0.71system 10:49.03elapsed 99%CPU (0avgtext+0avgdata 88532maxresident)k
96inputs+32outputs (1major+55066minor)pagefaults 0swaps
```
CB takes 645 seconds just to render this [recipe](https://github.com/conda/conda-build/files/2347465/llvm-compilers-linux.zip)

nehaljwani:
Fix #3107

### DIFF
--- a/conda_build/metadata.py
+++ b/conda_build/metadata.py
@@ -21,6 +21,7 @@ from .conda_interface import envs_dirs
 from .conda_interface import string_types
 
 from conda_build import exceptions, utils, variants, environ
+from conda_build.conda_interface import memoized
 from conda_build.features import feature_list
 from conda_build.config import Config, get_or_merge_config
 from conda_build.utils import (ensure_list, find_recipe, expand_globs, get_installed_packages,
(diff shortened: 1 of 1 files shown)


mingwandroid:
> CB takes 645 seconds just to render this recipe

What does it take after this?

msarahan:
This might mess up in creative ways.  I am pretty sure that it has been memoized in the past, but it bit us in some subtle way when a recipe path was the same for a second build, but was not the same recipe.  If that's in the tests (which I think it was), we should fix the tests.  This optimization is too important to not do.

PR #3109, opened 2018-09-04
tell pip to not go find things on PyPI (turn off downloading), simila…

msarahan:
…r to our setuptools monkeypatch

NOTE: right now, this is a bad idea, due to #3094.  Many packages erroneously think that a given thing is not installed.  This PR is a good idea in the long run, but we should probably wait to deploy it for a while, until the dust has settled from #3094

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -1867,6 +1867,10 @@ def write_build_scripts(m, script, build_file):
     # set PIP_CACHE_DIR to a path in the work dir that does not exist.
     env['PIP_CACHE_DIR'] = m.config.pip_cache_dir
 
+    # tell pip to not get anything from PyPI, please.  We have everything we need
+    # locally, and if we don't, it's a problem.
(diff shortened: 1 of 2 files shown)


mbargull:
> Many packages erroneously think that a given thing is not installed.

Do you mean `noarch: python` packages with misnamed `-py?.?.egg-info` folders with this or are there any other reasons?

msarahan:
I think this goes beyond noarch stuff.  I've been seeing it with numpy and scipy stuff that I've been working on.

mbargull:
Ah, interesting, thanks!

msarahan:
We can try it, I guess.  We may need to revert it, though.

msarahan:
strangely, it seems like it might break mercurial somehow?

https://travis-ci.org/conda/conda-build/jobs/462038881#L1725

I'll merge it and we'll see what happens...

PR #3123, opened 2018-09-07
add recipe_log_file and copy it into package (for ref about pkg recip…

msarahan:
…e history)


msarahan:
This is primarily for internal use, to build the RSS feed about what has changed.  It is meant to capture the recent git history of a recipe and include that with the package.  This history will then be parsed when creating the index, and commit messages newer than some threshold will be shown in the RSS feed.

PR #3133, opened 2018-09-11
allow channel auth when checking if package is built

### LINKED ISSUE #3127: skip-existing fails with basic auth channels
The `--skip-existing` option doesn't work for channels with basic authentication, i.e. those of the form `https://<user>:<pass>@<path>`. 

It fails to identify identical packages because the package index dictionary is indexed with canonical urls (without the `<user>:<pass>` bit), but the search url does contain `<user>:<pass>` section.

The culprit is this line: https://github.com/conda/conda-build/blob/520f858c2ce2d1cb447d4f1ded40db0845780378/conda_build/build.py#L2499

The `url` here contains the auth section, whereas the keys in the `index` dictionary are parsed from [channel urls wrapped in `Channel`](https://github.com/conda/conda/blob/21a8f3946d8cf9ce3803e2826be6e50400557f91/conda/models/channel.py#L493).

A relatively straightforward fix would be to do `str(Channel(url)) + '::' + metadata.dist())`, but `Channel` is not export by conda.

msarahan:
fixes #3127 

@teake - please test.  Please submit a PR with a test, too, if you can.  I don't have time for more than just throwing up what @kalefranz proposed in #3127

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -30,7 +30,7 @@ from bs4 import UnicodeDammit
 import yaml
 
 # used to get version
-from .conda_interface import env_path_backup_var_exists, conda_46
+from .conda_interface import env_path_backup_var_exists, conda_45, conda_46
 from .conda_interface import PY3
 from .conda_interface import prefix_placeholder
 from .conda_interface import TemporaryDirectory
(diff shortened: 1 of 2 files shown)


teake:
Add a test is a bit tricky, as I'll have to mock a simple HTTP file server with authentication.

I did test it though against my original use case, and it seems to work [after a few fixes](https://github.com/teake/conda-build/commit/a89da8ff26d77a8e4e63a2dcf727d8764e4f606f). Not sure how to get that into this PR, but I've created a PR-to-PR: msarahan/conda-build#11.

teake:
I forgot to mention that `update_index` gives a warning after this change:

```
WARNING:conda_build.index:The update_index function has changed to index all subdirs at once.  You're pointing it at a single subdir.  Please update your code to point it at the channel root, rather than a subdir.
```

It didn't before because it didn't get called as it was (wrongly?) indented differently.

msarahan:
Thanks @teake.  I'll fix the usage of update_index that's showing the warning.

msarahan:
Test fails are some unrelated screwup.  I have fixed the update_index warning elsewhere.  Merging.  Thanks @teake

PR #3136, opened 2018-09-12
source: If the local git cache can't be updated, delete it.

stuarteberg:
### Synopsis

In some cases, the a repo in the local `git_cache` can become corrupted in a way that prevents the user from building a recipe.  With this PR, conda-build will simply delete the locally cached repo if updating fails for any reason.  (If the error was due to corruption, maybe the recipe will build successfully on the next try.)

### Details

As a little motivating example, consider this (broken) recipe:

```yaml
package:
 name: foo
 version: 0.1

source:
  # Oops! This is a subdirectory, not the repo...
  git_url: https://github.com/stuarteberg/helloworld/helloworld
```

This will create a new directory `/minconda3/conda-bld/git_cache/github.com/stuarteberg/helloworld`, but then (of course) fail to populate it with a repo due to the incorrect `git_url`:

```
Cloning into bare repository '/minconda3/conda-bld/git_cache/github.com/stuarteberg/helloworld/helloworld'...
remote: Not Found
fatal: repository 'https://github.com/stuarteberg/helloworld/helloworld/' not found
```

So far, so good.  But now suppose the user fixes the recipe:

```yaml
  git_url: https://github.com/stuarteberg/helloworld
```

Since the git cache has been corrupted, the recipe can't be built:
```
BUILD START: ['foo-0.1-0.tar.bz2']
fatal: Not a git repository (or any of the parent directories): .git
Warning: failed to download source.  If building, will try again after downloading recipe dependencies.
Error was:
Command '['/usr/local/bin/git', 'fetch', 'origin', '+HEAD:_conda_cache_origin_head']' returned non-zero exit status 128.
Failed to download or patch source. Please see build log for info.
```

In this scenario, `conda-build` can't recover unless the user deletes the local cache herself.  (For example, with `conda clean --source-cache`.)

With this PR, the local cache is automatically deleted at the first sign of trouble.  (However, it is still incumbent on the user to re-try the recipe one more time.)

cc: @DocSavage

### DIFF
--- a/conda_build/source.py
+++ b/conda_build/source.py
@@ -203,19 +203,29 @@ def git_mirror_checkout_recursive(git, mirror_dir, checkout_dir, git_url, git_ca
     if not isdir(os.path.dirname(mirror_dir)):
         os.makedirs(os.path.dirname(mirror_dir))
     if isdir(mirror_dir):
-        if git_ref != 'HEAD':
-            check_call_env([git, 'fetch'], cwd=mirror_dir, stdout=stdout, stderr=stderr)
-        else:
-            # Unlike 'git clone', fetch doesn't automatically update the cache's HEAD,
-            # So here we explicitly store the remote HEAD in the cache's local refs/heads,
-            # and then explicitly set the cache's HEAD.
-            # This is important when the git repo is a local path like "git_url: ../",
-            # but the user is working with a branch other than 'master' without
-            # explicitly providing git_rev.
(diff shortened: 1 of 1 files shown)


stuarteberg:
The CI checks are complaining, but I think they aren't related to the changes in this PR.  Let me know if there's something I can do make them pass.

msarahan:
Thanks @stuarteberg - merging.

PR #3139, opened 2018-09-12
use recipe_log.json instead of .txt; plumb info from commits there in…

msarahan:
…to rss

This produces RSS feeds derived from user-set recipe info.  These recipe_log.json files are expected to contain:

```
{
  "recipe_origin": "your_git_url",
  "commits": [
     {"hash": _hash, "timestamp": _time,
      "author": _author, "description": _desc},
  ]
```

For Anaconda, we get this info from our build system at the time that a recipe is submitted:  https://github.com/conda/conda-concourse-ci/pull/102/files#diff-37ad8af277b4d4d1ca1c8474e11ad811R367

The automated index process produces RSS feeds like:
```
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE rss PUBLIC "-//Netscape Communications//DTD RSS 0.91//EN" "http://www.rssboard.org/rss-0.91.dtd">
<rss version="0.91">
  <channel>
    <title>conda-bld updates</title>
    <link>https://anaconda.org</link>
    <description>Updates in the last two weeks</description>
    <language>en</language>
    <copyright>Copyright 2018, Anaconda, Inc.</copyright>
    <pubDate>Wed, 12 Sep 2018 19:23:37 +0000</pubDate>
    <item>
      <title>conda-build (3.14.4)</title>
      <pubDate>Wed, 12 Sep 2018 19:02:47 +0000</pubDate>
      <link>git@github.com:AnacondaRecipes/conda-build-feedstock.git</link>
      <description>
        - '6c7ccc5 (Wed, 12 Sep 2018 19:02:47 +0000): blah'
        - '2604285 (Tue, 11 Sep 2018 17:44:35 +0000): 3.14.4'
        - '7852b80 (Tue, 11 Sep 2018 17:02:01 +0000): 3.14.3; add tqdm dep'
        - '1db403a (Fri, 07 Sep 2018 22:09:46 +0000): 3.14.2'
        - '0669763 (Thu, 06 Sep 2018 18:07:31 +0000): 3.14.1'
        - '96bf75f (Wed, 05 Sep 2018 00:09:59 +0000): 3.14.0'
      </description>
    </item>
  </channel>
</rss>
```

### DIFF
--- a/conda_build/api.py
+++ b/conda_build/api.py
@@ -369,7 +369,8 @@ def create_metapackage(name, version, entry_points=(), build_string=None, build_
 
 
 def update_index(dir_paths, config=None, force=False, check_md5=False, remove=False, channel_name=None,
-                 subdir=None, threads=None, patch_generator=None, verbose=False, progress=False, **kwargs):
+                 subdir=None, threads=None, patch_generator=None, verbose=False, progress=False,
+                 hotfix_source_repo=None, **kwargs):
     from locale import getpreferredencoding
     import os
     from .conda_interface import PY3
@@ -385,4 +386,5 @@ def update_index(dir_paths, config=None, force=False, check_md5=False, remove=Fa
                                 threads=threads, verbose=verbose, progress=progress)
         else:
             update_index(path, check_md5=check_md5, channel_name=channel_name,
-                         patch_generator=patch_generator, threads=threads, verbose=verbose, progress=progress)
+                         patch_generator=patch_generator, threads=threads, verbose=verbose, progress=progress,
+                         hotfix_source_repo=hotfix_source_repo)
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -361,14 +361,14 @@ def copy_license(m):
 def copy_recipe_log(m):
     # the purpose of this file is to capture some change history metadata that may tell people
     #    why a given build was changed the way that it was
-    log_file = m.get_value('about/recipe_log_file') or "recipe_log.txt"
+    log_file = m.get_value('about/recipe_log_file') or "recipe_log.json"
     # look in recipe folder first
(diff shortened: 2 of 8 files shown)


jjhelmus:
Will this require commit messages to contain RSS ready information and be committed before a build is submitted to concourse?  Neither of these seem desirable.

msarahan:
Define "RSS ready" - if you mean we shouldn't have crappy commit messages showing up in this log, I'd argue that we shouldn't be building packages from repos with crappy commit messages.

Yes, changes have to be committed.  There's stuff in the c3i PR that barfs if stuff is not committed for this exact reason.  You can disable it, but we should leave it enabled for things that are destined to be uploaded.

msarahan:
PS: yes, this is a major workflow change.  I think the ability to obviate the git commit step is worth it, but that's really a group decision to make.

jjhelmus:
If `conda-forge` is our upstream then a lot of the commit message will be coming from these.  Nearly all of these are merge commits, hopefully a good number from the regro bot.  I do not think this is a good source to base the RSS feed from.  

This seem to be moving the burden of the RSS message to the feedstock repos which at least for me is more work rather than less.

msarahan:
Well...  I have to say, if you don't like this, you need to suggest another scheme for automating that commit step.  This is already going to help us capture much more information than we've been putting into the git commit log.  Yes, there will be some visual noise, but I don't think it's enough to kibosh this whole approach.

I see your point about merge commits, but showing that we have synced with conda-forge is itself valuable information.

msarahan:
FWIW, we can also filter commits somehow on the c3i side.  We have total control over what goes into the recipe_log.json.

msarahan:
Merging, but we'll continue to re-evaluate where the recipe_log.json file info should come from.

PR #3152, opened 2018-09-17
Print the license file was packaged

jakirkham:
Provide feedback to the user that the license file was found and packaged.

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -353,6 +353,7 @@ def copy_license(m):
             utils.copy_into(src_file,
                             join(m.config.info_dir, 'LICENSE.txt'), m.config.timeout,
                             locking=m.config.locking)
+            print("Packaged license file.")
         else:
             raise ValueError("License file given in about/license_file does not exist in "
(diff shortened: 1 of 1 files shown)


jakirkham:
Not sure how this would cause CI failures. 😕

msarahan:
Unrelated.  Don't worry about it.  Do you think it would be helpful to output more info about exactly which license file gets found?  It's a bit superfluous in my opinion, but might be helpful to resolve any potential ambiguity.

jakirkham:
Ok, thanks for letting me know.

Thought about it. Was kind of on the fence for a few reasons. First wasn't sure if the path here was absolute, which might make things a bit noisy. Second the name of the license file is in `meta.yaml` anyways. Third we are only packaging one license file instead of multiple at this time. That said, I don't have strong feelings about it. If you would prefer to have it printed, would be happy to add it.

msarahan:
*shrug* - I'll merge this as-is and we'll see if anyone has more to say down the road.

jakirkham:
Thanks Mike :)

Issue #3128, opened 2018-09-10
error when using --variants CLI flag


ifmar:
When I use the --variants flag to specify variants via command line argument, as described in the documentation of build variants, I get an unrecognized arguments error:

`unrecognized arguments: --variants {'c_compiler':['gcc'], 'target_platform':['linux-aarch64']}`

Even the example from the docu produces the same error. I use conda version: 4.5.4 and conda-build version: 3.10.7.

I want to use the CLI flag to separate multiple build jobs for cross compiling in our CI instead of using a single job with a build matrix. Is there another way to achieve this?

mingwandroid:
You should put your variants in conda_build_config.yaml and pass that with -m to the command line.

mingwandroid:
Whether did you see those flag? The documentation is at https://conda.io/docs/user-guide/tasks/build-packages/variants.html

ifmar:
Ok thanks for your answer. I will use the -m flag with multiple conda_build_config.yaml files.

Since the flag is also not mentioned in conda build command reference, it seems to me that the --variants may not implemented.

msarahan:
Hmmm... it does appear that it got dropped at some point.  Sorry about that.  That was not intentional.

Anyway, I think I have some ideas for how to make it nicer.  We should be able to allow passing arbitrary arguments by using argparse's parse_known_args function, and handling any unknown args ourselves.  That would make your command something like:

``conda build recipe --target_platform linux-aarch64``

We'll put that feature request into our backlog.

gabm:
`conda build recipe --target_platform linux-aarch64` 

Would be very useful I think... very clear syntax ...

msarahan:
The crappy thing about it is that any typos would be totally uncaught.  Any unknown key at all would just become a variant.

``conda build recipe --pyhton 3.6``

would not raise an error...

gabm:
because we would interpret anything unknown as variant.. thats true, not good at all...

What about something similar like this:

`conda build recipe --variant target_platform=linux-aarch64`

gabm:
I created #3168 with a possible solution.. I think it looks quite neat and is safe

msarahan:
Thanks to @gabm this is re-implemented.  Closing this issue.

PR #3168, opened 2018-09-25
Add --variant interface to the commandline

gabm:
This PR tries to solve #3128, it adds the following commandline interface:

`conda-build recipe --variant target_platform=linux-aarch64,c_compiler_version=7.2.0`

This allows for easier build scripts, because otherwise we'd need to store this information in files and call conda-build with a path to these files...

The name `--variant` was chosen, because its already present in the `Config` class. To have minimal change, I chose to re-use it. Besides that, I would prefer `--variants` (plural) because it reflects better what it is. What do you think @msarahan ?

Only limitation is, that i can only hold values of 1 hierarchy level, but I think thats sufficient for now..

### DIFF
--- a/conda_build/cli/main_build.py
+++ b/conda_build/cli/main_build.py
@@ -29,6 +29,14 @@ on_win = (sys.platform == 'win32')
 
 logging.basicConfig(level=logging.INFO)
 
+# see: https://stackoverflow.com/questions/29986185/python-argparse-dict-arg
+class StoreDictKeyPair(argparse.Action):
+    def __call__(self, parser, namespace, values, option_string=None):
+        my_dict = {}
+        for kv in values.split(","):
+            k, v = kv.split("=")
+            my_dict[k] = v
+        setattr(namespace, self.dest, my_dict)
 
 def parse_args(args):
     p = get_render_parser()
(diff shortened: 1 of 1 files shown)


gabm:
Maybe `--variant` (singular) is even better though.. It is _one_ specific variant that we address...

msarahan:
I'm not actually picky about that.  Would it be useful to allow list values?  Like:

```
conda-build recipe --variant target_platform=linux-aarch64,linux-64;c_compiler_version=7.2.0
```

(delineate values with ``,`` and flags with ``;``)

msarahan:
I guess the tricky thing is finding proper delineators.  commas and to a lesser extent, semicolons, show up in desired values sometimes.

gabm:
sounds great.. I'd be glad to implement it... Can you give me examples for the delimiters in use? I don't really know about that...

Common and readable  are imho:

- comma
- colon
- semicolon
- equality sign

gabm:
I think in case we want lists, then we should go for abbreviated yaml (similar to what already had been there):

`--variant "{target_platform: [linux-aarch64,linux-64], c_compiler_version: 7.2.0}"`

https://docs.ansible.com/ansible/latest/reference_appendices/YAMLSyntax.html

This is portable, parseable and properly defined..

msarahan:
Yeah, that's probably best.  It is more annoying to have to quote things, but the safety is probably worth it to not have corner cases.

It was JSON before, but that's practically similar for these purposes.  I'm OK with either here.

If you make that change, I'll be happy to merge this.

gabm:
I implemented it as YAML which I think is more consistent which the other configuration files being YAML as well..

PR #3196, opened 2018-10-07
do not use name to split dist-info directory

### LINKED ISSUE #3195: dist-info check fails if package name is not the same as module name
The dist-info check in `post.py` uses the package name to partition the version string.  This results in an incorrect version being parsed if the package name is not the same as the name of the python module which in turn fails the build unnecessarily.  

The `dask-core` recipe from conda-forge exhibits this behavior.

jjhelmus:
The package name is not necessarily the same as the name of the Python library
and therefore should not be used when determining the version of the .dist-info
folder.

closes #3195

### DIFF
--- a/conda_build/post.py
+++ b/conda_build/post.py
@@ -242,7 +242,7 @@ def check_dist_info_version(name, version, files):
         if '.dist-info' in f:
             f_lower = f.lower()
             f_lower, _, _ = f_lower.rpartition('.dist-info')
-            _, _, f_lower = f_lower.rpartition(name + '-')
+            _, _, f_lower = f_lower.rpartition('-')
             if version != f_lower:
                 print("ERROR: dist-info version incorrect (is {}, should be {})".format(f_lower, version))
                 sys.exit(1)


jjhelmus:
The logic here seems a bit brittle but this fixes the immediate issue.  Perhaps a regex would be a better approach.

goanpeca:
>The logic here seems a bit brittle but this fixes the immediate issue. Perhaps a regex would be a better approach.

Yes a bit, would reading the Metadata help make this less brittle?

marscher:
The current solution makes it impossible to use Python pep440 versioning schemes. I guess @goanpeca solution sounds most reasonable, another approach would involve specifying a zoo of different versionings.

msarahan:
Is this still necessary after https://github.com/conda/conda-build/pull/3206/commits/7d682bd3ceaa99184197c953e0000a62d9330297 ?

msarahan:
I tested with bleach and dask-core.  Ray's commit is an adequate fix.  There may be a better way, and I'm open to it, but I'm closing this for now.

Issue #3189, opened 2018-10-05
Building RPMs


teake:
The documentation [currently](https://github.com/conda/conda-docs/blame/f0c1c41aa01c2dfa1f85c318ff53f45e0e50984b/docs/source/user-guide/tasks/build-packages/define-metadata.rst#L1198) [states](https://conda.io/docs/user-guide/tasks/build-packages/define-metadata.html#output-type) it is possible to build RPMs from recipes.

However, when I add:

```yaml
outputs:
  - type: rpm
```

to my `meta.yaml` I get a `KeyError: 'rpm'` coming from `conda_build.build.build()` ([here](https://github.com/conda/conda-build/blob/b0944ab229c9b80e84b4d2f2b684144abcec0187/conda_build/build.py#L1625) to be precise). Sure enough, the [`bundlers` dict](https://github.com/conda/conda-build/blob/b0944ab229c9b80e84b4d2f2b684144abcec0187/conda_build/build.py#L1117) only has conda bundler and a wheel bundler.

Am I not specifying the RPM output correctly, or is the doc incorrect and was this feature removed at some point?

msarahan:
doc is inaccurate.  That was wishful thinking.  We still want to implement it at some point, but it has never been added.

sscherfke:
I looked into this issue a bit more.

In theory, it should be no problem to build an RPM from a conda recipe since the recipe provides all the meta data needed for the RPM SPEC.

The problem is rather the filesystem layout:

- Conda packages put stuff into `$PREFIX/bin`, `$PREFIX/etc`. ..
- RPMs install stuff into `/usr/local/bin`, `/etc`, ...

So just doing `PREFIX=/usr/local` for RPM builds does not work for some files.

There might also be subtle differences for the requirements or build instructions.

Maybe these differences could be covered by putting a `build_rpm.sh` next to the `build.sh`.  The downside of this would be that a package had to be compiled twice.

mingwandroid:
If you compile it twice it's no longer going to be the same thing though.

msarahan:
The bigger issue here is the question of runtime requirements. Do they come from conda? If they come from the system package manager, there will almost certainly be incompatibilities. RPM as a file format with a fixed install path is a really insignificant detail here in comparison.

xref: https://github.com/pelson/conda-rpms

PR #3199, opened 2018-10-08
Remove mention of RPMs and .deb files from docs

teake:
These features are not implemented yet. See #3189.


teake:
The downstream docs in conda/conda-docs should also be updated. Will this happen automatically?

PR #3209, opened 2018-10-11
Fix _tar_xf_file definition to match usage

geremih:
The usage of _tar_xf_file doesn't match the definition. The definition assumes that it will be passed a list of entries, but the usage passes a single file path. This works most of time, but if the tar contains a file is the substring of the requested file, you get the wrong result (because using `in`)

### DIFF
--- a/conda_build/index.py
+++ b/conda_build/index.py
@@ -771,23 +771,20 @@ def _tar_xf(tarball, dir_path):
         libarchive.extract_file(tarball, flags)
 
 
-def _tar_xf_file(tarball, entries):
+def _tar_xf_file(tarball, file_path):
     if not os.path.isabs(tarball):
         tarball = os.path.join(os.getcwd(), tarball)
     result = None
-    n_found = 0
     with libarchive.file_reader(tarball) as archive:
         for entry in archive:
-            if entry.name in entries:
-                n_found += 1
+            if entry.name == file_path:
                 for block in entry.get_blocks():
                     if result is None:
                         result = bytes(block)
                     else:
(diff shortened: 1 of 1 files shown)


mingwandroid:
Thanks!

mingwandroid:
.. but I do want to support extraction of multiple files at once here eventually and there's other cases where I am passing a list. Can you change it to pass a list here at the calling site instead?

mingwandroid:
Alternatively use ensure_list.

mingwandroid:
Any update on this? I'll push a PR if I hear nothing back.

geremih:
hey @mingwandroid. I couldn't find usages of passing in a list and the expected return value (probably don't want to concatenate everything). 
Agree that it might be more convenient to open another PR with your intended usage.

mingwandroid:
Yeah, I have this function in various slightly different forms spread over a few of my forks of various things (conda-build, conda, conda-verify, conda-pack) at present. I'll clean it up sometime.

For this issue I have: https://github.com/conda/conda-build/pull/3210

geremih:
Sweet, closing this then. Though still unsure what's the usecase for concatenating all files though.

mingwandroid:
I wont be concatenating eventually. I'll return entries in a tuple for each file. The reason to do multiple at once is to avoid overhead reopening archives.

msarahan:
I incorporated @mingwandroid's ensure_list modification in the 3.16.1 tag.  Thanks @geremih for your contribution here, even if it didn't end up going in.

PR #3210, opened 2018-10-12
Use binsort to sort file list if found otherwise sort codefiles follo…

mingwandroid:
…wed by dirname

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -19,6 +19,7 @@ import string
 import subprocess
 import sys
 import time
+from tempfile import NamedTemporaryFile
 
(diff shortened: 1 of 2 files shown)


mingwandroid:
I am seeing up to 5% compression improvements when using this (but generally around 1 or 2%). Not earth shattering but worth having.

msarahan:
Is binsort available as a package on all platforms?  We'd need to add it as a new dependency for conda-build, right?

mingwandroid:
I've made it entirely optional for now. It's not simple to build on Windows though. It's a small codebase but fairly unixy.

PR #3212, opened 2018-10-12
Remove noarch binary file check

### LINKED ISSUE #3208: allow overriding the .so check for noarch: python?
`compas` is a pure-python package that could be made `noarch: python` except that it [ships some `.so` and `.dll` files](https://github.com/compas-dev/compas/tree/master/src/compas/numerical/fd/__fd_cpp) (that it doesn't compile, so they're the same across platforms). That's triggering [this check](https://github.com/conda/conda-build/blob/19f5d7847ee8a4c5b979680595103fa6cc21e4b1/conda_build/noarch_python.py#L60-L66) and killing the build ([PR](https://github.com/conda-forge/compas-feedstock/pull/6#issuecomment-429024394), [failed build](https://circleci.com/gh/conda-forge/compas-feedstock/37?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)).

The check definitely makes sense in general. But maybe there should be a way to override it for cases like this?

mandeep:
fixes #3208

### DIFF
--- a/conda_build/noarch_python.py
+++ b/conda_build/noarch_python.py
@@ -57,14 +57,6 @@ def handle_file(f, d, prefix):
     if f.endswith(('.egg-info', '.pyc', '.pyo')):
         os.unlink(path)
 
-    # The presence of .so indicated this is not a noarch package
-    elif f.endswith(('.so', '.dll', '.pyd', '.exe', '.dylib')):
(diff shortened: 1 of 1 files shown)


msarahan:
I haven't seen this error before: https://travis-ci.org/conda/conda-build/jobs/441310292#L2167

but it really doesn't seem related to your PR.  I'll merge this, and I'll have to sort out that error elsewhere.

PR #3217, opened 2018-10-15
Fix thread count when indexing

mingwandroid:
We use os.cpu_count() directly here instead of dividing it by two.

Mostly though this prevents max_threads=None getting passed to the
low level concurrency stuff. When that happens Python assumes that
the job is not CPU bound and uses cpu_count() * 5.


msarahan:
Good call.  Merging.

PR #3221, opened 2018-10-17
Swap pyldd for liefldd nearly everywhere

mingwandroid:
This allows us to figure out Windows dependencies. I still have some work
to do around whitelists for Windows system libraries and making sure that pyldd
is used as a fallback (we won't be able to figure out Windows stuff though
unless I do more work on pyldd, still, I think saying that you only get
this info if you use Python 3 in the base env is acceptable).


msarahan:
this has been incorporated into #3237; closing

Issue #3223, opened 2018-10-19
try_acquire_locks ignores already acquired locks


alexandersturm:
### Actual Behavior

Calling `try_acquire_locks` on an already acquired lock acts as if it successfully acquired the lock. This effectively breaks the entire locking mechanism, leading to all kinds of errors when running multiple builds in parallel.

### Expected Behavior

`try_acquire_locks` should raise if it can't acquire a lock. Parallel builds should work without issues.

If I'm reading the code correctly, [this break statement](https://github.com/conda/conda-build/blob/d392dc7ec39e52a5bd9423401c76805780dd7d82/conda_build/utils.py#L384) should be replaced with a `raise`. 

### Steps to Reproduce

```python
import filelock
import conda_build.utils
path = 'some_file'
with conda_build.utils.try_acquire_locks([filelock.FileLock(path)], 5):
    with conda_build.utils.try_acquire_locks([filelock.FileLock(path)], 5):
        print('Should not get here')
```

This snippet should raise an exception after 5 seconds. Instead, it prints "Should not get here" after 5 seconds.

##### `Output of conda info`
```
     active environment : None
       user config file : C:\Users\asturm\.condarc
 populated config files : C:\Users\asturm\.condarc
          conda version : 4.5.11
    conda-build version : 3.15.1
         python version : 3.6.6.final.0
       base environment : D:\conda (writable)
           channel URLs : https://repo.continuum.io/pkgs/main/win-64
                          https://repo.continuum.io/pkgs/main/noarch
                          https://repo.continuum.io/pkgs/free/win-64
                          https://repo.continuum.io/pkgs/free/noarch
                          https://repo.continuum.io/pkgs/r/win-64
                          https://repo.continuum.io/pkgs/r/noarch
                          https://repo.continuum.io/pkgs/pro/win-64
                          https://repo.continuum.io/pkgs/pro/noarch
                          https://repo.continuum.io/pkgs/msys2/win-64
                          https://repo.continuum.io/pkgs/msys2/noarch
          package cache : D:\conda\pkgs
                          C:\Users\asturm\AppData\Local\conda\conda\pkgs
       envs directories : D:\conda\envs
                          C:\Users\asturm\AppData\Local\conda\conda\envs
                          C:\Users\asturm\.conda\envs
               platform : win-64
             user-agent : conda/4.5.11 requests/2.19.1 CPython/3.6.6 Windows/7 Windows/6.1.7601
          administrator : False
             netrc file : None
           offline mode : False
```

PR #3224, opened 2018-10-19
Fix behavior of try_acquire_locks during lock contention

alexandersturm:
See #3223. The first commit (https://github.com/conda/conda-build/commit/dc5c81763259f69215ded1e35b95c323be00b7cf) is the minimal fix for that issue. The following commit (https://github.com/conda/conda-build/commit/d83e38a6af0c7c6b59ccc1d0568e508571d7d225) improves the behavior when trying to acquire multiple locks at once.


msarahan:
Thanks!  Looks like a great PR.  I'll wait on CI.  We have some outstanding failures, but those won't prevent a merge here.

mingwandroid:
This looks (on eyeballing the code) like a *fantastic* contribution to `conda-build`. I am looking forward to someone adding `--jobs/-j N` to conda-build, what do you think @msarahan?

alexandersturm:
Thanks! Looks like the CI is still running into some deadlocks though, so I'll have to spend some more time on getting everything fixed up :)

mingwandroid:
Fabuluous, please do!

alexandersturm:
I think this is good to go now (the fix became a bit more involved than I had initially hoped :). The remaining CI failures seem to be unrelated to my changes:
- `tests/test_api_build.py::test_legacy_noarch_python` also fails in master
- `tests/test_api_render.py::test_resolved_packages_recipe` passes locally

Let me know what you think.

conda_build/utils.py
```diff
@@ -101,6 +100,10 @@ def glob(pathname, recursive=True):
     from scandir import walk
 
 
+class BuildLockError(Exception):
```
msarahan: This probably belongs in the exceptions.py file, but I'm not too picky.

alexandersturm: Sure, let me move it.

PR #3225, opened 2018-10-19
Added emscripten selector

scopatz:
This is needed for building up the webassembly stack.

### DIFF
--- a/conda_build/metadata.py
+++ b/conda_build/metadata.py
@@ -121,6 +121,8 @@ def ns_cfg(config):
     for machine in non_x86_linux_machines:
         d[machine] = bool(plat == 'linux-%s' % machine)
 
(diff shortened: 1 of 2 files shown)


msarahan:
What if the EMSDK is part of a conda package?  Is there a better way to define this, perhaps like how the python version finding works?

mdboom:
(@scopatz and I were sprinting about this at PyData NYC)

I think the plan was that emsdk would be a conda package, but that we would still need a way to conditionally select on it from other packages.  I don't really know how all this works -- if there's a way for the emsdk package to add its own selector, that would work, too, of course.

msarahan:
All of the variables in conda_build_config.yaml are available as selectors.  I'd rather just leave it there and have it be more local configuration that way.  Is that OK?

PR #3232, opened 2018-10-23
Add r-impl to R packages

dsludwig:
@mingwandroid 

I needed this patch to conda-build for my R experiments to work (for the R environment variable to exist).

### DIFF
--- a/conda_build/environ.py
+++ b/conda_build/environ.py
@@ -34,6 +34,7 @@ from conda_build.variants import get_default_variant
 # these are things that we provide env vars for more explicitly.  This list disables the
 #    pass-through of variant values to env vars for these keys.
 LANGUAGES = ('PERL', 'LUA', 'R', "NUMPY", 'PYTHON')
+R_PACKAGES = ('r-base', 'mro-base', 'r-impl')
 
 
 def get_perl_ver(config):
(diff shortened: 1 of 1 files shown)


mingwandroid:
Can you explain the need for this @dsludwig?

mingwandroid:
If we need an R env var set why don't we set it in `r-base`'s activate script?

dsludwig:
I'm not sure why it's not there instead - I found I needed this because the cran skeletons depend on that environment variable.

I guess `r-impl` could set this in an activate script?

@mcg1969 might be able to provide more context.

mcg1969:
`r-impl` is a new metapackage that I proposed, for package to depend on when they can work with either Anaconda R or MRO. For each version of R, there would be two builds of `r-impl`: one that depends on each version of R. These platform-independent packages will depend on `r-impl` instead

mingwandroid:
What problem is this solving? You can select your R already with mro-base vs r-base.

mcg1969:
What if you want to build a package that depends on _either_ r-base or mro-base?

mingwandroid:
Ah are you talking about generic ones only here?

mcg1969:
For now, yes. Only when things work.

mingwandroid:
OK, LGTM.

Issue #1751, opened 2017-02-17
Add an interactive shell debug option for conda build


yesimon:
Recently I was struggling to debug an erroneous conda build and what I ended up doing was dropping into an interactive shell right before executing `build.sh`. This was really helpful since it allows simulating being in the same environment (environment variables, `$PATH` etc) as the actual build process. This feature exists in the Homebrew package manager as well. 

My suggestion is to add a `--shell` option to drop into an interactive shell right before the building process.

msarahan:
Good idea.  This is a small extension on top of an existing request that I think doesn't have an issue here: recording environment variables in build.sh as is done in bld.bat.

I don't think it'll take the form of dropping into a shell before exiting build.sh - that's just much more complicated.  Instead, what I'd recommend is:

1a) failed builds are already kept.  These live in (miniconda install root)/conda-bld/recipe_folder_name_(timestamp).
1b) for builds that exit successfully, but you still want to keep, use the ``--dirty`` argument.

2) cd into the 'work' folder in that timestamped folder
3) ``source activate ../_b(hit tab to autocomplete this long path)``

basically, a ``--shell`` option should script those actions - but should do so after the build has failed (or succeeded) and exited, and it should pick up where that build left off.  This will just just be simpler from the standpoint of the whole build execution flow.  I think the capability will be equivalent.

yesimon:
In my case the build was succeeding but the binary was incorrect (due to some improper compiler/linker flags). I hope that there's an option to drop in before the build process because sometimes the `build.sh` will change things irrevocably, like deleting/moving source files or changing the environment.

msarahan:
Sorry, I'm just not willing to spend the time to figure out nesting interactive shells in subprocesses.  I'll try to make this kind of feature still available, though, probably with the ``conda develop`` command.  It presently doesn't do very much helpful.  I would like it to create the build env specified by meta.yaml.  It could also set environment variables with activate.d scripts.

If you want to tackle figuring out nesting shells in subprocesses and put up a PR, I'd be glad to check it out.

yesimon:
It seemed to work okay with `subprocess.run("/bin/bash", ...' but I'll definitely investigate a bit deeper to make sure that everything is working properly and getting cleaned up.

epruesse:
[...]

PR #3237, opened 2018-10-28
WIP: add debug command for creating envs

msarahan:
The purpose of this feature is to aid in creating environments and activation scripts so that users can quickly create an environment just like conda-build creates for build/hot or test.  This feature creates the envs, splits the env activation stuff into a separate script that can be sourced without running a build, and outputs a line that can be copy-pasted to get into an activated shell that works the same as a build inside conda-build.

examples:

```
# create build/host envs and activation script
conda-debug my-test-recipe
# create test env and activation script
conda-debug /home/msarahan/miniconda/conda-bld/linux-64/somepkg-1.0-0.tar.bz2
```

Note that you have to have the dash between conda and debug.  That's because conda has those executables hard-coded, and ``conda debug`` may not work until we add an entry for it in conda.

fixes https://github.com/conda/conda-build/issues/1751


soapy1:
I haven't tested out everything yet but here are my initial notes:

- Initially had trouble getting it working - was my fault, didn’t have lib archive installed

- Tried debugging a package I had locally
    - conda-debug ~/miniconda3/pkgs/alabaster-0.7.11-py35_0.tar.bz2
- Did the step where you go into the environment
    - cd /Users/scastellarin/miniconda3/conda-bld/debug_1542233048885/test_tmp && source /Users/scastellarin/miniconda3/conda-bld/debug_1542233048885/test_tmp/conda_test_env_vars.sh
    - Ran the test with `./run_test.sh` and this killed my shell. Was somewhat unexpected. I guess the intended way to run it is `bash run_test.sh`, that works fine

- Tried debugging a recipe I had locally
    - conda-debug gdata-feedstock
- The first time I ran it it was good
- Ran it with `—test` flag and it killed my shell
- Ran it again with `—test` flag and it gave me an error `Debugging for test mode is only supported for package files that already exist. Please build your package first, then use it to create the debugging environment.`
- Tried it with `—path` that did and did not exist

Questions:
- the provided debug directory has 3 different testing files that seem to do the same thing. Eg. `conda_test_runner.sh`, `run_test.py`, `run_test.sh`. What is the purpose of having this duplication?
- Is there a way to purge the environment when the user is done?
- Do you think it would be beneficial if using condo-debug on a recipe also gave the user test runners

msarahan:
> Is there a way to purge the environment when the user is done?

Verified that ``conda build purge`` does clean these up

> Do you think it would be beneficial if using condo-debug on a recipe also gave the user test runners

Added text like:

```
################################################################################
Test environment created for debugging.  To enter a debugging environment:

cd /Users/msarahan/miniconda3/conda-bld/conda-build_1542302975704/work && source /Users/msarahan/miniconda3/conda-bld/conda-build_1542302975704/work/build_env_setup.sh

To run your tests, you might want to start with running the conda_test_runner.sh file.
################################################################################
```

after the path activation command printout.

soapy1:
Did a bit more testing on this:

- Installed locally creates package 
    - conda-debug ~/miniconda3/pkgs/alabaster-0.7.11-py35_0.tar.bz2

- Tried using -a flag
    - conda-debug ~/miniconda3/pkgs/alabaster-0.7.11-py35_0.tar.bz2 -a
- Nice!

- Tried setting path to a folder that already exists with stuff in it
```
$ ls test
Vagrantfile

$ conda-debug ~/miniconda3/pkgs/alabaster-0.7.11-py35_0.tar.bz2 -a -p test
```
- Then I did it again (to the same folder)
- Looks like it will overwrite it
- Tried with another package (a folder, not .tar.gz)
    - The output of this was a bit confusing
- Tried with another package (.tar.gz, not a folder)
    - Added the artifact

Steps to reproduce bug:
1. Run conda debug give it a path. Eg `conda-debug ~/miniconda3/pkgs/filelock-3.0.4-py36_0.tar.bz2 -p test`
2. Run conda debug for another package and give it the same path eg. `conda-debug ~/miniconda3/pkgs/alabaster-0.7.11-py35_0.tar.bz2 -p test`

- Gave it a directory that exists but is fulll of not packages
    - Returns on OSError `OSError: No meta.yaml or conda.yaml files found in /Users/scastellarin/Desktop/test`
- Gave it a directory filled with packages + recipes `OSError: More than one meta.yaml files found in /Users/scastellarin/miniconda3/pkgs`
 - did some more things kind of like this and got the same error message

minrk:
This is great! I'm not sure if it's the right thing to do, but I'd probably usually want to debug with a not-long host prefix if I could (parent/_host instead of parent/_h_placehold...). I know that would technically not test the issues that the long path is needed for, but the long paths genuinely make things harder to debug.

msarahan:
Thanks @minrk, that's a good idea, and should be easy to implement.

I had a couple of other ideas this morning:

1. for build/host debugging, after provisioning the source, try to git init/add files in the work dir.  Multiple sources will make this complicated, but the payoff would be huge in streamlining development of patches.
2. In addition to 1 here, add in commands or flags to conda-debug that streamline translation of changes in the debug work dir into patches that are entered into the original meta.yaml file.  This would be done with ``git format-patch`` and some parsing of the existing meta.yaml to understand what the next patch number should be.  It assumes that the conda-debug user would use "git commit" in the work dir for each thing that should be a patch.

minrk:
> In addition to 1 here, add in commands or flags to conda-debug that streamline translation of changes in the debug work dir into patches that are entered into the original meta.yaml file

I would **love** this. I regularly do `git init; git add .; git commit -m 'starting point'` so I can use format-patch, since I can't seem to ever remember the incantation to manually create patches. Making this easier would help me a ton.

btw, I've already used this to work out some issues in the petsc and openmpi repos, and it helped! A couple small weirdnesses I ran into:

- I did `source build_env_setup.sh` followed by `bash conda_build.sh`. The build failed because conda_build.sh sources the same env file, and apparently some of the flags being added twice caused builds to fail (I didn't bother debugging this too deeply)
- I messed myself up by debugging two builds from the same terminal session, so lots of flags inherited from the first debugging session were inherited in the second (I managed to accidentally compile petsc against the openmpi build I was debugging, which petsc caught in its configuration checks). Not sure what to do about that. Does conda-build normally *reset* things like CFLAGS, or does it always append? If it resets, then these env scripts should perhaps reset some envs explicitly as well. Alternately, a hint that starting a dedicated shell per debugging session is a good idea might be appropriate.

embray:
Thank you for working on this.  I'll try it out in a bit.  +1 to Min's comments that making builds easier for debugging (e.g. removing the long host prefix) would be best, as long as it is understood by the user that this is for creating debug builds of packages only, and not builds that should be pushed to users.

msarahan:
Going ahead with this.  There's one failing test:

tests/test_subpackages.py::test_subpackage_independent_hash

That test passes locally.  I'm not concerned about it now, though it would be nice to have green icons again...

ihnorton:
Thank you for implementing this! 👍  

I was a little bit surprised that I still needed to run `conda activate /path/to/_h_env/` to get `CC`, `CPPFLAGS`, etc. from the activation script for a package that is specified in my build requirements (`gxx_linux-64`). Would it make sense to add that activation as well, or at least a note in the documentation?

PR #3239, opened 2018-10-31
Change default build timeout to 900

Bezier89:
With a timeout of 90 seconds, we consistently get failures when performing multiple builds in parallel.

@msarahan I see that this setting is defined in config.py, but for the life of me I cannot figure out how to change this setting at runtime. There doesn't appear to be a corresponding condarc entry, nor a flag I can pass to conda build. Please let me know if I'm missing something. In the meantime, increasing the default timeout seems like a reasonable approach if this value cannot be easily changed at runtime.

### DIFF
--- a/conda_build/config.py
+++ b/conda_build/config.py
@@ -72,7 +72,7 @@ def _get_default_settings():
             Setting('verbose', True),
 
             Setting('debug', False),
-            Setting('timeout', 90),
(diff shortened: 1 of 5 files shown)


msarahan:
Seems fine to me.  It would be nice to have this be more of a setting, but I'll leave that for another day.

PR #3240, opened 2018-10-31
bugfix: append activation to the wheel script

tomashek:
Wheel creation as an output was not appending the environment activation to the wheel_output script. It looks like the metadata reference for the activation config was wrong. This PR fixed it for us internally. It now matches the reference in the bundle_conda method, line 867.


msarahan:
Thanks!  Waiting on CI, but looks fine to me.  How urgently do you need a release with this?

tomashek:
I can apply the fix internally until it is in a release, so no rush on your end.

PR #3091, opened 2018-08-21
Indexing rewrite

msarahan:
Most of this is adapted from https://github.com/ContinuumIO/conda-channel-clone/ (private project, sorry for curious outsiders)

Issue #309, opened 2015-01-22
python goes missing with conda-build 1.10.0


bgreen-litl:
Please see:

https://github.com/conda/conda-build/issues/308

This is totally reproducible for me with conda-build 1.10.0. Running "conda install -f -n _build python" only fixes the problem after first downgrading to conda-build 1.9.1

Let me know if you need more information, but it's like clockwork here and happens both on osx-64 and linux-64. (I happen to be building the pillow package in case that's relevant.)

PR #3249, opened 2018-11-07
about.json additions to support new channeldata.json index

mikecormier:
Add "tags", "identifiers", and "keywords" to about.json to support the new index for channeldata.json(https://github.com/conda/conda-build/pull/3091) 

@msarahan, this is to support the https://github.com/conda/conda/issues/7583 issue, and the recent index rewrite https://github.com/conda/conda-build/pull/309.

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -577,6 +577,9 @@ def write_about_json(m):
         d['root_pkgs'] = env.package_specs()
         # Include the extra section of the metadata in the about.json
         d['extra'] = m.get_section('extra')
+        d['tags'] = m.get_section('tags')
+        d['keywords'] = m.get_section('keywords')
+        d['identifiers'] = m.get_section('identifiers')
         json.dump(d, fo, indent=2, sort_keys=True)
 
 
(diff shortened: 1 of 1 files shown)


mikecormier:
@msarahan,

I don't think the test for tags and identifiers is correct. You added these test (https://github.com/conda/conda-build/commit/d50497167043add515af5cffe374c259a5ad8332#diff-44dd42c7e6fb5994d9dc00951530697c), however, I have tested my commits on multiple different packages and can validate that the additions I have created add the "tags", "identifiers", and "keywords"  into the about.json file. 

Any thoughts?

msarahan:
It looks like you are expecting ``tags``, ``identifiers``, and ``keywords`` to be top-level keys in meta.yaml.  I don't think that's a good idea.  My code has them as subkeys under the ``about`` section.  If you change your recipes so that your desired keys are under the about section, does it work?

mikecormier:
@msarahan.

It does work when those keys are under the about section. Thanks for the help. I will close this PR.

PR #3258, opened 2018-11-13
rewrite long prefix paths as $PREFIX, etc. in stdout

minrk:
makes build outputs more readable, easier to follow

I’m not 100% sure if this is a good idea, but it greatly improves the experience of debugging conda-build output for me, at least. By using environment variables, the results are still copy/pasteable for replication and testing.

Still need to confirm that the use of pipes/rewrites is correct on Windows


tests/test_build.py
```diff
@@ -179,3 +179,14 @@ def test_create_info_files_json_no_inodes(testing_workdir, testing_metadata):
     with open(files_json_path, "r") as files_json:
         output = json.load(files_json)
         assert output == expected_output
+
+
+def test_rewrite_output(testing_workdir, testing_config, capsys):
+    m = api.render(os.path.join(metadata_dir, "rewrite_env"), config=testing_config)[0][0]
```
msarahan: This could probably be just:

```
api.build(os.path.join(metadata_dir, "rewrite_env"), config=testing_config)
```

tests/test_build.py
```diff
@@ -179,3 +179,14 @@ def test_create_info_files_json_no_inodes(testing_workdir, testing_metadata):
...
+    build.build(m, stats=None)
+    captured = capsys.readouterr()
+    stdout = captured.out
+    assert "PREFIX=$PREFIX" in stdout
```
msarahan: I think you probably need an ``if`` block here, with this and the win syntax:

assert "PREFIX=%PREFIX%" in stdout

mingwandroid:
Great idea. I wanted this for a while.

mingwandroid:
It also allows running stuff pasted from logs (provided you run your activate) which is super handy.

msarahan:
@minrk let us know if you want us to finish this one for you.  I'm not sure how much access to a Windows machine you have.

minrk:
Help finishing up would be awesome. I've been using this locally on and off for a while and it's been pretty nice, especially with builds that use lots of prefix-referencing compiler flags. In particular, I'm less confident about using pipes on Windows in general because I know that full pipes causing subprocess hangs can be an issue, which could be mysterious and incredibly frustrating.

msarahan:
Thanks.  I will work on it today.  In the meantime, any feedback you can provide on #3237 would be very helpful.

tests/test_build.py
```diff
@@ -182,7 +182,7 @@ def test_create_info_files_json_no_inodes(testing_workdir, testing_metadata):
 
 
 def test_rewrite_output(testing_workdir, testing_config, capsys):
-    api.build(os.path.join(metadata_dir, "rewrite_env"), config=testing_config)
+    api.build(os.path.join(metadata_dir, "_rewrite_env"), config=testing_config)
```
msarahan: I moved this to _rewrite_env because it is a test that we want to manually run.  Things starting with _ are not run automatically with https://github.com/conda/conda-build/blob/master/tests/test_api_build.py#L82-L94

msarahan:
Going in.  Thanks @minrk!

PR #3264, opened 2018-11-20
avoid discarding build string during pin_run_as_build and ensure_valid_spec

minrk:
get_pin_from_build was discarding build string information when a given dep is in pin_run_as_build

This is coming up in attempts to ensure that packages built with a particular variant of a dependency have a runtime dependency on the same variant. It generally works, but when the dependency is in `pin_run_as_build`, the build string in requirements.run is discarded when the pinning is applied.

For example:

```yaml
# conda_build_config
pin_run_as_build:
  mpi:
    max_pin: x.x

# meta
requirements:
  host:
    - mpi * mpich
  run:
    - mpi * mpich
```

The result is runtime dependency on `mpi >=1.0,<2.0` without any mention of the `mpich` build string.

### DIFF
--- a/conda_build/render.py
+++ b/conda_build/render.py
@@ -428,7 +428,12 @@ def _simplify_to_exact_constraints(metadata):
 
         deps_list = []
         for name, values in deps_dict.items():
-            exact_pins = [dep for dep in values if len(dep) > 1]
+            exact_pins = []
+            for dep in values:
+                if len(dep) > 1:
+                    version, build = dep[:2]
+                    if not (any(c in version for c in ('>', '<', '*')) or '*' in build):
+                        exact_pins.append(dep)
             if len(values) == 1 and not any(values):
                 deps_list.append(name)
             elif exact_pins:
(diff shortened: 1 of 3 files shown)


minrk:
Additionally had to fix exact_pinning check, which assumed that any pinning that included a fragment of a build string was an exact pinning, which isn't true e.g. for `pkg * string`.

minrk:
Found another issue. `util.ensure_valid_spec` is also stripping build strings because the check:

```python
not hasattr(spec, 'build')
```

is always False because when a spec has a build string, it is not exposed as a build *attribute*. The build string is accessible as `spec.get('build')`.

The patches here appear to be required to use build string pinning in host requirements at all. This is currently blocking any packages from using the hdf5 mpi build, e.g. https://github.com/conda-forge/h5py-feedstock/pull/45 and https://github.com/conda-forge/fenics-feedstock/pull/73

minrk:
FWIW, with this branch, I have successfully built fenics with parallel hdf5 (https://github.com/conda-forge/fenics-feedstock/pull/73).

SylvainCorlay:
Thanks very much for this!

The same issue is blocking us from using variants of clang as host dependencies for cling.

msarahan:
Thanks @minrk

PR #3265, opened 2018-11-20
Make the --output-folder switch configurable via .condarc

teake:
This is convenient for e.g. passing build artifacts around during
CI. The root-dir option was already configurable via .condarc, but
since the root build dir also contains caches and non-purged builds,
it's cleaner to use to output-folder.

In a similar vain as #2074.


nehaljwani:
Please add these nice additions to docs as well :)
Please!

teake:
@nehaljwani Done! (Or at least, the PR).

PR #3266, opened 2018-11-20
Make the --long-test-prefix switch configurable via .condarc

teake:
Forgot to add this to #3265, sorry.

The reason I'd like this to be configurable via `.condarc` is simply because I need the switch and would like to do everything via `.condarc`.

Also, I think the original action of `--long-test-prefix` was wrong and changed it to `store_true`. This might cause some headaches though for people who used the switch as a feature bug.

### DIFF
--- a/conda_build/cli/main_build.py
+++ b/conda_build/cli/main_build.py
@@ -281,15 +281,17 @@ different sets of packages."""
         default=cc_conda_build.get('error_overlinking', 'false').lower() == 'true',
     )
     p.add_argument(
-        "--long-test-prefix", default=True, action="store_false",
+        "--long-test-prefix", action="store_true",
         help=("Use a long prefix for the test prefix, as well as the build prefix.  Affects only "
               "Linux and Mac.  Prefix length matches the --prefix-length flag.  This is on by "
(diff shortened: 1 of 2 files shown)


msarahan:
Looks right to me.

teake:
I'm not quite sure you meant the PR looks ok or if the existing behaviour does -- I'm assuming the latter.

I do think the existing `--long-test-prefix` action is wrong. Currently, if you give the option `--long-test-prefix` conda-build creates a short prefix for the test environment, which is exactly what it shouldn't do.
This is because the `store_false` action of argparse stores False if the option is given (which is not ok here), but defaults to True if the option is omitted (which is ok here). I've added a regression test to make sure the new behaviour is as intended. It failed on master.

Or it could be that I completely misunderstood the `--long-test-prefix` switch, in which case never mind.

msarahan:
No, I meant that your PR here looks correct.  The logic was wrong.  I expect and hope that not many people are using this flag.  It creates software that's very fragile in terms of where it can be installed - long path destinations will be prone to breaking with this flag.

Still, it's something that people wanted, so it should work correctly.  Thanks for your PR.

PR #3270, opened 2018-11-24
conda-index: don't assume presence of key: 'depends'

nehaljwani:
Packages built with very old versions of conda-build might not have the
'depends' keyword with an empty list for packages with no dependencies
in the index.json inside the tarball.


msarahan:
Thanks @nehaljwani

PR #3278, opened 2018-11-29
enable py37 selector for meta.yaml

grlee77:
The intention of this PR is to use `py37` in `meta.yaml` for cases like:

```
skip: True  # [py37]
```

I just tried to do a quick grep of the codebase to see where the existing ones were defined, but may have missed something.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -82,10 +82,10 @@ test_script:
   - conda create -n blarg -yq --download-only python cmake
   # remove all folders to avoid permission errors.  Leave root files (may have coverage info there)
   - for /d %%F in (c:\cbtmp\*) do rd /s /q "%%F"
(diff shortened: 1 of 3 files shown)


grlee77:
appveyor failures are seemingly unrelated:
```py.test: error: unrecognized arguments: --instafail```

**Comment by mingwandroid** [comment]
I added that option by mistake.. oops! Not sure why it's not working though. Please feel free to removed it here. I'll be busy today with other things.

**Comment by msarahan** [comment]
Sorry, we have explicitly decided not to add this.  I swear I've written this up, but I can't find it.  The reasoning is basically that continuing to support this requires ongoing PRs like yours.  We'd recommend a new pattern instead that will not require ongoing maintenance:

```
# [py == 37]
```

or ranges:

```
# [py >= 37]
```

grlee77:
fair enough. I had not been aware of the availability of ranges, but happened to come across it in some conda-smithy code shortly after making this PR.

PR #3282, opened 2018-11-30
Corrupt files result in `null` value in repodata.json

dsludwig:


msarahan:
Sorry, this didn't end up being the right fix.  The problem we had on the cloudflare mirror was that we weren't applying conda-forge's patch data at all.  It had been in an earlier script, but wasn't migrated to a newer script.  I'm going to go ahead and close this PR, because it has too many bad effects elsewhere.

Thanks, though!

PR #3291, opened 2018-12-04
Appveyor: Use py-lief only in py3 env

nehaljwani:


msarahan:
I think this is fine, but a bit superfluous as we have lief disabled on win for now completely?  Anyway, merging as it does no harm and will be useful in the future when we re-enable lief on win.

PR #3294, opened 2018-12-05
Fix #3292

### LINKED ISSUE #3292: conda-build 3.17 failing to build go-based packages
This issue is to pickup the parts of #3287 that did not get addressed, see:
  - `vault` issue: https://github.com/conda/conda-build/issues/3287#issuecomment-443922743
  - `ptvsd` issue: https://github.com/conda/conda-build/issues/3287#issuecomment-443953991
  - `go-bootstrap` issue: https://github.com/conda/conda-build/issues/3287#issuecomment-443965104
  - `git-lfs` issue: https://github.com/conda-forge/git-lfs-feedstock/pull/16

### Actual Behavior
Conda-build 3.17 fails to build packages that do not have dynamic linked libraries, e.g. `go`, `vault`, etc. 

### Expected Behavior
This behavior does not happen on conda-build 3.16 as can be seen when one pins conda-build to 3.16 within the conda-forge ci-scripts, see:
  - `git-lfs`: 
https://github.com/conda-forge/git-lfs-feedstock/pull/16/commits/2b5cb52fed265f4f37462e59a08b2ee89c384867
  - `vault`: 
https://github.com/conda-forge/vault-feedstock/pull/24/commits/6c4a10de48bbcf997edf401772df2946b19ae592
  - `go`: 
https://github.com/conda-forge/go-feedstock/pull/34/commits/9499072cbb18b24e5adc331b9cdcccbb1b5e6e63

### Steps to Reproduce
Build the packages above using conda-build 3.17. 

cc: @isuruf, @jakirkham

sodre:
There was a typo when getting the binary unique name when using py-lief which was triggered while building go-packages because those ELF executables do not have `dynamic_entries`. 

The code section that needs to be fixed is here https://github.com/conda/conda-build/blob/b845f7598fda4c754fa8773bddf70ca42cc11a72/conda_build/os_utils/liefldd.py#L237-L246

The typo is on line 246, since there is no `soname` attribute in an ELF binary according to [lief's  documentation](https://lief.quarkslab.com/doc/stable/api/python/elf.html#lief.ELF.Binary.section_from_offset).

The proposed fix is to remove lines 241 and 246. Once you inspect the body of the `241-for-loop`,  you see that the outside iteration variable `e` is never used. 

Finally, I added a simple `go-package` recipe as part of the tests so we can capture go-related errors earlier in the process.
 
cc: @msarahan, @mingwandroid

### DIFF
--- a/conda_build/os_utils/liefldd.py
+++ b/conda_build/os_utils/liefldd.py
@@ -238,12 +238,10 @@ def get_uniqueness_key(file):
          and  # noqa
          (binary.type == lief.ELF.ELF_CLASS.CLASS32 or binary.type == lief.ELF.ELF_CLASS.CLASS64)):
         dynamic_entries = binary.dynamic_entries
-        for e in dynamic_entries:
-            result = [e.name for e in dynamic_entries if e.tag == lief.ELF.DYNAMIC_TAGS.SONAME]
-            if result:
-                return result[0]
-            return binary.name
-        return binary.soname
+        result = [e.name for e in dynamic_entries if e.tag == lief.ELF.DYNAMIC_TAGS.SONAME]
+        if result:
+            return result[0]
+        return binary.name
     return binary.name
 
 
--- a/tests/test-recipes/go-package/hello.go
+++ b/tests/test-recipes/go-package/hello.go
@@ -1,0 +1,7 @@
+package main
+
+import "fmt"
+
+func main() {
+    fmt.Printf("Hello, World!\n")
+}
--- a/tests/test-recipes/go-package/meta.yaml
+++ b/tests/test-recipes/go-package/meta.yaml
@@ -1,0 +1,27 @@
+package:
+  name: go-hello
+  version: 1.0.0
+
+build:
+  number: 0
+  script:
+    - pushd ${RECIPE_DIR}  # [not win]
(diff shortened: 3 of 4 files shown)


conda_build/os_utils/liefldd.py
```diff
@@ -238,12 +238,10 @@ def get_uniqueness_key(file):
          and  # noqa
          (binary.type == lief.ELF.ELF_CLASS.CLASS32 or binary.type == lief.ELF.ELF_CLASS.CLASS64)):
         dynamic_entries = binary.dynamic_entries
-        for e in dynamic_entries:
-            result = [e.name for e in dynamic_entries if e.tag == lief.ELF.DYNAMIC_TAGS.SONAME]
-            if result:
-                return result[0]
-            return binary.name
-        return binary.soname
```
sodre: Go compiled binaries do not have dynamic_entries, so the for loop never executed, which caused line 246 to be triggered instead of 245.

If you look closely at lines 242-245 you will notice that it does not depend on the outside iterated variable `e` anyways and instead was relying on the list-comprehension in 242.

sodre:
@jjhelmus, @msarahan the fix to #3292 is ready for review

msarahan:
Thanks for figuring this out, @sodre.  Merging.

PR #3296, opened 2018-12-06
Expand instructions for setting up test environment.

jdblischak:
Without first installing anaconda-client, `py.test tests` fails because it cannot find binstar_client. I also added how to test a single file.


msarahan:
Thanks, docs improvements are always very helpful.

PR #3288, opened 2018-12-03
omit LIEF on win for now; fix behavior with no bld.bat on win

msarahan:
fixes #3287 
fixes #3285

PR #3298, opened 2018-12-07
log.warn does not accept `file=` keyword argument

dsludwig:
Introduced in https://github.com/conda/conda-build/pull/3288/files


msarahan:
Thanks @dsludwig

PR #3301, opened 2018-12-08
Reproduce AttributeError: 'NoneType' object has no attribute 'replace'

### LINKED ISSUE #3300: 3.17: AttributeError: 'NoneType' object has no attribute 'replace'
Please see here for details: https://github.com/conda-forge/maven-feedstock/pull/7#issuecomment-445413296

sodre:
Here is a minimal test case for reproducing the issue found by @dhirschfeld,
and @hadim using cb-3.17. 

My initial investigation leads me to believe that the bug is triggered when
you have a standalone .dll within a conda-package. Stand-alone here means that
it was not discovered by traversing some parent executable.

You can test that hypothesis with the following codebase

```python
from conda_build.os_utils import liefldd
liefldd.inspect_linkages_lief('jansi.dll')
```

fixes #3300


msarahan:
Thanks! This should be very helpful in narrowing it down. We'll try to get a fix out soon.

mingwandroid:
LIEF was a bit prematurely committed and I really full responsibility for that, sorry all.

Hopefully once I get back to coding the value of LIEF will become apparent as it allows us to trim our dependency trees to a big extent, which will make all the things faster, smaller and less rigid. This will also make downstream rebuilds less painful.

PR #3304, opened 2018-12-11
fix to ignore unsatisfiable pin_compatible calls for packages in othe…

bergtholdt:
…r outputs

by using a regular expression including the package_name in the check.


msarahan:
Makes sense.  Thanks @bergtholdt

PR #3307, opened 2018-12-12
Replace JSONDecodeError with Py version agnostic solution

Lnaden:
#3303 introduced except clauses looking for `json.decoder.JSONDecodeError` objects instead of relying on the Python 2/3 agnostic solution from the header of the index.py file.

This PR corrects those 2 lines to prevent crashing on Python 2.7 builds which have gone live with conda-build 3.17.3

I don't have a formal issue for this opened, would you like me to for documentation purposes or is this sufficient?

### DIFF
--- a/conda_build/index.py
+++ b/conda_build/index.py
@@ -169,7 +169,7 @@ def _download_channeldata(channel_url):
         try:
             with open(tf) as f:
                 data = json.load(f)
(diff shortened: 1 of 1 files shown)


msarahan:
Thanks! You need not open a separate issue. This is fine.

Issue #2255, opened 2017-08-03
SSL Crypto errors when building with 3.0.8


darkdreamingdan:
I have a basic NLTK downloader recipe, with a simple build.sh:

`python -m nltk.downloader -d $PREFIX/nltk_data book`

This works fine in conda-build 2.1.15.  After upgrading to 3.0.8, i get the following error:

```[nltk_data] Error loading book: <urlopen error [SSL:
/home/user/miniconda3/conda-bld/nltk_book_1501757062443/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho/lib/python3.6/runpy.py:125: RuntimeWarning: 'nltk.downloader' found in sys.modules after import of package 'nltk', but prior to execution of 'nltk.downloader'; this may result in unpredictable behaviour
[nltk_data]     CERTIFICATE_VERIFY_FAILED] certificate verify failed
[nltk_data]     (_ssl.c:748)>
```

I think this is to do with the pycrypto dependency that conda-build has.  For now, I am using 2.1.15

msarahan:
conda-build removed pycrypto as a dependency when we took out the package signing stuff, which was not implemented securely enough.  Maybe you just need to add pycrypto as a package build time requirement?

darkdreamingdan:
I just tried adding pycrypto as a dependency to by meta.yaml, installing it doesn't seem to help - the bug doesn't seem related to pycrypto after all.

What's really strange is that I can only only seem to reproduce with `conda build`, and not otherwise.  After the build hangs at this step, if i cd into the placeholder directory, and do `bin/python -m nltk.downloader book` it works fine, which is really strange.  It's not something to do with the environment layout itself, I think it's some variable that's not being set correctly during the conda build process.

msarahan:
yes, that might be the case.  Please see what you can find by comparing a dump of environment variables from your base prompt to what you should see embedded at the top of the generated conda_build.sh or build.sh file in the work folder.

darkdreamingdan:
So i digged into the conda_build.sh .   3.0.8 seems to export a _lot_ of variables while 2.1.15 only activates the environment.  While digging through the file, i found this:

`export SSL_CERT_FILE=""`

[...]

PR #2259, opened 2017-08-05
only set env vars in build scripts that actually have values

msarahan:

PR #3319, opened 2018-12-13
fix pip build isolation

### LINKED ISSUE #3318: PIP_NO_BUILD_ISOLATION not set in build environment
Despite the being specified in build.py, #3053, `PIP_NO_BUILD_ISOLATION` is not being set in the build/host environment.

``` yaml
package:
  name: foobar
  version: 1.0.0

build:
  script:
    - env | grep PIP_NO
    - exit 1
```

```
$ conda build --version
read filter "zstd" is not supported
write filter "zstd" is not supported
conda-build 3.17.4
$ conda build .
...
BUILD START: ['foobar-1.0.0-0.tar.bz2']
source tree in: /home/jhelmus/anaconda3/conda-bld/foobar_1544729597284/work
export PREFIX=/home/jhelmus/anaconda3/conda-bld/foobar_1544729597284/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_pla
export SRC_DIR=/home/jhelmus/anaconda3/conda-bld/foobar_1544729597284/work
PIP_NO_INDEX=True
PIP_NO_DEPENDENCIES=True
```

jjhelmus:
The environment variable PIP_NO_BUILD_ISOLATION is set to 'False' in the
build environment.  Previously this variable was not set in the environment
due to it having a falsey value. It is now set to the string 'False' so that it
has a truethy value in Python which results in in being included in the
environment file.

closes #3318

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -1877,7 +1877,7 @@ def write_build_scripts(m, script, build_file):
     env_file = join(m.config.work_dir, 'build_env_setup.sh')
     with open(env_file, 'w') as bf:
         for k, v in env.items():
-            if v:
+            if v != '':  # do not write empty strings
                 bf.write('export {0}="{1}"\n'.format(k, v))
 
         if m.activate_build_script:
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -1876,7 +1876,7 @@ def write_build_scripts(m, script, build_file):
     env_file = join(m.config.work_dir, 'build_env_setup.sh')
     with open(env_file, 'w') as bf:
         for k, v in env.items():
-            if v != '':  # do not write empty strings
+            if v != '' and v is not None:
                 bf.write('export {0}="{1}"\n'.format(k, v))
 
         if m.activate_build_script:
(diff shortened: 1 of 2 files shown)


dhirschfeld:
I'm actually curious that skipping falsey env vars is the right thing to do.

To satisfy my curiosity I tracked it down to https://github.com/conda/conda-build/pull/2259, which appears to have been in response to https://github.com/conda/conda-build/issues/2255.

I might be inclined to instead fix it by *only* skipping an empty string which I think was the original intention to resolve #2255 
```python
   with open(env_file, 'w') as bf:
        for k, v in env.items():
            if v == '':
                bf.write('export {0}="{1}"\n'.format(k, v))
```

Either way works though...

hmaarrfk:
Has a new release been cut since this has been pulled in?

We have been running into this issue:
https://github.com/conda-forge/scikit-image-feedstock/issues/36

hmaarrfk:
cc @jakirkham

Issue #282, opened 2014-12-22
conda-build has no offline mode


jf---:
today, the http://repo.continuum.io/ went down, which does bring about the question that it would be great if conda-build would support an offline mode ( when no network connection is available ), where only local packages are considered.

jakirkham:
Either this issue or issue ( https://github.com/conda/conda-build/issues/695 ) should be consolidated and closed.

PR #3328, opened 2018-12-20
Respect context.offline setting

qwhelan:
While `conda-build` does not have an `--offline` flag like other conda commands (see https://github.com/conda/conda-build/issues/282 and 
https://github.com/conda/conda-build/issues/695), it *does* respect the equivalent environment variable `$CONDA_OFFLINE`. This is due to the `conda.common.configuration.Configuration` class, which pulls in environment variables to the `conda.context` module.

The `asv` test suite runs a number of builds inside test cases to ensure its `conda` integration is working; those tests have been failing due to `conda-build` exiting with exit code 1 and the following message in stdout:
```
EnforceUnusedAdapter called with url https://repo.anaconda.com/pkgs/main/channeldata.json
This command is using a remote connection in offline mode.
```
The intent behind running with `$CONDA_OFFLINE=True` is to ensure all dependencies have been specified and downloaded prior to the tests starting (and I have confirmed that's the case).

As this is supported in much of the `conda` tooling, it appears like it could be a minimal change to extend this support by simply skipping downloading `channeldata.json` (which doesn't appear to directly make it into the build). If having `channeldata.json` accessible is necessary, a simple cache akin to `repodata.json` would also solve this case.

Additionally, this PR adds a test case that runs a build under `$CONDA_OFFLINE=True` for regression purposes.

### DIFF
--- a/conda_build/index.py
+++ b/conda_build/index.py
@@ -275,10 +275,11 @@ def get_build_index(subdir, bldpkgs_dir, output_folder=None, clear_cache=False,
                             retry += 1
                 else:
                     # download channeldata.json for url
-                    try:
-                        channel_data[channel.name] = _download_channeldata(channel.base_url + '/channeldata.json')
-                    except CondaHTTPError:
-                        continue
+                    if not context.offline:
+                        try:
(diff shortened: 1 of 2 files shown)


msarahan:
makes sense to me.  Thanks.

PR #3330, opened 2018-12-21
don't write bytecode when building noarch: python packages

msarahan:
Since we don't want to include this in noarch: python packages, it makes no sense to have python go through with creating pyc files when installing a package.  This cleans it up.  Might save a tiny bit of time.

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -1872,6 +1872,9 @@ def write_build_scripts(m, script, build_file):
     # locally, and if we don't, it's a problem.
     env["PIP_NO_INDEX"] = True
 
+    if m.noarch == "python":
+        env["PYTHONDONTWRITEBYTECODE"] = True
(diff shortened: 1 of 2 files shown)


mingwandroid:
Do we have any concerns about space requirement calculations being incorrect from `noarch: python` packages?

msarahan:
I personally do not.  I'd expect the upper bound for that to be some few MB.  At any rate, the packages created by conda-build shouldn't be including any pyc files, so any space requirement calculations would already be inaccurate, right?

mingwandroid:
I was wondering whether shipping with dummy pyc files (valid .pyc header, timestamp set to 1 second behind the .py files) that get deleted by the arch-ifying process would keep things more or less the same size.

.. but I'm not *too* concerned in the grand scheme of things either.

PR #3332, opened 2018-12-21
Fix mercurial TLS failures by switching from system mercurial/python 

qwhelan:
The Travis base image uses Ubuntu 14.04, which supports a maximum system python version of 2.7.6 via apt. Therefore, the version of mercurial installed via apt cannot get access to the TLS improvements available in python 2.7.9+ and will always fail with the following output:
>warning: connecting to bitbucket.org using legacy security technology (TLS 1.0); see https://mercurial-scm.org/wiki/SecureConnections for more info
>abort: error: _ssl.c:510: error:1408F10B:SSL routines:SSL3_GET_RECORD:wrong version number
>Warning: failed to download source.  If building, will try again after downloading recipe dependencies.
Error was: 
Command '['hg', 'clone', 'https://bitbucket.org/asmeurer/conda-build-hg-test', '/tmp/cb/popen-gw1/test_recipe_builds_source_hg_0/hg_cache/__bitbucket.org_asmeurer_conda-build-hg-test']' returned non-zero exit status 255.

Simply using the `conda`-provided `mercurial` package resolves this issue. The system `mercurial` could be used again once the base Ubuntu image is upgraded to a newer version than 14.04.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,6 @@ addons:
   apt:
     packages:
     - attr
-    - mercurial
 
 install:
   # show which shell we are running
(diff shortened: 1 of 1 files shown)


msarahan:
Yeah, thanks for attempting a fix. I think we should just xfail those tests
for now.

PR #3334, opened 2018-12-28
add support for conda pkg v2 format

msarahan:
Implements .conda package format, as described in https://docs.google.com/document/d/1HGKsbg_j69rKXPihhpCb1kNQSE8Iy3yOsUU2x68x8uw/edit?usp=sharing

This format's main purpose to Anaconda is to better support package signing as an enterprise feature, but we hope the improved compression algorithms and split, indexed file format will speed up metadata handling and install times.

TODO:
* [x] add index capability
* [ ] conda support for new packages (https://github.com/conda/conda/pull/8265)
* [ ] signing infrastructure plan


msarahan:
Right now, the internal packages have generic names.  It would be better if the internal packages had names that corresponded to the outer container, so that things aren't crazy.

For example:

* libabc-1.0-0.conda contains info.tar.zst and package.tar.zst
* it should contain info-libabc-1.0-0.tar.zst and package-libabc-1.0-0.tar.zst

Working with these packages is also not really great, given that zst isn't one of the standard tar flags (AFAIK).  It would be helpful to have a tool to extract these packages to a target location.

jakirkham:
Sorry to ask this on this PR, wasn't able to find the proper issue. Is there a good place for community feedback about this package format change?

jakirkham:
Sorry I didn't follow. Should we be discussing this in this PR or is there an issue that would be more appropriate for this discussion?

msarahan:
Sorry that wasn't clear.  This PR is fine.  Otherwise I'd just say "comment on the google doc" but I think discussion here is more visible and helpful.

msarahan:
I created a library that extracts and abstracts the creation and extraction of either format.  It's probably best now to move discussion to issues on that library: https://github.com/conda/conda-package-handling

conda_build/config.py
```diff
@@ -174,7 +174,13 @@ def _get_default_settings():
             Setting('extra_deps', []),
 
             # customize this so pip doesn't look in places we don't want.  Per-build path by default.
-            Setting('_pip_cache_dir', None)
+            Setting('_pip_cache_dir', None),
+
+            # set up compression algorithm used in new-style packages
+            Setting('compression_tuple', ('.tar.zst', 'zstd', 'zstd:compression-level=22')),
```
mariusvniekerk: we may want to default to using the multithreaded compression options inside zstd if we are planning on using compression level 22 which is very expensive to do.

msarahan: That's a good idea. Do you know if it affects determinism of the output, though? That and like a common trade-off, and here since the time for compression is on the build side, I think determinism will win out.

It's worth trying, to be sure.

msarahan: Looks like maybe, but it depends on the number of threads used for compression: https://github.com/facebook/zstd/issues/1077

PR #3342, opened 2019-01-04
document downstream testing

msarahan:


docs/source/define-metadata.rst
```diff
@@ -946,6 +946,50 @@ NOTE: Python .py and Perl .pl scripts are valid only
...
+.. code-block:: yaml
+
+   test:
+     downstreams:
+       - some_downstream_pkg
+       - other_downstream_pkg
+
+However, this does not mean that these packages are tested together. Rather,
+each of these are tested for satisfiability with your new package, then each of
+their test suites are run separately with the new package.
```
sodre: What happens when the test suite includes directives like source_files?

msarahan: That's independent of this - those source_files get included with the package, and thus become part of whatever downstream package that gets tested.

PR #3346, opened 2019-01-09
Conda 46 fixes

msarahan:


msarahan:
Seems like 4.6.0 is slower for conda-build.  We'll need to debug this.  It is fixed for the weird python upgrade failures, so I'm merging.

PR #3353, opened 2019-01-18
test: Fix run_test script name for multiple outputs

mingwandroid:


msarahan:
This appears to break the inclusion of .bat and .sh test files for noarch packages.  That needs to be fixed before this can be merged.

msarahan:
Yep, all better.  Timeouts aside, this is fine.  Merging.

PR #3356, opened 2019-01-23
attempt to use only released versions of conda (avoid auxlib version)

msarahan:


msarahan:
This didn't make a difference.  Closing.

PR #3386, opened 2019-02-10
Reduce length of test environement prefix

chrisburr:
I've seen a couple errors caused by the prefix replacement failing when the install path is shorter than the build path. This has been caused by the null bytes that are appended on the strings and could be fixed using `has_prefix_files` to explicitly mark them as text files.

I propose reducing the length of the testing environment path by 2 characters so these cases can be noticed by `conda-build`'s testing step.

### DIFF
--- a/conda_build/config.py
+++ b/conda_build/config.py
@@ -604,7 +604,9 @@ class Config(object):
         """The temporary folder where the test environment is created"""
         if on_win or not self.long_test_prefix:
             return self._short_test_prefix
-        return self._long_prefix(self._short_test_prefix)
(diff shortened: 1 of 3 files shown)


henryiii:
`test_long_test_prefix_length` looks for 80 chars, it needs to be changed to 78 too.

henryiii:
@chrisburr there's still a length issue.

chrisburr:
Thanks for noticing @henryiii 👍 

The failures are now the (normal?) timeouts and unrelated to this PR.

msarahan:
Yep, all good.  Thanks @chrisburr and @henryiii - merging.

PR #3387, opened 2019-02-10
Allow *_has_prefix_files to override the mode

chrisburr:
The documentation seems to imply that if a file is automatically detected as `binary` adding it to `text_has_prefix_files` should force it to be treated as `text` when replacing the prefix (and vice versa)

Currently an exception is raised `Did not detect hard-coded path in %s from text_has_prefix_files`.

Perhaps there is a different way of modifying this? If not, this PR modifies the behaviour to match what I expected from the documentation.


msarahan:
I think your interpretation of the docs makes sense.  Thanks for the fix.

PR #3389, opened 2019-02-11
Meta yaml example

rrigdon:

### DIFF
--- a/docs/source/define-metadata.rst
+++ b/docs/source/define-metadata.rst
@@ -12,7 +12,7 @@ Defining metadata (meta.yaml)
 All the metadata in the conda build recipe is specified in the
 ``meta.yaml`` file. See the example below: 
 
-.. code-block:: RST
+.. code-block:: YAML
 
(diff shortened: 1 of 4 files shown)


rrigdon:
Hey, I made some small changes. Are these worthy of the News from Rever process? @msarahan

docs/source/define-metadata.rst
```diff
@@ -10,9 +10,46 @@ Defining metadata (meta.yaml)
 
 
 All the metadata in the conda build recipe is specified in the
-``meta.yaml`` file, as in this simple example:
+``meta.yaml`` file. See the example below: 
 
-.. literalinclude:: ../../tutorials/meta.yaml
+.. code-block:: RST
```
msarahan: This is YAML, not RST.  I think the syntax highlighting might get messed up.

rrigdon: We tested it in the local html with both YAML and RST. RST was the one that looked the best.

PR #3397, opened 2019-02-14
[skip ci] Reorganized Resources, updated link scripts

rrigdon:
Massive reorganization of conda-build by adding a Resources section. I also updated the link scripts page per PR #619.


msarahan:
are there any redirects we need to put in place before this goes in?  Or will all of these keep any potential bookmarks or google hits working?

rrigdon:
This will likely break those links. I'll work on putting redirects in now.

rrigdon:
Redirects are in place.

msarahan:
fingers crossed!  good luck.

PR #3403, opened 2019-02-19
Removed duplicated code

gabrielcnr:
@msarahan - I came across this duplicated code. The expression remove is a copy of the one above.


msarahan:
Thanks!

PR #3415, opened 2019-02-25
Replace where command with which command

hrzafer:
AFAIK, there is no where command in unix. There is a whereis command and a which command. I think which is the "command to see which version comes first on PATH".


jjhelmus:
The `where` command exists on Windows.  On this platform the situation is reversed `where` exists but `which` does not.  Explaining this succinctly in this section would be best.  

@rrigdon Do you have any recommendation for wording around this issue?

hrzafer:
Added another commit on this PR. I don't know if I should create a new PR.

jjhelmus:
This looks good to me, @hrzafer thanks for the update.

PR #3422, opened 2019-02-28
Conda build reorg

rrigdon:


msarahan:
As for any reorg change, are redirects in place for all of these?

rrigdon:
Great point - let me check on the redirects.

rrigdon:
Redirects are in place.

PR #3436, opened 2019-03-08
Theme updates

rrigdon:
Updated the theme to be on brand with our [Distribution site.](https://docs.anaconda.com/)


docs/_static/css/custom.css
```diff
@@ -0,0 +1,57 @@
+@import url("theme.css");
+
+.wy-nav-content {
```
soapy1: Could you comment these to say what ui elements they change

docs/_static/css/custom.css
```diff
@@ -0,0 +1,57 @@
...
+.wy-nav-side {
+/* This relates to the entire color of the sidebar */
+   background-color:#EEEEEE;
+}
+
+.wy-menu-vertical {
+    color: black;
+}
+
+h1 {
```
soapy1: Could a class be attached to this?

soapy1:
These changes look good to me! Is there a timeline of when this change should go in? Like do we want to do a design review?

rrigdon:
It's been reviewed by Kelly and Laurel so I think it should be fine to publish this afternoon or first thing Monday.

PR #3465, opened 2019-04-02
Escape quotes in env variables.

beauby:
Quotes are not escaped when forwarding env variables.


marcelotrevisani:
Could you please add some tests to cover it?

beauby:
@marcelotrevisani Sure, any pointers?

marcelotrevisani:
Regarding the quotes you can see something in the ``test_get_summary`` and to test env variables you can see on ``test_environment_creation_preserves_PATH`` how to set and recover ``envs`` from the environment

PR #3471, opened 2019-04-10
Bump python version in CI from 3.6 -> 3.7

nehaljwani:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
-os: linux
+dist: xenial
 language: python
 python:
   - 2.7
-  - 3.6
+  - 3.7
 env:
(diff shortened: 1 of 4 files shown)


nehaljwani:
Can't use 3.7 on travis without switching to xenial. https://github.com/travis-ci/travis-ci/issues/9815

msarahan:
Is there anything blocking us updating to xenial?  I don't know of any reason not to.

msarahan:
The skip_build_id stuff is stuff that I changed to try to work around an infinite recursion problem that I hit with c3i.  It's fine to change this here, but we may need to revisit it in the future if we hit the infinite recursion problem again.

PR #3472, opened 2019-04-10
Introduce conda_build.config.get_channel_urls()

### LINKED ISSUE #3433: Conda-debug fails with local requirements
### Actual Behavior

Whenever a recipe has a build or host requirement on a locally build package, `conda-debug` fails with the error message

```
conda_build.exceptions.DependencyNeedsBuildingError: 
    Unsatisfiable dependencies for platform [...]: {'<name of dependency>'}
```

### Expected Behavior

Since `conda-build` works perfectly fine in these case, I expected `conda-debug` to also work.

### Steps to Reproduce

In `test_package_1/meta.yaml`:
```
package:
  name: test_package_1
  version: 1.0
```
In `test_package_2/meta.yaml`:
```
package:
  name: test_package_2
  version: 1.0

requirements:
  build:
    - test_package_1
```
Then do:
```
$ conda debug test_package_2
```
This fails the with error given above. 

Note however that when the local build directory is added to the channel list with `conda config`, it does work. Specifying the local build directory as a command line argument (either with `-c local` or `-c <conda_path>/conda-bld` does not work.

##### `Output of conda info`

```
     active environment : None
            shell level : 0
       user config file : ~/.condarc
 populated config files : ~/.condarc
          conda version : 4.6.7
    conda-build version : 3.17.8
         python version : 3.7.1.final.0
       base environment : ~/miniconda3  (writable)
           channel URLs : https://repo.anaconda.com/pkgs/main/osx-64
                          https://repo.anaconda.com/pkgs/free/osx-64
                          https://repo.anaconda.com/pkgs/free/noarch
                          https://repo.anaconda.com/pkgs/r/osx-64
          package cache : ~/miniconda3/pkgs
                          ~/.conda/pkgs
       envs directories : ~/miniconda3/envs
                          ~/.conda/envs
               platform : osx-64
             user-agent : conda/4.6.7 requests/2.19.1 CPython/3.7.1 Darwin/18.2.0 OSX/10.14.3
                UID:GID : 502:20
             netrc file : None
           offline mode : False
```
where I've unexpanded the home directory.

nehaljwani:
This patch introduces a function to reduce duplicate code and makes use
of it in conda_build.api.debug, conda_build.cli.main_render.execute and
conda_build.cli.main_build.execute.

This change makes conda-debug to also honor channels passed with the -c
argument.

Fixes #3433

### DIFF
--- a/conda_build/config.py
+++ b/conda_build/config.py
@@ -15,7 +15,7 @@ import time
 from .conda_interface import root_dir, root_writable
 from .conda_interface import binstar_upload
 from .variants import get_default_variant
-from .conda_interface import cc_platform, cc_conda_build, subdir
+from .conda_interface import cc_platform, cc_conda_build, subdir, url_path
 
 from .utils import get_build_folders, rm_rf, get_logger, get_conda_operation_locks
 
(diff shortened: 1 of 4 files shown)


conda_build/config.py
```diff
@@ -805,5 +805,22 @@ def get_or_merge_config(config, variant=None, **kwargs):
     return config
 
 
+def get_channel_urls(args):
```
nehaljwani: I am not so confident about this function signature. Is this okay, @msarahan @jjhelmus ?

msarahan: I think ideally you'd have it be "channels" and you'd call ``ensure_list`` on it to make sure it's a list.  However, given that it's already called so awkwardly with either ``channel`` or ``channels``, the way it is is probably the best you can do.  I wouldn't change this right now.  Another wart for the pile.

PR #3474, opened 2019-04-12
indexing creates repodata_from_packages.json file

jjhelmus:
When conda index is run `repodata_from_packages.json` and `repodata_from_packages.json.bz2 files` are created that contain the un-patched metadata as it appears in the packages. This information is helpful when determining what packages need to be extracted to obtain their metadata and which can used the cached metadata. Additionally, these files are helpful when writing and checking repodata hotfixes.

I welcome suggestions for alternative names for these files, `repodata_from_packages` is the first idea that popped into my head.


msarahan:
Merged with master to bring in newline normalization especially.  @jjhelmus please be sure to update your local notion of "master" to pick that up.  Your PR had awkward windows newlines, presumably from stuff that my windows computer originally introduced.  I think we've ironed that all out now, but we all need to be careful to not reintroduce it.

Issue #3475, opened 2019-04-12
Conda build ignores order of recipe directories (since 3.17.0)


tjd2002:
### Actual Behavior

Since 3.17.0, when providing multiple recipe directories on a command line, conda-build ignores the order in which the recipes are provided, and instead builds the recipes in (apparently) alphabetical order.

So, calling `conda build bbb aaa` now builds package 'aaa' before 'bbb'.

I had been relying on this behavior to build a set of packages with internal dependencies in a specific order. My workaround is to loop over the list of directories instead, calling conda-build once for each recipe. (If there's a better way to manage these dependencies, I'm all ears.)

I used `git bisect` to identify that this change in behavior was introduced by @msarahan in  1a934e5ad4d36b82609b5cef3c3c1a2c00fcb539. That commit is about adding debug functionality, so I suspect this was an accidental regression. The line to blame is:
https://github.com/conda/conda-build/blob/fc22ab8c161a6c79ed3d48fab0ffc954a5fa61c2/conda_build/utils.py#L1141

### Steps to Reproduce
Create 2 recipes, in 2 directories named aaa and bbb:

`aaa/meta.yaml`:
```meta.yaml
package:
  name: aaa
  version: 1.0
```

`bbb/meta.yaml`:
```meta.yaml
package:
  name: bbb
  version: 1.0
```
Call `conda build bbb aaa`...

conda builds package `aaa` first.

### Expected Behavior

I expect conda-build to first build package bbb, then package aaa. (Before 3.17.0, this is indeed what happened.)

##### `Output of conda info`

```
     active environment : base
    active env location : /home/tjd/anaconda3
            shell level : 1
       user config file : /home/tjd/.condarc
 populated config files : /home/tjd/.condarc
          conda version : 4.6.11
    conda-build version : 3.17.8
         python version : 3.6.8.final.0
       base environment : /home/tjd/anaconda3  (writable)
           channel URLs : https://repo.anaconda.com/pkgs/main/linux-64
                          https://repo.anaconda.com/pkgs/main/noarch
                          https://repo.anaconda.com/pkgs/free/linux-64
                          https://repo.anaconda.com/pkgs/free/noarch
                          https://repo.anaconda.com/pkgs/r/linux-64
                          https://repo.anaconda.com/pkgs/r/noarch
          package cache : /home/tjd/anaconda3/pkgs
                          /home/tjd/.conda/pkgs
       envs directories : /home/tjd/anaconda3/envs
[...]

PR #3477, opened 2019-04-12
Do not sort files after globbing (fixes 3475)

tjd2002:
This addresses #3475. 

Doesn't seem to cause any more tests to fail than were already failing.


nehaljwani:
@tjd2002 Did you change the line endings in this PR?

tjd2002:
Yes, I apologize. I will rebase.

PR #3483, opened 2019-04-20
Fix rendering fails if package/build not at top

epruesse:
The regular expression used in `metadata.MetaData.extract_package_and_build_text()` stalls with no characters consumed if the recipe starts with any of the excluded sections (e.g. `about:`). The result is then empty and api.render() raises an exception.

The regex also did not exclude the `source` section, leading to longer fragments than expected.

Honestly, I'm not sure the regex parsing is such a good idea - pretty hard to read and fragile. This bug looks like a copy/paste error from `extract_source_text` that "kind of worked" the way recipes are "commonly written".


epruesse:
Signing the CLA is difficult. I can give license under BSD or similar, or we can just take the few characters changed as "below the line for copyrightable". If in doubt, just take it as free advice and make a PR on your own.

PR #3490, opened 2019-05-01
[RFC] Add support for signing repodata

nehaljwani:
This is a draft, don't merge yet.

### DIFF
--- a/conda_build/cli/main_index.py
+++ b/conda_build/cli/main_index.py
@@ -66,6 +66,23 @@ def parse_args(args):
         action="store_true",
         dest="convert_if_not_present"
     )
+    p.add_argument(
+        "--gpg-sign",
+        help="Sign artifacts using GPG",
+        action="store_true",
+        dest="gpg_sign"
+    )
+    p.add_argument(
+        "--gpg-tool",
+        help="GPG program used to sign artifacts",
+        default="gpg",
+        dest="gpg_tool"
+    )
+    p.add_argument(
+        "--gpg-identity",
+        help="GPG identity used to sign artifacts",
+        dest="gpg_identity"
+    )
 
     args = p.parse_args(args)
     return p, args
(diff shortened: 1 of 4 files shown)


conda_build/utils.py
```diff
@@ -1768,3 +1768,38 @@ def sha256_checksum(filename, buffersize=65536):
         for block in iter(lambda: f.read(buffersize), b''):
             sha256.update(block)
     return sha256.hexdigest()
+
+
+class GPGWrapper(object):
```
nehaljwani: This will be eventually be used by Conda, CB and CPH. So, this should ideally go into cph.utils

conda_build/index.py
```diff
@@ -934,6 +936,13 @@ def index(self, patch_generator, hotfix_source_repo=None, verbose=False, progres
                 self._write_channeldata_rss(channel_data, package_mtimes, hotfix_source_repo)
                 self._write_channeldata(channel_data)
 
+                # Step 8. Sign the repodata
+                if gpg_wrapper:
+                    for subdir in subdirs:
+                        for _file in ["repodata.json.bz2", "repodata.json", "repodata2.json"]:
```
nehaljwani: I couldn't find any easy way to iterate over all the repodata files. Hence the hardcoding.

conda_build/cli/main_index.py
```diff
@@ -66,6 +66,23 @@ def parse_args(args):
...
+        "--gpg-tool",
+        help="GPG program used to sign artifacts",
+        default="gpg",
+        dest="gpg_tool"
+    )
+    p.add_argument(
+        "--gpg-identity",
+        help="GPG identity used to sign artifacts",
+        dest="gpg_identity"
+    )
```
nehaljwani: The argument options are similar to the 'twine' utility, but with slightly modified names.

PR #3500, opened 2019-05-10
implement new format indexing in its own repodata section

msarahan:
Sister PR to https://github.com/conda/conda/pull/8639

PR #3502, opened 2019-05-13
new .conda handling.  "current_repodata.json"

msarahan:
current_repodata.json is a subset of repodata.json.  It has only the most recent versions of packages in it.  It tries to include older packages where necessary to ensure satisfiability.  It can also use a file like conda_build_config.yaml to pull in more than one class of package (like python 2.7, 3.6, 3.7).

This PR includes and supersedes #3500


conda_build/api.py
```diff
@@ -378,22 +378,28 @@ def create_metapackage(name, version, entry_points=(), build_string=None, build_
...
-    from .conda_interface import PY3
+    from .conda_interface import PY3, string_types
     from conda_build.index import update_index
     from conda_build.utils import ensure_list
     dir_paths = [os.path.abspath(path) for path in _ensure_list(dir_paths)]
     # Don't use byte strings in Python 2
     if not PY3:
         dir_paths = [path.decode(getpreferredencoding()) for path in dir_paths]
 
+    if current_index_versions and isinstance(current_index_versions, string_types):
```
nehaljwani: Is the first condition needed? Unless I am missing something, `isinstance()` should on `NoneType` too.

conda_build/cli/main_index.py
```diff
@@ -60,11 +60,23 @@ def parse_args(args):
         "--no-progress", help="Hide progress bars", action="store_false", dest="progress"
     )
     p.add_argument(
-        "--convert",
```
nehaljwani: Is the removal of the `--convert` option intentional?

conda_build/index.py
```diff
@@ -37,11 +38,16 @@
 import libarchive
 import conda_package_handling.api
 
+#  BAD BAD BAD - conda internals
+from conda.core.subdir_data import SubdirData
+from conda.models.channel import Channel
```
nehaljwani: Shouldn't these also come via `conda_build.conda_interface`?

conda_build/cli/main_index.py
msarahan: Yeah, I think that's better done as an explicit step with CPH.  No reason to include it here.

conda_build/index.py
msarahan: Definitely, but they're not current exposed in conda's exports.  It's really bad, and this needs a better way to do things without reaching into conda's internals.  I'm not sure it should block this PR and 3.18.0, though.

conda_build/api.py
msarahan: I think you're right

PR #3506, opened 2019-05-15
Replace stdlib glob with utils.glob

### LINKED ISSUE #3504: test/source_files recursive glob doesn't work
The following recipe snippet doesn't work as I expect:
```yaml
test:
  source_files:
    - tests/**/*.py
```
Namely, it doesn't find the files of the form `tests/test_something.py`, but does find files similar to `tests/subdir/test_somethingelse.py`.

This is because the code in question, https://github.com/conda/conda-build/blob/63b1f0b98cfdff96e389e5015ba72b67ba06cc22/conda_build/build.py#L396
uses the stdlib glob (without recursion):
https://github.com/conda/conda-build/blob/63b1f0b98cfdff96e389e5015ba72b67ba06cc22/conda_build/build.py#L8
There's a PR that introduced glob2 (#2146) but that somehow overlooked the above line.

I myself am bit hesitant to write a PR that simply changes the above import to glob2 since it also touches other glob statements in `build.py`. Hence this issue.

As a side note, there's a generic `glob` definition in `conda-build/conda_build/utils.py` that turns recursion on by default. But using that won't work for Python < 3.5 since glob recursion was introduced in the stdlib in 3.5.

teake:
The stdlib glob doesn't use recursion (`**`) by default, but utils.glob
does. Note that `**` is only supported since Python 3.5, but since
anything older is already EOL (and because we fall back on glob2 for
Python 2) this isn't a problem.

Closes #3504.

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -5,7 +5,6 @@ from __future__ import absolute_import, division, print_function
 
 from collections import deque, OrderedDict
 import fnmatch
-from glob import glob
 import io
 import json
 import os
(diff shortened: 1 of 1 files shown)


msarahan:
Thanks!

PR #3509, opened 2019-05-18
Fix inconsistency in computation of changed files

neok-m4700:
When updating a file in channel with different build time and/or disk size `_calculate_update_set` incorrectly returns a zero length `update_set`.


msarahan:
Thanks!

PR #3512, opened 2019-05-18
Speed up fix post

mingwandroid:


msarahan:
Merging as a moderate improvement, but we still have a major speed problem.  We'll work on that soon.

PR #3515, opened 2019-05-20
Activate for test

mingwandroid:

### DIFF
--- a/ci/travis/run.sh
+++ b/ci/travis/run.sh
@@ -29,9 +29,9 @@ else
       conda create -n blarg3 -yq python=3.5
       conda create -n blarg4 -yq python nomkl numpy pandas svn
       conda create -n blarg5 -yq libpng=1.6.17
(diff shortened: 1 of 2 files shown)


mingwandroid:
Please review @msarahan.

msarahan:
This seems fine to me. I'm surprised that it's needed, though. It seems like we would have noticed the lack of activation before now? Activation is definitely necessary and this appears to be the right place. I just wonder if there is also somewhere deeper that is writing it, like in this function: https://github.com/conda/conda-build/pull/3515/commits/3da2b16a4b1204c9783811fe7062dd52b70d2bcb#diff-ac55d744935b1f37a3369258b88e8f1fR1970

mingwandroid:
I am fixing the bug I introduced [here](https://github.com/conda/conda-build/commit/c03f5a067e98b0a6c2e5f932e168d51e966f8159)

I didn't consider changing the behaviour as something I wanted to be doing here.

mingwandroid:
I assume there's no test coverage for this either. If you want me to add that, let me know.

PR #3525, opened 2019-06-03
Binary repackaging fixes and some speed ups.

mingwandroid:


soapy1:
Looks like the CI failures are not related to this PR. Tried this out a bit. Looks like conda index is quite a bit faster. With current conda build it takes more than 10 min to index conda-forge. With this PR it gets to sub 10 min 🚀

PR #3526, opened 2019-06-04
Split variant parsing function so that users can get an order preserving combined variant spec.

mariusvniekerk:
For complex build systems that rely heavily on variant specs to produce
and update a large set of packages (such as conda-forge or c3i), it is
useful to have some means of overriding the behavior that is used by
conda render when generating package variants.

Alternatively we could just provide a config value allowing an override for `combined_spec` to be specified at the config level.

cc @msarahan

### DIFF
--- a/conda_build/config.py
+++ b/conda_build/config.py
@@ -746,7 +746,7 @@ class Config(object):
 
     @post_variant_combine_hook.setter
     def post_variant_combine_hook(self, fn):
-        if not callable(fn):
+        if fn and not callable(fn):
             raise ValueError("fn is expected to be a callable of the form:"
                              "\ndef post_variant_combine_hook("
                              "\n        specs, combined_spec, config=config, log_output=log_output):"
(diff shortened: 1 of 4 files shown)


mariusvniekerk:
See also https://github.com/conda-forge/conda-smithy/pull/1072

conda_build/config.py
```diff
@@ -187,6 +187,11 @@ def _get_default_settings():
 
             # this can be set to different values (currently only 2 means anything) to use package formats
             Setting('conda_pkg_format', cc_conda_build.get('pkg_format', None))
```
msarahan:
```suggestion
            Setting('conda_pkg_format', cc_conda_build.get('pkg_format', None)),
```

conda_build/config.py
```diff
@@ -187,6 +187,11 @@ def _get_default_settings():
...
+
+            # This allows build orchestrators to introspect and change variant spec parsing behavior.
+            # This is an advanced feature only intended when using conda-build as a library.
+            Setting("post_variant_combine_hook", None)
```
msarahan:
```suggestion
            Setting("post_variant_combine_hook", None),
```

msarahan:
Seems fine.  Couple of minor syntax things.  One error, one in the hopes of avoiding an error like this for the next setting.

mariusvniekerk:
To see how this is used see https://github.com/mariusvniekerk/conda-smithy/pull/2

mariusvniekerk:
@msarahan This should be ready now

PR #3531, opened 2019-06-06
Drop `/usr` from CDT skeleton path

jakirkham:
This seems to already exist in RPMs. So by default will result in the `usr` directory being created again within `usr`, which causes problems for CDTs and anything using them. This fixes that by dropping `usr` from the path.

cc @mingwandroid

### DIFF
--- a/conda_build/skeletons/rpm.py
+++ b/conda_build/skeletons/rpm.py
@@ -105,8 +105,8 @@ BUILDSH = """\
 #!/bin/bash
 
 RPM=$(find ${PWD}/binary -name "*.rpm")
-mkdir -p ${PREFIX}/x86_64-conda_cos6-linux-gnu/sysroot/usr
-pushd ${PREFIX}/x86_64-conda_cos6-linux-gnu/sysroot/usr > /dev/null 2>&1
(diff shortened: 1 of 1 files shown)


jakirkham:
Should add this matched with what I saw in other CDTs. For example, [this from GTK2]( https://github.com/AnacondaRecipes/aggregate/blob/141b0e84415ff359d81d7ea8c4438b216d2f2281/gtk2-devel-cos6-i686/build.sh#L4-L5 ).

mingwandroid:
Sounds fine to me. BTW do we have things using gtk2 cdts?

jakirkham:
Not for what I'm doing currently. Was looking at making some CDTs for networking. So was checking what I had against other common CDTs to figure out issues I was running into locally.

msarahan:
Thanks @jakirkham

jakirkham:
I've been thinking about this more as some RPMs expect to be installed in `/usr` and others don't. Haven't found a hard and fast rule here. Maybe this information is in the SPEC file? I'm not seeing it there, but I may be missing it.

mbargull:
This might be fixed with https://github.com/conda/conda-build/issues/3595#issuecomment-509281317.

PR #3532, opened 2019-06-07
Update cran skeleton to not use toolchain for win

dbast:
Update cran skeleton to use m2w64 compilers for windows instead of toolchain.

The linter is telling since long: Using toolchain directly in this manner is deprecated.

(Instead of updating the liner: https://github.com/conda-forge/conda-smithy/pull/1062)

### DIFF
--- a/conda_build/skeletons/cran.py
+++ b/conda_build/skeletons/cran.py
@@ -1197,12 +1197,18 @@ def skeletonize(in_packages, output_dir=".", output_suffix="", add_maintainer=No
                 if need_c:
                     deps.append("{indent}{{{{ compiler('c') }}}}      {sel}".format(
                         indent=INDENT, sel=sel_src_not_win))
+                    deps.append("{indent}{{{{ compiler('m2w64_c') }}}}{sel}".format(
+                        indent=INDENT, sel=sel_src_and_win))
                 if need_cxx:
                     deps.append("{indent}{{{{ compiler('cxx') }}}}    {sel}".format(
                         indent=INDENT, sel=sel_src_not_win))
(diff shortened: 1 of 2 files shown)


jdblischak:
@dbast Looks good! I tested it with `r-dqnr`.

Here's what the indenting looks like currently:

```
requirements:
  build:
    - {{ compiler('c') }}              # [not win]
    - {{ compiler('m2w64_c') }}        # [win]
    - {{ compiler('cxx') }}            # [not win]
    - {{ compiler('m2w64_cxx') }}      # [win]
    - {{posix}}filesystem        # [win]
    - {{posix}}make
    - {{posix}}sed               # [win]
    - {{posix}}coreutils         # [win]
    - {{posix}}zip               # [win]
```

I'm not sure what the best aesthetic option is to deal with the fact that `{{ compiler('m2w64_cxx') }}` is so long. An alternative would be to have only it's line be out of sync.

```
requirements:
  build:
    - {{ compiler('c') }}        # [not win]
    - {{ compiler('m2w64_c') }}  # [win]
    - {{ compiler('cxx') }}      # [not win]
    - {{ compiler('m2w64_cxx') }}  # [win]
    - {{posix}}filesystem        # [win]
    - {{posix}}make
    - {{posix}}sed               # [win]
    - {{posix}}coreutils         # [win]
    - {{posix}}zip               # [win]
```

jdblischak:
Pinging @mingwandroid and @katietz

dbast:
@jdblischak Thanks for testing. I don't mind the indention... Idea was to but the compilers into their own indention block instead of indenting everything based on `{{ compiler('m2w64_cxx') }}`. Maybe not caring for cxx and fortran indention is the better option.

msarahan:
Good fix.  Thanks @dbast!

PR #3533, opened 2019-06-07
fix package_has_file function for .conda format (use cph)

msarahan:

### DIFF
--- a/conda_build/index.py
+++ b/conda_build/index.py
@@ -11,8 +11,8 @@ import json
 from numbers import Number
 import os
 from os.path import abspath, basename, getmtime, getsize, isdir, isfile, join, splitext, dirname
-from shutil import move
 import subprocess
+import sys
 import tarfile
 from tempfile import gettempdir
 import time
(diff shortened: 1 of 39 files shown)


conda_build/index.py
```diff
@@ -1430,48 +1233,22 @@ def _add_extra_path(extra_paths, path):
         _add_extra_path(extra_paths, join(subdir_path, REPODATA_JSON_FN + '.bz2'))
         _add_extra_path(extra_paths, join(subdir_path, REPODATA_FROM_PKGS_JSON_FN))
         _add_extra_path(extra_paths, join(subdir_path, REPODATA_FROM_PKGS_JSON_FN + '.bz2'))
-        _add_extra_path(extra_paths, join(subdir_path, "repodata2.json"))
+        # _add_extra_path(extra_paths, join(subdir_path, "repodata2.json"))
```
soapy1: why is repodata2.json getting removed?

msarahan: It has never been used. We can bring it back when we do namespaces. I think enough about the repodata will change with sharding that this stuff would need major rework anyway.

soapy1:
Tried this out, oh my, so much faster! 🚀

msarahan:
This seems mostly good to go.  We finally have a fully passing test suite with conda 4.7.  On Windows, we still have some permissions errors.  I think these are related to problems we've been seeing lately elsewhere (https://github.com/conda-forge/conda-feedstock/pull/95)

I think we should hold off on the next release until those are solved.  I'll take a look tomorrow.

PR #3540, opened 2019-06-13
Added dependency on setuptools

duncanmmacleod:
This PR adds an explicit dependency on `setuptools`, since `pkg_resources` module is a module-level import. I have not added a news entry, since this is a packaging-only change.


msarahan:
Good catch, thanks.

PR #3548, opened 2019-06-18
Rewrites CDT template without outputs

jakirkham:
Follows from the discussion in PR ( https://github.com/conda-forge/staged-recipes/pull/8560 ).

As there is only one output with CDTs, there isn't a need to use `outputs`. To simplify things a bit, this drops `outputs` and updates the CDT recipe to work without it.


jakirkham:
Thanks @msarahan! 😄

Issue #3550, opened 2019-06-18
Adding tests to CDTs


jakirkham:
Currently the [CDT template]( https://github.com/conda/conda-build/blob/9354eb9087ab0086170519903333379d1cdea393/conda_build/skeletons/rpm.py#L32-L61 ) lacks tests. It would be nice to include tests. Though this may be tricky given the wide variety of CDTs that can be encountered. Am raising this to discuss how we would like to handle this.

cc @mingwandroid

jakirkham:
As a first step, submitted PR ( https://github.com/conda/conda-build/pull/3551 ) that simply tests the `sysroot` directory exists in the package.

jakirkham:
Let's leave this open. There's also a PR above. Though as noted in it, there is another related issue ( https://github.com/conda/conda-build/issues/3553 ), which may need to be addressed first

PR #3551, opened 2019-06-18
Test CDTs contain sysroot directory

jakirkham:
Partially addresses https://github.com/conda/conda-build/issues/3550

This provides a fairly simplistic test that all CDTs should pass.

cc @mingwandroid

### DIFF
--- a/conda_build/skeletons/rpm.py
+++ b/conda_build/skeletons/rpm.py
@@ -52,6 +52,9 @@ outputs:
     build:
       noarch: generic
 {depends_run}
+    test:
+      commands:
+        - test -d "${PREFIX}/x86_64-conda_cos6-linux-gnu/sysroot"
     about:
       home: {home}
       license: {license}


conda_build/skeletons/rpm.py
```diff
@@ -52,6 +52,9 @@
     build:
       noarch: generic
 {depends_run}
+    test:
+      commands:
+        - test -d "${PREFIX}/x86_64-conda_cos6-linux-gnu/sysroot"
```
msarahan: Ray can/should correct me if I'm wrong here, but I think CDTs are not tied strictly to that folder, and we can have them for Power, too, for example.  Maybe we need selectors here for expected behavior on each platform that uses CDTs?

jjhelmus: @msarahan you are correct this path is different on non-x86_64 platforms like aarch64 or ppc64le.

jakirkham: That seems like a reasonable concern to raise. However there is already [hard-coding happening in the `build.sh`]( https://github.com/conda/conda-build/blob/a7ac3db661a59db7bd04d315b4a8ee739812f52c/conda_build/skeletons/rpm.py#L63-L64 ). Perhaps the right course of action is to raise an issue and address this concern separately?

jakirkham:
FWIW we may want to hold on this until issue ( https://github.com/conda/conda-build/issues/3553 ) is resolved.

conda_build/skeletons/rpm.py
jakirkham: Note: This diff is outdated, but the concern raised is not currently addressed.

jakirkham: Raised issue ( https://github.com/conda/conda-build/issues/3556 ) to discuss how we should address this.

jakirkham:
Could someone please take another look at this or the linked issue? Thanks 🙏

PR #3569, opened 2019-06-25
Make lief an optional dependency

jjhelmus:
Make lief an optional dependency by using patchelf to set the RPATHs and only calling function which depend on lief when it is available.


jjhelmus:
Concerning the change in 06a3fff, it might be preferable to allow the user to specify which tool should be used to patch RPATHS, lief or patchelf via a configuration parameter.  A benefit of this is that both patchelf and lief would be optional dependencies.  When binaries are produced one or the other must be installed but there would be no need for both.

mingwandroid:
Please don't I am already working on this.

PR #3572, opened 2019-06-25
Added info about how conda build uses a recipe to create a pkg

rrigdon:
Also added anchors in Defining Metadata.

### DIFF
--- a/docs/source/concepts/recipe.rst
+++ b/docs/source/concepts/recipe.rst
@@ -111,12 +111,12 @@ which shows the matrix of things to build against. The
 For example, defining a conda_build_config.yaml of the form
 and filling it defines a matrix of 4 packages to build::
 
-   python:
-     - 3.7
-     - 2.7
-   vs:
-     - 2008
-     - 2017
+   foo:
+     - 1.0
+     - 2.0
+   bar:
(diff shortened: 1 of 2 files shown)


docs/source/concepts/recipe.rst
```diff
@@ -86,6 +92,115 @@ The `conda skeleton <skeleton_ref>` command can help to make
...
+``conda build config`` determines how many builds it has to do.
+For example, defining a conda_build_config.yaml of the form
+and filling it defines a matrix of 4 packages to build::
+
+   python:
+     - 3.7
+     - 2.7
+   vs:
+     - 2008
+     - 2017
```
msarahan: This isn't really the best example, because these specific keys are coupled in practice.  I think it's mostly fine, but it might cause a bit of confusion for people who know too much to be dangerous.

rrigdon: Is there a better example we can put in here instead?

msarahan: Maybe just a totally generic one:

```
foo:
  - 1.0
  - 2.0
bar:
  - 1.2.0
  - 1.4.0
```

PR #3580, opened 2019-06-28
Drop source RPMs (as they are unused currently)

jakirkham:
Also community members have requested having checksums for the source RPMs as well. While there is some rough idea as to how this might be done, it appears to be a fair bit of work that no one has volunteered to take on. To streamline usage a bit, the suggestion here is to leave source RPMs out of the skeleton until we have solved this checksum issue.

cc @mingwandroid

xref: https://github.com/conda/conda-build/issues/3552
xref: https://github.com/conda-forge/staged-recipes/pull/8633#discussion_r296949547
xref: https://github.com/conda-forge/conda-forge.github.io/issues/783

### DIFF
--- a/conda_build/skeletons/rpm.py
+++ b/conda_build/skeletons/rpm.py
@@ -38,8 +38,6 @@ source:
   - url: {rpmurl}
     {checksum_name}: {checksum}
     folder: binary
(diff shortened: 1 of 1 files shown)


jakirkham:
Could we merge this instead?

PR #3596, opened 2019-07-09
add in fallback for run_exports when channeldata not available

msarahan:
fixes #3583

PR #3598, opened 2019-07-09
make pins for current_repodata additive - always newest, and pins are…

msarahan:
… additions to that


msarahan:
pytest errors here are being handled in #3596

PR #3603, opened 2019-07-15
Get run_exports the slow way

duncanmmacleod:
`channeldata.json` only contains one `run_exports` entry for each version of each package; some packages (looking at you `fftw`) have multiple builds of the same version with different `run_exports`. This PR naively undoes some recent work to force recipe rendering to use the slow, but robust method of accessing `run_exports`.

I'm not sure if this PR is well-motivated, or if there is a better solution to the `run_exports` problem, so I haven't added news entries, but will do if this change is acceptable.

### DIFF
--- a/conda_build/render.py
+++ b/conda_build/render.py
@@ -357,16 +357,12 @@ def get_upstream_pins(m, actions, env):
     run_exports = {}
     empty_run_exports = False
     for pkg in linked_packages:
-        channeldata = utils.download_channeldata(pkg.channel)
-        if channeldata:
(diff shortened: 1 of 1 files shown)


msarahan:
I don't think that's the right solution.  We should fix channeldata.json - it sounds like it is broken.  Can you add a test with your fftw package and your desired output from channeldata.json?

msarahan:
I think this is fixed in #3611 - closing.  I think we may need the build string differentiator as discussed on gitter, though.

jjhelmus:
Variants are not fixed by #3611

jezdez:
Not stale

Issue #3539, opened 2019-06-11
Disagreement in get_linkages warning due to double "/"


beenje:
### Actual Behavior

When building a recipe with conda-build 3.18.2 the following warnings are raised:

```
WARNING: Disagreement in get_linkages(filename=/opt/conda/conda-bld/tk_1560252847753/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_/lib/tdbc1.1.0/libtdbc1.1.0.so, resolve_filenames=True, recurse=False, sysroot=/opt/conda/conda-bld/tk_1560252847753/_build_env/x86_64-conda_cos6-linux-gnu/sysroot/, envroot=/opt/conda/conda-bld/tk_1560252847753/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_, arch=native):
 lief: {'/opt/conda/conda-bld/tk_1560252847753/_build_env/x86_64-conda_cos6-linux-gnu/sysroot//lib/libc.so.6'}
pyldd: {'/opt/conda/conda-bld/tk_1560252847753/_build_env/x86_64-conda_cos6-linux-gnu/sysroot/lib/libc.so.6'}
  (using lief)
```

The only difference is a double `/` after `sysroot` for lief.

### Expected Behavior

The double `/` shouldn't be there. There shouldn't be any warning.
There was no warning with conda-build 3.17.8

### Steps to Reproduce

Build the https://github.com/conda-forge/tk-feedstock recipe with conda-build 3.18.2

##### `Output of conda info`

```
$ conda info

     active environment : base
    active env location : /opt/conda
            shell level : 1
       user config file : /home/conda/.condarc
 populated config files : /opt/conda/.condarc
          conda version : 4.6.14
    conda-build version : 3.18.2
         python version : 3.7.3.final.0
       base environment : /opt/conda  (writable)
           channel URLs : https://conda.anaconda.org/conda-forge/linux-64
                          https://conda.anaconda.org/conda-forge/noarch
                          https://repo.anaconda.com/pkgs/main/linux-64
                          https://repo.anaconda.com/pkgs/main/noarch
                          https://repo.anaconda.com/pkgs/free/linux-64
[...]

PR #3604, opened 2019-07-15
Fixes spurious post build errors/warning messages

ddamiani:
This pull requests fixes to two cases where warnings/error messages would be printed by the post build checks when nothing was wrong:

- The check on the difference between the results of get_rpath_raw and patchelf in post.py would always fail because a list (get_rpath_raw) was being compared to a str (output from the subproccess call to pathelf). The first one is a fix for issue #3520 

- get_linkages sometimes throws a warning because the function ```inspect_linkages_lief``` doesn't strip trailing path separators from sysroot, but ```inspect_linkages_pyldd``` does. In cases were sysroot has a trailing separator the comparison of the results fail because some of the results from ```inspect_linkages_lief``` would have double separators. This one is a fix for #3539


msarahan:
Thank you!

PR #3620, opened 2019-07-19
Extracted the part responsible to deal with entry points in get_package_metadata

marcelotrevisani:
Refactoring of the get_package_metadata. Extract the part responsible to deal with the entry points.

The idea is to do a bigger refactoring on this part. But I particularly prefer to open merge request in small chunks so in that way would be easy and better to review the modifications and to spot new improvements.

### DIFF
--- a/conda_build/skeletons/pypi.py
+++ b/conda_build/skeletons/pypi.py
@@ -702,62 +702,13 @@ def get_package_metadata(package, d, data, output_dir, python_version, all_extra
                           setup_options=setup_options,
                           config=config)
 
-    setuptools_run = False
-
-    # Look at the entry_points and construct console_script and
-    #  gui_scripts entry_points for conda
-    entry_points = pkginfo.get('entry_points', [])
-    if entry_points:
-        if isinstance(entry_points, str):
-            # makes sure it is left-shifted
-            newstr = "\n".join(x.strip()
-                                for x in entry_points.splitlines())
-            _config = configparser.ConfigParser()
-            entry_points = {}
-            try:
-                _config.readfp(StringIO(newstr))
-            except Exception as err:
-                print("WARNING: entry-points not understood: ",
-                        err)
-                print("The string was", newstr)
-                entry_points = pkginfo['entry_points']
-            else:
-                setuptools_run = True
-                for section in _config.sections():
-                    if section in ['console_scripts', 'gui_scripts']:
-                        value = ['%s=%s' % (option, _config.get(section, option))
-                                    for option in _config.options(section)]
-                        entry_points[section] = value
-        if not isinstance(entry_points, dict):
-            print("WARNING: Could not add entry points. They were:")
-            print(entry_points)
-        else:
-            cs = entry_points.get('console_scripts', [])
-            gs = entry_points.get('gui_scripts', [])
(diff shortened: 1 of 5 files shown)


marcelotrevisani:
I am going to create the changelog for this as well.

conda_build/skeletons/pypi.py
```diff
@@ -829,25 +780,23 @@ def get_package_metadata(package, d, data, output_dir, python_version, all_extra
...
         elif data and 'license' in data:
             license_name = data['license']
         else:
             license_name = None
+
         if license_name:
-            if noprompt:
-                pass
-            elif '\n' not in license_name:
+            if not noprompt and '\n' not in license_name:
```
marcelotrevisani: I am going to revert this part

conda_build/skeletons/pypi.py
```diff
@@ -856,23 +805,109 @@ def get_package_metadata(package, d, data, output_dir, python_version, all_extra
...
+
+    if isinstance(entry_points, str):
+        # makes sure it is left-shifted
+        newstr = "\n".join(x.strip() for x in entry_points.splitlines())
+        _config = configparser.ConfigParser()
+
+        try:
+            # TODO: Should we use read_file instead of readfp?
+            #  Because readfp is deprecated
+            _config.readfp(StringIO(newstr))
```
marcelotrevisani: Should we replace the ``readfp`` by ``read_file`` ?

tests/test_api_skeleton.py
```diff
@@ -22,7 +24,9 @@
 
 repo_packages = [('', 'pypi', 'pip', '8.1.2'),
                  ('r', 'cran', 'acs', ''),
-                 ('r', 'cran', 'https://github.com/twitter/AnomalyDetection.git', ''),
+                 (
+                 'r', 'cran', 'https://github.com/twitter/AnomalyDetection.git',
+                 ''),
```
marcelotrevisani: Sorry, some modifications regarding the style were because of the linter which I am using. I am going to revert it

ocefpaf:
@msarahan and @jjhelmus you probably already interacted with @marcelotrevisani via conda-forge. He is interested in helping us create a unified and consistent way to regenerated recipes using the skeleton. This PR is only a start.

conda_build/skeletons/pypi.py
msarahan: Maybe worth trying, but I'm not sure what the knock-on effects might be.  Give it a try, but don't work too hard on this little detail.

marcelotrevisani: Sure! I might open another PR for that to see if it will impact.
This concern is just because I saw that ``readfp`` will be removed in the future from ``ConfigParser``

marcelotrevisani:
There are some tests failing but they do not seem to be related to this modification
These tests are failing:
``test_cmake_generator`` and  ``test_render_with_python_arg_CLI_reduces_subspace`` 

Is that right?

marcelotrevisani:
I just checked the master branch and it is failing the same tests.

You can merge this PR whenever you free comfortable to integrate it! :)

jjhelmus:
Tests failures are not related to this change.  Thanks for the addition @marcelotrevisani

Issue #3597, opened 2019-07-09
Error building package on Windows


Bezier89:
### Actual Behavior

The following error is produced (note that it happens rather early in the process, within a few seconds) when building any conda package. I've attached a [trivial recipe](https://github.com/conda/conda-build/files/3374762/conda.recipe.zip) where this still reproduces.

```
Traceback (most recent call last):
  File "[CONDA_ROOT]\lib\threading.py", line 916, in _bootstrap_inner
    self.run()
  File "[CONDA_ROOT]\lib\threading.py", line 864, in run
    self._target(*self._args, **self._kwargs)
  File "[CONDA_ROOT]\lib\concurrent\futures\process.py", line 270, in _queue_management_worker
    ready = wait([reader] + sentinels)
  File "[CONDA_ROOT]\lib\multiprocessing\connection.py", line 859, in wait
    ready_handles = _exhaustive_wait(waithandle_to_obj.keys(), timeout)
  File "[CONDA_ROOT]\lib\multiprocessing\connection.py", line 791, in _exhaustive_wait
    res = _winapi.WaitForMultipleObjects(L, False, timeout)
ValueError: need at most 63 handles, got a sequence of length 64
```

### Expected Behavior

No error when building conda package.

### Steps to Reproduce

Build any conda package. This doesn't seem to reproduce on every Windows machine. The common aspect among machines where I've seen it reproduce is that they're all Windows Server 2012 R2 machines. It does not reproduce on the Windows Server 2008 R2 or Windows Server 2016 machines I've tested, but there's possibly other aspects at play (e.g. number of cores on the machine).

##### `Output of conda info`

```
     active environment : None
          conda version : 4.7.5
    conda-build version : 3.18.5
         python version : 3.6.8.final.0
       virtual packages :
               platform : win-64
             user-agent : conda/4.7.5 requests/2.22.0 CPython/3.6.8 Windows/2012ServerR2 Windows/6.3.9600
          administrator : True
             netrc file : None
           offline mode : False
```

Bezier89:
Note: This issue is only present in conda-build 3.18. I've confirmed that the issue is resolved if I downgrade to conda-build 3.17.8.

cometchaser:
I observe the same error #3576

Bezier89:
@msarahan I was hoping https://github.com/conda/conda-build/commit/cd20cc6194cf07fb6316b971764e66e4f7a4e248 would fix this but the issue is still present in conda-build 3.18.7

[...]

PR #3622, opened 2019-07-19
Limit threads to 61 on Windows

Bezier89:
Fixes https://github.com/conda/conda-build/issues/3597

We could remove MAX_THREADS_DEFAULT and just use the default constructor for ProcessPoolExecutor if we were always on Python 3.7.4+, but I assume we want this to also work on older python versions :)


jjhelmus:
LGTM, thanks for the fix @Bezier89

PR #3632, opened 2019-07-25
Workaround for future deprecations (SafeConfigParser)

marcelotrevisani:

### DIFF
--- a/bdist_conda.py
+++ b/bdist_conda.py
@@ -204,7 +204,12 @@ class bdist_conda(install):
                 c = configparser.ConfigParser()
                 entry_points = {}
                 try:
-                    c.readfp(StringIO(newstr))
+                    import six
+
+                    if six.PY2:
+                        c.readfp(StringIO(newstr))
+                    else:
+                        c.read_file(StringIO(newstr))
                 except Exception as err:
                     # This seems to be the best error here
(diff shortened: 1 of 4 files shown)


bdist_conda.py
```diff
@@ -204,7 +204,12 @@ def run(self):
                 c = configparser.ConfigParser()
                 entry_points = {}
                 try:
-                    c.readfp(StringIO(newstr))
+                    import six
+
+                    if six.PY2:
```
ocefpaf: @msarahan is there a roadmap to drop Python 2 support?

marcelotrevisani: That would be good. Specially because a lot of packages already deprecated Python 2.7, such as ``PyTest``, ``Tensorflow``, ``scikit-learn``, ``numpy``, ``tornado``, ``scipy``, ``cython`` and so on
https://python3statement.org/

msarahan: I'll bring it up with the team. I have no idea what customer expectations might be. If conda build didn't need to live in the same env as conda, it would be easier to say that conda build can switch right away.

ocefpaf: If conda will be Python 2 for a while longer we cannot escape maintaining conda-build for Python 2 as well. Is this good to merge then?

jjhelmus:
LGTM

PR #3634, opened 2019-07-25
Actually fix https://github.com/conda/conda-build/issues/3597

Bezier89:
Looks like 60 was the magic number in this case. Was basing it off the python change which uses 61. Not sure where the extra thread comes from. Rather than trying to get this exact though, I figure it's better to have plenty of room to spare.


jjhelmus:
LGTM.

PR #3649, opened 2019-08-05
Unbreak skeleton cran about m2w64_... vs toolchain

katietz:

### DIFF
--- a/conda_build/skeletons/cran.py
+++ b/conda_build/skeletons/cran.py
@@ -1198,7 +1198,7 @@ def skeletonize(in_packages, output_dir=".", output_suffix="", add_maintainer=No
                 # m2w64_... recipes aren't imported from conda-forge
                 # so use old toolchain variant on windows
                 if need_c or need_cxx or need_f:
-                    deps.append("{indent}{{{{ native }}}}toolchain      {sel}".format(
+                    deps.append("{indent}{{{{native}}}}toolchain      {sel}".format(
                         indent=INDENT, sel=sel_src_and_win))
                 if need_cxx:
                     deps.append("{indent}{{{{ compiler('cxx') }}}}          {sel}".format(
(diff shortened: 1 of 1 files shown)


mingwandroid:
This needs to be discussed wirth @conda-forge/r

mingwandroid:
I thought you didn't want all compilers to get installed when only C/C++/Fortran compilers are needed? I think that's the point of this change.

katietz:
Well, the point is is less that we want to be more selective on required compiler parts.  The real issue is that current skeleton is broken on anconda.org.  And even by adding conda-forge channel, it doesn't resolve the problem.  So we need to fix the stuff for our users of anaconda.  Creating/important the required m2w64_ recipes to aggregate, and then we can change skeleton.

dbast:
Tagging @conda-forge/r is not possible here. So tagging @bgruening @jdblischak @johanneskoester @daler @cbrueffer @dpryan79 @ocefpaf

dbast:
This is basically reverting https://github.com/conda/conda-build/pull/3532

The conda-forge recipe linter does not like this ... hm

dbast:
What else could we do here? Maybe introduce a cli parameter to specify if the to be created recipe is targeting anaconda or conda-forge?

katietz:
We resolved that issze by importing toolchain met-package into Anconda.  By
this we can use it as is.

jdblischak:
> Maybe introduce a cli parameter to specify if the to be created recipe is targeting anaconda or conda-forge?

Long-term it would be nice to have a flag, e.g. `--conda-forge`, that would allow us to retire [conda_r_skeleton_helper](https://github.com/bgruening/conda_r_skeleton_helper) and make it easier for new contributors to submit R recipes to staged-recipes.

katietz:
Yeah, might be a good idea.  I noticed that we lack right now for R
skeleton to detect the need of llvm-openmp >=4.0.1 dependency for mac. On
linux it is always there, and for Windows it is a no-brainer too.  Not sure
if this actually a big deal, but finding a solution here would be indeed
helpful.

PR #3653, opened 2019-08-07
Install six when testing flake8

marcelotrevisani:
Six is missing on Travis when testing flake8.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,7 +65,7 @@ install:
   - if [[ "$TRAVIS_PYTHON_VERSION" == "2.7" ]]; then
       conda install -q futures scandir;
     fi
-  - conda install -q anaconda-client requests filelock contextlib2 jinja2 patchelf ripgrep pyflakes beautifulsoup4 chardet pycrypto glob2 psutil pytz tqdm conda-package-handling py-lief
+  - conda install -q anaconda-client six requests filelock contextlib2 jinja2 patchelf ripgrep pyflakes beautifulsoup4 chardet pycrypto glob2 psutil pytz tqdm conda-package-handling py-lief
   - pip install pkginfo
   - if [[ "$FLAKE8" == "true" ]]; then
       conda install -q flake8;
--- a/ci/travis/run.sh
+++ b/ci/travis/run.sh
@@ -8,7 +8,7 @@ if [[ "$FLAKE8" == "true" ]]; then
     conda build --help
     conda build --version
     conda build conda.recipe --no-anaconda-upload
-    conda create -n _cbtest conda-build glob2
+    conda create -n _cbtest conda-build glob2 six
     # because this is a file, conda is not going to process any of its dependencies.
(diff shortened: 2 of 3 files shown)


marcelotrevisani:
For some reason it is not installing ``six`` and because of that the build phase is failing.
https://travis-ci.org/conda/conda-build/jobs/568730356#L1239

I already added to the ``conda.recipe`` the ``six`` package.

it needs further investigation.
@msarahan , do you have any hints here?

msarahan:
That import is in the build itself, so six in the host section in conda.recipe _should_ take care of it...  I'll take more of a look today.

msarahan:
There is no host section: https://github.com/conda/conda-build/blob/master/conda.recipe/meta.yaml#L22

That's technically OK (it won't work for cross-compiling situations, but we're not doing that with this recipe).

What's not OK is that six is in run requirements, but it needs to also be in the build requirements, as it is both a build time requirement (for versioneer, apparently) and a run time requirement (used in conda-build a fair bit).

marcelotrevisani:
> There is no host section: https://github.com/conda/conda-build/blob/master/conda.recipe/meta.yaml#L22
> 
> That's technically OK (it won't work for cross-compiling situations, but we're not doing that with this recipe).
> 
> What's not OK is that six is in run requirements, but it needs to also be in the build requirements, as it is both a build time requirement (for versioneer, apparently) and a run time requirement (used in conda-build a fair bit).

I added ``six`` to the ``build`` requirements
https://github.com/conda/conda-build/pull/3653/files#diff-bd134fa847b11c75aba174f834eb80dfR26
But it is failing, in theory it should work fine with that, right? 

I also tried to replace the ``build`` by ``host`` but it is also failing.

Regarding ``six`` in ``run`` requirements, I used ``six`` in one of my merge requests because I thought that it was a depdnency already. However, I can replace it to avoid this dependency. I will open a PR regarding it.

marcelotrevisani:
Despite my modifications, ``six`` is still being in use in some places, such as:
https://github.com/conda/conda-build/blob/a25e927f7e5b305bddc620ded1d0f548bfac1631/conda_build/variants.py#L422
https://github.com/conda/conda-build/blob/a25e927f7e5b305bddc620ded1d0f548bfac1631/conda_build/skeletons/pypi.py#L226
https://github.com/conda/conda-build/blob/a25e927f7e5b305bddc620ded1d0f548bfac1631/conda_build/skeletons/pypi.py#L336

msarahan:
I don't think removing six usage is the right approach.  Six is fine to use, but something is going wrong that makes it not importable.  I will try to duplicate the error on my end.

msarahan:
I took another look at this.  I'm totally lost.  I added ``conda render conda.recipe`` to see what the rendering is coming up with.  From what I saw there, six is in every section.  It seems like there's a build in conda-build that is causing this, not a problem with the recipe.

msarahan:
Closing - @marcelotrevisani has a better fix in #3667

Issue #532, opened 2015-08-07
Allow to specify channels in a recipe


asmeurer:
like

```
channels:
  - asmeurer
```

Not to be confused with specifying channels on a package, which is https://github.com/conda/conda/issues/988.

asmeurer:
The problem with this is that the channels would only be added at build time, so it's only useful if you need them for your build dependencies. For the run dependencies, your users will still need to add those channels manually.

SlaterByte:
+1
I'd like to be able to send my whole package to a build machine and not have to set special channels. Most of my conda builds are agnostic scripts, except when it comes to channels that are required for build.

yohannLelievre:
@asmeurer 
Hi, I encounter exactly the same issue while I try to specify in my meta.yaml a dependency toward the perl package in dan_blanchard channel. Indeed, I'm no more able to load the perl package from the default channel...
What is the current state of this issue? Is there a way to get around it?

yohannLelievre:
I finally found a way to get around this situation.

When I build my package I use the following command :
`conda build myPackage -c theDependentChannel`

And when I install my package in a virtual environment :
`conda install myPackage -c myChannel -c theDependentChannel`

It's far from being clean... But it works.

alstuder:
@yohannLelievre 
Does not work for me. Whatever is found in default channel is takes precedence. Is your dependent package in the theDependentChannel also available in the default channel?

yohannLelievre:
@alstuder
I've made some tests with the following command:
`conda build myPackage -c theDependentChannel1 -c theDependentChannel2`
And, during the packaging, the precedence of my channels are respected for a package hosted in both channels. Here, the dependent packages from "myPackage" are first linked from "theDependentChannel1", if possible, and then from "theDependentChannel2".

When I executed the command below:
`conda build myPackage -c theDependentChannel2 -c theDependentChannel1`
I observed the expected change of precedence toward the channel "theDependentChannel2".

When I executed this command:
`conda build myPackage -c conda-forge`
With as dependent package "readline" (version 6.2), I linked readline from conda-forge.

And when I finally executed this command:
`conda build myPackage -c defaults -c conda-forge`
"readline" was linked from the defaults channel.

So maybe your issue is coming from your conda or conda-build version?
My conda version is the 4.3.15 and my conda-build version is the 2.1.8.

martin-sicho:
[...]

PR #3656, opened 2019-08-07
[WIP] Allow to specify channels in a recipe

mmathar:
Adds support for channels to meta.yaml recipes as discussed in #532 . This is achieved by adding the channels specified in the meta.yaml file to those given by the -c argument.

Example usage:
```
package:
  name: test
  version: "0.1"

requirements:
  channels:
    - anaconda
    - cdat-forge
  host:
    - cdat-forge::pylint
    - python=3.6
```
This will (in this case) install pylint from cdat-forge and all other dependencies (including python) from anaconda. This is effectively the same as calling
`conda build -c anaconda -c cdat-forge .`
on the same meta.yaml minus the channels section.

I noticed some failures in unrelated (?) tests locally. I'll take another look at this when the tests for this PR complete.

### DIFF
--- a/conda_build/metadata.py
+++ b/conda_build/metadata.py
@@ -23,7 +23,7 @@ from .conda_interface import string_types
 from conda_build import exceptions, utils, variants, environ
 from conda_build.conda_interface import memoized
 from conda_build.features import feature_list
-from conda_build.config import Config, get_or_merge_config
+from conda_build.config import Config, get_or_merge_config, get_channel_urls
 from conda_build.utils import (ensure_list, find_recipe, expand_globs, get_installed_packages,
                                HashableDict, insert_variant_versions)
 from conda_build.license_family import ensure_valid_license_family
@@ -390,6 +390,7 @@ default_structs = {
     'build/force_ignore_keys': list,
     'build/merge_build_host': bool,
     'build/msvc_compiler': text_type,
+    'requirements/channels': list,
     'requirements/build': list,
     'requirements/host': list,
     'requirements/run': list,
@@ -505,7 +506,7 @@ FIELDS = {
               },
     'outputs': {'name', 'version', 'number', 'script', 'script_interpreter', 'build',
                 'requirements', 'test', 'about', 'files', 'type'},
-    'requirements': {'build', 'host', 'run', 'conflicts', 'run_constrained'},
+    'requirements': {'requirements', 'build', 'host', 'run', 'conflicts', 'run_constrained'},
     'app': {'entry', 'icon', 'summary', 'type', 'cli_opts',
             'own_environment'},
     'test': {'requires', 'commands', 'files', 'imports', 'source_files', 'downstreams'},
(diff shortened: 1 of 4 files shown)


msarahan:
Thanks @mmathar!

We have generally avoided this because it limits the generality of recipes.  Someone may want to use their own pylint in your example, but your recipe will preclude it (without recipe modification, anyway).  I'm kind of inclined to say that channels might be allowable with exact specs, but at the same time, that's really the wrong direction for recipes to head.  Overspecifying things is painful and fragile.  I'm not turning down this PR, but we should have a discussion with the other conda-build maintainers.  Let's start here in the comments, but if we need higher bandwidth, we can schedule a call/video thing.

mmathar:
I'm not sure specifying a channels section in a recipe limits generality of recipes. As it is implemented in this PR any channel passed by -c will be preferred over the channels in the recipe (assuming no per-package channel specification). @msarahan 

On an unrelated note, it occurred to me it may be more transparent to only use the recipe's channels if the user does not specify any. Otherwise, some user specified channels and some recipe channels might be used which might lead to confusion.

> Someone may want to use their own pylint in your example, but your recipe will preclude it (without recipe modification, anyway). 

This may be wrong but isn't this the case already? Adding `cdat-forge::pylint` to a recipe basically forces the user to call conda-build with `-c cdat-forge` or the build will fail. As I understand the problem seems to be the channel specification on a per-package basis (which is already a thing)?

msarahan:
> As it is implemented in this PR any channel passed by -c will be preferred over the channels in the recipe (assuming no per-package channel specification)

You're sort of right.  There's two things here: the channel list, which affects specs that have no channel (this is where you're right), and specs with channels as part of them (this is where you're wrong - channels in meta.yaml will not override the channel in the spec.)

As implemented, the recipe-specified channels are the lowest priority.  That's going to be good in some ways and bad in others, especially when strict channel priority is in use.  I think you probably need the ability to both prepend and append channels.  Perhaps:

```
requirements:
  channels:
    prepend:
      - foo
    append:
      - bar
```

>  Adding cdat-forge::pylint to a recipe basically forces the user to call conda-build with -c cdat-forge or the build will fail

This is because you need some dependencies from cdat-forge.  If you had dependencies that could otherwise be satisfied, cdat-forge::pylint would be enough to use that channel for that package (but no dependencies)

mmathar:
> You're sort of right. There's two things here: the channel list, which affects specs that have no channel (this is where you're right), and specs with channels as part of them (this is where you're wrong - channels in meta.yaml will not override the channel in the spec.)

Just to make sure I understand this correctly: the channel list is passed by -c and specs with channels as part of them are e.g. channel::package? So the channels in the spec are meant to extend the channel list? I've thought of it more as the channel in the spec selecting a channel from the channel list.  I'm a bit confused now.

> I think you probably need the ability to both prepend and append channels.

Prepend and append to the channel list or do you mean something different?

> This is because you need some dependencies from cdat-forge. If you had dependencies that could otherwise be satisfied, cdat-forge::pylint would be enough to use that channel for that package (but no dependencies)

I'm not sure I understand this correctly. An example for illustration:
```
requirements:
  build:
    - intel::intel-openmp
```
intel-openmp is the only requirement and according to `conda search intel-openmp --info` does not have any dependencies. Is it expected that `conda build` works? Because it does not. Only when `-c intel` is added the package is found.

PR #3657, opened 2019-08-08
Fix some tests on windows

marcelotrevisani:
Fixed some tests on windows and removed one parametrization because it was failing almost every time.

Skipped one parametrization on ``test_cmake_generator`` for windows and python 3.5

Updated the subspace tests which were using python 3.5, now it is using python 3.6


msarahan:
Looks great if we had just a bit more CI time or some better speed.  We'll get there...

PR #3663, opened 2019-08-12
Split tests on CI

marcelotrevisani:
Added slow and sanity marks to tests. Some tests are taking too much time to finish and because of that, the CI is failing due to timeout.

The goal of this PR is to separate the tests in three jobs, each job will execute one "type" of tests such as ``slow`` tests, ``sanity`` tests and another one to execute the rest of them

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -10,16 +10,49 @@ environment:
       SYS_PYTHON_VERSION: "3.7"
       SYS_PYTHON_ARCH: "64"
       CONDA_CANARY: "False"
+      SLOW_TESTS: "True"
+
+    - SYS_PYTHON: "C:\\Miniconda37-x64"
+      SYS_PYTHON_VERSION: "3.7"
+      SYS_PYTHON_ARCH: "64"
+      CONDA_CANARY: "False"
+      SANITY: "True"
+
+    - SYS_PYTHON: "C:\\Miniconda37-x64"
+      SYS_PYTHON_VERSION: "3.7"
+      SYS_PYTHON_ARCH: "64"
+      CONDA_CANARY: "False"
+      SLOW_TESTS: "False"
 
     - SYS_PYTHON: "C:\\Miniconda-x64"
       SYS_PYTHON_VERSION: "2.7"
       SYS_PYTHON_ARCH: "64"
       CONDA_CANARY: "False"
+      SANITY: "True"
+
+    - SYS_PYTHON: "C:\\Miniconda-x64"
+      SYS_PYTHON_VERSION: "2.7"
+      SYS_PYTHON_ARCH: "64"
+      CONDA_CANARY: "False"
+      SLOW_TESTS: "False"
+
+    - SYS_PYTHON: "C:\\Miniconda-x64"
+      SYS_PYTHON_VERSION: "2.7"
+      SYS_PYTHON_ARCH: "64"
+      CONDA_CANARY: "False"
+      SLOW_TESTS: "True"
 
     - SYS_PYTHON: "C:\\Miniconda37-x64"
       SYS_PYTHON_VERSION: "3.7"
       SYS_PYTHON_ARCH: "64"
       CONDA_CANARY: "True"
+      SLOW_TESTS: "False"
+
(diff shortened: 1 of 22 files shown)


marcelotrevisani:
There are still some race conditions...
It needs more investigation

marcelotrevisani:
Just the test ``test_recipe_builds`` is taking almost 50min to finish...
Should we divide the tests in ``Recipe builds``, ``Slow`` and the rest?

marcelotrevisani:
Those ``test_cran_license`` are quite unstable.

marcelotrevisani:
Nice! Appveyor is passing... I discovered that we need to fix one of the tests and how the directories are handled in the tests. But I am going to do it in a second step.

appveyor.yml
```diff
@@ -105,5 +148,5 @@ test_script:
 #   - appveyor PushArtifact cbtmp.7z
 
 on_success:
-  - pip install codecov
+  - conda install -c conda-forge codecov -y
```
marcelotrevisani: Installing ``codecov`` from conda-forge to avoid any kind of problem regarding permissions when installing it using ``pip`` on appveyor

ci/travis/run.sh
```diff
@@ -26,11 +26,25 @@ else
...
+
+      if [[ "$SANITY" == "true" ]]; then
+          pip install git+https://github.com/conda/conda-verify.git
+          /opt/conda/bin/py.test -v -n auto --basetemp /tmp/cb --cov conda_build --cov-append --cov-report xml -m "sanity and not slow and not serial" tests
+          /opt/conda/bin/py.test -v -n 0 --basetemp /tmp/cb_serial --cov conda_build --cov-append --cov-report xml -m "sanity and not slow and serial" tests
+      else
+          /opt/conda/bin/py.test -v -n auto --basetemp /tmp/cb --cov conda_build --cov-append --cov-report xml -m "not serial $SLOW_MARK and not sanity" tests
+          # install conda-verify from its master branch, at least for a while until it's more stable
+          pip install git+https://github.com/conda/conda-verify.git
+          if [[ $SLOW_MARK == "and not slow" ]]; then
```
marcelotrevisani: This is a workaround because ``pytest`` will exit with code ``5``if no tests are selected.
As there is no test with markers ``slow`` and ``serial` this script will return exit code ``5`` and it will fail on CI

tests/utils.py
```diff
@@ -8,9 +8,16 @@
...
 from conda_build.metadata import MetaData
-from conda_build.utils import on_win, FileNotFoundError
+from conda_build.utils import on_win
 
-thisdir = os.path.dirname(os.path.realpath(__file__))
+
+def get_root_dir():
+    import conda_build
+    conda_build_dir = os.path.realpath(os.path.dirname(conda_build.__file__))
+    return os.path.abspath(os.path.join(conda_build_dir, ".."))
```
marcelotrevisani: temporary fix, that will be replaced in a second step by fixtures to handle it properly

marcelotrevisani:
I squashed the whole commits because I did a lot of tests to achieve this, haha

appveyor.yml
ocefpaf: @jjhelmus maybe `codecov` should be in `defaults` :wink:

mingwandroid:
meh.

marcelotrevisani:
> meh.

Yeah, this is not ideal indeed! :(

It is a workaround to be able to run the whole tests on travis and appveyor. We might split part of those tests to run on azure as well, in that way it will speed up a bit the testing phase.

If we had a longer timeout on travis/appveyor (something like 3h) that would not be necessary. As we are limited by 50min on travis and 1h on appveyor that seems to be one solution for now.

However, if you have a better solution ou suggestion please feel free to share and I am going to try to implement it! :)

appveyor.yml
jjhelmus: Building this out now

jjhelmus:
I agree @mingwandroid and @marcelotrevisani, splitting tests into groups to get them to fit into the CI time is not great but I don't see a better method without major work.  This improves immediately improves the tests so I'm in favor.

Issue #994, opened 2016-06-03
conda skeleton cpan --recursive Test::Deep


jerowe:
Error:

```
Using url https://cpan.metacpan.org/authors/id/R/RJ/RJBS/Test-Deep-1.120.tar.gz (38973) for Test-Deep.
Processing dependencies for Test-Deep.....Using Anaconda Cloud api site https://api.anaconda.org
Fetching package metadata: ....
.An unexpected error has occurred, please consider sending the
following traceback to the conda GitHub issue tracker at:


https://github.com/conda/conda-build/issues


Include the output of the command 'conda info' in your report.

Traceback (most recent call last):
  File "/home/jillian/.local/easybuild/software/anaconda/2-4.0.0/bin/conda-skeleton", line 5, in <module>
    sys.exit(main())
  File "/home/jillian/.local/easybuild/software/anaconda/2-4.0.0/lib/python2.7/site-packages/conda_build/main_skeleton.py", line 298, in main
    args_func(args, p)
  File "/home/jillian/.local/easybuild/software/anaconda/2-4.0.0/lib/python2.7/site-packages/conda_build/main_build.py", line 493, in args_func
    args.func(args, p)
  File "/home/jillian/.local/easybuild/software/anaconda/2-4.0.0/lib/python2.7/site-packages/conda_build/main_skeleton.py", line 315, in execute
    cpan.main(args, parser)
  File "/home/jillian/.local/easybuild/software/anaconda/2-4.0.0/lib/python2.7/site-packages/conda_build/cpan.py", line 249, in main
    processed_packages)
  File "/home/jillian/.local/easybuild/software/anaconda/2-4.0.0/lib/python2.7/site-packages/conda_build/cpan.py", line 407, in deps_for_package
    dep_version, perl_version, dependency=True)
  File "/home/jillian/.local/easybuild/software/anaconda/2-4.0.0/lib/python2.7/site-packages/conda_build/cpan.py", line 539, in get_release_info
    core_version = core_module_version(orig_package, perl_version)
  File "/home/jillian/.local/easybuild/software/anaconda/2-4.0.0/lib/python2.7/site-packages/conda/utils.py", line 118, in **call**
    value = self.func(_args, *_kw)
  File "/home/jillian/.local/easybuild/software/anaconda/2-4.0.0/lib/python2.7/site-packages/conda_build/cpan.py", line 340, in core_module_version
    if LooseVersion(first_version) > LooseVersion(version):
[...]

PR #3671, opened 2019-08-13
Revert to using corelist and LooseVersion

### LINKED ISSUE #3661: conda skelton cpan YAML::XS incorrectly identified as core module
### Actual Behavior

```
(base) spa309:conda/spa-4/test_master % conda skeleton cpan YAML::XS
We found core module perl-yaml-xs. Skipping recipe creation.        
--dirty flag and --keep-old-work not specified. Removing build/test folder after successful build/test.

INFO:conda_build.config:--dirty flag and --keep-old-work not specified. Removing build/test folder after successful build/test.

```

### Expected Behavior

Should create a recipe for `perl-yaml-libyaml` because that's the distribution containing `YAML::XS`

### Steps to Reproduce

From a freshly installed miniconda with `conda-build` installed
```
conda skelton cpan YAML::XS
```

##### `Output of conda info`

```

     active environment : base
    active env location : /home/ubuntu/miniconda3
            shell level : 1
       user config file : /home/ubuntu/.condarc
 populated config files : 
          conda version : 4.7.11
    conda-build version : 3.18.9
         python version : 3.7.3.final.0
       virtual packages : 
       base environment : /home/ubuntu/miniconda3  (writable)
          package cache : /home/ubuntu/miniconda3/pkgs
                          /home/ubuntu/.conda/pkgs
       envs directories : /home/ubuntu/miniconda3/envs
                          /home/ubuntu/.conda/envs
               platform : linux-64
             user-agent : conda/4.7.11 requests/2.22.0 CPython/3.7.3 Linux/4.15.0-1044-aws ubuntu/18.04.1 glibc/2.27
                UID:GID : 1000:1000
             netrc file : None
           offline mode : False
```

timsnyder:
corelist is the cannonical method for determining whether
a perl module is included in the perl core library. The
metacpan community does not have a method for their API
to convey whether a module is in the core or not (
see metacpan/metacpan-api#925).  Until metacpan can
tell us whether a module is in the core, revert back to
using the corelist script that ships with perl.

Also revert to using distutils.version.LooseVersion for
version comparisons.  LooseVersion implements versioning
that is free-form like Perl and the pkg_resources.parse_version
will incorrectly sort versions compared that align with
PEP-440 against those that don't.  (Reading the code
pkg_resources assumes that all versions that do not comply
with PEP-440 occur before those that do. Since nothing in
the Perl world will intentionally comply with PEP-440
using pkg_resources will lead to confusing results.

When moving back to LooseVersion, we must fix issues like
the one reported in #994:
  TypeError: expected string or buffer
and it's Python3 equivalent that uses bytes instead of buffer.
This TypeError is encountered when we pass an int into
LooseVersion instead of a string or bytes.  Metacpan has
an open issue (metacpan/metacpan-api#789) that reports
inconsistancy in the returned type of versions in the json.

To address this, we now wrap distutils.version.LooseVersion
in our own LooseVersion helper that coerces the supplied
version parameter into a string.  We also strip any leading
'v' to avoid TypeError about comparing int against str.

fixes #3661

### DIFF
--- a/conda_build/skeletons/cpan.py
+++ b/conda_build/skeletons/cpan.py
@@ -176,6 +176,7 @@ class PerlTmpDownload(TmpDownload):
 
             return dst
 
+
 def LooseVersion(version):
     """
     Wrap distutils.version.LooseVersion, coercing passed version into str
@@ -314,7 +315,7 @@ def skeletonize(packages, output_dir=".", version=None,
 
         # Add Perl version to core module requirements, since these are empty
         # packages, unless we're newer than what's in core
-        if (core_version is not None and 
+        if (core_version is not None and
            ((version in [None, '']) or (core_version >= LooseVersion(version)))):
             if not write_core:
                 print('We found core module %s. Skipping recipe creation.' %
@@ -490,7 +491,6 @@ def core_module_version(module, version, config):
               the module is actually part of the Perl core, the version of Perl
               passed in will be used as the module version.
     '''
-
 
     # In case we were given a dist, convert to module
     module = module.replace('-', '::')
@@ -733,9 +733,9 @@ def get_release_info(cpan_url, package, version, perl_version, config,
     # request
 
     version_str = str(version)
-    if (version is not None and 
-        LooseVersion('0') != LooseVersion(version_str) and
-        LooseVersion(rel_dict['version']) != LooseVersion(version_str)):
+    if (version is not None and
+            LooseVersion('0') != LooseVersion(version_str) and
+            LooseVersion(rel_dict['version']) != LooseVersion(version_str)):
         author = rel_dict['author']
         try:
             new_rel_dict = get_cpan_api_url('{0}/release/{1}/{2}-{3}'.format(cpan_url,
--- a/conda_build/skeletons/cpan.py
+++ b/conda_build/skeletons/cpan.py
@@ -303,12 +303,12 @@ def skeletonize(packages, output_dir=".", version=None,
 
         # Fetch all metadata from CPAN
         core_version = core_module_version(package, perl_version, config=config)
-        if version in [None, ''] and core_version is None:
+        if not version and not core_version:
             release_data = latest_release_data
         else:
             release_data = get_release_info(meta_cpan_url, package,
                                             (LooseVersion(version) if
-                                             version not in [None, ''] else
+                                             version else
                                              core_version),
                                              perl_version,
                                              config=config)
@@ -316,7 +316,7 @@ def skeletonize(packages, output_dir=".", version=None,
         # Add Perl version to core module requirements, since these are empty
         # packages, unless we're newer than what's in core
         if (core_version is not None and
-           ((version in [None, '']) or (core_version >= LooseVersion(version)))):
+           (not version or core_version >= LooseVersion(version))):
             if not write_core:
                 print('We found core module %s. Skipping recipe creation.' %
                       packagename)
@@ -494,7 +494,7 @@ def core_module_version(module, version, config):
 
     # In case we were given a dist, convert to module
     module = module.replace('-', '::')
-    if version in [None, '']:
+    if not version:
         version = LooseVersion(config.variant['perl'])
     else:
         version = LooseVersion(version)
@@ -580,7 +580,7 @@ def deps_for_package(package, release_data, perl_version, output_dir,
     for dep_dict in release_data['dependency']:
         # Only care about requirements outside the develop phase
         try:
-            if dep_dict['relationship'] == 'requires' and dep_dict['phase'] not in ['develop']:
+            if dep_dict['relationship'] == 'requires' and dep_dict['phase'] != 'develop':
                 print('.', end='')
                 sys.stdout.flush()
                 # Format dependency string (with Perl trailing dist comment)
--- a/conda_build/skeletons/cpan.py
+++ b/conda_build/skeletons/cpan.py
@@ -5,12 +5,12 @@ Tools for converting CPAN packages to conda recipes.
 from __future__ import absolute_import, division, print_function
 
 import codecs
-from pkg_resources import parse_version
+import distutils.version
 from glob import glob
 import gzip
 import json
 import os
-from os import makedirs
+from os import makedirs, environ
 from os.path import basename, dirname, join, exists
 import subprocess
 import sys
@@ -26,7 +26,6 @@ from conda_build.config import get_or_merge_config
 from conda_build.utils import on_win, check_call_env
 from conda_build.variants import get_default_variant
 
-import requests
 
 CPAN_META = """\
 {{% set name = "{packagename}" %}}
@@ -131,8 +130,6 @@ IF exist Build.PL (
 :: for a list of environment variables that are set during the build process.
 """
 
-perl_core = []
-
 
 class InvalidReleaseError(RuntimeError):
 
@@ -179,9 +176,24 @@ class PerlTmpDownload(TmpDownload):
 
             return dst
 
+def LooseVersion(version):
+    """
+    Wrap distutils.version.LooseVersion, coercing passed version into str
+    and preserving None passed in
 
-def loose_version(ver):
-    return str(parse_version(str(ver)))
+    metacpan is sometimes bad about returning strings or numbers for versions
+    see https://github.com/metacpan/metacpan-api/issues/789
+
+    we also sometimes get version numbers prefixed with 'v' and that causes
+    TypeError about comparing ints against strings
+
+    :param version: version string or number
+    :return: distutils.version.LooseVersion(str(version).lstrip('v'))
+    """
+    if version is None:
+        return version
+    else:
+        return distutils.version.LooseVersion(str(version).lstrip('v'))
 
 
 def get_cpan_api_url(url, colons):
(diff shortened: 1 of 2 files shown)


conda_build/skeletons/cpan.py
```diff
@@ -5,12 +5,12 @@
 from __future__ import absolute_import, division, print_function
 
 import codecs
-from pkg_resources import parse_version
+import distutils.version
```
marcelotrevisani: I would recommend to import using ``from`` here
```suggestion
from distutils import version
```

marcelotrevisani: Or just import ``distutils``

conda_build/skeletons/cpan.py
```diff
@@ -180,8 +177,24 @@ def __enter__(self):
             return dst
 
 
-def loose_version(ver):
-    return str(parse_version(str(ver)))
+def LooseVersion(version):
```
marcelotrevisani: Following PEP8 and snake case, that should be
```suggestion
def loose_version(version):
```

conda_build/skeletons/cpan.py
```diff
@@ -289,21 +302,21 @@ def skeletonize(packages, output_dir=".", version=None,
                                                'import_tests': ''})
 
         # Fetch all metadata from CPAN
-        if version is None:
+        core_version = core_module_version(package, perl_version, config=config)
+        if version in [None, ''] and core_version is None:
```
marcelotrevisani: In Python you can just test the variable and it will already test if it is ``None`` or empty. Such as:
```suggestion
        if not version and not core_version:
```

conda_build/skeletons/cpan.py
```diff
@@ -289,21 +302,21 @@ def skeletonize(packages, output_dir=".", version=None,
...
         else:
             release_data = get_release_info(meta_cpan_url, package,
-                                            parse_version(version),
-                                            perl_version,
-                                            config=config)
-
-        # Check if recipe directory already exists
-        dir_path = join(output_dir, packagename, release_data['version'])
+                                            (LooseVersion(version) if
+                                             version not in [None, ''] else
```
marcelotrevisani: same thing here
```suggestion
                                             version else
```

conda_build/skeletons/cpan.py
```diff
@@ -289,21 +302,21 @@ def skeletonize(packages, output_dir=".", version=None,
...
+                                             core_version),
+                                             perl_version,
+                                             config=config)
 
         # Add Perl version to core module requirements, since these are empty
         # packages, unless we're newer than what's in core
-        if metacpan_api_is_core_version(meta_cpan_url, package):
-
+        if (core_version is not None and
+           ((version in [None, '']) or (core_version >= LooseVersion(version)))):
```
marcelotrevisani:
```suggestion
           (not version or core_version >= LooseVersion(version))):
```

conda_build/skeletons/cpan.py
```diff
@@ -471,12 +474,74 @@ def latest_pkg_version(pkg):
...
         pkg_list = None
     if pkg_list:
-        pkg_version = parse_version(pkg_list[-1].version)
+        pkg_version = LooseVersion(pkg_list[-1].version)
     else:
         pkg_version = None
     return pkg_version
 
 
+@memoized
```
marcelotrevisani: There is no ``memoized`` in Python ``2.7``, as we still support it we cannot use it

conda_build/skeletons/cpan.py
```diff
@@ -471,12 +474,74 @@ def latest_pkg_version(pkg):
...
+
+    :returns: The version of the specified module that is currently available
+              in the specified version of Perl. If the version is `undef`, but
+              the module is actually part of the Perl core, the version of Perl
+              passed in will be used as the module version.
+    '''
+
+    # In case we were given a dist, convert to module
+    module = module.replace('-', '::')
+    if version in [None, '']:
```
marcelotrevisani:
```suggestion
    if not version:
```

conda_build/skeletons/cpan.py
```diff
@@ -515,16 +578,16 @@ def deps_for_package(package, release_data, perl_version, output_dir,
         return build_deps, build_core_deps, run_deps, run_core_deps, packages_to_append
 
     for dep_dict in release_data['dependency']:
-        # Only care about requirements
+        # Only care about requirements outside the develop phase
         try:
-            if dep_dict['relationship'] == 'requires':
+            if dep_dict['relationship'] == 'requires' and dep_dict['phase'] not in ['develop']:
```
marcelotrevisani: As there is just one element there, you can just compare it
```suggestion
            if dep_dict['relationship'] == 'requires' and dep_dict['phase'] != 'develop':
```

marcelotrevisani:
Hi @timsnyder , thanks for opening this PR! 
I commented about some improvements in the code, I hope you don't mind.  :)

Could you please write a test for this new behavior? Is it possible? I would really appreciate that!

conda_build/skeletons/cpan.py
timsnyder: Yes.  I agree we should make this a simple compare to keep this PR separate from the next one where I add to the list and give new functionality.

conda_build/skeletons/cpan.py
timsnyder: Good point.  It was this way in the old code and I agree we should clean it up.

conda_build/skeletons/cpan.py
timsnyder: I was trying to make it clear that we are wrapping the `LooseVersion` class creation by naming it this way, also that I was moving back to using `LooseVersion` instead of the `loose_version` function introduced in PR #2156.

If others agree, I'm fine doing the name change.  I'll do it offline and push a new commit since it will need to be changed throughout the file.

conda_build/skeletons/cpan.py
```diff
@@ -131,8 +130,6 @@
 :: for a list of environment variables that are set during the build process.
 """
 
-perl_core = []
```
timsnyder: This is code gardening.  `perl_core` was dead, unused throughout cpan.py.

conda_build/skeletons/cpan.py
```diff
@@ -26,7 +26,6 @@
 from conda_build.utils import on_win, check_call_env
 from conda_build.variants import get_default_variant
 
-import requests
```
timsnyder: `requests` is no longer used directly, as before PR #2156

conda_build/skeletons/cpan.py
timsnyder: I was trying to do the minimal required import to wrap `LooseVersion`.  We aren't using all of `distutils`, only parts that are inside `distutils.version`.  

I think that `from distutils import version` will conflict with the many uses of `version` inside `def`s below, won't it?

If I end up changing `def LooseVersion` to `def loose_version` as you suggest below, I'll change this back to `from distutils.version import LooseVersion` as it was pre PR #2156

conda_build/skeletons/cpan.py
timsnyder: It's made available from `conda_build.conda_interface` and used in many other places in cpan.py.  This use is reintroduction of `def core_module_version` from before PR #2156 becuase we need to use the Perl `corelist` utility as I explained in my commit message and the description of this PR.

If we need to clean these out of cpan.py, I'd like to do that in a separate PR that removes the import and all other uses.

timsnyder:
Thanks for the great feedback @marcelotrevisani .  I appreciate it!

I definitely agree that I should add tests.  I wanted to get the code out here for others to look at and digest.  Please note that I'm bringing back some of the behavior from before PR #2156 with this PR.  I probably should have been more explicit about that in my commit message and description for this PR.

timsnyder:
@msarahan should we invite any of the previous contributors to cpan.py to review this?  I'm not sure if you've heard from Jillian or Dan recently.

conda_build/skeletons/cpan.py
msarahan: Hmmm... this is a tiny bit of a gray area in my opinion.  I'd be inclined to keep LooseVersion, since the usage really matches the instantiation of a class anyway.  If changing it to PEP8/snake case, I think the function name should be more of a verb than a noun.  `get_loose_version` or `loosen_version` or something like that.

conda_build/skeletons/cpan.py
```diff
@@ -180,8 +177,24 @@ def __enter__(self):
...
+    metacpan is sometimes bad about returning strings or numbers for versions
+    see https://github.com/metacpan/metacpan-api/issues/789
+
+    we also sometimes get version numbers prefixed with 'v' and that causes
+    TypeError about comparing ints against strings
+
+    :param version: version string or number
+    :return: distutils.version.LooseVersion(str(version).lstrip('v'))
+    """
+    if version is None:
```
msarahan: This block could be just:

```
if version:
    return  distutils.version.LooseVersion(str(version).lstrip('v'))
```

because python's implicit return type is None anyway.

conda_build/skeletons/cpan.py
msarahan: It's defined in conda.  I think it's fine here.

msarahan:
I haven't seen the other contributors to this stuff around lately.  I'm not sure if Bioconda people would be interested in reviewing - they use more perl than most of us.  I will point them to this PR.

conda_build/skeletons/cpan.py
timsnyder: @msarahan good point.  I'll change this.

conda_build/skeletons/cpan.py
timsnyder: I'll just keep LooseVersion in this PR then.  There are quite a few other improvements that I'm saving for future PR's to break things into smaller sets of changes.

conda_build/skeletons/cpan.py
timsnyder: Resolved in https://github.com/conda/conda-build/pull/3671/commits/fa7f928e22ab8374de923e427f3f0a46790e6b6d

timsnyder:
@msarahan one of the tests is failing now that I added `corelist` back into the cpan skeleton:
```
__________________________ test_repo[perl-cpan-Moo-] ___________________________‌
[gw1] linux2 -- Python 2.7.16 /usr/share/miniconda/bin/python
  File "/home/vsts/work/1/s/tests/test_api_skeleton.py", line 42, in test_repo
    config=testing_config)
  File "/home/vsts/work/1/s/conda_build/api.py", line 290, in skeletonize
    recursive=recursive, config=config, **kwargs)
  File "/home/vsts/work/1/s/conda_build/skeletons/cpan.py", line 303, in skeletonize
    core_version = core_module_version(package, perl_version, config=config)
  File "/usr/share/miniconda/lib/python2.7/site-packages/conda/exports.py", line 198, in __call__
    value = self.func(*args, **kw)
  File "/home/vsts/work/1/s/conda_build/skeletons/cpan.py", line 512, in core_module_version
    'environment.') % ' '.join(cmd))
SystemExit: Error: command failed: corelist -v 5.26.0 Moo
Please make sure you have the perl conda package installed in your default environment.
```

Do you want to make `perl` a run requirement or only a test requirement?  I can see how either one might make sense.  Having it be a run requirement will make the cpan skeleton work always.  However, I know not many people use it.

jerowe:
I get a ton of github alerts and I must have missed this. I'm dedicating some time the next few weeks to bioconda perl things.

PR #3686, opened 2019-08-23
case sensitivity in binary inspection breaks some win packages (JIRA …

katietz:
…CB-787)


jjhelmus:
LGTM, waiting on CI before merging.

PR #3687, opened 2019-08-23
install codecov from defaults not conda-forge

jjhelmus:


jjhelmus:
@marcelotrevisani @ocefpaf codecov is now in defaults

marcelotrevisani:
I am working on azure pipelines which might improve our pipeline, but I think it may take some time

Thanks to added ``codecov`` to the main channel! :)

Regarding the tests, it seems there is a race condition, one of them got stuck and was killed by ``travis`` due to no response for more than 10 min

PR #3691, opened 2019-08-29
render: Permit --file and --output to be used simultaneously

stuarteberg:
The following command does not behave as expected because `conda render` does not respect both `--file` and `--output` simultaneously.  (If `--output` is present, then `--file` is ignored.)

```
conda render --output --file=rendered-meta.yaml myrecipe > rendered-tarball-name.txt
```

With this PR, both options can be used at the same time.

### DIFF
--- a/conda_build/cli/main_render.py
+++ b/conda_build/cli/main_render.py
@@ -24,6 +24,7 @@ from conda_build.variants import get_package_variants, set_language_env_vars
 from conda_build.utils import LoggingContext
 
 on_win = (sys.platform == 'win32')
+log = logging.getLogger(__name__)
 
 
 # see: https://stackoverflow.com/questions/29986185/python-argparse-dict-arg
@@ -196,11 +197,18 @@ def execute(args, print_results=True):
                                  no_download_source=args.no_source,
                                  variants=args.variants)
 
+    if args.file and len(metadata_tuples) > 1:
+        log.warning("Multiple variants rendered. "
+                    "Only one will be written to the file you specified ({}).".format(args.file))
+
     if print_results:
         if args.output:
             with LoggingContext(logging.CRITICAL + 1):
                 paths = api.get_output_file_paths(metadata_tuples, config=config)
                 print('\n'.join(sorted(paths)))
+            if args.file:
+                m = metadata_tuples[-1][0]
+                api.output_yaml(m, args.file, suppress_outputs=True)
         else:
             logging.basicConfig(level=logging.INFO)
             for (m, _, _) in metadata_tuples:
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -116,6 +116,20 @@ def test_render_output_build_path(testing_workdir, testing_metadata, capfd, capl
     assert error == ""
 
 
+def test_render_output_build_path_and_file(testing_workdir, testing_metadata, capfd, caplog):
+    api.output_yaml(testing_metadata, 'meta.yaml')
+    rendered_filename = 'out.yaml'
+    args = ['--output', '--file', rendered_filename, os.path.join(testing_workdir)]
+    main_render.execute(args)
+    test_path = os.path.join(sys.prefix, "conda-bld", testing_metadata.config.host_subdir,
+                             "test_render_output_build_path_and_file-1.0-1.tar.bz2")
+    output, error = capfd.readouterr()
+    assert output.rstrip() == test_path, error
+    assert error == ""
+    rendered_meta = yaml.safe_load(open(rendered_filename, 'r'))
+    assert rendered_meta['package']['name'] == 'test_render_output_build_path_and_file'
+
+
 def test_build_output_build_path(testing_workdir, testing_metadata, testing_config, capfd):
     api.output_yaml(testing_metadata, 'meta.yaml')
     testing_config.verbose = False
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -107,7 +107,7 @@ def test_no_filename_hash(testing_workdir, testing_metadata, capfd):
 
 def test_render_output_build_path(testing_workdir, testing_metadata, capfd, caplog):
     api.output_yaml(testing_metadata, 'meta.yaml')
-    args = ['--output', os.path.join(testing_workdir)]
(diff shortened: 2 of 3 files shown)


conda_build/cli/main_render.py
```diff
@@ -196,11 +197,18 @@ def execute(args, print_results=True):
...
+        log.warning("Multiple variants rendered. "
+                    "Only one will be written to the file you specified ({}).".format(args.file))
+
     if print_results:
         if args.output:
             with LoggingContext(logging.CRITICAL + 1):
                 paths = api.get_output_file_paths(metadata_tuples, config=config)
                 print('\n'.join(sorted(paths)))
+            if args.file:
+                (m, _, _) = metadata_tuples[-1]
```
marcelotrevisani:
```suggestion
                m = metadata_tuples[-1][0]
```

marcelotrevisani:
Hi @stuarteberg ,
Thanks for your contribution! :)

Just one thing, is it possible for you to create a test to cover this scenario? I would appreciate that! :)

conda_build/cli/main_render.py
stuarteberg: When I first wrote this code, I couldn't make up my mind between these two choices.  I thought it might be nice to use same syntax that the for-loop below uses.  But OK.  I've changed it now.

stuarteberg:
@ marcelotrevisani 

>is it possible for you to create a test to cover this scenario

Good point. I've now added a test.

stuarteberg:
@marcelotrevisani 

>is it possible for you to create a test to cover this scenario

Good point. I've now added a test.

tests/test_cli.py
```diff
@@ -116,6 +116,22 @@ def test_render_output_build_path(testing_workdir, testing_metadata, capfd, capl
     assert error == ""
 
 
+def test_render_output_build_path_and_file(testing_workdir, testing_metadata, capfd, caplog):
+    api.output_yaml(testing_metadata, 'meta.yaml')
+    
+    with TemporaryDirectory() as tmpdir:
```
marcelotrevisani: Instead of using ``TemporaryDirectory`` you can just use a ``pytest`` fixture called ``tmpdir``

tests/test_cli.py
```diff
@@ -116,6 +116,22 @@ def test_render_output_build_path(testing_workdir, testing_metadata, capfd, capl
     assert error == ""
 
 
+def test_render_output_build_path_and_file(testing_workdir, testing_metadata, capfd, caplog):
```
marcelotrevisani:
```suggestion
def test_render_output_build_path_and_file(testing_workdir, testing_metadata, capfd, caplog, tmpdir):
```

tests/test_cli.py
marcelotrevisani: the ``pytest`` fixture ``tmpdir`` is a ``py.path .local``
Please see
http://doc.pytest.org/en/latest/tmpdir.html#the-tmpdir-fixture

marcelotrevisani: Probably it will simplify a bit the test :)
If you need any help please ping me!

stuarteberg:
>Instead of using `TemporaryDirectory` you can just use a `pytest` fixture called `tmpdir`

OK, it turns out that the `testing_workdir` fixture (used by this test and most others in `test_cli.py`) uses `tmpdir` internally, and it also `chdir`s into the created `tmpdir` for the duration of the test.

Therefore, it would be unnecessary for me to add the `tmpdir` fixture explicitly:

```python
def test_render_output_build_path_and_file(testing_workdir, testing_metadata, capfd, caplog, tmpdir):
    assert testing_workdir == os.getcwd() == tmpdir # This is true!
```

So I've changed this test to not bother with yet another temporary directory.  Instead, it just writes `out.yaml` into the current directory.

tests/test_cli.py
```diff
@@ -116,6 +116,20 @@ def test_render_output_build_path(testing_workdir, testing_metadata, capfd, capl
...
+    api.output_yaml(testing_metadata, 'meta.yaml')
+    rendered_filename = 'out.yaml'
+    args = ['--output', '--file', rendered_filename, os.path.join(testing_workdir)]
```
marcelotrevisani: is that right?
``os.path.join(testing_workdir)``

stuarteberg: >is that right?
>`os.path.join(testing_workdir)`

Good point.  I created this test by copying/modifying the test above it.  I have no idea why that test is calling `join()` with only a single argument.  I agree with you -- it seems like a mistake.  Now I've pushed a commit to fix my test and the two others in the file making the same mistake.

tests/test_cli.py
```diff
@@ -116,11 +116,25 @@ def test_render_output_build_path(testing_workdir, testing_metadata, capfd, capl
...
+    api.output_yaml(testing_metadata, 'meta.yaml')
+    rendered_filename = 'out.yaml'
+    args = ['--output', '--file', rendered_filename, testing_workdir]
+    main_render.execute(args)
+    test_path = os.path.join(sys.prefix, "conda-bld", testing_metadata.config.host_subdir,
+                             "test_render_output_build_path_and_file-1.0-1.tar.bz2")
+    output, error = capfd.readouterr()
+    assert output.rstrip() == test_path, error
+    assert error == ""
+    rendered_meta = yaml.safe_load(open(rendered_filename, 'r'))
```
marcelotrevisani: Just another thing here (sorry for not spotted it earlier)
It would be better to use a context manager here with the ``open``. Something like:
```suggestion
    with open(rendered_filename, 'r') as rendered_yaml:
        rendered_meta = yaml.safe_load(rendered_yaml)
```

stuarteberg: >It would be better to use a context manager here with the `open`

OK, I've changed all the tests in the file not to use `open()` without a context manager.

stuarteberg:
The CI is still timing out, even after a forced retry.  Presumably, that's unrelated to this PR.

FWIW, `pytest tests/test_cli.py` passes on my machine (mac, python-3.7).  BTW, `pytest-mock` is required to run the test suite, so I added it as a test requirement in `conda.recipe/meta.yaml`.

Anyway, I think this is ready for merge.

PR #3714, opened 2019-09-05
conda skeleton CRAN ignores CRAN's OS_type information (CB-810)

katietz:

### DIFF
--- a/conda_build/skeletons/cran.py
+++ b/conda_build/skeletons/cran.py
@@ -1115,7 +1115,7 @@ def skeletonize(in_packages, output_dir=".", output_suffix="", add_maintainer=No
             else:
                 d['homeurl'] = ' https://CRAN.R-project.org/package={}'.format(package)
 
-        if not use_noarch_generic or cran_package.get("NeedsCompilation", 'no') == 'yes' or os_type != '':
+        if not use_noarch_generic or cran_package.get("NeedsCompilation", 'no') == 'yes':
             d['noarch_generic'] = ''
         else:
             d['noarch_generic'] = 'noarch: generic'
--- a/conda_build/skeletons/cran.py
+++ b/conda_build/skeletons/cran.py
@@ -1166,8 +1166,6 @@ def skeletonize(in_packages, output_dir=".", output_suffix="", add_maintainer=No
         if 'R' not in dep_dict:
             dep_dict['R'] = ''
 
-        if 'rJava' in dep_dict:
-            print("Java is in dependencies")
         os_type = cran_package.get("OS_type", '')
         if os_type != 'unix' and os_type != 'windows' and os_type != '':
             print("Unknown OS_type: {} in CRAN package".format(os_type))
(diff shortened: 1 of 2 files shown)


marcelotrevisani:
Hi @katietz , could you please write some tests to cover this scenario?

conda_build/skeletons/cran.py
```diff
@@ -1165,6 +1166,23 @@ def skeletonize(in_packages, output_dir=".", output_suffix="", add_maintainer=No
         if 'R' not in dep_dict:
             dep_dict['R'] = ''
 
+        if 'rJava' in dep_dict:
+            print("Java is in dependencies")
```
mingwandroid: Will this handle transitive dependencies on rJava?

mingwandroid: also is it just debug text?

katietz: Yes,it will add a skip on non Linux hosts for java.  Actually not sure if we want that within that commit, but logicial it belongs here IMHO.  I don't touch here the 'noarch' settings for now.  But logically the package isn't noarch, if we just build for a specific target,

katietz: Removed that debug print

conda_build/skeletons/cran.py
```diff
@@ -1165,6 +1166,17 @@ def skeletonize(in_packages, output_dir=".", output_suffix="", add_maintainer=No
         if 'R' not in dep_dict:
             dep_dict['R'] = ''
 
+        os_type = cran_package.get("OS_type", '')
+        if os_type != 'unix' and os_type != 'windows' and os_type != '':
+            print("Unknown OS_type: {} in CRAN package".format(os_type))
+            os_type = ''
+        if os_type == 'unix':
+            d['skip_os'] = 'skip: True  # [not unix]'
```
marcelotrevisani: I think you need to add ``d["noarch_generic"] = ""`` if any of ``OS_type`` is present
```suggestion
            d['skip_os'] = 'skip: True  # [not unix]'
            d["noarch_generic"] = ""
```

katietz: I added testcase for it.  Overriding here the 'noarch: generic' is IMHO correct, but I guess @mingwandroid  has here a strong opinion

conda_build/skeletons/cran.py
```diff
@@ -1165,6 +1166,18 @@ def skeletonize(in_packages, output_dir=".", output_suffix="", add_maintainer=No
...
+            d["noarch_generic"] = ""
+        if os_type == 'windows':
+            d['skip_os'] = 'skip: True  # [not win]'
```
marcelotrevisani: I think you will need to add the ``d["noarch_generic"] = ""`` here as well
```suggestion
            d['skip_os'] = 'skip: True  # [not win]'
            d["noarch_generic"] = ""
```

If @mingwandroid has a better idea that is fine for me! :)

marcelotrevisani:
Is it ready?

marcelotrevisani:
Thanks for adding tests!
I think that is good to go

Issue #3716, opened 2019-09-06
Update conda skeleton pypi command


marcelotrevisani:
Should we use 
```
{{ PYTHON }} -m pip install . -vv
```
https://github.com/conda-forge/staged-recipes/#7-when-or-why-do-i-need-to-use--python---m-pip-install---vv

Instead of 
```
{{ PYTHON }} -m pip install . --no-deps --ignore-installed -vv
```

on the conda skeleton pypi to generate the recipes?

Related to 
https://github.com/conda-forge/staged-recipes/pull/9444#discussion_r321696835

cc: @CJ-Wright

CJ-Wright:
I'm thinking of doing something on the bot side which will also update existing pip commands using https://github.com/regro/cf-scripts/pull/577

marcelotrevisani:
> I'm thinking of doing something on the bot side which will also update existing pip commands using [regro/cf-scripts#577](https://github.com/regro/cf-scripts/pull/577)

I might take a look at it as well. Thanks!

marcelotrevisani:
Closing this in favor of #3717

PR #3717, opened 2019-09-06
Update script command on conda skeleton pypi

marcelotrevisani:
Update script command on conda skeleton pypi to use `{{ PYTHON }} -m pip install . -vv`

### DIFF
--- a/conda_build/skeletons/pypi.py
+++ b/conda_build/skeletons/pypi.py
@@ -344,9 +344,9 @@ def skeletonize(packages, output_dir=".", version=None, recursive=False,
             if noarch_python:
                 ordered_recipe['build']['noarch'] = 'python'
 
-            ordered_recipe['build']['script'] = ('{{ PYTHON }} -m pip install . --no-deps '
-                                                 '--ignore-installed -vv ' +
-                                                 ' '.join(setup_options))
+            ordered_recipe['build']['script'] = (
+                    '{{ PYTHON }} -m pip install . -vv' + ' '.join(setup_options)
+                )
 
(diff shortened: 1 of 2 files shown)


marcelotrevisani:
Related to #3716 

cc: @CJ-Wright

conda_build/skeletons/pypi.py
```diff
@@ -339,7 +339,9 @@ def skeletonize(packages, output_dir=".", version=None, recursive=False,
             ordered_recipe['source']['sha256'] = d['hash_value']
 
             if d['entry_points']:
-                ordered_recipe['build']['entry_points'] = d['entry_points']
+                ordered_recipe['build']['script'] = (
```
marcelotrevisani: geez. copy paste on the wrong place,

ocefpaf:
Thanks! This will match closely to what we are using in conda-forge and will reduce the number of edits :smile: 

@marcelotrevisani, on a related note, conda-skeleton is "quoting" all the lines and that breaks the bot updates. I guess that it would be nice to have at least the checksum unquoted.

marcelotrevisani:
> Thanks! This will match closely to what we are using in conda-forge and will reduce the number of edits smile
> 
> @marcelotrevisani, on a related note, conda-skeleton is "quoting" all the lines and that breaks the bot updates. I guess that it would be nice to have at least the checksum unquoted.

Sure! I can take a look at it as well!
Could you just open a new issue (just to have some track and to not forgot) for it an give me an example just to be sure that is what I am thinking? haha

Issue #3685, opened 2019-08-22
conda index 3.18.* crashes on corrupt package with only an obscure error message


y2kbugger:
### Actual Behavior
```
hash & extract packages for win-64:   0%|          | 0/1 [00:00<?, ?it/s]Exception in thread Thread-2:
Traceback (most recent call last):
  File "C:\ProgramData\CooperConda\lib\threading.py", line 916, in _bootstrap_inner
    self.run()
  File "C:\ProgramData\CooperConda\lib\threading.py", line 864, in run
    self._target(*self._args, **self._kwargs)
  File "C:\ProgramData\CooperConda\lib\concurrent\futures\process.py", line 272, in _queue_management_worker
    result_item = reader.recv()
  File "C:\ProgramData\CooperConda\lib\multiprocessing\connection.py", line 251, in recv
    return _ForkingPickler.loads(buf.getbuffer())
TypeError: __init__() missing 1 required positional argument: 'msg'
```

### Expected Behavior

Not crash, and give error message similar to the 3.17.* series. Downgrading is how I finally determined that their was a problem package.

>     Package win-64/bokeh-1.3.4-py37_0.tar.bz2 appears to be corrupt.  Please remove it and re-download it
>     ERROR:conda_build.index:Package` win-64/bokeh-1.3.4-py37_0.tar.bz2 appears to be corrupt.
>     Please remove it and re-download it

### Steps to Reproduce

[Ex.zip](https://github.com/conda/conda-build/files/3530275/Ex.zip)

1. Extract the zip file.
2.     $ cd Ex
       $ conda index .

##### `Output of conda info`
```
$ conda info

     active environment : base
    active env location : C:\ProgramData\CooperConda
            shell level : 1
       user config file : C:\Users\zdkohler\.condarc
 populated config files : C:\ProgramData\CooperConda\.condarc
          conda version : 4.7.11
    conda-build version : 3.18.9
         python version : 3.6.9.final.0
       virtual packages : __cuda=9.1
       base environment : C:\ProgramData\CooperConda  (writable)
           channel URLs : http://technicalwebdev/CooperConda/Internal/win-64
                          http://technicalwebdev/CooperConda/Internal/noarch
                          http://technicalwebdev/CooperConda/External/win-64
                          http://technicalwebdev/CooperConda/External/noarch
          package cache : C:\ProgramData\CooperConda\pkgs
                          C:\Users\zdkohler\.conda\pkgs
[...]

PR #3719, opened 2019-09-06
improve error handling around invalid archives in indexing

msarahan:
fixes https://github.com/conda/conda-build/issues/3685

Note that the exception imported from conda-package-handling is a reasonably new addition to CPH.  We may need to bump up the lower bound of CPH in the conda-build requirements.


msarahan:
Don't need a new constraint; the exception was added in 1.3.0

PR #3726, opened 2019-09-07
Fix flake8 problems

marcelotrevisani:

### DIFF
--- a/conda_build/index.py
+++ b/conda_build/index.py
@@ -50,6 +50,11 @@ from .conda_interface import TemporaryDirectory
 from .conda_interface import Resolve
 from .utils import glob, get_logger, FileNotFoundError, JSONDecodeError
 
+# try:
+#     from conda.base.constants import CONDA_TARBALL_EXTENSIONS
+# except Exception:
+#     from conda.base.constants import CONDA_TARBALL_EXTENSION
+#     CONDA_TARBALL_EXTENSIONS = (CONDA_TARBALL_EXTENSION,)
 
 # TODO: better to define this in conda; doing it here because we're implementing it in conda-build first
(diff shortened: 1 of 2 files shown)


conda_build/index.py
```diff
@@ -38,28 +36,20 @@
...
 
-
 from . import conda_interface, utils
 from .conda_interface import MatchSpec, VersionOrder, human_bytes, context
 from .conda_interface import CondaError, CondaHTTPError, get_index, url_path
 from .conda_interface import TemporaryDirectory
 from .conda_interface import Resolve
 from .utils import glob, get_logger, FileNotFoundError, JSONDecodeError
 
-# try:
```
mcg1969: Did flake8 really complain about these commented lines? It did not for me—and I would suggest we shouldn't remove it without deliberate intent. From the blame it seems @msarahan put this in here deliberately, as a reminder for future development. I'd like to recommend this be restored.

marcelotrevisani: Flake8 did not complain about it indeed. 
However, I just do not see the necessity of commenting code, since you can create an issue an relate to it and explain better why it should change and how it should be changed. From my experience (and according to the literature), those commented codes stay in the project almost forever (for example that one is 6+ month old, I have seen other older than that). I don't mind to revert it. It is just that I usually try to follow the principles of clean code. But that is fine for me, I can revert it then. :)

mcg1969: Thanks. I generally agree with your position, but since neither of us wrote this code, and because the stated intent of this PR is simply to fix flake8 issues, it's better for now to avoid.

marcelotrevisani: Sure thing! No problem from my side. I already reverted it! :)

mcg1969: Thanks! Will approve as soon as we're green

PR #3727, opened 2019-09-07
Version support for CRAN skeleton

mcg1969:
Adding support for version selection for `conda skeleton cran`.

To do this, we've replaced the internal index `conda_metadata` with a simpler index `conda_index`. We never used most of the information in `conda_metadata`; it really exists just to provide a lowercase-to-fullycased package map. The new `conda_index` keys are lowercase names of packages as before, and the values are `(name, version)` pairs, where `name` is the fully-cased package name, and `version` is the active version of the package. What is important is that _archived_ packages are included in this list, with `version = None`. Because the `PACKAGES` file does not include information about archived packages or older versions, this index is constructed by scraping the HTML from `/src/contrib/` and `/src/contrib/Archive`.

If a specific version is requested, and that version is not in the current package set, the function `get_cran_archive_versions` retrieves that from a directory listing of `/src/contrib/Archive/<package>`.

The metadata for the package is fetched by downloading the source package into the source cache, and extracted from there, using the `get_archive_metadata` function.

### DIFF
--- a/conda_build/skeletons/cran.py
+++ b/conda_build/skeletons/cran.py
@@ -546,6 +546,25 @@ def get_package_metadata(cran_url, package, session):
     return d
 
 
+def get_archive_metadata(cran_url, package, version, cache, verbose=True):
+    filename_rendered = package + '_' + version + '.tar.gz'
+    package_url = cran_url + '/src/contrib/Archive/' + package + '/' + filename_rendered
+    if verbose:
+        print("Downloading {} from {}".format('source', package_url))
+    cached_path, _ = source.download_to_cache(cache, '',
+        {'url': package_url, 'fn': 'source' + '-' + filename_rendered})
+    with tarfile.open(cached_path) as tf:
+        try:
+            member = tf.getmember(package + '/DESCRIPTION')
+        except KeyError:
+            print("logic error, file {} should have a DESCRIPTION".format(package_url))
+            sys.exit(1)
+        DESCRIPTION = tf.extractfile(member).read().decode('utf-8')
+    d = dict_from_cran_lines(remove_package_line_continuations(DESCRIPTION.splitlines()))
+    d['orig_description'] = DESCRIPTION
+    return d
+
+
 def get_latest_git_tag(config):
     # SO says to use taggerdate instead of committerdate, but that is invalid for lightweight tags.
     p = subprocess.Popen(['git', 'for-each-ref',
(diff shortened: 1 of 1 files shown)


mcg1969:
The travis failure is due to flake issues on files not touched here. #3726 or equivalent will fix that

mcg1969:
@mingwandroid @msarahan IMO, this is ready to merge. I've been using it heavily in the last few days to build out some packages for Ford—they have approved specific, archived versions, so I'm making full use of the explicit version functionality.

Note that I have chosen to append the version number to the package directory when the version is explicitly specified—this way it is straightforward to have multiple recipes in the same directory for the same package, including an un-versioned directory for the latest version.

mcg1969:
@katietz you might be interested in this too

conda_build/skeletons/cran.py
```diff
@@ -146,7 +147,7 @@
 if {source_pf_bash}; then
   export DISABLE_AUTOBREW=1
   mv DESCRIPTION DESCRIPTION.old
-  grep -v '^Priority: ' DESCRIPTION.old > DESCRIPTION
+  grep -va '^Priority: ' DESCRIPTION.old > DESCRIPTION
```
mcg1969: This is a small but important change independent of the version support. Without the `-a` option, DESCRIPTION files with accented characters will not be processed correctly, and will very likely break the build.

conda_build/skeletons/cran.py
```diff
@@ -594,17 +611,41 @@ def get_session(output_dir, verbose=True):
...
+            versions.append(version)
+    return versions
+
+
+def get_cran_index(cran_url, session, verbose=True):
     if verbose:
-        print("Fetching metadata from %s" % cran_url)
-    r = session.get(cran_url + "/src/contrib/PACKAGES")
+        print("Fetching main index from %s" % cran_url)
+    r = session.get(cran_url + "/src/contrib/")
```
mcg1969: This is a much lighter-weight download, and quicker to process. At the stages where we use this information we only need package name and version anyway.

conda_build/skeletons/cran.py
```diff
@@ -620,8 +661,10 @@ def make_array(m, key, allow_empty=False):
     return result
 
 
-def existing_recipe_dir(output_dir, output_suffix, package):
+def existing_recipe_dir(output_dir, output_suffix, package, version):
     result = None
+    if version:
```
mcg1969: When a version is explicitly requested—even if it happens to be the latest available version—we append that to the directory name. I'm using this now because I'm building multiple versions of the same package against the same base R version for a customer.

conda_build/skeletons/cran.py
```diff
@@ -1276,7 +1331,7 @@ def skeletonize(in_packages, output_dir=".", output_suffix="", add_maintainer=No
                             lower_name = name.lower()
                             if lower_name not in package_dicts:
                                 inputs_dict = package_to_inputs_dict(output_dir, output_suffix,
-                                                                     git_tag, lower_name)
+                                                                     git_tag, lower_name, None)
```
mcg1969: Recursive skeleton creations never use the explicit version.

conda_build/skeletons/cran.py
katietz: That's a good catch!

conda_build/skeletons/cran.py
katietz: Ok, have you made measurements for this?

mcg1969: Just my own subjective experience with this; it is building recipes is noticeably faster now. To be fair though I didn't A/B this one by itself. The speed difference in the use-binaries code path is significantly greater. There's a functional reason why this change is important, too—because I need to incorporate the _archived_ package into this index as well, and that can't be done by parsing a PACKAGES file.

msarahan:
Thanks!

PR #3728, opened 2019-09-08
Fix binary recipe building for cran

mcg1969:
While working on the CRAN versioning work I discovered that the `--use-binaries-ver` option wasn't working. This fixes that. And while I was at it I ripped out the use of bs4 for parsing in place of a simple `re.findall`, which speeds things up tremendously.

### DIFF
--- a/conda_build/skeletons/cran.py
+++ b/conda_build/skeletons/cran.py
@@ -721,30 +721,15 @@ def package_to_inputs_dict(output_dir, output_suffix, git_tag, package):
 
 
 def get_available_binaries(cran_url, details):
-    import requests
-    from bs4 import BeautifulSoup
-
-    def get_url_paths(url, ext='', params={}):
-        response = requests.get(url, params=params)
-        if response.ok:
-            response_text = response.text
-        else:
-            return response.raise_for_status()
-        soup = BeautifulSoup(response_text, 'html.parser')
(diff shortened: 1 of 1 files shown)


mcg1969:
The travis failure is due to flake issues on files I did not touch; can rebase once #3726 or equivalent is merged

conda_build/skeletons/cran.py
```diff
@@ -966,11 +951,12 @@ def skeletonize(in_packages, output_dir=".", output_suffix="", add_maintainer=No
                     archive_details['cranurl'] = archive_details['binaries'][package][-1][1]
                     archive_details['conda_version'] = archive_details['binaries'][package][-1][0]
                     archive_details['cran_version'] = archive_details['conda_version'].replace('_', '-')
-                    avaliable_artefact = True
+                    available_artifact = True
```
mingwandroid: https://writingexplained.org/artefact-vs-artifact-difference

mingwandroid:
Cool, I approve of getting rid of all BS :-)

Are you repackaging binaries though? Eek!

conda_build/skeletons/cran.py
mcg1969: Ah! That was muscle memory. My original intent was to correct the misspelling of "available"

mcg1969:
No, I'm not repackaging binaries currently. I just happened to notice that it wasn't working because my other PR is changing `cran_metadata`.

PR #3734, opened 2019-09-16
Added info about hashes being in build strings

rrigdon:

### DIFF
--- a/docs/source/resources/define-metadata.rst
+++ b/docs/source/resources/define-metadata.rst
@@ -294,6 +294,19 @@ string plus the build number.
      number: 1
      string: abc
 
+Hashes will appear when binary compatibility matters, but not when it doesn't.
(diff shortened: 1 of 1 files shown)


docs/source/resources/define-metadata.rst
```diff
@@ -294,6 +294,19 @@ string plus the build number.
      number: 1
      string: abc
 
+Hashes will appear when binary compatibility matters, but not when it doesn't.
```
msarahan: This is the overall idea, but it may not be strictly true.  The true rule is that a hash will appear when the package is affected by one or more variables from the conda_build_config.yaml file.  The hash is made up from the "used" variables - if anything is used, you have a hash.  If not, you don't.  There are a few special cases that do not affect the hash: python, r, etc - things that already had a place in the build string.

PR #3738, opened 2019-09-18
WIP: fix CRAN url in recipe; fetch all packages

mcg1969:
I was taking a look at Bioconductor and noted that each "epoch" is structured identically to CRAN, with one significant exception: it doesn't serve a directory index for the `/src/contrib/` path. So I had to drop back to grabbing the PACKAGES file to build the CRAN index again—but at least I still only need the name and version; no need to parse the entire file. Interestingly, `/src/contrib/Archive/` _does_ provide a simple directory index so the `--version` command still works.

And then I realized I might at least want to optionally fix the CRAN URL in the recipe for these cases, instead of building a variant file, so I added a `--fix-cran-url` option.

Finally, I thought, "why not create all of the recipes in one go?" So I added a `--fetch-all` option. To be used with caution. But if anyone wants a complete set of skeletons for Bioconductor 3.9  (@katietz, perhaps?), I know someone who has one.

### DIFF
--- a/conda_build/skeletons/cran.py
+++ b/conda_build/skeletons/cran.py
@@ -14,6 +14,7 @@ import re
 import subprocess
 import sys
 import hashlib
+import gzip
 
 import requests
 import tarfile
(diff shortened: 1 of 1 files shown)


mcg1969:
Example usage:
```
conda skeleton cran --fetch-all --cran-url https://bioconductor.org/packages/3.9/bioc/ --fix-cran-url --use-noarch-generic --r-interp 'r-base >=3.6,<3.7.0a0'
```

PR #3741, opened 2019-09-20
Add `extra` to `outputs` in `FIELDS`

jakirkham:
Adding it here should also avoid some lint that I'm [running into]( https://github.com/conda-forge/ucx-split-feedstock/pull/14#issuecomment-533550603 ).

### DIFF
--- a/conda_build/metadata.py
+++ b/conda_build/metadata.py
@@ -504,7 +504,7 @@ FIELDS = {
               'pre-link', 'post-link', 'pre-unlink', 'missing_dso_whitelist',
               },
     'outputs': {'name', 'version', 'number', 'script', 'script_interpreter', 'build',
-                'requirements', 'test', 'about', 'files', 'type', 'run_exports'},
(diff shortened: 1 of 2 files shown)


marcelotrevisani:
Hi @jakirkham , is it possible to also add a test to cover this scenario? I would appreciate that! :)

jakirkham:
Not sure how `FIELDS` is tested now TBH, but have pushed a test recipe that leverages the `extra` field. Does that help?

jakirkham:
@marcelotrevisani , is there anything else to do here?

marcelotrevisani:
> Not sure how `FIELDS` is tested now TBH, but have pushed a test recipe that leverages the `extra` field. Does that help?

I was just expecting one recipe with those fields, so we can be sure that at least ``conda-build`` is supporting it. But I think that is fine then. I will merge it! thanks! :)

jakirkham:
Ah ok thanks for clarifying. Also thanks for the review. 😀

PR #3742, opened 2019-09-20
Fix pkg_format 2 (.conda) package builds.

asford:
_Based off #3743 for ci fix._

- [x] Add an api test covering direct builds of v2 (`.conda`) format packages, specified via `conda config --set conda_build.pkg_format 2`.
- [x] Fix error in v2 format package builds, explicitly specifying the proper bundler for `conda_v2` type packages.
- [x] Fix detection and extraction of `.conda` packages in the test phase, enabling test execution for `.conda` packages.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ install:
   - conda info
   # avoids a python 3.7 problem
   - conda install -q cytoolz
-  - conda install -q anaconda-client requests filelock contextlib2 jinja2 patchelf ripgrep pyflakes beautifulsoup4 chardet pycrypto glob2 psutil pytz tqdm conda-package-handling py-lief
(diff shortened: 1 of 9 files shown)


asford:
@msarahan This is a follow up to the discussion in https://github.com/conda/conda/pull/9254, fixing a build error when v2 packages are specified via `.condarc`.

asford:
@msarahan, pulling the conversation here from https://github.com/conda/conda-package-handling/issues/53.

The config format you mentioned there would be a good target for this PR.  Adding a loop over package formats is likely out of scope, but we could clean up the configuration values from (None or "2") to (".tar.bz2", ".conda" or a list of containing one of those values). Thoughts?

msarahan:
This seems fine.  As I mentioned in the other issue, I really don't want the current "2" behavior documented.  I don't want its use to proliferate.  Before this config value is documented, we really need the config value cleanup and looping.  The looping should be trivial.  Just replace this bit with the loop over the config values, and provide a default config in conda_build/config.py: https://github.com/conda/conda-build/blob/master/conda_build/build.py#L1024-L1025

gilescope:
@msarahan time flew by, are we safe now to use the current "2" behavior?

PR #3743, opened 2019-09-21
Add python-libarchive-c to package and CI dependencies.

asford:
It appears that python-libarchive-c was previously being pulled in as a transitive dependency during CI test runs, but is not longer being implicitly installed in the test environment. 

Adds python-libarchive-c, which is a hard requirement for conda-build, into the conda recipe run requirements, matching the existing dependency specified in setup.py.


msarahan:
Thanks!

PR #3753, opened 2019-10-02
Hide values of environment variables

saraedum:
these sometimes contain secrets that leak when conda-build is used in a CI environment, see https://github.com/conda-forge/conda-smithy/issues/1162.

### DIFF
--- a/tests/test_api_build.py
+++ b/tests/test_api_build.py
@@ -1408,3 +1408,26 @@ def test_ignore_verify_codes(testing_config):
     # this recipe intentionally has a license error.  If ignore_verify_codes works,
     #    it will build OK.  If not, it will error out.
     api.build(recipe_dir, config=testing_config)
+
+
+def test_script_env_warnings(testing_config, recwarn):
+    recipe_dir = os.path.join(metadata_dir, '_script_env_warnings')
+    token = 'CONDA_BUILD_PYTEST_SCRIPT_ENV_TEST_TOKEN'
+    def assert_keyword(keyword):
+        messages = [str(w.message) for w in recwarn.list]
+        assert any([token in m and keyword in m for m in messages])
+        recwarn.clear()
+
+    api.build(recipe_dir, config=testing_config)
+    assert_keyword('undefined')
+
+    os.environ[token] ='SECRET'
+    try:
+        api.build(recipe_dir, config=testing_config)
+        assert_keyword('SECRET')
+
+        testing_config.suppress_variables = True
+        api.build(recipe_dir, config=testing_config)
+        assert_keyword('<hidden>')
+    finally:
+        os.environ.pop(token)
--- a/conda_build/cli/main_build.py
+++ b/conda_build/cli/main_build.py
@@ -344,8 +344,7 @@ different sets of packages."""
 
     p.add_argument('--suppress-variables',
                    action='store_true',
-                   help=("Do not display value of environment variables specified in build.script_env."),
-    )
+                   help=("Do not display value of environment variables specified in build.script_env."), )
 
     add_parser_channels(p)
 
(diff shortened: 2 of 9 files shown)


saraedum:
Should I add a news entry for a small change like this?

jjhelmus:
Can this be suppression be placed behind a configuration option or flag?  I find the listing of environment variables contents very useful when debugging and would prefer to keep them but understand the need to suppress them in cases when leaking secrets is a concern.

jjhelmus:
An alternative is add `export PYTHONWARNINGS=ignore` before calling `conda build`.  This will suppress printing of all warning messages.

saraedum:
@jjhelmus: Thanks for the feedback. What do you think about something like this? I could go all the way and allow `script_env` to optionally be a dictionary `{ name: … , hidden: true }` but I am not sure if that would be overengineering it a bit here?

saraedum:
> An alternative is add `export PYTHONWARNINGS=ignore` before calling `conda build`. This will suppress printing of all warning messages.

I guess it's the same argument here that this might hide some warnings that people would like to see. But sure, it's a workaround.

marcelotrevisani:
Maybe an option which will be passed by the command line would be good. What do you think?
Something like
```
conda build --suppress-variables
```

saraedum:
If you guys don't mind that I add another command line switch that also works of course.

saraedum:
@marcelotrevisani @jjhelmus Does this look alright to you now?

marcelotrevisani:
> @marcelotrevisani @jjhelmus Does this look alright to you now?

I think that would be a nice addition. However, could you please fix those failing tests? They are related to the ``--suppress-variables``

saraedum:
Note that `build.sh` needs to be around for the warning to print. Not sure why.

saraedum:
@marcelotrevisani would you mind having another look now?

tests/test_api_build.py
```diff
@@ -1408,3 +1408,26 @@ def test_ignore_verify_codes(testing_config):
     # this recipe intentionally has a license error.  If ignore_verify_codes works,
     #    it will build OK.  If not, it will error out.
     api.build(recipe_dir, config=testing_config)
+
+
+def test_script_env_warnings(testing_config, recwarn):
```
marcelotrevisani: Thanks for adding this test! I really appreciate that! :)

tests/test_api_build.py
```diff
@@ -1408,3 +1408,26 @@ def test_ignore_verify_codes(testing_config):
...
+    def assert_keyword(keyword):
+        messages = [str(w.message) for w in recwarn.list]
+        assert any([token in m and keyword in m for m in messages])
+        recwarn.clear()
+
+    api.build(recipe_dir, config=testing_config)
+    assert_keyword('undefined')
+
+    os.environ[token] ='SECRET'
+    try:
```
marcelotrevisani: I just think that the ``try`` here is not necessary, is it?

saraedum: I don't know pytest well enough. Just thought I'd clean up after myself.

saraedum: So, how does this work? I don't need to cleanup or I should just drop the `try - finally` part of it?

marcelotrevisani: I think that is fine then, as it is used just here should be ok
However, you can create a fixture and do the setup and teardown there. But as you are just adding and removing an env variable that is ok
:p

marcelotrevisani:
For me that is fine! As @jjhelmus was reviewing it as well, I would like to see if he also agree with the modifications

msarahan:
Thanks @saraedum @jjhelmus @marcelotrevisani for the hard work on this.

PR #3758, opened 2019-10-07
Initial artifact signing code

awwad:
Draft of code to perform artifact signing during indexing.

**Please note** that the `authenticate.py` module will not be included in the final conda-build code, as it will instead be imported via `conda.exports` (from `conda.common.authenticate`).

### DIFF
--- a/conda_build/index.py
+++ b/conda_build/index.py
@@ -70,14 +70,6 @@ CONDA_TARBALL_EXTENSIONS = ('.conda', '.tar.bz2')
 log = get_logger(__name__)
 
 
-
-
-# # DEBUG ONLY
-# test_dir = ''
-
-
-
-
 # use this for debugging, because ProcessPoolExecutor isn't pdb/ipdb friendly
 class DummyExecutor(Executor):
     def map(self, func, *iterables):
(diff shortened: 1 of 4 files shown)


conda_build/index.py
```diff
@@ -244,9 +251,11 @@ def _ensure_valid_channel(local_folder, subdir):
...
 
 
-def update_index(dir_path, check_md5=False, channel_name=None, patch_generator=None, threads=MAX_THREADS_DEFAULT,
-                 verbose=False, progress=False, hotfix_source_repo=None, subdirs=None, warn=True,
-                 current_index_versions=None, debug=False):
+def update_index(
+        dir_path, check_md5=False, channel_name=None, patch_generator=None,
+        threads=MAX_THREADS_DEFAULT, verbose=False, progress=False,
+        hotfix_source_repo=None, subdirs=None, warn=True,
+        current_index_versions=None, debug=False, indexer_signing_key=None):
```
soapy1: How does a user specify an index_signer_key to use for creating an index?

soapy1:
LGTM. Would be nice to sort out getting the authenticate module from conda before merging this in though

awwad:
Definitely -- just wanted a sense of whether things made sense.  TY.

Issue #3754, opened 2019-10-04
load_setup_py_data should cd to the directory containing setup.py


jakirkham:
### Actual Behavior

Currently `load_setup_py_data` does not appear to `cd` to where the `setup.py` file lives. This is a non-issue if the `setup.py` is in the top-level of the working directory as in PR ( https://github.com/conda-forge/metawrap-feedstock/pull/9 ). However if the source directory is a subdirectory as in PR ( https://github.com/conda-forge/metawrap-feedstock/pull/10/files ), the lack of using `cd` results in files appearing as missing. The result is things like versioneer end up failing when a subdirectory is involved and `load_setup_py_data` is used.

### Expected Behavior

It would be nice if we `cd`ed to the directory where the `setup.py` file lives before running it.

### Steps to Reproduce

This recipe encounters the issue. Dropping the `src` folder "fixes" the issue.

```yaml
# filename: recipe/meta.yaml

{% set name = "metawrap" %}
{% set data = load_setup_py_data(setup_file="src/setup.py") %}
{% set version = data.get("version") %}

package:
  name: {{ name|lower }}
  version: {{ version }}

source:
  git_url: https://github.com/jakirkham/metawrap
  git_rev: HEAD
  folder: src

build:
  noarch: python
  number: 1
  script: python -m pip install --no-deps --ignore-installed .

requirements:
  host:
    - pip
    - python
    - setuptools

  run:
    - python

test:
  source_files:
    - tests

  imports:
    - metawrap

  commands:
    - python -m unittest discover -s .

about:
  home: http://github.com/jakirkham/metawrap
  license: BSD 3-Clause
  license_family: BSD
  license_file: LICENSE.txt
  summary: A collection of wrappers for functions and classes.
  doc_url: https://metawrap.readthedocs.io/
  dev_url: http://github.com/jakirkham/metawrap

extra:
  recipe-maintainers:
    - jakirkham
```

##### `Output of conda info`

```
$ conda info

     active environment : base
    active env location : /opt/conda
            shell level : 1
       user config file : /home/conda/.condarc
 populated config files : /home/conda/.condarc
          conda version : 4.7.12
    conda-build version : 3.18.9
         python version : 3.7.3.final.0
       virtual packages : 
       base environment : /opt/conda  (writable)
           channel URLs : https://conda.anaconda.org/conda-forge/linux-64
                          https://conda.anaconda.org/conda-forge/noarch
                          https://repo.anaconda.com/pkgs/main/linux-64
[...]

PR #3759, opened 2019-10-07
Test with `setup.py` in a subdirectory

jakirkham:
Adds a test case like the one in issue ( https://github.com/conda/conda-build/issues/3754 ).

cc @marcelotrevisani

### DIFF
--- a/tests/test-recipes/metadata/source_setup_py_data_subdir/bld.bat
+++ b/tests/test-recipes/metadata/source_setup_py_data_subdir/bld.bat
@@ -1,0 +1,11 @@
+if not exist .git exit 1
+git config core.fileMode false
+if errorlevel 1 exit 1
+git describe --tags --dirty
+if errorlevel 1 exit 1
+for /f "delims=" %%i in ('git describe') do set gitdesc=%%i
+if errorlevel 1 exit 1
+echo "%gitdesc%"
+if not "%gitdesc%"=="1.21.0" exit 1
+echo "%PKG_VERSION%"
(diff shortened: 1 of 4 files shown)


jakirkham:
If there is a way to mark this as an xfail, would be happy to make that change so we can get it in. 🙂

marcelotrevisani:
> If there is a way to mark this as an xfail, would be happy to make that change so we can get it in. slightly_smiling_face

Sure, I can mark it as Xfail if you want to. Just to be sure, this PR is to cover the issue which you described, right? Do you want to fix it or just want to have this test case that reproduces the error to another person to fix it?

jakirkham:
That would be great! Thanks.

Yes. I haven’t managed to fix it yet, but maybe the test helps someone fix it in the future.

jakirkham:
Looks like some the AppVeyor tests are hanging and thus timing out on the CI. Is this expected? Doesn't seem related to this PR.

marcelotrevisani:
> Looks like some the AppVeyor tests are hanging and thus timing out on the CI. Is this expected? Doesn't seem related to this PR.

It is not related to this PR indeed. Sometimes some tests take more time to finish and we have some strict timeout on ``appveyor``. I was trying to run the tests using azure for windows as well, but I am seeing some problems regarding the ``git`` env variables and ``cython`` which for some reason is not using the MSVC correctly.

marcelotrevisani:
is it ready to be integrate this test case?

jakirkham:
Yep. Thanks @marcelotrevisani! 😀

jakirkham:
Thanks for the review 🙂

PR #3764, opened 2019-10-08
FIX: bug using multiple outputs

neok-m4700:
Because `m.config.host_prefix` is wipped out when using multiple outputs, use `prefix_files_backup` instead.

Reproduce the bug using following recipe: I expect the archive `my-pkg-1.0-0.tar.bz2` to contain `bin/exe` and `lib/liba.so`.

**meta.yaml**
```yaml
{% set name = 'my-pkg' %}
{% set version = '1.0' %}

package:
  name: {{name}}
  version: {{version}}

build:
  script:
    - mkdir -p $PREFIX/bin $PREFIX/lib
    - touch $PREFIX/bin/exe
    - touch $PREFIX/lib/liba.so

outputs:
  - name: {{name}}
    run_exports:
      - lib{{name}}
  - name: lib{{name}}
    files:
      - lib
```

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -1542,7 +1542,7 @@ def build(m, stats, post=None, need_source_download=True, need_reparse_in_env=Fa
 
                 if (top_level_meta.name() == output_d.get('name') and not (output_d.get('files') or
                                                                            output_d.get('script'))):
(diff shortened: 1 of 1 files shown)


mingwandroid:
Thank you for the PR, I may be running into this, testing your fix.

neok-m4700:
Likely outdated.

Issue #3774, opened 2019-10-16
file warning given on a directory


scopatz:
### Actual Behavior

```
base scopatz@artemis ~/staged-recipes/recipes systemd $ conda build systemd/
/home/scopatz/miniconda/lib/python3.7/site-packages/conda_build/utils.py:1209: UserWarning: RECIPE_PATH received is a file. File: systemd/meta.yaml
It should be a path to a folder. 
Forcing conda-build to use the recipe file.
  UserWarning
```

### Expected Behavior

The warning should not be there, since I passed in a dir path, not a path to meta.yaml

##### `Output of conda info`

```
$ conda info                                                                       

     active environment : base
    active env location : /home/scopatz/miniconda
            shell level : 1
       user config file : /home/scopatz/.condarc
 populated config files : /home/scopatz/.condarc
          conda version : 4.7.12
    conda-build version : 3.18.10
         python version : 3.7.3.final.0
       virtual packages : __cuda=9.1
       base environment : /home/scopatz/miniconda  (writable)
           channel URLs : https://conda.anaconda.org/conda-forge/linux-64
                          https://conda.anaconda.org/conda-forge/noarch
                          https://repo.anaconda.com/pkgs/main/linux-64
                          https://repo.anaconda.com/pkgs/main/noarch
                          https://repo.anaconda.com/pkgs/r/linux-64
                          https://repo.anaconda.com/pkgs/r/noarch
          package cache : /home/scopatz/miniconda/pkgs
                          /home/scopatz/.conda/pkgs
       envs directories : /home/scopatz/miniconda/envs
                          /home/scopatz/.conda/envs
               platform : linux-64
             user-agent : conda/4.7.12 requests/2.22.0 CPython/3.7.3 Linux/4.18.0-25-generic ubuntu/18.10 glibc/2.28
                UID:GID : 1000:1000
             netrc file : None
           offline mode : False
```

michaelaye:
interestingly, this warning isn't given when running `condo-debug`.

scopatz:
Haha weird.  Thanks for fixing this @marcelotrevisani!

PR #3786, opened 2019-10-28
Fix UserWarnings regarding recipe path

marcelotrevisani:
Related to #3774


jjhelmus:
LGTM, thanks for the fix @marcelotrevisani

PR #3793, opened 2019-11-02
Use normpath instead of fnmatch where appropriate

chrisburr:
In #3686 `fnmatch` was used to improve the handling paths of case insensitive filesystems. I think this is buggy as files containing `*`/`?`/`[something]` will now be matched dynamically instead of literally. See https://github.com/conda/conda-build/pull/3782#discussion_r337590375 for more details.

This changes the matching to be equivalent to `normcase(a) == normcase(b)` instead. The performance is much better than `fnmatch` and completely negligible on case sensitive filesystems.

### DIFF
--- a/conda_build/inspect_pkg.py
+++ b/conda_build/inspect_pkg.py
@@ -7,10 +7,9 @@
 from __future__ import absolute_import, division, print_function
 
 from collections import defaultdict
-import fnmatch
 import json
 from operator import itemgetter
-from os.path import abspath, join, dirname, exists, basename
+from os.path import abspath, join, dirname, exists, basename, normcase
 import os
 import re
 import sys
@@ -40,11 +39,10 @@ def which_package(in_prefix_path, prefix):
     the conda packages the file came from.  Usually the iteration yields
     only one package.
     """
-    norm_ipp = in_prefix_path.replace(os.sep, '/')
-    match = re.compile(fnmatch.translate(norm_ipp)).match
+    norm_ipp = normcase(in_prefix_path.replace(os.sep, '/'))
     for dist in linked(prefix):
         dfiles = dist_files(prefix, dist)
-        if any(map(match, dfiles)):
+        if any(norm_ipp == normcase(w) for w in dfiles):
             yield dist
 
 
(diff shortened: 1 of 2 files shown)


marcelotrevisani:
Hi @chrisburr , thanks for adding this!
I think that is good to go!

mingwandroid:
I have done issues with normcase here guys. I'll report back later. I want to be able to process Linux and macOS packages on windows by using case sensitive directories. This will prevent that AFAICT.

marcelotrevisani:
> I have done issues with normcase here guys. I'll report back later. I want to be able to process Linux and macOS packages on windows by using case sensitive directories. This will prevent that AFAICT.

Oh, you are considering sensitive directories then? uuhhnn.. I didn't know that. So, I will revert this modification then. Is that enought for now @mingwandroid ?

mingwandroid:
I'm working on conda build quite heavily atm. I will figure out what to do here soon. OK with having this merged for now so long as we think carefully about it before any release.

marcelotrevisani:
> I'm working on conda build quite heavily atm. I will figure out what to do here soon. OK with having this merged for now so long as we think carefully about it before any release.

Sure! Sorry about this! :sweat_smile: 
If you think it is better to revert it for now, please feel free to merge the
https://github.com/conda/conda-build/pull/3794
Which reverts this modifications

mingwandroid:
For example I have recently made sure you can test Linux packages on macOS (including e.g ncurses which has files of the same name in different cases). I've also enabled rerunning the post stuff on these tests, optionally. I see no reason the same shouldn't work on windows given it has better case sensitivity capabilities now (and anyone running docket or sharing folders with vms should use this).

I'm thinking to add detection of when we're case sensitive and to obey that. Then we can use normcase otherwise. I hope not to degrade the performance in either case. We'll see.

marcelotrevisani:
> For example I have recently made sure you can test Linux packages on macOS (including e.g ncurses which has files of the same name in different cases). I've also enabled rerunning the post stuff on these tests, optionally. I see no reason the same shouldn't work on windows given it has better case sensitivity capabilities now (and anyone running docket or sharing folders with vms should use this).
> 
> I'm thinking to add detection of when we're case sensitive and to obey that. Then we can use normcase otherwise. I hope not to degrade the performance in either case. We'll see.

I understand it better now, thanks for the explanation! :)

Issue #3360, opened 2019-01-25
Document new features


sscherfke:
Conda-build gained a lot of new features in the lasts month but not all of them were properly documented, for example, the `missing_dso_whitelist` and `runpath_whitelist` keys for the `meta.yaml`'s build-section. And now, with v3.17 and py-lief there seems to be an additonal setting (`prefix_owners`) which is also not documented.

These things are important for people outside of conda-forge that don't build their own c-compiler but rely on some low-level system libs (like libc).

scopatz:
Bumping this! `missing_dso_whitelist` is still undocumented

rrigdon:
https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#whitelisting-shared-libraries

Thanks for the feedback!

PR #3797, opened 2019-11-04
Added whitelist info to build section

rrigdon:
- Is the heading of "Whitelisting shared libraries" correct? What should it be?
- What is the background/value/purpose of these two keys? 
- What are the example values that we should show in the code blocks?

### DIFF
--- a/docs/source/resources/define-metadata.rst
+++ b/docs/source/resources/define-metadata.rst
@@ -825,7 +825,7 @@ examining linkage information.
 
 During the post-build phase, the shared libraries in the newly created
 package are examined for linkages which are not provided by the
-package's requirements or a predefined list of libraries. If such
+package's requirements or a predefined list of system libraries. If such
 libraries are detected, either a warning ``--no-error-overlinking``
 or error ``--error-overlinking`` will result.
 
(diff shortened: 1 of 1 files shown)


rrigdon:
This is addressing this issue: https://github.com/conda/conda-build/issues/3360

docs/source/resources/define-metadata.rst
```diff
@@ -816,6 +816,38 @@ pinned down to the build string level. This will
...
+Whitelisting shared libraries
+-----------------------------
+
+The ``missing_dso_whitelist`` build key is a list of globs for
+dynamic shared object (DSO) files that should be ignored when
+examining linkage information.
+
+During the post-build phase, the shared libraries in the newly created
+package are examined for linkages which are not provided by the
+package's requirements or a predefined list of libraries. If such
```
mingwandroid: > or a predefined list of libraries

Well, this is the whitelist itself!

mingwandroid: Maybe 'a predefined list of system libraries' would be better, because conda-build does have that concept too and yup, it a DSO is in either list (in code the fixed whitelist is called `whitelist`, but I may change it to `sysroot_whitelist` to disambiguate that from `missing_dso_whitelist`).

PR #3800, opened 2019-11-06
On ASCII locales, allow UTF-8 text in scripts

zeehio:
The noarch check for binary scripts uses the locale preferred encoding to guess if the script is binary or text based.

On some CI systems, (e.g. gitlab continuous integration using the `conda/miniconda3` docker image), the default encoding is ascii (expressed as "ANSI_X3.4-1968").

In general, in any system with an ascii default encoding conda assumes "if it is not ascii then it is binary". I argue it makes more sense to say "if it is not utf-8 then it is binary". Since ascii is a subset of UTF-8, we will reduce the false positives with this change.

Other encodings are not affected by this.

### DIFF
--- a/conda_build/noarch_python.py
+++ b/conda_build/noarch_python.py
@@ -26,7 +26,11 @@ def rewrite_script(fn, prefix):
 
     # Load and check the source file for not being a binary
     src = join(prefix, 'Scripts' if ISWIN else 'bin', fn)
-    with io.open(src, encoding=locale.getpreferredencoding()) as fi:
+    encoding = locale.getpreferredencoding()
+    # if default locale is ascii, allow UTF-8 (a reasonably modern ASCII extension)
+    if encoding == "ANSI_X3.4-1968":
+        encoding = "UTF-8"
+    with io.open(src, encoding=encoding) as fi:
         try:
             data = fi.read()
         except UnicodeDecodeError:  # file is binary


zeehio:
CLA signed

msarahan:
Cla confirmed. Thanks! We'll work on a review very soon.

msarahan:
All good here.  Thanks for the explanation of this change!

mingwandroid:
I think it's better to determine binary vs text via looking for a NULL byte. That's what we do basically everywhere else. Am I missing something here? FWIW, I don't believe any other method of attempting this distinction is valid.

mingwandroid:
Note, I'm fine with this PR, as it's definitely an improvement, but I maintain the code is wrong-headed.

IMHO `locale.getpreferredencoding()` should not be used for any internal processing in `conda` nor in `conda-build`. These functions return different results based on locale, and we *never* want that. Besides, this code is inspecting Python scripts and I'm sure the interpreter already has code for reading the encoding information at the top of the files and telling us the encoding, respecting what the author has identified at the top of each file (should they decide to do that).

zeehio:
> I think it's better to determine binary vs text via looking for a NULL byte. That's what we do basically everywhere else. Am I missing something here? FWIW, I don't believe any other method of attempting this distinction is valid.

I don't know if there is any conda package using text based scripts written in UCS-2, UTF-16 or any other multibyte encoding that uses zero bytes. Those encodings are not valid on python scripts, since they would not be valid according to [PEP 263](https://www.python.org/dev/peps/pep-0263/) but I don't know about other languages used in conda packages.

PR #3802, opened 2019-11-08
Master variant values as dicts

mingwandroid:

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -316,9 +316,6 @@ def regex_files_rg(files, prefix, tag, rg, regex_rg, replacement_re,
         try:
             if utils.on_win:
                 args = [a.decode('utf-8') for a in args]
-                print(' '.join(args))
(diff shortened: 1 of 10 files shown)


mingwandroid:
@msarahan, @jjhelmus this should be ready for review. The sole Windows failure appears unrelated (permissions stuff).

@conda-forge/core : the test recipe's conda-build-config.yaml sketches out how I intend this to be used. Your thoughts/reviews would be welcome.

msarahan:
You can't ping conda-forge/core here - it only works within an org.  Probably better to announce it to them on gitter.

msarahan:
All looks good to me.  We could use tqdm for progress bars, but we can add that later.

mingwandroid:
OK, I'll ping them when the last part of this is hooked up. Trying to decide what to do about CBC.yaml files passed on the command line when running --test.

PR #3804, opened 2019-11-08
Use the generalized non-deferred regex replacement code when CBC.yaml carries it

mingwandroid:


mingwandroid:
I cannot reproduce the Linux py27 failure.

marcelotrevisani:
> I cannot reproduce the Linux py27 failure.

Hi Ray,

I saw that the test which is failing is ``test_convert_no_dependencies[package0-win]`` which it seems that it should be executed just on the ``windows`` platform. Am I right?
Because the line which triggered the problem was:
[Azure Build](https://dev.azure.com/anaconda-conda/conda-build/_build/results?buildId=52&view=logs&jobId=6c8b0c6a-f542-59e7-f531-6482f0e0cdec&taskId=b9041b71-fced-5394-e600-021f78f9e249&lineStart=1449&lineEnd=1450&colStart=1&colEnd=1)
``conda_build/convert.py``
```python
shutil.move(os.path.join(temp_dir, 'Lib'), dest_dir)
```
I think that folder ``Lib`` is just for windows right?

marcelotrevisani:
If it is a flaky error we can mark that test as ``serial``, that might be some kind of race condition which the tests are facing.

Another thing which I am thinking is to use the plugin [pytest-replay](https://github.com/ESSS/pytest-replay) to record the order of each ``pytest`` worker is executing the tests on our CI
and we can upload that as artifacts on ``azure``, for the next time it will be easier reproduce those errors locally.

msarahan:
@marcelotrevisani pytest-replay seems helpful for tracking down some of the race condition stuff.  Let's add it, or at least keep it in mind for sometime soon.

mingwandroid:
Thanks!

PR #3808, opened 2019-11-13
Some trivial fixes and Xcode avoidance (preferring our tools if found, in build prefix or base env) also warn about Apple's stub exes

mingwandroid:

### DIFF
--- a/conda_build/os_utils/macho.py
+++ b/conda_build/os_utils/macho.py
@@ -46,13 +46,13 @@ def is_macho(path):
     return bool(head in MAGIC)
 
 
-def is_dylib(path):
+def is_dylib(path, build_prefix):
     return human_filetype(path) == 'DYLIB'
 
 
-def human_filetype(path):
-    ot = find_preferably_prefixed_executable('otool')
-    output = check_output((ot, '-h', path)).decode('utf-8')
+def human_filetype(path, build_prefix):
+    otool = find_apple_cctools_executable('otool', build_prefix)
+    output = check_output((otool, '-h', path)).decode('utf-8')
     lines = output.splitlines()
     if not lines[0].startswith((path, 'Mach header')):
         raise ValueError(
@@ -146,6 +146,24 @@ def _get_matching_load_commands(lines, cb_filter):
     return result
 
 
+def find_apple_cctools_executable(name, build_prefix, nofail=False):
+    tool = find_preferably_prefixed_executable(name, build_prefix)
+    try:
+        if '/usr/bin' in tool:
+            with open(tool, 'rb') as f:
+                s = f.read()
+            if s.find(b'usr/lib/libxcselect.dylib') != -1:
+                print("WARNING :: Found `{}` but is is an Apple Xcode stub executable.".format(tool))
+                # This is not the real tool, but Apple's irritating GUI dialog launcher.
+                raise
+    except Exception as _:  # noqa
+        print("ERROR :: Failed to run `{}`.  Please use `conda` to install `cctools` into your base environment.\n"
+              "         An alternative option for users of macOS is to install `Xcode` or `Command Line Tools for Xcode`."
+              .format(tool))
+        sys.exit(1)
+    return tool
+
+
 def otool(path, build_prefix=None, cb_filter=is_dylib_info):
     """A wrapper around otool -l
 
@@ -163,7 +181,8 @@ def otool(path, build_prefix=None, cb_filter=is_dylib_info):
     Any key values that can be converted to integers are converted
     to integers, the rest are strings.
     """
-    lines = check_output([find_preferably_prefixed_executable('otool', build_prefix), '-l', path],
+    otool = find_apple_cctools_executable('otool', build_prefix)
+    lines = check_output([otool, '-l', path],
                           stderr=STDOUT).decode('utf-8')
     # llvm-objdump returns 0 for some things that are anything but successful completion.
     lines_split = lines.splitlines()
@@ -205,7 +224,7 @@ def _chmod(filename, mode):
 
 
 def install_name_tool(args, build_prefix=None, verbose=False):
-    args_full = [find_preferably_prefixed_executable('install_name_tool', build_prefix)]
+    args_full = [find_apple_cctools_executable('install_name_tool', build_prefix)]
     args_full.extend(args)
     if verbose:
         print(' '.join(args_full))
--- a/conda_build/post.py
+++ b/conda_build/post.py
@@ -498,25 +498,27 @@ def mk_relative_linux(f, prefix, rpaths=('lib',), method='LIEF'):
                 elf, method))
             method = 'LIEF'
         else:
-            if have_lief:
-                if existing != [existing_pe]:
-                    print('ERROR :: get_rpaths_raw()={} and patchelf={} disagree for {} :: '.format(
-                        existing, [existing_pe], elf))
-                # Use LIEF if method is LIEF to get the initial value?
-                existing = existing_pe.split(os.pathsep)
+            existing_pe = existing_pe.split(os.pathsep)
+    if have_lief:
+        existing2, _, _ = get_rpaths_raw(elf)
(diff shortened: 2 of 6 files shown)


mingwandroid:
@isuruf, included in this PR is a preference for our Apple toolchain binaries, either in the build prefix or in the base env.

conda_build/post.py
```diff
@@ -485,31 +485,40 @@ def mk_relative_linux(f, prefix, rpaths=('lib',), method='LIEF'):
...
 
+    existing, _, _ = get_rpaths_raw(elf)
     patchelf = external.find_executable('patchelf', prefix)
-    try:
-        existing = check_output([patchelf, '--print-rpath', elf]).decode('utf-8').splitlines()[0]
-    except CalledProcessError:
-        print('patchelf: --print-rpath failed for %s\n' % (elf))
-        return
+    if not patchelf:
+        print("ERROR :: You should install patchelf, will proceed with LIEF for {} (was {})".format(elf, method))
```
msarahan: Is it actually an error?  If not, we should change ERROR to WARNING
```suggestion
        print("WARNING :: You should install patchelf, will proceed with LIEF for {} (was {})".format(elf, method))
```

conda_build/post.py
```diff
@@ -485,31 +485,40 @@ def mk_relative_linux(f, prefix, rpaths=('lib',), method='LIEF'):
...
+        method = 'LIEF'
+    else:
+        try:
+            existing_pe = check_output([patchelf, '--print-rpath', elf]).decode('utf-8').splitlines()[0]
+        except CalledProcessError:
+            print("ERROR :: `patchelf --print-rpath` failed for {}, will proceed with LIEF (was {})".format(
```
msarahan:
```suggestion
            print("WARNING :: `patchelf --print-rpath` failed for {}, will proceed with LIEF (was {})".format(
```

conda_build/post.py
mingwandroid: Maybe if you explicitly ask for patchelf (rather than leaving it as default/None), which you'd do because LIEF is known to be failing, then I think we probably should error out?

msarahan: makes sense.  My main point is: if you show a message and continue, that's a warning.  If you show a warning and stop, that's an error.

conda_build/os_utils/macho.py
```diff
@@ -146,6 +146,23 @@ def _get_matching_load_commands(lines, cb_filter):
...
+    tool = find_preferably_prefixed_executable(name, build_prefix)
+    try:
+        if '/usr/bin' in tool:
+            with open(tool, 'rb') as f:
+                s = f.read()
+            if s.find(b'usr/lib/libxcselect.dylib') != -1:
+                # This is not the real tool, but Apple's irritating GUI dialog launcher.
+                raise
+    except Exception as e:
+        print("ERROR :: Failed to run `{}`.  Please use `conda` to install `cctools` into your base environment.\n" \
```
mingwandroid: I should change this to say "into your build or base environment" I think?

msarahan: "please add cctools to your recipe's `build` section" is the nicest, IMHO

msarahan: but if that needs to be in effectively every recipe, then we'll be better off making it a run dep of conda-build on mac.

msarahan: or maybe not of conda-build, but of the compiler package on mac

mingwandroid: It is a dep of the compiler package! Packages that do not use compilers shouldn't be inflicted with them IMHO, not even in the base env.

msarahan: Is cctools necessary if people have no compiled code?  I appreciate separating them, I just wonder if we can exclude this.

mingwandroid: No, but there's no way to get into the code flow for making DSOs relative if you don't have macOS DSOs, and if you do have them (regardless of whether self-compiled or repackaged) then we need to be able to do the required path fixups. If you don't have DSOs then no need for compilers or binutilsy things, no, but you'd never see this message. This will only trip people who do need these things but haven't taken the necessary steps to provide them.

mingwandroid: Hmm, thinking about this, I think we should say something like "please use the `compiler` jinja2 function (link) as a build requirement or else, if you are sure you do not need compilers, add `cctools` instead".

.. or something worded better than that?

msarahan: I thought about it a bit more, and this absolutely has to be in the base env, NOT in the build env.  The build env is not available to conda-build for doing stuff with.  IMHO, we should just make this a CB dep on mac.

mingwandroid: I've not seen it not be available when needed so far, can you detail how that could happen? I'd rather not introduce our compiler decisions into our build tool so deeply if I can avoid it.

mingwandroid: This code is careful to *also* fall back to Apple's tools (which is in fact all we ever ran here before, with no warnings or information .. this is all improvement already, and if the build env is always available at this point, as I think it is, then I'd rather we talk about build envs here).

I actually wanted to go further and prefer our tools for everything conda-build needs (including things like git, but that's too difficult as you need git etc before you know your deps - for post stuff we already know them), but again, it would only *ever* be a preference change iff the tools are found where we look for them.

jakirkham:
Would it be possible to update the title/description to give an overview of what is being changed here? 🙂

conda_build/os_utils/macho.py
msarahan: Your post-build link checking stuff uses this, right?  At the stage where that happens, you can't assume that cctools is still around in the build env.  You shouldn't assume that the build env is around at all.  I have always just installed xcode right away, as most others will have done, too, I imagine.  I think this is for a better future world, where people won't need xcode, right?  In that world, if they only have cctools in their build env, then your tool will (should) fail to find cctools, unless you separately install it in the base env (or wherever conda-build lives or otherwise on PATH)

mingwandroid: So do note, anyone with Xcode or Xcode command line tools (currently everyone running conda-build) will not see this message. This frees us from needing Xcode.

msarahan: Yeah, we're saying the same thing, I think.  The message is fine.  Error out if people need to install that thing.  Their choices are either installing xcode or install our package.

mingwandroid: Also our activation scripts for cctools etc need to be free to set env vars should we decide to. Having env vars in the base env for our core tools is a line I'd like not to cross.

mingwandroid: (full disclosure, our clang package sets these env vars for now, but that's because I was lazy!)

mingwandroid: .. and yes, we could indirect it through a cctools activation (or implementation) package to avoid this problem.

mingwandroid:
@jakirkham, The commits are detailed in both comments and the commit message. I don't feel the need to summarize them in the PR title.

conda_build/os_utils/macho.py
mingwandroid: One quick thing, should we have different exceptions for these errors? BuildToolMissing, BuildToolIsAppleStub?

mingwandroid: Under what condition can we not assume it's around. Post to me implies it gets run just after we do the build, and if any env destruction happened in-between then that would seem incorrect to me. Now moving build before test is totally fine and to be expected.

jakirkham:
Thanks @mingwandroid! 😄

mingwandroid:
I think this is in a good shape for merging now.

mingwandroid:
(I have another PR that is exactly this branch - at present - plus WIP macOS Azure Pipeline addition over at https://github.com/conda/conda-build/pull/3810. Help welcome.

mingwandroid:
Merging this. If there's any rough spots we can revist.

Issue #3756, opened 2019-10-04
Run tests on OSX


marcelotrevisani:
That might be done on azure pipelines as well, similar to linux.

mingwandroid:
I think we can make more of the azure pipelines yaml go away to be replaced by calls to common scripts.

marcelotrevisani:
I agree! We can definitely do some improvements on our ``yaml``!

PR #3810, opened 2019-11-14
Macos azure

mingwandroid:

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -26,13 +26,13 @@ jobs:
   - script: |
       sudo apt update
       sudo apt install attr -y
+      source ci/azurepipelines/activate_conda "/usr/share/miniconda/bin/python"
       conda info
       ps -ef | grep $$
       conda install python=$PYTHON_VERSION -y
       mkdir -p /usr/share/miniconda/locks
       mkdir -p /usr/share/miniconda/bin
       chmod -w /usr/share/miniconda/locks
-      source /usr/share/miniconda/bin/activate
       if [ "$CONDA_VERSION" = "release" ]; then
         conda update -y conda;
       else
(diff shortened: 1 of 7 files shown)


marcelotrevisani:
Hi @mingwandroid ,
I just saw that some tests are failing because of the ``otool``, which I really don't know how it works on ``osx``, ``osx`` is not my cup of tea. :sweat_smile:
```
ERROR :: Failed to run `/usr/bin/otool`.  Please use `conda` to install `cctools` into your base environment.
         An alternative option for users of macOS is to install `Xcode` or `Command Line Tools for Xcode`.
```
Should we install that ``cctools`` or should we skip the tests which need it?

I just saw that you are appending the conda path to each step, right? Doing the 
https://github.com/conda/conda-build/pull/3810/files#diff-fec826feae04e51c0d94076385408bdcR121
``eval "$($(Build.StagingDirectory)/miniconda/bin/python -m conda shell.bash hook)"``

If you want to improve it you can do something similar which we are doing for ``windows``:
https://github.com/conda/conda-build/blob/480aa03ef3826e537be8dd84699590a3d31f5b51/azure-pipelines.yml#L133
`` Write-Host "##vso[task.prependpath]$ENV:UserProfile\Miniconda3;$ENV:UserProfile\Miniconda3\Scripts;$ENV:UserProfile\Miniconda3\Library\bin"``

or if you want to you can let it as it is and I will modify it in a couple of weeks, because I have in my todo list some improvements to do on our ``Azure`` pipelines.

mingwandroid:
Is Xcode installed on these images? Does anyone know? It is correct to fail if not, and we should add an xfail test for that probably, that's skipped if Xcode is installed (and not the stub exe), probably just run the test always though so we can do these inspections (does it link with that xcselect dylib?) without creating a new fixture?

mingwandroid:
Can I do proper activation rather than prependingpath? I'm not interested in avoiding activation, it's important we do this correctly.

marcelotrevisani:
> Can I do proper activation rather than prependingpath? I'm not interested in avoiding activation, it's important we do this correctly.

Sure, as you want to do the activation is better to keep your code then

marcelotrevisani:
> Is Xcode installed on these images? Does anyone know? It is correct to fail if not, and we should add an xfail test for that probably, that's skipped if Xcode is installed (and not the stub exe), probably just run the test always though so we can do these inspections (does it link with that xcselect dylib?) without creating a new fixture?

Just for future references, the whole description of what is installed on ``osx 10.13``
https://github.com/microsoft/azure-pipelines-image-generation/blob/master/images/macos/macos-10.13-Readme.md

mingwandroid:
This is ready to go now from my side. A few failures on macOS python 3.8 @jjhelmus .. think they could be showing an issue with our builds.

mingwandroid:
Permissions thing on Windows, ok to merge @msarahan, @marcelotrevisani?

marcelotrevisani:
> Permissions thing on Windows, ok to merge @msarahan, @marcelotrevisani?

I think so!
I will mark that test as serial in a couple of days.

marcelotrevisani:
I think that is good to go from my side.
Very nice job @mingwandroid ! Thank! :smile: 
for me ``osx`` is always a pain to make it work properly

It closes #3756

PR #3815, opened 2019-11-18
Add pytest-replay and pytest-rerunfailures

marcelotrevisani:
Add [pytest-replay](https://github.com/ESSS/pytest-replay/) to help us to investigate when a test suddenly fails on CI, which might be related to some race condition. Thereby, we will be able to see which tests were running in parallel.

The only problem here is, the output of ``pytest-replay`` is different between version lower than ``1.0.0`` and versions greater or equal then ``1.0.0`` are available just for Python 3+. 
The release ``1.2.0`` added more information to the output, such as the time which the test started and finished, outcome and node id. Nonetheless, older versions just list the order of the tests which were executed by each worker.

https://github.com/ESSS/pytest-replay/#features

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -56,17 +56,17 @@ jobs:
       pip install --no-deps .
       conda info -a
       mkdir $BUILD_ARTIFACTSTAGINGDIRECTORY/pytest-replay
-      echo "##vso[task.setvariable variable=PYTEST_REPLAY_OPTIONS]"
       PYTEST_REPLAY_OPTIONS="--replay-record-dir=$BUILD_ARTIFACTSTAGINGDIRECTORY/pytest-replay"
       if [[ "$PYTHON_VERSION" != "2.7" ]]; then
         PYTEST_REPLAY_OPTIONS+=" --replay-base-name=Linux-serial-$CONDA_VERSION-Py$PYTHON_VERSION"
       fi
+      echo "##vso[task.setvariable variable=PYTEST_REPLAY_OPTIONS]$PYTEST_REPLAY_OPTIONS"
     displayName: Preparing test environment
 
   - script: |
       source ci/azurepipelines/activate_conda "/usr/share/miniconda/bin/python"
       conda install conda-verify -y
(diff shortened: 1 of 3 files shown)


azure-pipelines.yml
```diff
@@ -50,29 +50,43 @@ jobs:
                        ripgrep pyflakes beautifulsoup4 chardet pycrypto glob2 psutil pytz tqdm  \
                        conda-package-handling py-lief python-libarchive-c perl conda-forge::shellcheck
       conda install -q pytest pip pytest-cov pytest-forked pytest-xdist nomkl numpy mock pytest-mock
+      conda install -q pytest-replay -c conda-forge
```
marcelotrevisani: @msarahan , can we have ``pytest-replay`` on the main channel?

msarahan: I'll add it to the queue

azure-pipelines.yml
```diff
@@ -50,29 +50,42 @@ jobs:
                        ripgrep pyflakes beautifulsoup4 chardet pycrypto glob2 psutil pytz tqdm  \
                        conda-package-handling py-lief python-libarchive-c perl conda-forge::shellcheck
       conda install -q pytest pip pytest-cov pytest-forked pytest-xdist nomkl numpy mock pytest-mock
+      conda install -q pytest-replay pytest-rerunfailures -c conda-forge
```
marcelotrevisani: @msarahan 
``pyteste-rerunfailures`` would be nice to have on the main channels as well 
I add it to rerun some tests which were failing because of loss of connection to the ``cran`` repositories
I think that is ready to be integrated, just need someone to review it :)

tests/test_api_skeleton.py
```diff
@@ -458,6 +458,7 @@ def test_pypi_section_order_preserved(testing_workdir):
 
 @pytest.mark.slow
 @pytest.mark.parametrize("package, license_id, license_family, license_files", cran_packages)
+@pytest.mark.flaky(reruns=5, reruns_delay=1)
```
marcelotrevisani: We might use the same strategy for other tests which fails with http error

PR #3816, opened 2019-11-20
[WIP] Added placeholder page about source/no_hoist

rrigdon:

### DIFF
--- a/docs/source/user-guide/source-no-host.rst
+++ b/docs/source/user-guide/source-no-host.rst
@@ -1,0 +1,2 @@
+This page contains content that will be worked on and incorporated into
+another space.


mingwandroid:
code is here: https://github.com/conda/conda-build/blob/496808082439fc4337e8f71b73d1ae83752b6910/conda_build/source.py#L160-L163

When conda-build unpacks a source archive, if checks to see if that archive contains a single folder in its top level directory structure. If so, as a convenience to recipe scripts, this top level folder is discarded and its contents are `hoisted` into the top level unpack folder. Technically this is destructive of information (the top level folder name) which can cause problems, particularly when repackaging other projects. In the case of `conda skeleton rpm`, some packages suffered hoisting (e.g. single level /lib64 folder) while others did not (a /lib64 and a /share folder), making them mutually incompatible.

mingwandroid:
So now, you can specify `source/no_hoist` and conda-build will not perform this check, unpacking this source entry literally.

rrigdon:
The work on this is going to be handled in this PR: https://github.com/conda/conda-build/pull/3817

PR #3817, opened 2019-11-20
[WIP] Regex replacements and source/no_hoist

rrigdon:

### DIFF
--- a/docs/source/concepts/recipe.rst
+++ b/docs/source/concepts/recipe.rst
@@ -145,6 +145,30 @@ Conda-build will use this information to identify dependencies to
 link to and identify the run requirements for the package. This allows
 conda-build to understand what is needed to install the package.
 
+
+Source/no_hoist
+---------------
+
+When conda-build unpacks a source archive, if checks to see
+if that archive contains a single folder in its top-level
+directory structure. If so, as a convenience to recipe scripts,
+this top-level folder is discarded and its contents are hoisted
+into the top-level unpack folder. Technically this is destructive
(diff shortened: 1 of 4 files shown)


mingwandroid:
We often need to, after building a recipe but before packaging it, replace some strings in some (usually) generate files. On linux and macOS we often find the build prefix or the sysroot getting baked into files which causes problems because those folders are temporary. Instead, in general, we want to replace those with a token or and environment variable.

It is possible to add keys to your conda_build_config.yaml (variants stuff) called replacements/all_replacements which is a list of replacement instructions for conda-build to perform during its post build phase. Here is an example:

https://github.com/conda/conda-build/blob/496808082439fc4337e8f71b73d1ae83752b6910/tests/test-recipes/variants/31_variant_subkeys/conda_build_config.yaml#L1-L20

mingwandroid:
`regex_re` is in `PCRE2` format and is strictly as tight or tighter matching that `regex_rg` (rg==`ripgrep`). `regex_rg` is optional. If only `regex_re` is provided then it must be compatible with both `PCRE2` and `ripgrep` regex formats. If your regex is not compatible with both then you must design a simpler, possibly looser one for `regex_rg`.

`tag` is informational, conda-build will print this out.

`replacement_re` is a PRCE2 regex and *can* contain references to groups mentioned in `regex_re` to allow more complex substitutions (example not given).

`glob_patterns` is a list of glob wildcards to consider for this prefix replacement.

rrigdon:
This is waiting to be merged until future release of conda-build.

PR #3823, opened 2019-12-06
Removed features, updated track_features

rrigdon:


rrigdon:
This work adds on to work completed in https://github.com/conda/conda-build/pull/3820

PR #3826, opened 2019-12-09
WIP: CRAN skeleton map dependency

katietz:

### DIFF
--- a/conda_build/mapdeps.py
+++ b/conda_build/mapdeps.py
@@ -1,0 +1,58 @@
+"""
+Tools for doing dependency mappings for recipes
+"""
+
+import sys
+import unicodedata
+import yaml
+
+def mapdeps_load(fn):
+    rslt = []
+    try:
+      with open(fn, 'r') as stream:
+        rslt = yaml.safe_load(stream)
+    except:
+      print("Unable to load {} as yaml".format(fn))
+    mapdeps_get_for_cran(rslt, 'rcpparmadillo')
+    return rslt
+
+def mapdeps_get_for_cran(deps, crandep):
+    rslt = []
+    try:
+      v = deps['dependencies']
+      for i in v:
+        try:
+          fc = i['for_cran']
+          if crandep in fc:
+              rslt.append(i)
+              ln = mapdeps_get_dep_name(i)
+              print('{}'.format(ln))
+        except:
+          pass
+    except:
+      pass
+    return rslt
+
+def mapdeps_get_dep_name(dep):
+   name = dep['dep']
+   ver = ''
+   try:
+     ver = dep['version']
+   except:
+     pass
+   if ver != '':
+     name += ' >=' + ver
+   for_os = mapdeps_get_oslimit(dep)
+   if for_os != '':
+     name += '  # [' + for_os + ']'
+   return name
+
+def mapdeps_get_oslimit(dep):
+   rslt = ''
+   try:
+     rslt = dep['oslimit']
+   except:
+     pass
+   return rslt
+
+#end 
--- a/conda_build/mapdeps.py
+++ b/conda_build/mapdeps.py
@@ -6,17 +6,50 @@ import sys
 import unicodedata
 import yaml
 
-def mapdeps_load(fn):
+def load(fn):
     rslt = []
     try:
       with open(fn, 'r') as stream:
         rslt = yaml.safe_load(stream)
     except:
       print("Unable to load {} as yaml".format(fn))
-    mapdeps_get_for_cran(rslt, 'rcpparmadillo')
+    # get_for_cran(rslt, 'rcpparmadillo')
     return rslt
 
-def mapdeps_get_for_cran(deps, crandep):
+def get_sysreqs(deps, sysreqs):
+    rslt = []
+    strs = []
+    for d in sysreqs:
+      try:
+        v = deps['systemreq']
+        for i in v:
+           try:
+             name = i['dep'].lower()
+             if name in d.lower():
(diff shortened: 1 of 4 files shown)


conda_build/mapdeps.py
```diff
@@ -0,0 +1,107 @@
...
+     name += '  # [' + for_os + ']'
+   return name
+
+def get_oslimit(dep):
+   rslt = ''
+   try:
+     rslt = dep['oslimit']
+   except:
+     pass
+   return rslt
```
marcelotrevisani: I think you may remove those ``try`` ``except`` (as in this implementation is not treating them) and use the ``get`` method
for this specific case (which can be applied to other parts as well) that might be something like:
```python
def get_oslimit(dep):
    return dep.get("oslimit", default="")
```

or am I missing something about this implementation? I hope that helps

conda_build/mapdeps.py
```diff
@@ -0,0 +1,107 @@
...
+            if fcn in strs:
+              rslt.append(i)
+              break
+        except:
+          pass
+    except:
+      pass
+    return rslt
+
+def get_for_cran(deps, crandep):
```
marcelotrevisani: I think you can do the job here just using one list comphreension.
Something like:
```python
def get_for_cran(deps, cran_dep):
    return [i for i in deps.get("dependencies", []) if cran_dep in i.get("for_cran", "")]
```

msarahan:
@marcelotrevisani Thanks!  Kai and I are reviewing this here in person.  I asked him to put it up so I could get at the code and poke at it.  Your revisions are good suggestions.

katietz:
I modified code for making it more dense. Happy to get further comments

conda_build/skeletons/cran.py
```diff
@@ -1165,6 +1179,29 @@ def skeletonize(in_packages, output_dir=".", output_suffix="", add_maintainer=No
         if 'R' not in dep_dict:
             dep_dict['R'] = ''
 
+        os_type = cran_package.get("OS_type", '')
+        if os_type != 'unix' and os_type != 'windows' and os_type != '':
```
marcelotrevisani: I think you may use a ``if``  ``elif`` and this part can be the final ``else`` and I also think that ``print`` should be a ``log.warn`` maybe?
It will be something like:
```python
        if os_type == 'unix':
            d['skip_os'] = 'skip: True  # [not unix]'
            d["noarch_generic"] = ""
        elif os_type == 'windows':
            d['skip_os'] = 'skip: True  # [not win]'
            d["noarch_generic"] = ""
        elif os_type == '':
            d['skip_os'] = '# no skip'
        else:
            log.warn("Unknown OS_type: {} in CRAN package".format(os_type))
            d['skip_os'] = '# no skip'
            os_type = ''
```

conda_build/skeletons/cran.py
```diff
@@ -1165,6 +1179,29 @@ def skeletonize(in_packages, output_dir=".", output_suffix="", add_maintainer=No
...
+            d["noarch_generic"] = ""
+        if os_type == '':
+            d['skip_os'] = '# no skip'
+        skips = ''
+        # print("seek for {} in blg_mapdeps".format(package))
+        skips = get_for_cran_skip(glb_mapdeps, package, '')
+        for n in dep_dict:
+            skips = get_for_cran_skip(glb_mapdeps, n, skips)
+        if skips != '':
+            skips.strip()
```
marcelotrevisani: I think you forgot to reassign this to ``skips`` maybe? or maybe is it a leftover?
It is because ``skips.strip()`` return a new string and ``skips`` will remain the same. Looking for the logic, maybe that can be something like
```python
if skips.strip():
    skips = "  # [" + skips + "]"
    d["skip_os"] = "skip: True" + skips
```

conda_build/skeletons/cran.py
```diff
@@ -1284,6 +1321,14 @@ def skeletonize(in_packages, output_dir=".", output_suffix="", add_maintainer=No
                                 package_dicts.update({lower_name: {'inputs': inputs_dict}})
                                 package_list.append(lower_name)
 
+            for name in dep_dict:
+               # print('seek {} in for_cran'.format(name))
```
marcelotrevisani: Could you please delete these commented codes?

tests/test_api_skeleton.py
```diff
@@ -463,3 +463,17 @@ def test_cran_license(package, license_id, license_family, license_files, testin
...
         assert os.path.basename(m_license_file) in license_files
+
+# CRAN packages to test skip entry.
+# (package, skip_text)
+cran_os_type_pkgs = [('bigReg', 'skip: True  # [not unix]'),
+                     ('blatr',  'skip: True  # [not win]')
+                    ]
+
+@pytest.mark.parametrize("package, skip_text", cran_os_type_pkgs)
+def test_cran_os_type(package, skip_text, testing_workdir, testing_config):
```
marcelotrevisani: Thanks for the tests! :)

conda_build/skeletons/cran.py
katietz: Yes, thanks. The intention is here to have whitespace as indicator for "all".  But well, we could introduce here also the special value "all" instead to indicate that
I addressed it in current logic

conda_build/skeletons/cran.py
katietz: Removed the commented code lines

conda_build/skeletons/cran.py
katietz: Adjusted logic. thanks

Issue #3833, opened 2019-12-18
"conda index" leads to "JSONDecodeError" in "conda install|update"


sscherfke:
### Actual Behavior

Since we’ve upgraded to conda 4.7 and conda-build 3.18, we ran into lots of `JSONDecodeError`s in our build pipelines, because the repositories’ `current_repodata.json` (or other JSON files) could not be decoded.  This happened especially often in CI pipelines, when other pipelines were uploading new packages.

### Expected Behavior

Package uploading should not corrupt the Conda index for installs/updates in other pipelines.

### Analysis and Proposed Fix

The problem above should only exist if the JSON files don’t get updated in an atomic operation.

In a first attempt, I wrapped all `conda index` calls on our forge and and all HTTP GET requests for `.json` files with file locks.  This fixed the issue (but we ran into deadlocks very quickly :cat: ).

I then tried to fix the issue directly in `conda index`. This is the relevant code:

```python
# conda_build/index.py:421+16
def _maybe_write(path, content, write_newline_end=False, content_is_binary=False):
    temp_path = join(gettempdir(), str(uuid4()))

    if not content_is_binary:
        content = ensure_binary(content)
    with open(temp_path, 'wb') as fh:
        fh.write(content)
        if write_newline_end:
            fh.write(b'\n')
    if isfile(path):
        if utils.md5_file(temp_path) == utils.md5_file(path):
            # No need to change mtimes. The contents already match.
            os.unlink(temp_path)
            return False
    # log.info("writing %s", path)
    utils.move_with_fallback(temp_path, path)
    return True

# conda_build/utils.py:591+9
def move_with_fallback(src, dst):
    try:
        shutil.move(src, dst)
    except PermissionError:
        try:
            copy_into(src, dst)
            os.unlink(src)
        except PermissionError:
            log = get_logger(__name__)
            log.debug("Failed to copy/remove path from %s to %s due to permission error" % (src, dst))
```
The crux of the problem lies in [shutil.move()](https://docs.python.org/3/library/shutil.html#shutil.move) (or the `copy_into()` fallback):

> If the destination is on the current filesystem, then os.rename() is used. Otherwise, src is copied to dst using copy_function and then removed.

The index of our forge lives in `$HOME` and `/tmp` is usually a different filesystem, so there is no *atomic move* op, but a *copy into* which leads to the error described above.

IMHO, the easiest solution for this would be to place the temporary file next to its destination.  That way, we can always do an atomic move:

```diff
[...]

PR #3834, opened 2019-12-19
Prevent non-atomic writes to repodata JSON files

sscherfke:
Fixes: #3833

### DIFF
--- a/conda_build/index.py
+++ b/conda_build/index.py
@@ -14,7 +14,6 @@ import os
 from os.path import abspath, basename, getmtime, getsize, isdir, isfile, join, splitext, dirname
 import subprocess
 import sys
-from tempfile import gettempdir
 import time
 from uuid import uuid4
 
@@ -419,7 +418,9 @@ def _get_jinja2_environment():
 
 
 def _maybe_write(path, content, write_newline_end=False, content_is_binary=False):
-    temp_path = join(gettempdir(), str(uuid4()))
+    # Create the temp file next "path" so that we can use an atomic move, see
+    # https://github.com/conda/conda-build/issues/3833
+    temp_path = '%s.%s' % (path, uuid4())
 
     if not content_is_binary:
         content = ensure_binary(content)
--- a/news/index-atomic-write.rst
+++ b/news/index-atomic-write.rst
@@ -1,0 +1,4 @@
+Bug fixes:
+----------
+
+* Prevent non-atomic writes to repodata JSON files  #3833


sscherfke:
The link to the CLA stuff in the README seems to be broken: https://www.clahub.com/agreements/conda/conda-build

sscherfke:
Hm, `os.rename()` does not or can not overwrite an existing file on Windows.  I’ll look into it.

sscherfke:
Using `utils.move_with_fallback()` again …

Also signed the CLA.

CHANGELOG.txt
```diff
@@ -79,7 +85,7 @@ Enhancements:
 -------------
 
 * add --use-channeldata argument to conda render/build.
-* Extract the part in the skeletons pypi responsible to get the package metadata to a free function. 
+* Extract the part in the skeletons pypi responsible to get the package metadata to a free function.
```
marcelotrevisani: Could you please revert this file?
You can add the description and you changelog here
https://github.com/conda/conda-build/blob/master/news/

and you can also use our template
https://github.com/conda/conda-build/blob/master/news/TEMPLATE

:)

marcelotrevisani: Just add a new ``rst`` file on that folder using our template and it should be fine

conda_build/index.py
```diff
@@ -419,7 +418,9 @@ def _filter_add_href(text, link, **kwargs):
 
 
 def _maybe_write(path, content, write_newline_end=False, content_is_binary=False):
-    temp_path = join(gettempdir(), str(uuid4()))
+    # Create the temp file next "path" so that we can use an atomic move, see
+    # https://github.com/conda/conda-build/issues/3833
+    temp_path = '%s.%s' % (path, uuid4())
```
marcelotrevisani: is that possible to reproduce the error using a test?
It would be interesting to have a test for it.

CHANGELOG.txt
sscherfke: Will do that :)

conda_build/index.py
sscherfke: I’m afraid that’s no easily possible. 

The error was that `path` and the `temp_path` were on two different file systems and `temp_file` was big enough that there could happen a read on `path` while `temp_path` was being copied into it.

Now, both files are always in the same folder and thus on the same filesystem.

CHANGELOG.txt
sscherfke: Does it look okay now?

sscherfke:
Can this be merged? :)

CHANGELOG.txt
marcelotrevisani: The file is still here =p

marcelotrevisani: Could you please revert it? After that, I think that should be good to go

marcelotrevisani:
> Can this be merged? :)

Sorry for the delay @sscherfke , because of the end of the year, holidays and all that stuff we might be a bit slower than usual :smile:

CHANGELOG.txt
sscherfke: Oh, I thought I had reverted it. Tried it again. :)

sscherfke:
The changelog should now be clean.

marcelotrevisani:
Thanks for your contribution @sscherfke ! :tada:

sscherfke:
Thanks for your kind feedback 🙂

PR #3836, opened 2019-12-21
Add --r-interp-build argument

mcg1969:
Adds a new argument to `conda skeleton cran`, `--r-interp-build`. This argument has the same format as `--r-interp`, except that it is applied only to the build step.

This is useful for R because if we build for the earliest `x.x` release of R, it is generally compatible with all `x.x.y` sub-versions of R. For instance, building a package for `R 3.5.0` allows it to be compatible for all `R 3.5.*` interpreters.

If not supplied, it defaults to using the same value as `--r-interp`.

### DIFF
--- a/conda_build/skeletons/cran.py
+++ b/conda_build/skeletons/cran.py
@@ -354,6 +354,11 @@ def add_parser(repos):
         help="Declare R interpreter package",
     )
     cran.add_argument(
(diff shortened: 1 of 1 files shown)


mcg1969:
Closing this because I've discovered that the built-in `r-base` pinning does the right thing.

PR #3838, opened 2019-12-22
Optionally add full unit tests to cran recipes

mcg1969:
This implements an `--include-tests` option for `conda skeleton cran`, which adds the full unit tests to CRAN package builds. This is done by
- copying the contents of the source package's `tests/` directory into the R path
- adding the `Suggests:` dependencies to the test phase
- calling `tools::testInstalledPackage` instead of just importing the package

Windows isn't finished yet, hence the WIP

### DIFF
--- a/conda_build/skeletons/cran.py
+++ b/conda_build/skeletons/cran.py
@@ -66,6 +66,7 @@ TEST_META_WITH_UNIT = """\
   commands:
     - cp -r tests $PREFIX/lib/R/library/{cran_packagename}/  # [not win]
     - $R -e "tools::testInstalledPackage('{cran_packagename}')"  # [not win]
+    - xcopy /E tests %PREFIX%\\lib\\R\\library\\{cran_packagename}\\tests\\  # [win]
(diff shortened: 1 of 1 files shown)


mcg1969:
Windows tests added. The change to the skeleton code is non-trivial, but surprisingly the change to the _recipe_ generated is surprisingly simple.

conda_build/skeletons/cran.py
```diff
@@ -1226,28 +1252,31 @@ def skeletonize(in_packages, output_dir=".", output_suffix="", add_maintainer=No
             d['skip_os'] = '# no skip'
 
         need_git = is_github_url
-        if cran_package.get("NeedsCompilation", 'no') == 'yes':
```
mcg1969: This scan was run originally only for files that explicitly claimed no compilation. But since I need to scan this package for test files regardless, I removed this if/then. This might have the side effect that some packages that _claim_ no compilation are nevertheless given compilation dependencies because of the presence of C/Fortran files in the package. I suspect this is not an issue, but if the reviewer disagrees what I would suggest is to put a check _after_ the pass through the file to zero out `need_c`, etc. for pure-R packages

PR #3839, opened 2019-12-28
Fetch the latest CRAN archived version if requested

mcg1969:
This implements an `--allow-archived` argument for `conda skeleton cran`. When selected, it will build a recipe for the latest version of an archived package instead of failing. For instance, `conda will return the error
```
ERROR: Package graph is archived; to build, use --allow-archived or supply a --version value
```
but will succeed if `conda skeleton cran graph --allow-archived` is used.

As the error message implies, it will also work to supply an explicit version; e.g., `conda skeleton cran graph --version 1.30.0`. This is already working in master; the error message has simply been adjusted to include the `--allow-archived` option.

### DIFF
--- a/conda_build/skeletons/cran.py
+++ b/conda_build/skeletons/cran.py
@@ -400,6 +400,12 @@ def add_parser(repos):
         help="Don't include an Archive download url.",
     )
     cran.add_argument(
+        '--allow-archived',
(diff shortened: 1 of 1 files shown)


marcelotrevisani:
LGTM! Thanks for that @mcg1969  !

Issue #3835, opened 2019-12-19
Publish py38 build of conda-build?


aldanor:
Now that conda 4.8.0 is available for py38, could conda-build be possibly released as py38 as well (so e.g. it could be installed into py38 root environments).

(Skimming over the list of dependencies, it looks like the only package that will need updating is lief / py-lief)

aldanor:
@msarahan Any plans in regards to this? Given that conda is now published as py38 in anaconda channel, could conda-build catch up as well? Thanks.

marcelotrevisani:
I think we can release it to Python 3.8 as well.
However, it would be better to enable our tests for Python 3.8 first. After that, I think we can generate the packages for Python 3.8

PR #3841, opened 2020-01-02
Enable Python 3.8 on Azure Pipelines

marcelotrevisani:
Related to #3835

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -21,6 +21,12 @@ jobs:
       py37-release:
         PYTHON_VERSION: '3.7'
         CONDA_VERSION: 'canary'
+      py38-canary:
+        PYTHON_VERSION: '3.8'
+        CONDA_VERSION: 'release'
+      py38-release:
+        PYTHON_VERSION: '3.8'
+        CONDA_VERSION: 'canary'
   timeoutInMinutes: 480
   steps:
   - script: |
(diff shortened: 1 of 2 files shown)


aldanor:
Shouldn't lief / py-lief be built for py38 first? They seem to only exist for py37 atm.

marcelotrevisani:
> Shouldn't lief / py-lief be built for py38 first? They seem to only exist for py37 atm.

Yeap, you are right! We should build it for py38 first!

cc: @msarahan

mingwandroid:
Unfortunately LIEF doesn't work with Windows Python 3.8 yet and I'm not sure why. I'm working on enabling debug Python interpreters to be built with the python-feedstock recipe at present so I can debug this (and things like it) more easily.

aldanor:
@mingwandroid Any further ideas on this?

(especially now that conda's py38 builds are out, this becomes the stumbling block...)

mingwandroid:
Yup, thanks all!

aldanor:
Great stuff, thanks! (would the py38 builds appear in the default channel at some point?)

azure-pipelines.yml
```diff
@@ -21,6 +21,12 @@ jobs:
       py37-release:
         PYTHON_VERSION: '3.7'
         CONDA_VERSION: 'canary'
+      py38-canary:
+        PYTHON_VERSION: '3.8'
+        CONDA_VERSION: 'release'
+      py38-release:
+        PYTHON_VERSION: '3.8'
+        CONDA_VERSION: 'canary'
```
mrakitin, line 29: Is not it swapped? The section name does not match the conda version...

PR #3842, opened 2020-01-03
[CRAN Skeleton] fix when directory indexes are not allowed (e.g., bioconductor)

mcg1969:
This modifies the `get_cran_index` function to parse the `src/contrib/PACAKGES.gz` file instead of relying on the web server's directory index of `src/contrib/` itself. This is necessary to retrieve packages from Bioconductor. To test, try this:
```
conda skeleton cran --cran-url https://www.bioconductor.org/packages/release/bioc biocgenerics
```
Of course we should confirm that direct CRAN packages still work:
```
conda skeleton cran tergm
```
There is more work to make Bioconductor packages skeletonize perfectly—e.g., right now, the default home page for the package defaults to an invalid CRAN url. But this gets a buildable artifact.

### DIFF
--- a/conda_build/skeletons/cran.py
+++ b/conda_build/skeletons/cran.py
@@ -14,6 +14,7 @@ import re
 import subprocess
 import sys
 import hashlib
+import gzip
 
 import requests
 import tarfile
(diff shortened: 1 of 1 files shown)


conda_build/skeletons/cran.py
```diff
@@ -644,14 +645,22 @@ def get_cran_archive_versions(cran_url, session, package, verbose=True):
...
+        if package is None:
+            if line.startswith('Package: '):
+                package = line.rstrip().split(' ', 1)[1]
+        elif line.startswith('Version: '):
+            records[package.lower()] = (package, line.rstrip().split(' ', 1)[1])
+            package = None
     r = session.get(cran_url + "/src/contrib/Archive/")
+    if r.status_code in (403, 404):
+        if verbose:
+            print("Cannot fetch an archive index from %s" % cran_url)
```
marcelotrevisani: I think it would be better to use
```suggestion
            warnings.warn("Cannot fetch an archive index from %s" % cran_url, UserWarning)
```
and you will need to import 
```python
import warnings
```

marcelotrevisani: Just one more thing, do you think you can create a test to cover this scenario please? That would be awesome!

mcg1969: Will work on it!

mcg1969: I'm not sure I agree with the use of `warnings.warn` here. I really would consider this more informational than a warning. Some sources of R packages just don't have an Archive directory, _by design_. And if the Archive directory is absent it should be treated as if it was there, but empty.

Put another way, this module is screaming for a proper `logging` implementation—and this would be `logging.INFO` or maybe even `logging.DEBUG`, not `logging.WARN`.

mcg1969: Here's an example of a source of R packages with no Archive directory—http://datacube.wu.ac.at/

PR #3844, opened 2020-01-04
BUG warn on dangling symlinks

beckermr:
Closes #3844 

@isuruf for viz

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -1263,6 +1263,7 @@ def path_type(path):
 def build_info_files_json_v1(m, prefix, files, files_with_prefix):
     log = utils.get_logger(__name__)
 
+    no_link_files = m.get_value('build/no_link')
     files_json = []
     files_inodes = get_inodes(files, prefix)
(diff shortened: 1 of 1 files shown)


conda_build/build.py
```diff
@@ -1280,6 +1282,7 @@ def build_info_files_json_v1(m, prefix, files, files_with_prefix):
         elif not isfile(path):
             # this is a softlink that points to nowhere, so is zero bytes
             file_info["size_in_bytes"] = 0
+            log.warn('file %s is a symlink with no target', path)
```
marcelotrevisani:
```suggestion
            warnings.warn('file %s is a symlink with no target', path)
```
should be fine

marcelotrevisani: you will need to import warnings as well
```python
import warnings
```

beckermr:
Done @marcelotrevisani !

marcelotrevisani:
Thanks again for your contribution! :)

PR #3863, opened 2020-01-31
Be robust against spaces in %SRC_DIR%.

anntzer:
The build directory simply needs to be quoted.
This can be tested by trying to build any recipe after installing conda
in a path that contains spaces (e.g., if your username contains spaces).

### DIFF
--- a/conda_build/windows.py
+++ b/conda_build/windows.py
@@ -234,7 +234,7 @@ def write_build_scripts(m, env, bld_bat):
             data = fi.read()
         with codecs.getwriter('utf-8')(open(work_script, 'wb')) as fo:
             fo.write('IF "%CONDA_BUILD%" == "" (\n')
-            fo.write("    call {}\n".format(env_script))
+            fo.write('    call "{}"\n'.format(env_script))
             fo.write(')\n')
             fo.write("REM ===== end generated header =====\n")
             fo.write(data)
--- a/news/spaces-in-path.rst
+++ b/news/spaces-in-path.rst
@@ -1,0 +1,5 @@
+Bug fixes:
+----------
+
+* Improved support for cases where conda is installed on Windows in a path that
+  contains a space.


anntzer:
Kindly bumping.
The test failure seems spurious, and the CLA was signed on Jan. 31st.

jjhelmus:
@cla-bot check

anntzer:
Kindly bumping again...

mingwandroid:
Thanks for the PR and the ping, sorry it took so long to get back to you!

My stance is that many build tools (e.g. GNU make) do not handle paths in spaces and therefore conda-build should never be performed in such a compilation root. You should use the --croot option to specify a build directory that does not have this issue.

If you want to update this PR so that it *also* warns strongly when a space is detected then I am "OK" with this, even if it sends out a signal that we think it is valid to try to attempt to compile any significant percentage of software (on *any* OS) in a path containing spaces.

anntzer:
> Thanks for the PR and the ping, sorry it took so long to get back to you!

No worries, thanks for the review.

> My stance is that many build tools (e.g. GNU make) do not handle paths in spaces

My understanding is that this just depends on whether the makefile handles quotation carefully, although I may be wrong.

> and therefore conda-build should never be performed in such a compilation root. You should use the --croot option to specify a build directory that does not have this issue.  If you want to update this PR so that it also warns strongly when a space is detected then I am "OK" with this, even if it sends out a signal that we think it is valid to try to attempt to compile any significant percentage of software (on any OS) in a path containing spaces.

Thanks for pointing out the existence of `--croot`.  It may be cleaner to just detect at the very beginning if there's spaces in the workdir, and if so, immediately error out cleanly with a suggestion to use `--croot`?

mingwandroid:
> My understanding is that this just depends on whether the makefile handles quotation carefully, although I may be wrong

I am afraid in general this is untrue. No one takes the care (either humans or tools like autotools ad cmake) take care of this when generating Makefiles. The same is likely true of ninja, meson and the other Makefile generators too but I'm not sure. Either way, autotools is incapable of supporting this use-case.

mingwandroid:
> Thanks for pointing out the existence of --croot. It may be cleaner to just detect at the very beginning if there's spaces in the workdir, and if so, immediately error out cleanly with a suggestion to use --croot?

For me, maybe an error (configurable but defaulting to true, override-able from both cmdline and from within meta.yaml (e.g. `build/can_handle_spaces`), or this PR plus a strong warning would be great.

anntzer:
Given that I now know the workaround :) I'll leave the new tasks to someone else for now if anyone wants to pick them up, but may revisit this in the future if no one does.  Feel free to take over the PR, or to close it.

mingwandroid:
OK thanks!

PR #3875, opened 2020-02-11
allow passing in post-metadata-augmenting function

msarahan:
This is for Anaconda-internal use.  We use it to add additional metadata beyond what the normal post-metadata adds.


jjhelmus:
This is no longer needed, closing.

PR #3889, opened 2020-03-02
Support VAR=VAL in build/script_env

mingwandroid:


mingwandroid:
New macOS failure is because we've now enabled / fixed the parallel tests on that platform, still it is something I should try to fix. I'll do that in the background, but this PR represents an improvement on the status quo on CI even if it means we're not fully green on CI because of it.

PR #3890, opened 2020-03-02
Less os.path in post.py

mingwandroid:


mingwandroid:
Failures are same on macOS as previously and an intermittent permissions thing on Windows. Unlikely due to this change so merging.

PR #3891, opened 2020-03-02
Fix for globbing test files on Windows

mingwandroid:


mingwandroid:
Flakiness on macOS is on my radar.

PR #3893, opened 2020-03-03
Change behaviour of build/binary_has_prefix_files

mingwandroid:
If specified as a list, then only those files in that list are
considered eligible for binary prefix replacement.

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -947,8 +947,9 @@ def get_files_with_prefix(m, files_in, prefix):
 
 def record_prefix_files(m, files_with_prefix):
 
+    filtered = []
     if not files_with_prefix:
-        return
+        return filtered
 
     # Copies are made to ease debugging. Sorry.
     binary_has_prefix_files = m.binary_has_prefix_files()[:]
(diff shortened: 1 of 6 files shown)


mingwandroid:
Please review @jjhelmus, @forrestwaters, @msarahan, @isuruf if you have some spare clocks.

The reason to want this feature is because otherwise binary prefix replacement is an all or nothing thing on Unix, and when it's an all thing (in general, by default), conda is prevented from hardlinking any binaries in the package that contain the prefix even though this may be harmless.

I want to reduce the core cost of a miniconda env, then a conda-build env after that.

msarahan:
Seems reasonable to me.. you might want a minor or major version bump, just in case this is still something that anyone uses for whatever bizarre reason.

jjhelmus:
LGTM.  I agree with @msarahan this should be released as conda-build 3.19.0 or 4.0.0, my vote would be 3.19.0.  In addition this new behavior should be added to the documentation.

The failing unit tests look un-related.

mingwandroid:
Yes, I agree. FWIW, my uncontroversial changes since Nov 2019 were committed to the latest release tag. I've held off on ones that change behavior or add features deliberately so we can think about version numbers.

mingwandroid:
Thanks for the reviews!

mingwandroid:
Regarding documentation, good point, so I took a look, it seems the docs document the behavior this PR implements anyway:

```
* Any binary file containing the build prefix can automatically
  be registered in ``info/has_prefix`` using
  ``build/detect_binary_files_with_prefix`` in ``meta.yaml``.
  \Alternatively, individual binary files can be registered by
  listing them in ``build/binary_has_prefix_files`` in
  ``meta.yaml``. The registered files will have their build
  prefix replaced with the install prefix at install time. This
  works by padding the install prefix with null terminators, such
  that the length of the binary file remains the same. The build
  prefix must therefore be long enough to accommodate any
  reasonable installation prefix. On macOS and Linux, conda-build
  pads the build prefix to 255 characters by appending
  ``_placehold``\s to the end of the build directory name.
```

conda_build/build.py
```diff
@@ -1194,9 +1198,9 @@ def create_info_files(m, files, prefix):
     write_info_files_file(m, files)
 
     files_with_prefix = get_files_with_prefix(m, files, prefix)
+    files_with_prefix = record_prefix_files(m, files_with_prefix)
```
mbargull: This broke `prefix_placeholder` recording in `paths.json`.
I don't know what the proper fix is and can't investigate this currently.
xref: https://github.com/conda-forge/conda-build-feedstock/pull/131

mbargull: Also, when someone fixes this, I strongly suggest to add a test that checks for the `prefix_placeholder` entries to avoid regressions.

mingwandroid:
Looking into this next. Sorry for the trouble. It does need more tests, yes.

PR #3897, opened 2020-03-05
Use patchelf by default

mingwandroid:


stuarteberg:
Somewhat related: It seems like `get_rpaths_raw()` does not return a list-of-lists as expected.  That results in spurious warnings like this one:
https://github.com/conda-forge/staged-recipes/pull/10979#issuecomment-595403326

(Not sure if this PR is the right place for a fix, though.)

PR #3901, opened 2020-03-11
Clarify 'deletes the build env' in concepts/recipe

timsnyder:
I don't see anywhere in the docs where it is explained why the SOURCE_DIR, and build/host environments are deleted at the beginning of test phase (or in between build and test phase).  This commit captures that reasoning so that I can point here for new people learning about conda-build.

- [x] I'll push a NEWS file in just a sec.  I made this edit with the web ui and I don't see an easy way to create a new news file there.


jjhelmus:
This looks great, thanks for putting it together.  Since this is a docs fix I'm not worried about the CI failures.

PR #3903, opened 2020-03-13
 Fix issue 3899

mingwandroid:

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -37,9 +37,9 @@ jobs:
       mkdir -p /usr/share/miniconda/bin
       chmod -w /usr/share/miniconda/locks
       if [ "$CONDA_VERSION" = "release" ]; then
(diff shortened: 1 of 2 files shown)


mingwandroid:
I cannot figure out why the AP builds are not being run here. Do you have any idea @jjhelmus?

marcelotrevisani:
That is odd, indeed
Let's try to close and reopen this PR to see it is going to trigger it

marcelotrevisani:
@mingwandroid 
I think I fixed it with
https://github.com/conda/conda-build/pull/3904/files

But the explanation of why this way is not working anymore... I really don't know.
Microsoft might drop the way to trigger branches

mingwandroid:
Failures are connection things. Please review @jjhelmus

jjhelmus:
LGTM, thanks for the fix @mingwandroid

PR #3912, opened 2020-03-25
Ignore patching Mach-O files if not running on macOS

chrisburr:
As found in https://github.com/conda-forge/cmake-feedstock/pull/115#issuecomment-603472268 the Linux CMake build produces a Mach-O binary which then triggers this assertion and causes the build to fail.

This is a regression in 3.18 as this function used to never be called on other platforms: https://github.com/conda/conda-build/pull/3808/files#diff-60400bbbdb51124a0a8a06ca17ecec5cL1105

cc @jakirkham

### DIFF
--- a/conda_build/post.py
+++ b/conda_build/post.py
@@ -396,7 +396,9 @@ def osx_ch_link(path, link_dict, host_prefix, build_prefix, files):
 
 
 def mk_relative_osx(path, host_prefix, build_prefix, files, rpaths=('lib',)):
-    assert sys.platform == 'darwin'
+    if sys.platform != 'darwin':
+      log.warn("Found Mach-O file but patching is only supported on macOS, skipping: %s", path)
+      return
     prefix = build_prefix if exists(build_prefix) else host_prefix
     names = macho.otool(path, prefix)
     s = macho.install_name_change(path, prefix,
--- a/news/dont-fix-macho-files-on-other-platforms.rst
+++ b/news/dont-fix-macho-files-on-other-platforms.rst
@@ -1,0 +1,25 @@
+Enhancements:
+-------------
+
+* <news item>
+
+Bug fixes:
+----------
+
+* Show a warning instead of failing if a Mach-O file is prouduced by a build running on a platform other than macOS.
+
+Deprecations:
+-------------
+
+* <news item>
+
+Docs:
+-----
+
+* <news item>
+
+Other:
+------
+
+* <news item>
+


mingwandroid:
Would removing the assert and building install_name_tool in Linux not be a better approach here?

jakirkham:
This seems pretty clean and simple to me. Thanks @chrisburr 😀

Why do we need `install_name_tool` on Linux?

mingwandroid:
Because seems to be the case here we d are build broken macOS software on Linux.

chrisburr:
> Would removing the assert and building install_name_tool in Linux not be a better approach here?

Definitely and conda-forge already has it in `cctools`, though the binaries are prefixed like `x86_64-apple-darwin13.4.0-otool` and `x86_64-apple-darwin13.4.0-install_name_tool`. I'm not sure how best to handle the prefix? Also what if a Mach-O binary appears during a build on Windows?

As it stands the import is commented out so other changes will also be needed:
https://github.com/conda/conda-build/blob/79da1715907bdbcaadd896f6517e3d398896f038/conda_build/post.py#L39-L40

> Because as seems to be the case here we are building broken macOS software on Linux.

In the case of CMake it's actually okay as the binary in question doesn't need to be modified to make it relocatable so we're not actually building broken software. That said, there is the potential to but I think it's still better than the current situation where it's impossible to build CMake.

mingwandroid:
Thanks!

jakirkham:
Thanks Chris and Ray! 😄

conda_build/post.py
```diff
@@ -396,7 +396,9 @@ def osx_ch_link(path, link_dict, host_prefix, build_prefix, files):
 
 
 def mk_relative_osx(path, host_prefix, build_prefix, files, rpaths=('lib',)):
-    assert sys.platform == 'darwin'
+    if sys.platform != 'darwin':
+      log.warn("Found Mach-O file but patching is only supported on macOS, skipping: %s", path)
+      return
```
mrakitin, line 401: Is it a standard style (2 blank space indents) here in conda-build? I saw 4 space indents in many other codes of the code base.

jakirkham: It should be 4 spaces. Can you please send a PR to fix it?

jjhelmus: PR #3925 adjusts the indentation and fixes another bug introduced here.

mrakitin: Thank you!

PR #3928, opened 2020-04-06
jsonify any files in RECIPE_DIR/info_yaml.d

mingwandroid:
Rather than forcing people to always use `extras` metadata to augment packages, give another alternative. `*.yaml` in `join(dirname(meta.yaml), 'info_yaml.d')` are turned into `info/info_json.d/*.json`).

Ping @kalefranz

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -736,6 +736,30 @@ def copy_readme(m):
                   "as README.md and README.rst", file=sys.stderr)
 
 
+def jsonify_info_yamls(m):
+    iyd = "info_yaml.d"
+    ijd = "info_json.d"
+    src = join(dirname(m.meta_path), iyd)
+    res = []
+    if os.path.exists(src) and isdir(src):
+        for root, dirs, files in os.walk(src):
+            for file in files:
+                file = join(root, file)
+                bn, ext = os.path.splitext(os.path.basename(file))
+                if ext == '.yaml':
+                    dst = join(m.config.info_dir, ijd, bn+'.json')
+                    try:
+                        os.makedirs(os.path.dirname(dst))
+                    except:
+                        pass
+                    with open(file, 'r') as i, open(dst, 'w') as o:
+                        import yaml
+                        yaml = yaml.full_load(i)
+                        json.dump(yaml, o, sort_keys=True, indent=2, separators=(',', ': '))
+                        res.append(join(os.path.basename(m.config.info_dir), ijd, bn+'.json'))
+    return res
+
+
 def copy_license(m):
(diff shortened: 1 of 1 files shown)


conda_build/build.py
```diff
@@ -736,6 +736,30 @@ def copy_readme(m):
                   "as README.md and README.rst", file=sys.stderr)
 
 
+def jsonify_info_yamls(m):
+    iyd = "info_yaml.d"
+    ijd = "info_json.d"
```
kalefranz, line 741: I think `info/extra.d/*.json` would be better than `info/info_json.d/*.json`. I think you mentioned that `extra` implies the contents aren't as important. To me, `extra` means more "overflow" for information that's not more-strictly encoded in the files directly in the `info/` directory. A contrived scenario might be if someone creates an `info/info_json.d/index.json` file? How would a sophisticated yet new-to-the-ecoystem user know how to interpret that? If it were me, I think I'd interpret that `info/index.json` is really the source of truth and `info/extra.d/index.json` is something else. Perhaps we can find a better word than `extra` though?

conda_build/build.py
```diff
@@ -736,6 +736,30 @@ def copy_readme(m):
...
+    if os.path.exists(src) and isdir(src):
+        for root, dirs, files in os.walk(src):
+            for file in files:
+                file = join(root, file)
+                bn, ext = os.path.splitext(os.path.basename(file))
+                if ext == '.yaml':
+                    dst = join(m.config.info_dir, ijd, bn+'.json')
+                    try:
+                        os.makedirs(os.path.dirname(dst))
+                    except:
```
kalefranz, line 753: Use `except EnvironmentError` (equivalent to `except (IOError, OSError)`). If you want a catch-all, best practice is to always use `except Exception`. A naked `except` catches `KeyboardInterrupt` and several other exceptions that you really don't want to be catching.

conda_build/build.py
```diff
@@ -736,6 +736,30 @@ def copy_readme(m):
...
+                        pass
+                    with open(file, 'r') as i, open(dst, 'w') as o:
+                        import yaml
+                        yaml = yaml.full_load(i)
```
kalefranz, line 757: What the reason not to just use `safe_load()` here? Maybe best to add the reasoning as a comment in the code.

mingwandroid:
Thanks Kale, all good suggestions. Let's field the naming question though?

kenodegard:
superseded by https://github.com/conda/conda-build/pull/3952 (https://github.com/conda/conda-build/pull/3952/commits/5c3cf7e4febd411b4c45a7657221d70dcc07df6a)

PR #3942, opened 2020-04-19
conda index cache

kalefranz:
Refactors the conda index cache to create a single cache directory per package, rather that splitting individual metadata files out among multiple disparate directories.

Makes `cytoolz` and `scandir` (py27) hard dependencies of conda-build for performance reasons.

Once released, this PR will induce a one-time re-caching for all existing long-lived conda-index caches. One mitigating factor is that now, we extract and cache the _whole_ `info/` directory, so that within the cache we have access to the full metadata of a package. Even if we need to add extra information into an aggregate file like repodata.json or channeldata.json (or a new one) in the future, we won't have to re-extract all packages ever again.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -264,7 +264,7 @@ jobs:
       python -c "import sys; print(sys.executable)"
       python -c "import sys; print(sys.prefix)"
       call conda update -q --all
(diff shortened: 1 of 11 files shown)


dholth:
This is a good idea but we're preparing to replace the cache with a per-subdir sqlite3 database. In that branch we try to extract as few files from info/ as possible for speed.

PR #3952, opened 2020-05-12
12 3.19.4 bits

mingwandroid:

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -1025,7 +1025,7 @@ def record_prefix_files(m, files_with_prefix):
                     if fn in text_has_prefix_files:
                         text_has_prefix_files.remove(fn)
                 else:
-                    ignored_because = " :: Not in build/%s_has_prefix_files" % (mode)
+                    ignored_because = " (not in build/%s_has_prefix_files)" % (mode)
 
                 print("{fn} ({mode}): {action}{reason}".format(fn=fn, mode=mode,
                                                                action="Ignoring" if ignored_because else "Patching",
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -2474,6 +2474,8 @@ def _write_test_run_script(metadata, test_run_script, test_env_script, py_files,
             test_env_script=test_env_script))
         if utils.on_win:
             tf.write("IF %ERRORLEVEL% NEQ 0 exit 1\n")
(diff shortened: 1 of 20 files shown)


mingwandroid:
@jjhelmus preview, I will also check some community PRs once you (and ci) are happy with all this!

conda_build/build.py
```diff
@@ -973,12 +998,13 @@ def record_prefix_files(m, files_with_prefix):
 
         print("Files containing CONDA_PREFIX")
         print("-----------------------------")
-        detect_binary_files_with_prefix = m.get_value('build/detect_binary_files_with_prefix', False)
+        detect_binary_files_with_prefix = m.get_value('build/detect_binary_files_with_prefix',
+                                                      not len_binary_has_prefix_files and not utils.on_win)
```
jjhelmus, line 1002:
```suggestion
                                                      len_binary_has_prefix_files and not utils.on_win)
```
The default should be True when there are binary prefix files and the platform is not windows

mingwandroid: > when there are binary prefix files

This is not the thing being checked here. `not len_binary_has_prefix_files` means the user hasn't specified any binary files, so we need to check them all (on non Windows).

The goal is to be able to control which binaries get prefix replaced and which do not. I think this change is consistent with the documentation: https://docs.conda.io/projects/conda-build/en/latest/resources/make-relocatable.html#making-packages-relocatable

> Any binary file containing the build prefix can automatically be registered in info/has_prefix using build/detect_binary_files_with_prefix in meta.yaml. Alternatively, individual binary files can be registered by listing them in build/binary_has_prefix_files in meta.yaml. The registered files will have their build prefix replaced with the install prefix at install time. This works by padding the install prefix with null terminators, such that the length of the binary file remains the same. The build prefix must therefore be long enough to accommodate any reasonable installation prefix. On macOS and Linux, conda-build pads the build prefix to 255 characters by appending _placeholds to the end of the build directory name.

I will look at any test failures soon.

jjhelmus: Yes, you are completely correct.  I mis-understood the `len_binary_has_prefix_files` variable.

mingwandroid:
I think this is good to go now. If you have time and agree, I'll look over community PRs after some food.

PR #3953, opened 2020-05-14
12 3.19.4 bits

mingwandroid:


mingwandroid:
@jjhelmus some clean-up. Should we release to canary first? I would like to hook this up to our ci for adventurous people so we can test some package builds.

I also want to flip the default (for us, not globally) to error for overlinking/overdepending, then switch that globally in 3.20 if it seems to be working well.

msarahan:
probably better to bump minor version when making that change. If it ends up breaking people, it's easier to tell them to stick with the earlier minor release.

msarahan:
oh, sorry Ray, I misread that. I restated exactly what you said. =(

jjhelmus:
For the last few releases I have been using `rever` to update the CHANGLOG.txt file from entries in the `news` folder automatically.  Prior to a release, I've added files to the news folder for change that were missing as needed, e.g.  f1e1852bf0ec879512c70c9f5725f983c7f05f56.  

I would be okay if we moved away from `rever` for releases but it does automate much of the hassle of making a release.

jjhelmus:
> I also want to flip the default (for us, not globally) to error for overlinking/overdepending, then switch that globally in 3.20 if it seems to be working well.

Builds submitted to concourse have had the `--error-overlinking` option added except on Windows for some time.  Were you thinking of extending to include `--error-overdepnding` on all platforms before switching the default in conda build.  I think conda-forge has a similar configuration.

The CLI does indicate the `--error-overlinking` will be the default in conda-build 4.0.  If we make the switch I think we should move the major version as is documented.

jjhelmus:
> Should we release to canary first?

Releasing to canary sounds good.  conda-forge will pick up the release as well and can decide if they want to release it or not.

mingwandroid:
Please review @jjhelmus and @chenghlee

I'll add news.d entries soon then take some time off.

Thanks for taking care of the final release.

mingwandroid:
The failures look intermittent too. Let's see if the news.d entries trigger greener builds.

PR #3957, opened 2020-05-21
ENH make sure packages in current repo data w/ features have versions without features

beckermr:
This PR makes sure that packages in the current repo data which have features also have version without features. This change makes sure that feature minimization works.

cc @isuruf @jjhelmus

### DIFF
--- a/conda_build/index.py
+++ b/conda_build/index.py
@@ -718,7 +718,7 @@ def _add_prev_ver_for_features(new_r, orig_r):
                     keep_m = _m
                     break
             if keep_m is not None:
-                expanded_groups[g_name].add(keep_m)
+                expanded_groups[g_name] = set([keep_m]) | set(expanded_groups.get(g_name, []))
 
     return [pkg for group in expanded_groups.values() for pkg in group]
 
--- a/conda_build/index.py
+++ b/conda_build/index.py
@@ -720,8 +720,6 @@ def _shard_newest_packages(subdir, r, pins=None):
             for m in matches
             if not (m.track_features or m.features)
         )
-        for m in matches:
-            print(m, m.track_features, m.features, m in groups[g_name])
 
     new_r = _get_resolve_object(subdir, precs=[pkg for group in groups.values() for pkg in group])
     return set(_add_missing_deps(new_r, r))
@@ -730,7 +728,6 @@ def _shard_newest_packages(subdir, r, pins=None):
 def _build_current_repodata(subdir, repodata, pins):
     r = _get_resolve_object(subdir, repodata=repodata)
     keep_pkgs = _shard_newest_packages(subdir, r, pins)
-    print([k.name for k in keep_pkgs])
     new_repodata = {k: repodata[k] for k in set(repodata.keys()) - set(['packages', 'packages.conda'])}
     packages = {}
     conda_packages = {}
--- a/conda_build/index.py
+++ b/conda_build/index.py
@@ -705,12 +705,12 @@ def _add_prev_ver_for_fratures(new_r, orig_r):
 
         # versions are sorted here so this is the latest
         latest_version = VersionOrder("%s" % new_r.groups[g_name][0].version)
-        if g_name in orig_r.grouos:
+        if g_name in orig_r.groups:
             # now we iterate through the list to find the next to latest
             # without a feature
             keep_m = None
             for i in range(len(orig_r.groups[g_name])):
-                _m = orig_r.grouos[g_name][i]
+                _m = orig_r.groups[g_name][i]
                 if (
                     VersionOrder(_m.version) < latest_version and
                     not (_m.track_features or _m.features)
--- a/conda_build/index.py
+++ b/conda_build/index.py
@@ -742,7 +742,7 @@ def _shard_newest_packages(subdir, r, pins=None):
             for pin_value in pins[g_name]:
                 version = r.find_matches(MatchSpec('%s=%s' % (g_name, pin_value)))[0].version
                 matches.update(r.find_matches(MatchSpec('%s=%s' % (g_name, version))))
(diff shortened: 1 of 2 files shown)


beckermr:
So the osx test that failed passes locally for me and the windows failures appear unrelated but I don't quite follow them. 

@marcelotrevisani do you know what is going on here?

beckermr:
Ah it appears all of these tests are failing on master too. So I think this one is ready to go!

conda_build/index.py
```diff
@@ -715,14 +715,22 @@ def _shard_newest_packages(subdir, r, pins=None):
...
                 version = r.find_matches(MatchSpec('%s=%s' % (g_name, pin_value)))[0].version
                 matches.update(r.find_matches(MatchSpec('%s=%s' % (g_name, version))))
-        groups[g_name] = matches
+        groups[g_name] = set(
+            m
+            for m in matches
+            if not (m.track_features or m.features)
+        )
+        for m in matches:
+            print(m, m.track_features, m.features, m in groups[g_name])
```
beckermr, line 724:
```suggestion
```

conda_build/index.py
```diff
@@ -715,14 +715,22 @@ def _shard_newest_packages(subdir, r, pins=None):
             for pin_value in pins[g_name]:
                 version = r.find_matches(MatchSpec('%s=%s' % (g_name, pin_value)))[0].version
                 matches.update(r.find_matches(MatchSpec('%s=%s' % (g_name, version))))
-        groups[g_name] = matches
+        groups[g_name] = set(
+            m
+            for m in matches
+            if not (m.track_features or m.features)
+        )
+        for m in matches:
```
beckermr, line 723:
```suggestion
```

conda_build/index.py
```diff
@@ -715,14 +715,22 @@ def _shard_newest_packages(subdir, r, pins=None):
...
+            print(m, m.track_features, m.features, m in groups[g_name])
+
     new_r = _get_resolve_object(subdir, precs=[pkg for group in groups.values() for pkg in group])
     return set(_add_missing_deps(new_r, r))
 
 
 def _build_current_repodata(subdir, repodata, pins):
     r = _get_resolve_object(subdir, repodata=repodata)
     keep_pkgs = _shard_newest_packages(subdir, r, pins)
+    print([k.name for k in keep_pkgs])
```
beckermr, line 733:
```suggestion
```

beckermr:
Bump here! I fixed one bug in the pr and it appears to be working now in my tests of the new sysroot stuff for conda forge.

jjhelmus:
I'm not certain excluding packages with features entirely from `current_repodata.json` is a good idea.  Excluding packages from `current_repodata` could result in packages without features having un-met dependencies because the dependency has a feature.

I think a better solution is to treat features similar to "pins". That is, if package with features is included in `current_repodata` the newest version package without features should also be include, provided that one exists.

beckermr:
> I'm not certain excluding packages with features entirely from current_repodata.json is a good idea. Excluding packages from current_repodata could result in packages without features having un-met dependencies because the dependency has a feature.

So, in this case, the solver will fall back to the full repodata, but that is a performance hit. Is the performance the concern here?

beckermr:
> I think a better solution is to treat features similar to "pins". That is, if package with features is included in current_repodata the newest version package without features should also be include, provided that one exists.

This should work.

jjhelmus:
> So, in this case, the solver will fall back to the full repodata, but that is a performance hit. Is the performance the concern here?

I'm not worried about performance, more my worry is that if all of the packages for a particular name have a feature then that entire package is excluded `current_repodata` which may cause the solver to prefer the solution from `current_repodata` over something that is more logical.  

As an example:  Currently `conda create -n test blas=*=openblas numpy` will give a solution using `current_repodata`.  With this proposed change the `blas-1.0-openblas` package would not be included in `current_repodata` and conda should fall back to the full repodata.  

My concern is that there will be some other solution that conda does find in current_repodata for these edge cases that will make no sense.  That said, I cannot provide a non-contrived concrete example.  Expanding, `current_repodata` to always include a non-feature package seems like a lower risk than removing all feature packages.

beckermr:
Ahhhhh thank you! That makes sense to me!

beckermr:
ok @jjhelmus this is ready for another look!

conda_build/index.py
```diff
@@ -696,6 +696,33 @@ def _add_missing_deps(new_r, original_r):
...
 
+def _add_prev_ver_for_features(new_r, orig_r):
+    expanded_groups = copy.deepcopy(new_r.groups)
+    for g_name in new_r.groups:
+        if not any(m.track_features or m.features for m in new_r.groups[g_name]):
+            # no features so skip
+            continue
+
+        # versions are sorted here so this is the latest
+        latest_version = VersionOrder("%s" % new_r.groups[g_name][0].version)
```
beckermr, line 707:
```suggestion
        latest_version = VersionOrder(new_r.groups[g_name][0].version)
```

conda_build/index.py
```diff
@@ -696,6 +696,33 @@ def _add_missing_deps(new_r, original_r):
...
+        if g_name in orig_r.groups:
+            # now we iterate through the list to find the next to latest
+            # without a feature
+            keep_m = None
+            for i in range(len(orig_r.groups[g_name])):
+                _m = orig_r.groups[g_name][i]
+                if (
+                    VersionOrder(_m.version) <= latest_version and
```
beckermr, line 715:
```suggestion
                    VersionOrder(str(_m.version)) <= latest_version and
```

beckermr:
Bump here @jjhelmus

beckermr:
Bump here

jjhelmus:
LGTM thanks again for putting this together @beckermr

PR #3972, opened 2020-06-14
Add tests for non-utf-8 charsets

hmaarrfk:
Adds tests for binary files that contain many charsets, utf16 and utf32

xref: https://github.com/conda/conda/pull/9946
xref: https://github.com/conda/conda/issues/4298


hmaarrfk:
@mingwandroid thanks for the easy pointer.

I hope that this test is satisfactory. I left the xfail commented out so that you may inspect the failure logs.

PR #3976, opened 2020-06-23
Clarify usage

oleg-alexandrov:
To use 'root' as an env is a bit confusing, given that it invokes suggestions of admin rights on Unix, and the whole point of conda is to avoid using the system libraries. 

Here making clear that 'root' is just a name.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -16,7 +16,7 @@ Installation
     # Display information about current conda install
     conda info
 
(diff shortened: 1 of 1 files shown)


jjhelmus:
@cla-bot check

jjhelmus:
Thanks for the change.

PR #3982, opened 2020-07-03
Teake/git describe hash on branches

teake:
GIT_DESCRIBE_HASH was not set when there is no tag reachable
from the head of the git repo being build. (This is because
`git describe --tags --long` fails in that case). With this change,
a `git describe --all --long` command is used to fetch the short
hash and store it in GIT_DESCRIBE_HASH. That way,
GIT_DESCRIBE_HASH will be available to use in recipes
also in cases when there is no tag reachable.

I opted to use `git describe --all --long` for this, and not
`git rev-parse --short` because the latter omits the leading `g`
that is prefixed with git-describe.

Note that the resulting `conda_build.environ.get_git_info()` function
is not super DRY.

### DIFF
--- a/conda_build/environ.py
+++ b/conda_build/environ.py
@@ -200,6 +200,23 @@ def get_git_info(git_exe, repo, debug):
         )
         log.debug(msg)
 
+    # If there was no tag reachable from HEAD, the above failed and the short hash is not set.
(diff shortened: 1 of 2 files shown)


jjhelmus:
LGTM, thanks for the fix @teake

PR #3987, opened 2020-07-08
Remove conda-forge from test deps

mingwandroid:
They dropped Python 2.7 support earlier than AD did so pytest etc are too old


jjhelmus:
LGTM, thanks @mingwandroid

PR #4014, opened 2020-08-14
Allow python=3.8 for pypi skeletons

PhilReinhold:
Just what it says on the lid.


marcelotrevisani:
There is an alternative for ``conda skeleton`` if you would like to try. [grayskull](https://github.com/marcelotrevisani/grayskull) generates recipes for projects on pypi
=p

PR #4017, opened 2020-08-18
get tests passing again

jjhelmus:
planemo package requires Python>=3.5


jjhelmus:
a623427 is needed as a follow up to #4013

jjhelmus:
This PR does not get all the tests passing but makes some progress.  Merging and will return to fix the rest later.

PR #4010, opened 2020-08-10
14 for new release

mingwandroid:
@jjhelmus, @ktietz, still quite WIP (final content is the same as 13-New-features.3, so we have features that are still WIP) .. I have squashed the commits into sensible commits that can be judged independently though.

PR #4026, opened 2020-08-22
Various changes

jjhelmus:
cherry-picked commits from PR #4010 that are ready to be merged.

### DIFF
--- a/conda_build/source.py
+++ b/conda_build/source.py
@@ -52,7 +52,7 @@ def download_to_cache(cache_folder, recipe_path, source_dict, verbose=False):
     for hash_type in ('md5', 'sha1', 'sha256'):
         if hash_type in source_dict:
             if source_dict[hash_type] in (None, ""):
-                raise ValueError('Empty %s hash provided for %s'.format(hash_type, fn))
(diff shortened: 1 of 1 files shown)


wolfv:
I believe that the call to `create_info_files` in line 2853 will fail right now because the API changed (needs a replacement object).

wolfv:
I don't understand what this replacements stuff is about actually, and why it was necessary to change the API in this way? I understand that these are internal functions but still :)

Is there some docs on the replacement?

mingwandroid:
You can just ignore it until it's documented.

mingwandroid:
Without the CBC keys to drive it it does nothing, yes the code is used, but for the old prefix replacement stuff, which still works as before.

I will change internal APIs occasionally in the interest of cleanup, sorry.

mingwandroid:
Ah I see what you mean about line 2853, will fix it, thanks

mingwandroid:
https://github.com/conda/conda-build/pull/4041

wolfv:
just FYI we're using the same functions in boa, that's why I noticed that it now fails :)

mingwandroid:
OK, sorry about the breakage! Good luck with boa.

PR #4044, opened 2020-09-08
Master 3.20.2 release

mingwandroid:

### DIFF
--- a/.authors.yml
+++ b/.authors.yml
@@ -414,7 +414,7 @@
 - name: Ray Donnelly
   email: mingw.android@gmail.com
   github: mingwandroid
-  num_commits: 558
+  num_commits: 567
   first_commit: 2016-03-06 18:52:51
 - name: Paul Madden
   email: maddenp@colorado.edu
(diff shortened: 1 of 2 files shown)


mingwandroid:
Ping @jjhelmus. If this is merged, then we can tag 3.20.2 (though merging might do that? I dunno, tags I never fully grok!)

jjhelmus:
For releasing with rever the CHANGELOG.txt should not be edited directly.  Rather create a file in the news directory.  `rever` will take care of creating the tag, updating the CHANGELOG.txt and uploading a static tarball to the GH release entry.

mingwandroid:
It wasn't edited directly, I just skipped adding news entries to git and did it locally. Does it matter?

mingwandroid:
The only workflow difference between this and 3.20.0 was that I made a PR this time instead of just merging to master.

jjhelmus:
> The only workflow difference between this and 3.20.1 was that I made a PR this time instead of just merging to master.

I'm following now.  I think it it better push to master.  A PR will create a merge commit which will not align with the tag that rever creates.

kenodegard:
[conda-build 3.20.2](https://github.com/conda/conda-build/releases/tag/3.20.2) was released in September 2020, this is no longer needed

PR #4045, opened 2020-09-08
Refactor recipe search & conda_build.api.build

njalerikson:
- Both `conda_build.api.build` and `conda_build.utils.find_recipe` perform checks for *.AppleDouble* and `os.path.isfile`, removed checks from former and rely on `IOErrors` raised in latter.
- Extracted a constant `VALID_METAS` that contains all valid conda recipe meta filenames (I went ahead and included the option to specify both `.yaml` or `.yml` extensions, I understand this doesn't make this possible to use these alternative extensions yet as `meta.yaml` is still hardcoded in a number of places).
- Updated `conda_build.utils.rec_glob` to include the option to pass in an `ignores` argument which removes directories we do not wish to traverse (this simplifies ignoring *.AppleDouble*).
- Updated `conda_build.utils.find_recipe` to better check for base level meta files and ensured that we return the full path to the meta file (not the directory as was previously done if an actual meta file was provided).
- To simplify `conda_build.api.build` make sure recipe paths are absolute prior to searching for meta file such that all recipes returned are already absolute.
- Added ensure_list in `conda_build.utils.expand_globs` such that we can pass in single paths, a minor convenience improvement.
- All of this allows `conda_build.api.build` to be refactored with streamlined and simplified logic and fully harness the duck typing made possible by updates to `conda_build.utils.find_recipe`.

### DIFF
--- a/conda_build/api.py
+++ b/conda_build/api.py
@@ -158,7 +158,6 @@ def build(recipe_paths_or_metadata, post=None, need_source_download=True,
 
     If recipe paths are provided, renders recipe before building.
     Tests built packages by default.  notest=True to skip test."""
-
     import os
     from conda_build.build import build_tree
     from conda_build.conda_interface import string_types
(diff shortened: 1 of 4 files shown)


njalerikson:
Wow, I wasn't aware of how hardcoded `meta.yaml` has become in conda-build. This PR doesn't magically fix that issue but I will continue to work towards that end in future PRs.

mingwandroid:
Many thanks for the many recent PRs @njalerikson, your efforts are appreciated.

They are on my radar and I will review them as soon as I can.

njalerikson:
No rush, I'm just slowly stepping my way through conda_build trying to fully understand what's going on. I know there are some significant PRs currently open that suggest y'all are working to finalize a release so I don't expect my changes to be looked at until those are addressed.

My bad if my force pushes are pinging people, just trying to make sure that the PRs don't get too stale.

mingwandroid:
3.20.* was a fairly big release. We're thinking about what to do for 4 though we need to think about deprecation cycles for some things we'd like to clean out. Your thoughts on deprecations are welcome. The old `py=37` stuff is on my list of things to get rid off, along with merging build and host envs and some odd behaviour around envs when you don't specify a host env that date back to supporting 'legacy' recipes.

njalerikson:
@mingwandroid just wanted to see if there's anything I can do to help move PRs forward

mingwandroid:
Thanks @njalerikson, much appreciated.

PR #4053, opened 2020-09-14
Switch to using CONDA_PACKAGE_EXTENSIONS

njalerikson:
I see that `conda` switched from using `CONDA_TARBALL_EXTENSIONS` to `CONDA_PACKAGE_EXTENSIONS` over a year ago.

This change doesn't alter anything in `conda_build` beyond renaming the old `CONDA_TARBALL_EXTENSIONS` variables and removing duplicate code.

### DIFF
--- a/conda_build/api.py
+++ b/conda_build/api.py
@@ -399,7 +399,7 @@ def debug(recipe_or_package_path_or_metadata_tuples, path=None, test=False,
     import time
     from conda_build.conda_interface import string_types
     from conda_build.build import test as run_test, build as run_build
-    from conda_build.utils import CONDA_TARBALL_EXTENSIONS, on_win, LoggingContext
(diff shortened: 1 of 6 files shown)


njalerikson:
Not 100% what's causing the macOS py37 failure, while it looks like it may be related to changes in this PR as it has to do with the tar ball:

```
conda_package_handling.exceptions.InvalidArchiveError: Error with archive /private/tmp/cb/popen-gw2/test_env_creation_fail_exits_b0/osx-64/_post-link-fails-1.0-0.tar.bz2.  You probably need to delete and re-download or re-create this file.  Message from libarchive was:
Truncated tar archive
```

None of the other tests are failing and I don't see how any of the changes in this PR could cause this to occur without seeing additional issues in a number of other tests?

mingwandroid:
Thanks @njalerikson

PR #4055, opened 2020-09-16
Refactor conda_build.build.get_all_replacements

njalerikson:
Refactor to make code readable and to remove duplicate logic. Also removed what appears to be debugging print statements that are no longer in use.

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -859,52 +859,34 @@ def write_hash_input(m):
         json.dump(recipe_input, f, indent=2)
 
 
-def get_all_replacements(config_or_variant):
-    # This function tests that our various
-    '''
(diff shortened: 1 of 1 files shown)


mingwandroid:
The debugging code is there because I suspect there is still a bug in here .. I'd like to defer merging this one or else I'll need to add debugging back to it, which isn't too bad..

njalerikson:
Yeah, that's what I was suspecting when I saw how new the code was, no hard feelings if you reject this then :)

mingwandroid:
It's no problem, I appreciate the help and will give this a proper review when I can.

mingwandroid:
It is still plenty debuggable like this, thanks @njalerikson

PR #4062, opened 2020-09-18
ENH raise runtimeerror instead of calling sys.exit

### LINKED ISSUE #4061: conda build calls `sys.exit` when downloading the source
### Actual Behavior

If you use the `provide` function and a patch cannot be applied to the source, conda build calls `sys.exit`.

See https://github.com/conda/conda-build/blob/master/conda_build/source.py#L615

### Expected Behavior

conda-build should raise a suitable error.

##### `Output of conda info`

This happens for the latest version of conda build in my CI jobs so I do not have the info output.

```

```

cc @mingwandroid @isuruf @scopatz @CJ-Wright

beckermr:
Closes #4061 

cc @mingwandroid @jjhelmus

### DIFF
--- a/conda_build/source.py
+++ b/conda_build/source.py
@@ -612,7 +612,7 @@ def apply_patch(src_dir, path, config, git=None):
 
     exception = None
     if not isfile(path):
-        sys.exit('Error: no such patch: %s' % path)
+        raise RuntimeError('Error: no such patch: %s' % path)
 
     if config.verbose:
         stdout = None


conda_build/source.py
```diff
@@ -612,7 +612,7 @@ def patch_or_reverse(patch, patch_args, cwd, stdout, stderr):
 
     exception = None
     if not isfile(path):
-        sys.exit('Error: no such patch: %s' % path)
+        raise RuntimeError('Error: no such patch: %s' % path)
```
ocefpaf, line 615: You can raise `SystemExit` and keep "similar" to what happened before with a nicer way to catch. However, I agree that this is more like a run time error instead.

beckermr: Right and agreed a `RuntimeError` is better. The issue is if a user is catching errors, they cannot use `Exception` since `SystemExit` doesn't inherit from that. Asking users to catch `SystemExit` is dangerous and definitely a python anti-pattern.

mingwandroid:
Thanks @beckermr

PR #4066, opened 2020-09-23
Do not clobber config argument in conda_build.build.build_tree

njalerikson:
Instead of clobbering `config` passed into `conda_build.build.build_tree` use a per iteration `cfg` variable.


mingwandroid:
Thanks @njalerikson

Issue #4054, opened 2020-09-14
conda-build should/could detect already applied patches and issue errors or warnings about them instead of reversing them


beckermr:
Sometimes a patch will get applied in reverse. I am not 100% clear on when this happens, but it happens pretty silently and is usually not the desired behavior. This behavior can generate bad packages. See eg https://github.com/AnacondaRecipes/conda-build-feedstock/pull/7.

So my questions are 

1. Is this the desired behavior?
2. Should we add a flag to turn this off if desired?
3. Do some recipes rely on this?

cc @isuruf

mingwandroid:
Patching is complicated. Even modern tools like git implement unfortunate behavior WRT patching. I had to rewrite the patching code in conda-build quite a lot at 3.20 time.

The reasons were:

In unix-land, all line-endings are LF and everything is fine. On Windows, batch files and VS project files and more often have line-endings ending in CRLF. In general, the old patching code could handle a patch with uniform LF or CRLF just fine; conda-build tries various attempts to get a patch to apply changing line-endings from one type to another and trying again. This generally worked fine. However if it encountered a patch where some bits changed a file with LF and other bits changed a file with CRLF then trouble would ensue. The patch would apply partially, and other bits would not apply. The process would fail and the source tree would be left in an indeterminate state.

That's ok, provided we reverse the patch, so I think what you are seeing is the logs reporting that has happened. The failed patch gets reversed (which correctly deletes .rej files and undoes the bits that managed to get applied). The patch gets reversed and various other attempts are made to make it work.

One thing I took particular care over is to ensure that if a patch is entirely 'native', i.e. it correctly contains patch-blocks that contain LF when appropriate and CRLF when appropriate (i.e. potentially mixed), then it will apply first time without problem.

For a concrete example of a patch that is mixed in LF/CRLF patching-ness (and was therefore conda-build before 3.20 could not apply) see the python 3.8 (master) recipe on AnacondaRecipes.

Maybe I should hide the messages but I'd rather point all this stuff out, there's no need for us to have messy patches.

I have a really poor python script that wraps git format-patch to post-correct the line-endings. It does this by finding the first one in the given source file and doesn't allow for patching blocks in source files that are *also* mixed in LF / CRLF, but it works for my purposes: https://github.com/AnacondaRecipes/aggregate/tree/master/git-format-patch-le.py

mingwandroid:
I am not sure what happens in the situation where a patch has already been applied, that could trip up my new code.

[...]

PR #4067, opened 2020-09-27
Use dry-run when testing if patches apply

mingwandroid:


mingwandroid:
Should fix https://github.com/conda/conda-build/issues/4054

PR #4068, opened 2020-09-28
Master.dry run patch

mingwandroid:

### DIFF
--- a/conda_build/source.py
+++ b/conda_build/source.py
@@ -585,27 +585,19 @@ def apply_patch(src_dir, path, config, git=None):
         #
         import tempfile
         temp_name = os.path.join(tempfile.gettempdir(), next(tempfile._get_candidate_names()))
-        patch_args.append('-r')
-        patch_args.append(temp_name)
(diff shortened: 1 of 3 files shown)


mingwandroid:
@jjhelmus @isuruf, fixes https://github.com/conda/conda-build/issues/4054

jjhelmus:
LGTM, thanks @mingwandroid

mbargull:
The comment
```
        # 3. When patch indicates failure, we call it with -R to reverse the damage.
```
is now out of sync ;).
I haven't checked: there is some test coverage for failing patches, right?

But the code looks fine to me -- happy to see that the simple `--dry-run` seems to work out!

realthunder:
I got a recent build failure which probably relate to the change here. Please note that `patch --dry-run` may fail if the patch modifies the same file more than once, even if the patch applies fine when you do it for real. See [here](https://savannah.gnu.org/bugs/?25860), which also mentions the complication when doing reverse patch. Modifying the same file multiple times is common when squashing multiple commits into one patch file.

PR #4070, opened 2020-09-29
Fix comment and function name to try_apply_patch

mingwandroid:


mingwandroid:
cc @mbargull, @jjhelmus 

may as well correct this while we are here.

jjhelmus (approved):

mbargull (approved):

PR #4075, opened 2020-09-30
Refactor dict_of_lists_to_lists_of_dict

njalerikson:
Refactor code to eliminate overly verbose and complex code. 6-12 times faster implementation.

I sought to eliminate a number of private helper functions that did not aid in simplifying the code. I also ended up adding a new function `conda_build.utils.islist` which we may find valuable to use elsewhere.

Yes, I like using list/dictionary comprehension and when formatted properly I believe they are easier to read - I totally understand how subjective this is so let me know if you disagree and want it coded differently.

I personally wish to rename `dict_of_lists_to_lists_of_dict` to `expand_variants` (and `lists_of_dict_to_dict_of_lists` to `collapse_spec`) in order to make the functions more self documented as well as making them less of a mouthful. @mingwandroid what do you think? I have no problem reverting this change.

I also moved all spec validation checks to the `validate_spec` function instead of having them spread out in various locations. This may require that `validate_spec` is rerun after combining specs in places like this:
https://github.com/conda/conda-build/blob/d8fa0b3e4edf3d98452f8fcfc381853e9955a47f/conda_build/variants.py#L517-L525

Not 100% sure some of the assumptions I made are valid, need to dig a little further into corner cases. For example, instead of joining zip_keys on some special char ("#") and then splitting the keys after the Cartesian Product, I opted to use `tuples` as the keys and at least with my test cases I get the correct result but this is likely only safe to do if all of specs keys are strings to start with.

### DIFF
--- a/conda_build/utils.py
+++ b/conda_build/utils.py
@@ -1175,12 +1175,74 @@ def package_has_file(package_path, file_path, refresh=False):
 
 
 def ensure_list(arg):
-    if (isinstance(arg, string_types) or not hasattr(arg, '__iter__')):
-        if arg is not None:
-            arg = [arg]
-        else:
-            arg = []
-    return arg
+    """
+    Ensure the object is a list. If not return it in a list.
+
+    :param arg: Object to ensure is a list
+    :type arg: any
+    :return: `arg` as a `list`
+    :rtype: `list`
+    """
+    if arg is None:
+        return []
+    elif islist(arg):
+        return arg
+    else:
+        return [arg]
+
+
+def islist(arg, uniform=False):
+    """
+    Check whether `arg` is a `list`. Optionally determine whether the list elements
+    are all uniform.
+
+    When checking for generic uniformity (`uniform=True`) we check to see if all
+    elements are of the first element's type (`type(arg[0]) == type(arg[1])`). For
+    any other kinds of uniformity checks are desired provide a uniformity function:
+
+    .. code-block:: python
+        # uniformity function checking if elements are str and not empty
+        >>> truthy_str = lambda e: isinstance(e, str) and e
+        >>> islist(["foo", "bar"], uniform=truthy_str)
+        True
+        >>> islist(["", "bar"], uniform=truthy_str)
+        False
+        >>> islist([0, "bar"], uniform=truthy_str)
+        False
+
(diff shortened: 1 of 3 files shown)


mingwandroid:
I approve very much @njalerikson! Thanks for the {clean,speed}-ups!

I would like some more reviewers on this one though if possible, @msarahan and @jjhelmus. I've always found this code a bit scary to work with personally!

njalerikson:
I see that 2 tests are failing:

The first is failing because of the aforementioned change where all spec validation occurs separately from `dict_of_lists_to_lists_of_dict` in `validate_spec`.

The second is failing as zip_keys specify keys that are not defined in `spec`. This too is a condition that is caught in `validate_spec`.

I don't want to update these tests before I get confirmation that these overall changes are desired.

msarahan:
Looks like a great refactor to me. Thanks for taking the time to clean it up.

conda_build/utils.py
```diff
@@ -1175,12 +1175,74 @@ def package_has_file(package_path, file_path, refresh=False):
...
+
+    :param arg: Object to ensure is a list
+    :type arg: any
+    :return: `arg` as a `list`
+    :rtype: `list`
+    """
+    if arg is None:
+        return []
+    elif islist(arg):
+        return arg
```
njalerikson, line 1189: So I'm returning any iterable here (which is what the old `ensure_list` also did), this results in a number of places where we end up doing `list(ensure_list(obj))` to avoid generator issues and the likes.

Do we want to simply wrap this in `list` and just get it over with?

Doing so was also proposed in PR #3979.

mingwandroid: Sounds fine to me.

njalerikson:
Ok, I've dug further and better understand how variants are processed from config files or from the command line argument `--variants`. I haven't found anything that invalidates the assumptions I've made with this PR but I do see room for further improvement. I have a similarly large PR almost ready to go once this is accepted that reworks `conda_build.variants.combine_specs` to include much of what is in `conda_build.variants.validate_specs` to better handle clobbering/extending errors.

Are there additional changes/validations needed for this PR to move forward?

mingwandroid:
No changes needed. Thanks a lot. Keen to see the next wave!

PR #4079, opened 2020-10-03
Integrate validate_spec with combine_specs

kenodegard:
Since validations are equally important after combining specs, perform the validations while combining in order to alert the user as to which spec is causing the issue for more helpful debugging. Otherwise sought to standardize terminology (e.g. use explode instead of expand).

Changes:
- ~~added `include_dict` argument to `utils.islist` to control whether a `dict` is considered a list or not~~ (see #4088)
- ~~added some terminology definitions at the top of variants, hopefully, this will make variants less daunting to understand by future devs (read me)~~ (see #4086)
- `variants.validate_spec` was combined with `variants.combine_specs` as I found that combining necessitates similar checks being done, as a backwards compatible solution `variants.validate_spec` is now just a call to `variants.combine_specs` with a single spec
- ~~a slight update to `variants.find_config_files` inorder to make the call from `variants.combine_specs` cleaner~~ (see #4086)

### DIFF
--- a/conda_build/variants.py
+++ b/conda_build/variants.py
@@ -29,7 +29,7 @@ DEFAULT_VARIANTS = {
     'pin_run_as_build': OrderedDict(python=OrderedDict(min_pin='x.x', max_pin='x.x')),
     'ignore_version': [],
     'ignore_build_only_deps': ['python', 'numpy'],
-    'extend_keys': ['pin_run_as_build', 'ignore_version', 'ignore_build_only_deps', 'extend_keys'],
+    'extend_keys': ['pin_run_as_build', 'ignore_version', 'ignore_build_only_deps'],
     'cran_mirror': "https://cran.r-project.org",
 }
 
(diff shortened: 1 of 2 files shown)


conda_build/variants.py
```diff
@@ -120,195 +137,213 @@ def parse_config_file(path, config):
...
+                "      (src:{}) {}".format(k, seen[k]["src"], tuple(seen[k]["zg"]), src, tuple(zg))
+                if k in seen else seen.setdefault(k, {}).update({"src": src, "zg": zg})
+                for zg in zk
+                if zg not in zip_keys
+                for k in zg
+            )
+            zip_keys.update({zg: src for zg in zk})
+    # only include zip_keys if they are explicitly defined
+    if zip_keys:
+        combined["zip_keys"] = list(zip_keys)
```
njalerikson, line 257: I personally do not see any danger in passing through an empty zip_keys and would rather not have this if clause.

I added this check here to pass `tests/test_variants::test_later_spec_priority`, the same is done for extend_keys below.

conda_build/variants.py
```diff
@@ -120,195 +137,213 @@ def parse_config_file(path, config):
...
+
+    # extract extend_keys from all specs
+    extend_keys = {k for s in specs.values() for k in _get_extend_keys(s, include_defaults=False)}
+    # only include extend_keys that are explicitly defined
+    if extend_keys:
+        combined["extend_keys"] = list(extend_keys)
+    extend_keys.update(DEFAULT_VARIANTS['extend_keys'])
```
njalerikson, line 264: Continuing from the comment above regarding the zip_keys if clause, here too I don't see any danger in passing through "empty" extend_keys. I would much rather want to do this:

```python
# extract extend_keys from all specs
extend_keys = {k for s in specs.values() for k in _get_extend_keys(s)}
combined["extend_keys"] = extend_keys
```

Again this was added inorder to pass `tests/test_variants::test_later_spec_priority`. Thoughts?

njalerikson:
@mingwandroid phew it has taken me a while to get back here (yay for other projects consuming all of my time...), I'm not necessarily in love with this implementation (mostly due to the high code complexity) so I would like some feedback on whether this is a good direction

PR #4087, opened 2020-10-07
remove spurious error message from shutil_move_more_retrying

randomphrase:
- also throw exception when retries are exhausted
- cleaned up logging, now includes actual error encountered


randomphrase:
Currently getting explicit confirmation from my employer.

PR #4088, opened 2020-10-08
Helper functions to extract keys

njalerikson:
Introduce new private functions for extracting special types of keys from `spec`. Otherwise, revert to explode terminology used prior to #4075.

This is an attempt to split PR #4079 into smaller changes.

### DIFF
--- a/conda_build/utils.py
+++ b/conda_build/utils.py
@@ -1180,24 +1180,26 @@ def package_has_file(package_path, file_path, refresh=False):
     return content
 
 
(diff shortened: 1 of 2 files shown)


njalerikson:
Here too the two failures appear entirely unrelated.

mingwandroid:
Do you mind if I hold off on this PR until the next conda-build release after the one I'm about to do @njalerikson?

mingwandroid:
Thanks again!

PR #4095, opened 2020-10-14
Fall back to shutil.copy if shutil.copy2 fails when copying patches

mingwandroid:


jezdez:
Not clear what this is for.

PR #4099, opened 2020-10-16
Master.c3i

mingwandroid:


mingwandroid:
Single failure looks like a one-off, and doesn't happen on py 3.8 (Windows).

PR #4102, opened 2020-10-18
Fix windows build rename error

yxlao:
Previously it shows "Failed to rename work directory despite sleeping and retrying..." even when the rename is successful. This causes the Windows `conda-build` to fail.

Fix: on successful rename, break out of the while loop.


yxlao:
CLA signed

jezdez:
@anaconda-issue-bot check

PR #4105, opened 2020-10-19
Remove whitespace causing flake8 error

njalerikson:
Not surprisingly this causes pipelines to fail across the board.


njalerikson:
I see this has been long resolved.

PR #4107, opened 2020-10-23
fix indexing by file

cjmartian:
This should fix indexing by file. The previous way would effectively truncate the existing repodata.json and create a new one with only the artifacts in the file.
I am deliberately making the ignore and remove set empty here. If there's a better approach, I'm open to suggestions.


mingwandroid:
Done in a rebased PR.

PR #4124, opened 2020-11-13
Fix index.json:"platform" in cross-platform builds

mbargull:
If we cross-build on different platforms and not just different architectures of the same platform, the `platform` key in `index.json` is not set to the target platform.
refs:
  - https://gitter.im/conda-forge/conda-forge.github.io?at=5faa7d4b8a236947ba9813eb
    > I'm having an odd upload issue at https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=237099 .
    > The build went fine and uploads for `linux-64` and `osx-64` were successful. But the `win-64` one does not upload (restarted once).
    > There is one peculiarity I did: The `win-64` one is built on `linux-64` -- but that shouldn't cause any issues, shouldn't it? (Given that we upload `osx-arm64` packages built from `osx-64`.)
  - https://gitter.im/conda-forge/conda-forge.github.io?at=5faad739c6fe0131d4e912df
    > I got it:
    > I replaced `"platform": "linux",` with `"platform": "win",` in `info/index.json` and then the upload worked.
    > This would also explain why we didn't see this with other cross-compiles before since we had `osx-arm64` on `osx-64`, i.e., unchanged `"platform": "osx",`. So, `conda-build` needs a little change to do this then.

### DIFF
--- a/conda_build/metadata.py
+++ b/conda_build/metadata.py
@@ -1371,8 +1371,8 @@ class MetaData(object):
             version=self.version(),
             build=self.build_id(),
             build_number=self.build_number() if self.build_number() else 0,
-            platform=self.config.platform if (self.config.platform != 'noarch' and
-                                              arch != 'noarch') else None,
+            platform=self.config.host_platform if (self.config.host_platform != 'noarch' and
+                                                   arch != 'noarch') else None,
             arch=ARCH_MAP.get(arch, arch),
             subdir=self.config.target_subdir,
             depends=sorted(' '.join(ms.spec.split())
--- a/news/gh-4124-info_index-host_platform.rst
+++ b/news/gh-4124-info_index-host_platform.rst
@@ -1,0 +1,25 @@
+Enhancements:
+-------------
+
+* <news item>
+
+Bug fixes:
+----------
+
+* Set "platform" in index.json to the target platform for cross-platform builds.
+
+Deprecations:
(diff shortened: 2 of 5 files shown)


mbargull:
(Adding the actual fix and news entry after seeing the CI fail for the reported issue.)

mbargull:
```python
________________________ test_cross_info_index_platform ________________________
[gw1] darwin -- Python 3.8.3 /Users/runner/work/1/a/miniconda/bin/python
  File "/Users/runner/work/1/s/tests/test_api_render.py", line 173, in test_cross_info_index_platform
    assert metadata.config.host_platform == info_index['platform']
AssertionError: assert 'win' == 'osx'
  - osx
  + win
```
:tada: (as intended).

mbargull:
NB: Of course there are more hurdles for general purpose "Windows on Unix" building (linking, prefix replacement, ...).
The intend from me is to use it for "simple" cases, i.e., only-text-file packages and completely statically linked binaries with no prefix references in them.
refs: https://github.com/conda-forge/containers-common-feedstock/pull/1 , https://github.com/conda-forge/skopeo-feedstock/issues/9

mbargull:
(Failures on macOS are unrelated, something storage-related or the like.)

mingwandroid:
Thanks @mbargull

PR #4140, opened 2020-11-25
Fix applying patches to read-only files

mbargull:
If we have a patch file that modifies read-only sources, `conda-build` currently tries to create temporary read-only files twice (e.g., if we have Git patches with `---a/file +++b/file` entries).

The actual fix and news-entry will follow in subsequent commits; testing failure case on CI first.


mingwandroid:
Thanks!

mingwandroid:
Will merge after I check the CI runs.

mbargull:
Yeah, I'll have to check what the CI spits out. I'm a little confused because from the previous CI log it seemed to rather fail on the tests in the build script but not on the patching itself -- running the test locally I could see it fail at https://github.com/conda/conda-build/pull/4140/files#diff-b1edd70d35f342df704bf1985eeb0e2c6febc5d539d6fd991374ad5236395f76R684 prior to adding the `set( ... )` change.
Plus, I don't know if the `attrib` test on Windows works or not; in theory it would, but IDK whether the permissions are set correctly on Windows too.

mbargull:
> Yeah, I'll have to check what the CI spits out. I'm a little confused because from the previous CI log it seemed to rather fail on the tests in the build script but not on the patching itself \[...\]

Oh, yeah, Git doesn't record the non-write permission of the file XD. Circumvented this via a tarball source.
Let's give the CI another meant-to-fail run and then I'll add the other commits back.

mbargull:
CI logs still seemed to indicate a build rather than patch failure :shrug:. Well, then let's just see how the run with the fixes turns out.

mbargull:
Dammit, nvm. I made a `chmod` typo before and pushed a faulty tarball..

mbargull:
Unix tests finally failed with the expected
```python
  File "/home/vsts/work/1/s/conda_build/source.py", line 682, in copy_to_be_patched_files
    shutil.copy2(os.path.join(src_dir, file), dst)
  File "/usr/share/miniconda/lib/python3.8/shutil.py", line 432, in copy2
    copyfile(src, dst, follow_symlinks=follow_symlinks)
  File "/usr/share/miniconda/lib/python3.8/shutil.py", line 261, in copyfile
    with open(src, 'rb') as fsrc, open(dst, 'wb') as fdst:
PermissionError: [Errno 13] Permission denied: '/tmp/tmpedtj2csy/src_dir/somefile'
```
. Windows ones did not -- I simply commented the `attrib` check in the test out for now (that part shouldn't be critical, plus, debugging on Windows is too time consuming for me, currently).

mingwandroid:
Thanks! I set this ready for review.. I am happy to merge it now though.

mbargull:
> Thanks! I set this ready for review.. I am happy to merge it now though.

Thanks, yep, ready to go; newly added (reduced) tests passed on Windows and other failures are unrelated. :tada:

PR #4147, opened 2020-12-04
Smarter build_numbers

### LINKED ISSUE #4146: MetaData.build_number()
Looking at the `build_number` method:
https://github.com/conda/conda-build/blob/cbc2428bdc09bf9eb0d855dc476fb7c0b164dcea/conda_build/metadata.py#L1174-L1181

I've reviewed the history of this method and see that it's been improved over the years to better work with jinja. I also see that in the current state most uses of this method (4 out of 6) do something like `meta.build_number() or 0`.

I have 2 observations:
1. In the current implementation it would appear that if a completely bad build value is provided (e.g. 'abba') then it will silently be accepted (unless there's a check elsewhere I haven't found yet). Is this desired or should a smarter check be done? 
2. I think it would be wise to return 0 instead of an empty string when receiving a falsy value.

Something like if build number is `None`/missing then return 0 (this seems to satisfy stuff happening with jinja templating, see blame history) else if build number is non-`int` raise error.

njalerikson:
In order to decrease duplicate code and to make the build_number method "smarter" we explicitly check for falsy build numbers and otherwise raise a ValueError for otherwise bad values (e.g. "abba").

Resolves #4146

### DIFF
--- a/conda_build/environ.py
+++ b/conda_build/environ.py
@@ -491,7 +491,7 @@ def meta_vars(meta, skip_build_id=False):
     # use `get_value` to prevent early exit while name is still unresolved during rendering
     d['PKG_NAME'] = meta.get_value('package/name')
     d['PKG_VERSION'] = meta.version()
-    d['PKG_BUILDNUM'] = str(meta.build_number() or 0)
+    d['PKG_BUILDNUM'] = str(meta.build_number())
     if meta.final and not skip_build_id:
         d['PKG_BUILD_STRING'] = str(meta.build_id())
         d['PKG_HASH'] = meta.hash_dependencies()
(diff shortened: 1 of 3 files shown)


njalerikson:
I see that a number of tests are failing.

Both `Linux py27-release` and `macOS py27-release-xcode` are failing with a bunch of conflicts being raised as it attempts to configure and install dependencies into the base env, this seems unrelated to this change?

Errors in `Linux py37-release`, `Linux py38-release`, `maxOS py37-releasse-xcode`, `maxOS py38-releasse-xcode`,  are all the same 3 failed tests and appear to be caused by HTTP issues, also unrelated to this change?

Errors in `Windows py*-release` are all raising the same HTTP issue, so also unrelated to this change?

njalerikson:
Here too I see that rebasing cleared up prior test failures, @mingwandroid anything else needed here?

PR #4149, opened 2020-12-10
LIEF: Allow parsing static libs to fail

mingwandroid:


mingwandroid:
Merging as I need this to build Python 3.9.1

Issue #4118, opened 2020-10-30
`conda render` has a problem with `GIT_*` variables on patched sources


ajelinski:
There is a bug in `conda-build` v3.20.4 and v3.20.5 which causes never-ending `conda render`. Funny thing is, there's no bug if you add `--verbose` (which unfortunately won't work for e.g. `conda render --output`).
`conda-build` v3.20.3 doesn't have such problems.

### Actual Behavior

Virtually never-ending rendering with output:
```
write() argument must be str, not bytes
write() argument must be str, not bytes
write() argument must be str, not bytes
write() argument must be str, not bytes
write() argument must be str, not bytes
write() argument must be str, not bytes
write() argument must be str, not bytes
write() argument must be str, not bytes
write() argument must be str, not bytes
```

### Expected Behavior

Rendered metadata should be sent to the `stdout`.

### Steps to Reproduce

1. Install `conda-build` v3.20.4 or v3.20.5
2. Create a directory with `meta.yaml`:
```
package:
  name: bsdiff4
  version: {{ GIT_DESCRIBE_TAG }}

source:
  git_url: https://github.com/ilanschnell/bsdiff4.git
  git_rev: 1.1.4
  patches:
    - test.patch
```
and `test.patch`:
```
diff --git a/README.rst b/README.rst
index ac92874..30c5ab4 100644
--- a/README.rst
+++ b/README.rst
@@ -2,6 +2,8 @@
 bsdiff4: binary diff and patch using the BSDIFF4-format
 =======================================================

+
+
 The code is mostly derived from cx_bsdiff (written by Anthony Tuininga,
 http://cx-bsdiff.sourceforge.net/).  The cx_bsdiff code in turn was derived
 from bsdiff, the standalone utility produced for BSD which can be found
```
3. Run `conda render .` in the created directory.

##### `Output of conda info`

Not applicable. Tested on a few versions of conda with different configurations on Linux and macOS.

cgr71ii:
This problems still exists with conda build 3.21.4. When using GIT_* variables in `build: number:` or `build: string:`, some binaries are not correctly patched. BTW this bug is really annoying... It was really hard to find out that the problem was related to the use of environmental variables...

I will not open a new issue because I think that with the following example, which should work as expected (i.e. it fails), it will be clear that the problem might be the same.

Just in case: this problem was not happening in December of 2020 to me.

## Example
meta.yaml
```yaml
{% set name = "bitextor" %}
{% set version = "8" %}

package:
[...]

PR #4152, opened 2020-12-12
Fix "write() argument must be str, not bytes" error

chrisburr:
When rendering I see a stream of errors this seems to fix the problem:
```
write() argument must be str, not bytes
write() argument must be str, not bytes
write() argument must be str, not bytes
write() argument must be str, not bytes
write() argument must be str, not bytes
write() argument must be str, not bytes
write() argument must be str, not bytes
write() argument must be str, not bytes
write() argument must be str, not bytes
```

I think this fixes https://github.com/conda/conda-build/issues/4118.


mingwandroid:
Thanks @chrisburr!

PR #4153, opened 2020-12-12
Allow directories as license_file source

### LINKED ISSUE #4121: Support for (collected) license files via directories, globs, list files
### Actual Behavior

Currently, we can specify either single files or a list of files for the `about: license_file:` field.

In some language ecosystems like Rust or Go it is common to vendor dependencies.
The license files of these dependencies are then included in the source distributions but often/usually there is no readily composed list of them available. Hence, it would make sense to use helper scripts during the build to gather a list of those license files and use that list as an input for the `about: license_file:` field.

### Expected Behavior

Allow one or more of the following:
1. Allow entries in `license_file` to point to directories and recursive include all files in them as "license files".
    (Appending a suffix `/` could be made a requirement to make this behavior opt-in.)
2. Allow (recursive) globs in `license_file` entries to be given instead of only static file names.
3. Allow some additional syntax to recognize entries in `license_file` not as "license files" directly, but just as list files.
   (e.g., have a `gather_licenses` script create a `${SRC_DIR}/thirdparty-licenses` file in which every line references the actual license files and have `license_file: <some new syntax>thirdparty-licenses` expand to the file list.)

All of the options above bare the possibility to accidentally include additional non-license files. But I believe it can be rather subjective to judge if one or the other is more prone to it. Hence, I think all of them more or less equally good and I would probably go with the one that is easiest to use and document\*, i.e., option 1.

(\* usage instruction could be simple as "move thirdparty license files to a `$SRC_DIR/additional-licenses` folder and add to `license_file` an entry `additional-licenses/` (ending with `/` to denote  a directory for which all files are included recursively)".)

mbargull:
Resolves #4121.

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -768,12 +768,19 @@ def copy_license(m):
         if license_file == "":
             continue
         src_file = join(m.config.work_dir, license_file)
-        if not os.path.isfile(src_file):
+        if not os.path.isfile(src_file) and not os.path.isdir(src_file):
             src_file = os.path.join(m.path, license_file)
-        if os.path.isfile(src_file):
+        if os.path.isdir(src_file) and not src_file.endswith("/"):
+            raise ValueError(
+                "License entry in about/license_file ({}) points to a directory but does not "
+                "end with a '/'. Make sure the directory only contains license files and "
+                "append a '/' to include the folder and all of its content as license "
+                "information.".format(src_file)
+            )
+        if os.path.isfile(src_file) or os.path.isdir(src_file):
             # Rename absolute file paths or relative file paths starting with .. or .
             if os.path.isabs(license_file) or license_file.startswith("."):
-                filename = "LICENSE{}.txt".format(count)
+                filename = "LICENSE{}{}".format(count, ".txt" if os.path.isfile(src_file) else "")
                 count += 1
             else:
                 filename = license_file
--- a/docs/source/resources/define-metadata.rst
+++ b/docs/source/resources/define-metadata.rst
@@ -1507,13 +1507,19 @@ License file
 Add a file containing the software license to the package
 metadata. Many licenses require the license statement to be
 distributed with the package. The filename is relative to the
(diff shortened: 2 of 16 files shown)


conda_build/build.py
```diff
@@ -768,12 +768,19 @@ def copy_license(m):
...
+        if not os.path.isfile(src_file) and not os.path.isdir(src_file):
             src_file = os.path.join(m.path, license_file)
-        if os.path.isfile(src_file):
+        if os.path.isdir(src_file) and not src_file.endswith("/"):
+            raise ValueError(
+                "License entry in about/license_file ({}) points to a directory but does not "
+                "end with a '/'. Make sure the directory only contains license files and "
+                "append a '/' to include the folder and all of its content as license "
+                "information.".format(src_file)
+            )
```
mbargull, line 779: This restriction can be removed whenever needed. I added it for now because it's easier to lift restrictions than to add them afterwards.

mbargull:
@mingwandroid, if you deem this change as not very controversial, it would be nice to slip it into the next release.
It would allow me to replace copies of https://github.com/conda-forge/podman-feedstock/blob/48f1c466949e6631146a357e78166c3af342b60e/recipe/build.sh#L18-L56 I have in several recipes by a one-liner :).

mingwandroid:
I am fine with trying to get this into the next release. What do you think about the test failures? Not all are 404 errors, some look license related.

mbargull:
The CI failures seem unrelated, though I'm working on fixes some at gh-4154

mbargull:
If CI from gh-4154 runs through successfully and it's merged, we can rebase this commit.

mbargull:
> I am fine with trying to get this into the next release.

Thanks!

> If CI from [gh-4154](https://github.com/conda/conda-build/pull/4154) runs through successfully and it's merged, we can rebase this commit.

Done and all green; ready to go :).

mingwandroid:
Thanks again @mbargull

PR #4160, opened 2020-12-22
Assume non-revisible patches

mingwandroid:


ci/azurepipelines/install_conda_build_test_deps
```diff
@@ -11,6 +11,9 @@ function install_conda_build_test_deps_fn()
   _PKGS+=(py-lief pytest-mock)
   if [[ "$PYTHON_VERSION" != "2.7" ]] || [[ ! $(uname) =~ M* ]]; then
     _PKGS+=(pytest-replay pytest-rerunfailures)
+  else
+    # Workaround https://github.com/nedbat/coveragepy/issues/895
+    _PKGS+=('coverage!=5.0')
```
mbargull, line 16: I copied this constraint to the Windows part in `azure-pipelines.yml` after I saw that `ci/azurepipelines/install_conda_build_test_deps` isn't run for the Window CI.

----
The issue that `coverage==5.0` caused was that it printed a stack trace with a `... % ( ...` line in it and `conda`'s faulty `CondaError.__str__` (from the expected `LinkError` in a test) picked that percent character as a formatting token. Using any other `coverage` version avoids that (and we don't have any `>5.0` build for Py2.7 which is why the test picked `5.0` in the first place).

mingwandroid:
Thanks @mbargull!

PR #4162, opened 2020-12-31
docs: move tl;dr on separation of build & host

ltalirz:
The TL;DR on how to separate build and host sections was placed at the
end of the host sections, somewhat defying its purpose.
This commit moves the basic instructions directly to the build section.
It also tries to remove some duplication and to simplify the flow for a
couple of sentences.


jezdez:
@anaconda-issue-bot check

PR #4168, opened 2021-01-10
Make post-link run_export/library_nature determination less gross

mingwandroid:

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -1357,8 +1357,11 @@ def build_info_files_json_v1(m, prefix, files, files_with_prefix):
         }
         if file_info["path_type"] == PathType.hardlink:
             file_info["size_in_bytes"] = os.stat(path).st_size
+        elif isdir(path):
+            # this is a symlink to a directory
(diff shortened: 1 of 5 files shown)


conda_build/build.py
```diff
@@ -1357,13 +1357,21 @@ def build_info_files_json_v1(m, prefix, files, files_with_prefix):
...
+        elif file_info["path_type"] == PathType.softlink:
+            dest = os.path.normpath(os.path.join(os.path.dirname(path), os.readlink(path)))
             # softlink that points somewhere
-            file_info["size_in_bytes"] = os.stat(path).st_size
+            if isdir(dest) and not islink(dest):
+                file_info["size_in_bytes"] = 0
+            elif isfile(dest) and not islink(dest):
+                file_info["size_in_bytes"] = os.stat(path).st_size
+            elif islink(dest):
+                sys.exit("TODO :: Nested symlinks not handled yet")
```
speleo3, line 1368: Was this intentional? Why should symlinks to non-existing files not be allowed? This is a breaking change.

Issue #4173, opened 2021-01-11
new conda-build breaks building packages


mingwandroid:
From: https://github.com/conda-forge/conda-build-feedstock/issues/161

```
2021-01-11T05:10:56.5597132Z + conda build /home/conda/recipe_root -m /home/conda/feedstock_root/.ci_support/linux_64_.yaml --suppress-variables --clobber-file /home/conda/feedstock_root/.ci_support/clobber_linux_64_.yaml
2021-01-11T05:10:57.2259107Z No numpy version specified in conda_build_config.yaml.  Falling back to default numpy value of 1.16
2021-01-11T05:10:57.2260405Z WARNING:conda_build.metadata:No numpy version specified in conda_build_config.yaml.  Falling back to default numpy value of 1.16
2021-01-11T05:10:57.2971092Z Adding in variants from internal_defaults
2021-01-11T05:10:57.2972473Z Adding in variants from /home/conda/recipe_root/conda_build_config.yaml
2021-01-11T05:10:57.2972995Z Adding in variants from /home/conda/feedstock_root/.ci_support/linux_64_.yaml
2021-01-11T05:10:57.2973558Z INFO:conda_build.variants:Adding in variants from internal_defaults
2021-01-11T05:10:57.2974070Z INFO:conda_build.variants:Adding in variants from /home/conda/recipe_root/conda_build_config.yaml
2021-01-11T05:10:57.2974892Z INFO:conda_build.variants:Adding in variants from /home/conda/feedstock_root/.ci_support/linux_64_.yaml
2021-01-11T05:10:57.3268816Z Traceback (most recent call last):
2021-01-11T05:10:57.3287508Z   File "/opt/conda/lib/python3.8/site-packages/conda/models/version.py", line 38, in __call__
2021-01-11T05:10:57.3288173Z     return cls._cache_[arg]
2021-01-11T05:10:57.3288968Z KeyError: '3.6.* *_cpython'
2021-01-11T05:10:57.3289310Z 
2021-01-11T05:10:57.3289680Z During handling of the above exception, another exception occurred:
2021-01-11T05:10:57.3290211Z 
2021-01-11T05:10:57.3290516Z Traceback (most recent call last):
2021-01-11T05:10:57.3293650Z   File "/opt/conda/bin/conda-build", line 11, in <module>
2021-01-11T05:10:57.3294264Z     sys.exit(main())
[...]

PR #4174, opened 2021-01-11
Fix https://github.com/conda/conda-build/issues/4173

mingwandroid:


mingwandroid:
I am going to merge this in-spite of the test suites failing because there were no failures previous to this change and the new failures do not even touch this code. So the problem appears to be to do with packages in the ecosystem that we are testing against.

PR #4179, opened 2021-01-15
Metadata Schema

kenodegard:
Instead of having a highly decentralized process of validating the contents and format of the metadata this introduces a schema that is used to validate/standardize the metadata prior to storage/usage.

Using this schema obsoletes the need to check whether a section is an `OPTIONALLY_ITERABLE_FIELDS` and other optional styles of defining the metadata (e.g. "build/run_exports" which can be a simple `list` which implicitly means those values are "weak", or a `dict` explicitly specifying "weak", "strong", etc. values).

This further helps to centralize the checking of the metadata to where it happens all at once upon parsing the metafile and not randomly throughout the build process and at least theoretically should help users to more easily catch metadata issues.

I'm leaving the cleanup process of removing obsolete code to future PRs.

### DIFF
--- a/conda_build/metadata.py
+++ b/conda_build/metadata.py
@@ -13,7 +13,7 @@ import time
 
 from bs4 import UnicodeDammit
 
-from .conda_interface import iteritems, PY3, text_type
+from .conda_interface import PY3, text_type
 from .conda_interface import md5_file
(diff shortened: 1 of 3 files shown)


mingwandroid:
Thank you for your work again @njalerikson. I look forward to reviewing it as soon as I get some time.

PR #4183, opened 2021-01-16
Replace number stringification hack with custom YAML loader

kenodegard:
This change was prompted by issues identified in #4179.

Instead of monkeypatching the yaml Loader every time it is used in an attempt to avoid parsing numbers make a custom Loader where the float/int tags are entirely removed.

This change also catches a number of cases that were previously missed. For example, pyyaml defined float resolvers for "-+0123456789." and int resolvers for "-+0123456789" and since the prior implementation only monkeypatched for "0123456789" it was still possible for numbers to be missed (e.g. ".0", "-1", "+1").


njalerikson:
Yay! All done here!

@mingwandroid - no rush just let me know if anything here needs further work/discussion.

kenodegard:
@jezdez added a few simple tests that are just checking to see if any int/floats occur in the parsed yaml, the zero, positive, and negative tests all fail on master.

jezdez (approved):

Issue #4185, opened 2021-01-19
test_build_expands_wildcards appears to haphazardly fail


njalerikson:
See any recently run pipeline (even the one for 3.21.4: [d2048ea](https://github.com/conda/conda-build/runs/1709273055)).

I believe that the wildcard test is failing haphazardly since "whether or not the [glob] results are sorted depends on the file system" ([Python Docs: glob.glob](https://docs.python.org/3/library/glob.html#glob.glob)).

So I envision that the correct fix for this test is to either modify `conda_build.utils.expand_glob` to sort its glob results or to modify the test to generate the list of expected files using `expand_glob` or `glob`.

Thoughts?

mingwandroid:
Sounds reasonable @njalerikson. I will not get to any conda-build things for a week or more now, but your PRs are always appreciated!

njalerikson:
resolved in #4186

PR #4186, opened 2021-01-19
Fix glob ambiguity

njalerikson:
This is intended to address #4185.

Unfortunately, I still have not been able to replicate the issue on a local machine so we'll have to see how Azure does.


njalerikson:
Looks like this fixes it, I'll leave it up to others to decide if it's an acceptable fix.

mingwandroid:
Great! Thanks @njalerikson

PR #4197, opened 2021-02-19
Implement negative matching for `files` in outputs

### LINKED ISSUE #4196: Negative matching 'files' for outputs
### Actual Behavior

There is currently no good way to exclude one or more files from being included in an output package that uses `files`.

### Expected Behavior

Support for some sort of negative match condition that is collected and applied _after_ the positive glob expressions are resolved. Anything analogous to the `%exclude` macro you can use in RPM spec files would be great.

Perhaps something as simple as a leading exclamation `!` would work:

```yaml
outputs:
  - name: mpyackage
    files:
      - bin/mypackage*
      - "!bin/mypackage-optional*"

  - name: mypackage-optional
    files:
      - bin/mypackage-optional
```

[AFAIK the exclamation is special in YAML, so it would have to be quoted always]

I'm happy to contribute to development, if other people think this is a good idea.

duncanmmacleod:
This PR implements #4196 by adding supporting for a negative match expression for the `files` list in `outputs`.

Closes #4196.

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -1617,12 +1617,20 @@ def bundle_conda(output, metadata, env, stats, **kw):
     elif files:
         # Files is specified by the output
         # we exclude the list of files that we want to keep, so post-process picks them up as "new"
-        include = [f for f in files if not f.startswith('!')]
-        exclude = [f[1:] for f in files if f.startswith('!')]
-        keep_files = {os.path.normpath(pth)
-                         for pth in utils.expand_globs(include, metadata.config.host_prefix)}
-        exclude_files = {os.path.normpath(pth)
-                            for pth in utils.expand_globs(exclude, metadata.config.host_prefix)}
+        if isinstance(files, dict):
+            include = files.get("include", [])
+            exclude = files.get("exclude", [])
+        else:
+            include = list(files)
+            exclude = []
+        keep_files = {
+            os.path.normpath(pth)
+            for pth in utils.expand_globs(include, metadata.config.host_prefix)
+        }
+        exclude_files = {
+            os.path.normpath(pth)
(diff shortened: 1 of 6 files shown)


duncanmmacleod:
@mingwandroid, I think this is ready for a review (sorry if there's a better @ ping for the conda-build team). The macOS jobs are failing, but the logs suggest it's not my fault..

duncanmmacleod:
Ping for anyone to review this.

conda_build/build.py
```diff
@@ -1619,8 +1619,13 @@ def bundle_conda(output, metadata, env, stats, **kw):
     elif files:
         # Files is specified by the output
         # we exclude the list of files that we want to keep, so post-process picks them up as "new"
+        include = [f for f in files if not f.startswith('!')]
+        exclude = [f[1:] for f in files if f.startswith('!')]
```
mbargull, line 1623: We also need to consider files which themselves start with a `!`. For example with `gitignore` you can start entries with `\!` for this.
Personally, I'd rather suggest to have a more explicit form to avoid these custom syntax and special cases  (also wrt. YAML tags). I.e., I'd rather have `files` optionally be a dictionary with two keys:
```yaml
  files:
    include:
      - *.txt
    exclude:
      - do-not-want.txt
```
.

mbargull (changes requested):
Needs to consider filenames starting with `!`.

jezdez (changes requested):
I concur with @mbargull's comment, this should be done with a new `exclude` section instead of prepending the filename with `!`.

conda_build/build.py
jezdez: Agreed, @mbargull!

PR #4221, opened 2021-04-19
Address "WARNING :: Failed to get_static_lib_exports(*.a)"

mbargull:
Currently, `conda-build` prints out `WARNING :: Failed to get_static_lib_exports(*.a)` whenever static libraries are using in the builds.
This is because of a mismatched type of `lief.*.parse(raw=...)` (explicitly needs `List[int]` rather than `bytes`).
When the parameter type is fixed, it fails because `lief.Binary.exported_functions` gives an iterable of `lief.Function` objects (or just names, i.e., `str`s, in older versions) which do not have `.exported`/`.imported` members.

I can't say with any confidence, I know what I'm doing here because I did not dig in too deep of LIEF, the surrounding code and everything. Silver lining: If the changed code now raises another exception, the behavior would be identical because the previous one always failed (and the exception is caught and replaced the the mentioned warning). The unknown for me is just in case no exception is raised, IDK what difference a non-empty list of returned "imported by not exported" symbols affects.


conda_build/os_utils/liefldd.py
```diff
@@ -701,7 +701,7 @@ def _parse_ar_hdr(content, index):
                 # syms_b = get_symbols(obj, defined=True, undefined=False)
                 # print(syms_b)
             else:
-                obj = lief.ELF.parse(raw=content[obj_start:obj_end])
+                obj = lief.ELF.parse(raw=list(content[obj_start:obj_end]))
```
mbargull, line 704: `lief.parse` accepts `raw: bytes` or `raw: List[int]`, but `lief.*.parse` only accepts `raw: List[int]`.

conda_build/os_utils/liefldd.py
```diff
@@ -890,8 +890,10 @@ def get_symbols(file, defined=True, undefined=True, notexported=False, arch='nat
         except:
             pass
     res = []
+    imported_function_names = set()
     if len(binary.exported_functions):
         syms = binary.exported_functions
+        imported_function_names = {s.name for s in binary.imported_functions}
```
mbargull, line 896: I don't really know if this is the correct thing to do. Just made sense given the names of `lief.Binary`'s members.

conda_build/os_utils/liefldd.py
```diff
@@ -890,8 +890,10 @@ def get_symbols(file, defined=True, undefined=True, notexported=False, arch='nat
         except:
             pass
     res = []
+    imported_function_names = set()
     if len(binary.exported_functions):
```
mbargull, line 894: NB: Due to lack of deeper knowledge of LIEF, I have no idea what the implications are in prioritizing `binary.exported_functions` over `binary.symbols` over `binary.static_symbols`. So, IDK if it makes sense to use a (maybe?) smaller set of symbols from `.exported_functions` rather than just `.symbols`.

mbargull:
@isuruf, @xhochy, apart from Ray, you are the ones that I know of who had a look at LIEF or related code. Could one of you take a look at this?
Of course, this is supposed to be a proper fix if possible. But personally, I'd mainly want to de-clutter conda-build's logs to better see its DSO analysis output :).

xhochy:
> @isuruf, @xhochy, apart from Ray, you are the ones that I know of who had a look at LIEF or related code. Could one of you take a look at this? Of course, this is supposed to be a proper fix if possible. But personally, I'd mainly want to de-clutter conda-build's logs to better see its DSO analysis output :).

I sadly don't have any expertise either that would go beyond running this code and seeing whether it passes. So I cannot add something on top of what other conda-build reviewers would do here.

mbargull:
pre-commit.ci autofix

beckermr:
Not stale

PR #4236, opened 2021-06-14
Add activate_in_script to metadata

bethune-bryant:
The `activate_in_script` field was causing `conda_build.api.check` to fail.

This PR adds it to the list of valid  `metadata` fields.


bethune-bryant:
Will be resubmitted by @bnemanich who has been approved and signed the CLA.

PR #4239, opened 2021-06-21
Fix Jinja render example for resolved_packages

ChristofKaufmann:
Small documentation fix regarding the Jinja example for `resolved_packages('host')`. The template code occurs only in the `run` section and not in the `host` section. So the `host` section should not change by the rendering.


ChristofKaufmann:
Signed the CLA.

jezdez:
@anaconda-issue-bot check

PR #4260, opened 2021-08-05
Fix macOS test failure by updating test recipe

chenghlee:
Include `pip` as a build requirement in the `source_setup_py_data` test recipe to ensure the test does not fail when conda's `add_pip_as_python_dependency` configuration option is set.


jezdez:
Closing since this was done in #4277

PR #4273, opened 2021-08-20
use scandir() to avoid separate is_dir() syscall

dholth:
This should avoid one syscall per directory entry for the (modest) `channel_root` listing.


jezdez:
@anaconda-issue-bot check

jezdez (approved):

PR #4274, opened 2021-08-20
update references to "conda build" command

arlenk:
page currently references "conda-build" command, which I think should be a reference to "conda build"


jezdez:
@anaconda-issue-bot check

mbargull:
Thanks for contributing to the project, @arlenk!
The actual command is indeed `conda-build` and the fact that `conda build` also works is just due to a feature of `conda` to pick up `conda-...` named executables as "subcommands". So, `conda-build` is the direct invocation of `conda-build` and `conda build` just happens to work the same.\*
If there are places in the docs where `conda build` is used, those should probably reference `conda-build` rather than the other way around ;).

(\* Well, they just mostly work the same, to be pedantic; there is an edge case if `conda-build` is not in the `base` environment, then the `conda` shell function tries to pick up `conda-build` from `base` instead of the one from the activated env which the `conda-build` command would run...)

PR #4275, opened 2021-08-23
Docs/build package install instructions

truh:
I wasn't able to find information on how to install a package without uploading it
to anaconda.org anywhere in the documentation so wrote a section about it.

Not sure if this is the correct part of the documentation but it was about as far as I got before wanting to install the package the first time so it seems appropriate to me. Feel free to move it to a different place or edit/rewrite it if necessary.


jezdez:
@anaconda-issue-bot check

truh:
I'm going to close the PR in its current form since there are other places in the docs that tell you to install the package with `conda install --use-local` but I do think the approach to install the the package from file should be mentioned as well. Might rework it and open a new PR.

PR #4277, opened 2021-09-01
CI fixes 2021-09

mbargull:
- Fix 2 tests on macOS:
  `load_setup_py_data` runs `setup.py` in the `host` env and as such always needs `setuptools` to be installed there.
  Previously, the macOS parallel tests weren't run at all because those serial tests failed.
- Actually test Python 3.7/3.8 on the corresponding Linux/macOS CI jobs.
  Previously, 3.9 was installed unconditionally on all those jobs.
- Split CI jobs into serial/parallel ones.
  Splitting for all platforms results in 18 jobs (3 platforms times 3 Python version times 2 serial/parallel tests) but only 10 jobs can run in parallel for this project on Azure (in sum for all PRs/branches, IIRC). So, far from optimal, but on average the time to finish is reduced that way.
  The jobs are also reordered in hope of having the longer ones be scheduled first (which in reality may or may not happen, depending on available CI resources).
- Moved the `vmImage` to `ubuntu-18.04` since `ubuntu-16.04` will be removed in a couple of weeks.


mbargull:
@jezdez, in the second to last community meeting you mentioned the plan to overhaul the CI setup (GitHub Actions, container-based, more uniform setup across the different platforms, etc.).
This PR is just to "keep us above water" until then. There are still some flaky parallel tests on macOS/Windows (at one point a macOS CI run even stalled?!), at least partially I'd assume it has to do with `conda` not handling concurrent package cache writes correctly. Also, some long running tests need some weeding out/adjustments since the overall test duration is pretty/too long, even if split over multiple jobs.

Because of the `ubuntu-16.04` removal on Azure, we should not wait too long with merging, BTW.

jezdez:
Yeah, this is a good intermediate step, it's been bothering me for a while that the test suite is failing. Plus I think this will make it easier to refactor and move to GHA as well, once we have the time.

jezdez (approved):

PR #4294, opened 2021-10-09
Use `pip`'s in-tree build feature

jakirkham:
This is future behavior of `pip` to not copy to a temporary directory first. As this will soon become default. Opt-in on older `pip` versions for consistency. Uses `pip`'s doc's guidance for setting environment variables (linked below).

ref: https://pip.pypa.io/en/stable/topics/configuration/#environment-variables
xref: https://github.com/pypa/pip/issues/7555


jakirkham:
The other option here is to maintain out-of-tree builds by setting `PIP_USE_DEPRECATED="out-of-tree-build"`, but this will presumably go away at some point

PR #4303, opened 2021-10-22
Add extra meta to about.json

cjmartian:
Allows arbitrary information to be added to about.json
This might be helpful for embedding build system log IDs into the conda package for better traceability.

PR #4304, opened 2021-10-25
Add extra git info

cjmartian:
Adds optional git info to about.json when builds happen. 

Signed-off-by: Connor Martin <connormartin7@gmail.com>


conda_build/build.py
```diff
@@ -1102,6 +1102,53 @@ def sanitize_channel(channel):
...
+    p = subprocess.Popen(
+            command,
+            stdin=subprocess.PIPE,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            )
+    out, _ = p.communicate()
+    branch = out.decode("utf-8")[:-1]
+    if not branch:
+        raise
```
jezdez, line 1122: It would make sense to have some exception here to know what's going on.

conda_build/build.py
```diff
@@ -1102,6 +1102,53 @@ def sanitize_channel(channel):
...
+            )
+    out, _ = p.communicate()
+    origin_url = out.decode("utf-8")[:-1]
+
+    git_info = {
+            "origin_url": origin_url,
+            "commit": remote_commit
+            }
+
+    return git_info
```
jezdez, line 1149: This all should use [subprocess.run](https://docs.python.org/3/library/subprocess.html#subprocess.run) instead to simplify the code and improve error handling.

conda_build/build.py
```diff
@@ -1177,10 +1224,18 @@ def write_about_json(m):
             d['conda_private'] = conda_private
         except (KeyError, AttributeError):
             pass
+        extra = m.get_section('extra')
+        if m.config.extra_info:
+            try:
+                git_info = get_git_info(m)
+                extra.update(git_info)
+            except Exception as e:
+                print(f"Unable to get git info. Skipping adding extra_info to about.json: {e}")
```
jezdez, line 1233:
```suggestion
            except Exception as exc:
                print(f"Unable to get git info. Skipping adding extra_info to about.json: {exc}")
```

jezdez: Is using print the only way to report back to the user?

conda_build/build.py
```diff
@@ -1102,6 +1102,53 @@ def sanitize_channel(channel):
...
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            )
+    out, _ = p.communicate()
+    origin_url = out.decode("utf-8")[:-1]
+
+    git_info = {
+            "origin_url": origin_url,
+            "commit": remote_commit
+            }
```
jezdez, line 1147:
```suggestion
    git_info = {
        "origin_url": origin_url,
        "commit": remote_commit,
    }
```

conda_build/config.py
```diff
@@ -48,6 +48,7 @@ def set_invocation_time():
 ignore_verify_codes_default = []
 exit_on_verify_error_default = False
 conda_pkg_format_default = None
+extra_info_default = 'false'
```
jezdez, line 51: Why is this quoted but other boolean values not?

jezdez (changes requested):
Just some general nit-picking, if we'd already have a plugin API, this would be a neat extension to separate out of conda-build.

conda_build/config.py
cjmartian: I was actually kind of confused by this too, it seems like `exit_on_verify_error_default` is the only one that isn't [quoted](https://github.com/conda/conda-build/pull/4304/files/b4e527a832cb70fccbbb30bf22ad820072321c2c#diff-eaaf464431ee59c1575fecf5aa57498e16bb0694fd2cb6ee1638d3ebd682e68fR41-R51).

dbast:
Given that https://github.com/conda/conda-build/pull/4303 is merged, do we really need/want this PR merged?
* It requires git as run time requirement
* Build systems typically already know all the data to be burned in without a additional cli call (git in this PR)
* 4303 is more generic and already allows to do `conda-build ... --extra-meta repo-url=$REPO_URL --extra-meta commit-sha1=$COMMIT_SHA1 --extra-meta ci-run-id=$CI_RUN_ID`

jezdez:
> Given that #4303 is merged, do we really need/want this PR merged?
> 
> * It requires git as run time requirement
> * Build systems typically already know all the data to be burned in without a additional cli call (git in this PR)
> * 4303 is more generic and already allows to do `conda-build ... --extra-meta repo-url=$REPO_URL --extra-meta commit-sha1=$COMMIT_SHA1 --extra-meta ci-run-id=$CI_RUN_ID`

That makes sense to me and fits with keeping the code base lean (or at least start somewhere!).

ScottBailey:
Since #4303 exists and is merged, this is far less important; however, it seems like there would be some value for us - and conda-forge - to have this data (probably more specifically as `git_url`, `git_sha`, and `git_dirty`) in `info/about.json` and/or `info/recipe/conda_build_config.yaml`. I say this because there's no requirement a channel is populated via the prefect build tool and, for example, conda-forge does not use prefect. Having this functionality baked into cond-build increases the likely hood it's used and also follows the DRY philosophy.

I think this is the sort of thing that should be enabled by _default_; however, just having it as a CLI option is a step in the right direction for traceability, in my opinion. As people see it's value it will become enabled by default. Or not, as the case may be.

@chenghlee @jezdez @dbast

PR #4327, opened 2021-11-19
Initiate a first pre-commit CI pass

kenodegard:
Instead of cluttering other benign PRs with the first pre-commit pass we do this.


kenodegard:
pre-commit.ci autofix

kenodegard:
@jezdez I know we are still experiencing broken tests on `conda-build`, are these failures of concern here?

setup.cfg
```diff
@@ -1,6 +1,6 @@
 [flake8]
 max-line-length = 180
-ignore = E122,E123,E126,E127,E128,E731,E722,W504,W605
+ignore = E122,E123,E126,E127,E128,E203,E731,E722,W503,W504,W605
```
jezdez, line 3: Those ignore statements are a big worrying TBH, can we remove them?

jezdez:
> @jezdez I know we are still experiencing broken tests on `conda-build`, are these failures of concern here?

It seems as if there is a dependency problem with urllib3 and brotlipy which makes sense to fix, I don't think they are triggered by this though.

pyproject.toml
```diff
@@ -0,0 +1,9 @@
+[tool.black]
+line-length = 180
```
jezdez, line 2: Hmm, 180 seems way too much, is that how many characters some lines have?

jezdez (changes requested):
I think this is a good first step but I'd rather have us strengthen the flake8 config a bit.

setup.cfg
kenodegard: I would love to remove most of these (eventually) but I think that would be out of scope for this PR?

pyproject.toml
kenodegard: This was copied from the prior line-length set for flake8. I personally don't like conda or conda-build's line length policies and would much rather we switch to black's standard 88 line length. Would such a change be out of scope here too?

pyproject.toml
```diff
@@ -0,0 +1,9 @@
...
+target-version = ['py36', 'py37', 'py38']
+extend-exclude = '''
+^/(
+    conda/_vendor
+  | devenv
+)/
+'''
```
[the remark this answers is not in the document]
kenodegard: Good catch!

setup.cfg
kenodegard: As for the ignores I explicitly added (both are needed for black and flake8 to work together):
[E203](https://www.flake8rules.com/rules/E203.html): whitespace before ':'
[W503](https://www.flake8rules.com/rules/W503.html): line break occurred before a binary operator

jezdez: Hmm, I'm wondering, mostly to better understand the impact, if we were to remove most of the ignores, would we see an immediate increase in code changes here in the PR?

pyproject.toml
jezdez: I think this might be okay for this PR TBH since line length should only apply to a small number of lines in total.

kenodegard: True, do we want to use conda's line length or just use the default from black?

setup.cfg
kenodegard: No, but pre-commit CI's flake8 will probably blowup with complaints. Pre-commit CI by default runs pre-commit on all files, not just the changes, we can change this to be consistent with the local pre-commit behavior if we want.

these are changes that `black`/`darker` would address:
- [E122](https://www.flake8rules.com/rules/E122.html): under-indented continuation lines
- [E128](https://www.flake8rules.com/rules/E128.html): under-indented continuation lines
- [E126](https://www.flake8rules.com/rules/E126.html): over-indented continuation lines
- [E127](https://www.flake8rules.com/rules/E127.html): over-indented continuation lines
- [W503](https://www.flake8rules.com/rules/W503.html): line break occurred before a binary operator
- [W504](https://www.flake8rules.com/rules/W504.html): line break occurred after a binary operator
- [E123](https://www.flake8rules.com/rules/E123.html): align closing bracket with opening bracket's line
- [E203](https://www.flake8rules.com/rules/E203.html): whitespace before ':'

other changes we will need to manually fix:
- [E731](https://www.flake8rules.com/rules/E731.html): do not assign lambda, use def
- [E722](https://www.flake8rules.com/rules/E722.html): do not use bare except
- [W605](https://www.flake8rules.com/rules/W605.html): invalid escape sequence

pyproject.toml
jezdez: Hm, I don't have a strong opinion other than preferring one standard than many. If both conda and conda-build seem unreasonable, we should default to black's default.

setup.cfg
jezdez: Oh interesting, I assumed the CI would also only just run against the changes, having consistency between dev and CI makes sense to me.

kenodegard: Hm actually I'm not finding the option to change this (was very sure I'd seen it somewhere...). I believe pre-commit CI is effectively running `pre-commit run --all-files`. This has no impact on how `darker` works but means that `flake8` gets run on the entire code base. If we remove these undesired ignore statements flake8 will report an additional 700+ errors. Most will be relatively simple to address with the exception of E722 errors.

pyproject.toml
kenodegard: Ok I went with black's default then, as black suggests we should also consider using [Bugbear's B950](https://github.com/PyCQA/flake8-bugbear#:~:text=tutorial%20for%20details.-,B950,-%3A%20Line%20too%20long) for linting.

jezdez (approved):
Thank you @kenodegard! 🚀

conda_build/launcher_sources/cpython-launcher-c-mods-for-setuptools.3.6.patch
```diff
@@ -3,7 +3,7 @@
 @@ -8,11 +8,72 @@
   * Curt Hagenlocher (job management)
   */
- 
+
```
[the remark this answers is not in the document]
kenodegard: Good point, will open PR for this.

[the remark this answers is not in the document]
mbargull: Thanks!

PR #4336, opened 2021-11-23
Do not auto format .patch/.diff

kenodegard:
Revert changes to `.patch` and prevent auto formatting in the future.

https://github.com/conda/conda-build/pull/4327#discussion_r755266340


jezdez (approved):
Good catch @mbargull!

mbargull (approved):
LGTM, thanks!

PR #4341, opened 2021-12-02
Fix keeping old work when workdir does not match package name

Lnaden:
#4167 changed the way `compute_build_id` was computed to be based on the package name. However, it also made the assumption that the old `work_dir` would be equal to `{croot}/{package name}/work` which is not true if someone manually passes in a `croot` and a `build_id` to the `Config` constructor; either manually or as a pass through on `api.render`.

The result was old work would be lost if the original `work_dir` did not match the specific string, instead of just looking where the original work_dir was. This PR fixes the bug and adds a test to make sure the work is preserved and correctly moved to the new folder.

This is a very niche edge case found when directly using the API at omnia-md/conda-dev-recipes#212

### DIFF
--- a/conda_build/config.py
+++ b/conda_build/config.py
@@ -590,7 +590,7 @@ class Config:
         else:
             # Maybe call set_invocation_time() here?
             pat_dict['t'] = invocation_time
-            test_old_dir = join(self.croot, package_name, 'work')
+            test_old_dir = self.work_dir
             old_dir = test_old_dir if os.path.exists(test_old_dir) else None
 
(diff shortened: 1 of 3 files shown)


kenodegard (approved):
Nice! Thanks for taking the time to cleanup the tests a little!

Lnaden:
I'm not sure the one failing test has anything to do with this PR. Looks like an upstream problem.

If the issue with `brotli` is something introduced by this PR, I'm open to any suggestions on resolving it.

kenodegard:
@Lnaden agreed, the failing test is not due to these changes.

PR #4338, opened 2021-11-29
WIP: Initial commit for migrating to github actions for tests

rchord:
Migrating the tests away from Azure Pipelines. Adding the github workflow file for tests.

PR #4344, opened 2021-12-07
Initial commit for migrating to github actions for tests

jezdez:
Replaces #4338.


ci/github/github_setup_pytest_replay
```diff
@@ -0,0 +1,5 @@
+mkdir ./pytest-replay
+declare -a PYTEST_REPLAY_OPTIONS=()
+PYTEST_REPLAY_OPTIONS+=("--replay-record-dir=./pytest-replay")
+PYTEST_REPLAY_OPTIONS+=("--replay-base-name=Linux-$CONDA_VERSION-Py$PYTHON_VERSION")
+echo "${PYTEST_REPLAY_OPTIONS[@]}"
```
kenodegard, line 5: Yes, I see that this is run in the `Source Scripts` job but I don't think it's actually used?

.github/workflows/tests.yml
```diff
@@ -0,0 +1,99 @@
...
+          PYTHON_VERSION: ${{ matrix.python-version }}
+          CONDA_VERSION: ${{ matrix.conda-version }}
+
+      - name: Run Serial tests
+        if: matrix.test-type == 'serial'
+        run: |
+          pytest --color=yes -vv -n 0 --basetemp /tmp/cb_serial --cov conda_build --cov-report xml -m "serial" tests --test-run-title="${{ matrix.os }}-SerialTests" --test-run-title="${{ matrix.os }}-SerialTests-${{ matrix.conda-version }}-Py${{ matrix.python-version }}" $PYTEST_REPLAY
+        env:
+          PYTHON_VERSION: ${{ matrix.python-version }}
+          CONDA_VERSION: ${{ matrix.conda-version }}
```
kenodegard, line 80: I don't think these are used?

.github/workflows/tests.yml
```diff
@@ -0,0 +1,99 @@
...
+          PYTEST_CMD: ${{ env.pytest-serial-cmd }}
```
kenodegard, line 81: This also appears unused?

.github/workflows/tests.yml
```diff
@@ -0,0 +1,99 @@
...
+          PYTEST_REPLAY: ${{ env.pytest-replay }}
```
kenodegard, line 82: I don't think this is necessary, the global env is already exposed to each individual job.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,99 @@
...
+      - name: Run Parallel tests
+        if: matrix.test-type == 'parallel'
+        run: |
+          pytest --color=yes -vv -n auto --basetemp /tmp/cb --cov conda_build --cov-append --cov-report xml -m "not serial" tests --test-run-title="${{ matrix.os }}-ParallelTests" --test-run-title="${{ matrix.os }}-ParallelTests-${{ matrix.conda-version }}-Py${{ matrix.python-version }}" $PYTEST_REPLAY
+          ls -la pytest-replay
+        env:
+          PYTHON_VERSION: ${{ matrix.python-version }}
+          CONDA_VERSION: ${{ matrix.conda-version }}
+          PYTEST_CMD: ${{ env.pytest-parallel-cmd }}
+          PYTEST_REPLAY: ${{ env.pytest-replay }}
```
kenodegard, line 93: Ditto

.github/workflows/tests.yml
```diff
@@ -0,0 +1,99 @@
...
+          conda info -a
+          source ci/github/github_setup_pytest_replay
+        env:
+          PYTHON_VERSION: ${{ matrix.python-version }}
+          CONDA_VERSION: ${{ matrix.conda-version }}
+
+      - name: Run Serial tests
+        if: matrix.test-type == 'serial'
+        run: |
+          pytest --color=yes -vv -n 0 --basetemp /tmp/cb_serial --cov conda_build --cov-report xml -m "serial" tests --test-run-title="${{ matrix.os }}-SerialTests" --test-run-title="${{ matrix.os }}-SerialTests-${{ matrix.conda-version }}-Py${{ matrix.python-version }}" $PYTEST_REPLAY
```
kenodegard, line 77: Part of me wants to break these over several lines to make it more readable but I'm fine either way:

```suggestion
        run: >
          pytest
              --color=yes
              -vv
              -n 0
              --basetemp /tmp/cb_serial
              --cov conda_build
              --cov-report xml
              -m "serial" tests
              --test-run-title="${{ matrix.os }}-SerialTests"
              --test-run-title="${{ matrix.os }}-SerialTests-${{ matrix.conda-version }}-Py${{ matrix.python-version }}"
              $PYTEST_REPLAY
```

rchord: Yes! I like this. I thought of breaking them but then I got more focused on something else.

.github/workflows/tests.yml
rchord: If I need to call it in the "run" portion I need to add it as a variable like this. I know, I thought this too and was sad when it complained. Makes it really clunky feeling

.github/workflows/tests.yml
rchord: oops

.github/workflows/tests.yml
rchord: They are used as part of the "--test-run-title"

ci/github/github_setup_pytest_replay
rchord: Its being sourced at in the "Source Scripts" job of the tests.yml

.github/workflows/tests.yml
kenodegard: But you didn't use the env vars above in `--test-run-title`?

rchord: Oh yeah... huh? How'd that happen...  yeah what the jank. Let me fix that. I am so confused because this goes against what I said before about how I have to call it in the "env:" to use it in the script. Let me research this if I am remembering it right. I guess it has been 2 weeks since I looked at this.

rchord: Welp it appears I got my head all muddled. I removed all the "env:" that duplicate a variable that has already been declared. This is exactly why we do reviews lol

ci/github/github_setup_pytest_replay
rchord: This is how they did it in the azure pipeline so I just duplicated mostly what was done within the Github structure, but should we be sourcing scripts like this? Or should we keep this scripting within the action yaml itself?

.github/workflows/tests.yml
rchord: disregard that statement. It's not true in this case and I am removing it.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,104 @@
...
+              -m "serial" tests \
+              --test-run-title="${{ matrix.os }}-SerialTests" \
+              --test-run-title="${{ matrix.os }}-SerialTests-${{ matrix.conda-version }}-Py${{ matrix.python-version }}" \
+              ${{ env.pytest-replay }}
+
+      - name: Run Parallel tests
+        if: matrix.test-type == 'parallel'
+        run: |
+          pytest \
+              --color=yes -vv \
```
kenodegard, line 90:
```suggestion
              --color=yes \
              -vv \
```

.github/workflows/tests.yml
```diff
@@ -0,0 +1,104 @@
...
+              -n auto \
+              --basetemp /tmp/cb \
+              --cov conda_build \
+              --cov-append --cov-report xml \
```
kenodegard, line 94:
```suggestion
              --cov-append \
              --cov-report xml \
```

ci/github/github_setup_pytest_replay
kenodegard: Both of the replay options defined here are identical to those defined as `env.pytest-replay` so those appears to be duplicates.

The only command that appears to be a non-duplicate is the `mkdir` command. I'd be inclined to nuke this setup script and move the `mkdir` command into the workflow itself.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,104 @@
...
+
+      - name: Install Conda-Canary if canary version
+        if: matrix.conda-version == 'canary'
+        run: |
+          conda install -y -c conda-canary conda
+
+      - name: Install default Conda if release version
+        if: matrix.conda-version == 'release'
+        run: |
+          conda install -y conda
```
kenodegard, line 62: Yup looks good to me.

kenodegard (approved):

beeankha (approved):
@rchord walked me through this PR, everything looks good to go! 👍

PR #4353, opened 2022-01-10
Run Tests Against Windows in GitHub Actions

rchord:
Adding Windows to Github Actions for tests.


beeankha:
This is now ready for review! 

1. The Windows tests are running properly via GitHub Actions
2. All of the tests are passing 🎉 

Two tests have been marked as `xfail` but the reason is known and fixing those tests up will be part of work done in the future related to fixing tech debt.

kenodegard (approved):
🎉 This is amazing! Very happy to see that we were able to recover previous xfail tests!

Issue #4366, opened 2022-01-26
bdist_conda does not work with setuptools >=60.0.0


dbast:
### Actual Behavior

setuptools >=60.0.0 now defaults to their local version of distutils, see https://setuptools.pypa.io/en/latest/history.html#v60-0-0

Setting `SETUPTOOLS_USE_DISTUTILS=stdlib` makes the bdist_conda test work again, as tested here https://github.com/conda-forge/conda-build-feedstock/pull/175

While pinning down the version to `setuptools <60.0.0` works for now, we should figure out how to get https://github.com/conda/conda-build/blob/master/bdist_conda.py working with setuptools >=60.0.0 in the long run.

### Expected Behavior

conda-build being compatible with setuptools >=60.0.0

### Steps to Reproduce

see https://github.com/conda-forge/conda-build-feedstock/pull/172 + https://github.com/conda-forge/conda-build-feedstock/pull/175

##### `Output of conda info`

```

```

PR #4368, opened 2022-01-26
Support bdist_conda also with setuptools >=60.0.0

dbast:
The distutils included in setuptools >=60.0.0 are enabled by default, when the setuptools package is imported, see https://setuptools.pypa.io/en/latest/history.html#v60-0-0

ref: https://github.com/conda-forge/conda-build-feedstock/pull/172

fixes: https://github.com/conda/conda-build/issues/4366


cjmartian (approved):

conda.recipe/build.sh
```diff
@@ -3,3 +3,9 @@
 python setup.py install --single-version-externally-managed --record=record.txt
 
 cp bdist_conda.py ${PREFIX}/lib/python${PY_VER}/distutils/command
+
+# required for distutils included in setuptools >=60.0.0
+# see also https://setuptools.pypa.io/en/latest/history.html#v60-0-0
+SETUPTOOLS_DISTUTILS="${PREFIX}/lib/python${PY_VER}/site-packages/setuptools/_distutils/command/"
+mkdir -p "${SETUPTOOLS_DISTUTILS}"
+cp bdist_conda.py "${SETUPTOOLS_DISTUTILS}"
```
jezdez, line 11: Forgive my lack of knowledge, will this be run during installation? There is no other way to add additional bdist commands to setuptools?

dbast: yeah... those lines also look wired ... I did not find any distutils plugin mechanism (like pytest has) except the solution chosen in the past where something is placed into the distutils command folder ... 

Seems like we doing this since while, see line 5 in that file https://github.com/conda/conda-build/pull/4368/files#diff-2f7714f3308cb222f2b5b3bba3d7395b040ef39f5303b5917175630860f8099eR5

`cp bdist_conda.py ${PREFIX}/lib/python${PY_VER}/distutils/command`

All we do with this PR is to also copy the same file into the distutils folder that got vendored by setuptools (once setuptools are imported you can not import distutils from python anymore.)

dbast: refs:
* https://docs.python.org/3/distutils/extending.html
* https://setuptools.pypa.io/en/latest/userguide/extension.html

Maybe an setup.py entry similar to this would work:
```
setup(
    # ...
    entry_points={
        "distutils.commands": [
            "bdist_conda = conda_build.bdist_conda:bdist_conda",
        ],
    },
)
```
Requires to move `bdist_conda.py` to somewhere in `./conda_build/...` which should be possible.

It also requires, that the bdist_conda class in conda_build.bdist_conda is a setuptools.Command subclass, which is the case.

jezdez: If it works, it would be a more future-proof approach than modifying the distutils package (under the setuptools package). So I'd be in favor of fixing the problem that way, but I also see it increasing the scope of your change a bit.

dbast: agree, will give it a try.

jezdez:
@chenghlee Do you have any thoughts on this, e.g. any good reason to NOT use setuptools entrypoints for this subcommand?

conda.recipe/test_bdist_conda_setup.py
```diff
@@ -0,0 +1,8 @@
+from distutils.core import setup, Extension
+import distutils.command.bdist_conda
```
jezdez, line 2: Is this needed?

jezdez:
> The entry_point solution (done via the last commit) does not work... tested locally via `conda-build conda.recipe`

How did it fail specifically? With a command not found error?

> The initial solution (initial commit) does work... any ideas what else to try?

I could take a closer look tomorrow..

dbast:
yep. failed with
```
python test_bdist_conda_setup.py bdist_conda --help
  File "/Users/dbast/miniconda3/envs/cb/conda-bld/conda-build_1643299389910/test_tmp/test_bdist_conda_setup.py", line 3, in <module>
    import distutils.command.bdist_conda
ModuleNotFoundError: No module named 'distutils.command.bdist_conda'

```

conda.recipe/test_bdist_conda_setup.py
dbast: good point... that is the old test file from the conda-forge-feedstock... those import are not required / can be changed for setup.py entry_point.

setup.py
```diff
@@ -52,7 +52,11 @@
                             'conda-render = conda_build.cli.main_render:main',
                             'conda-skeleton = conda_build.cli.main_skeleton:main',
                             'conda-debug = conda_build.cli.main_debug:main',
-                            ]},
+                            ],
+        "distutils.commands": [
+            "bdist_conda = conda_build.bdist_conda:bdist_conda",
```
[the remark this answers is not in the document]
dbast: that is a command class and should be fine.

dbast:
It works now!

conda.recipe/test_bdist_conda_setup.py
dbast: Thanks for pointing me at that... with the entry_point it is now enough to do a `from setuptools import setup` ... that gets it working.

dbast:
the command works, but not its args... something to look into.

dbast:
command + args work now.

ready for review again. thanks!

conda.recipe/test_bdist_conda_setup.py
```diff
@@ -0,0 +1,8 @@
+from setuptools import setup
+import conda_build.bdist_conda
+
+setup(
+    name="package",
+    version="1.0.0",
+    distclass=conda_build.bdist_conda.CondaDistribution
```
jezdez, line 7:
```suggestion
    distclass=conda_build.bdist_conda.CondaDistribution,
```

jezdez (approved):
Awesome, thank you!

I think this should warrant a 3.22 release given that it's deprecating a feature? The versioning scheme of conda-build is a bit unclear to me.

jezdez:
> @jezdez This is more or lest just a formal deprecation ... if bdist_conda.py is copied to the distutils folder by the feedstock recipe, it would still work with setuptools<=60.0.0, see e.g. https://github.com/conda-forge/conda-build-feedstock/blob/master/recipe/build.sh#L5
> 
> Would thus do a minor version increment and ensure it is available via the old and new way via the feedstock build files, but still speak about deprecation in the changelog to create awareness about it.

Right, the entrypoint in setuptools to add a new subcommand existed for years, this is mostly to make sure we help existing users to continue to use it in its new place. Let's announce the pending deprecation, and make sure to ask them to change their import statement to the correct import path, and then remove it in the release after the next one.

Oh to be clear, 3.22.0 *is* a minor release 😄 I'm ambivalent about marking a feature as deprecated in a patch releases like 3.21.9. @chenghlee do you have opinions around this deprecation?

BTW, I checked on GitHub's new search and found ~30 instances of the `distutils.command.bdist_conda` import: https://cs.github.com/?scopeName=All+repos&scope=&q=distutils.command.bdist_conda

PR #4381, opened 2022-02-14
Add --append option to conda-index

cjmartian:
Adds `--append` option to `conda-index` in order to simply add newly built packages to `repodata.json`. The motivation is to be able add to an existing repodata without needing every file locally and regenerating it.

### DIFF
--- a/conda_build/index.py
+++ b/conda_build/index.py
@@ -787,7 +787,6 @@ class ChannelIndex:
         else:
             level = logging.ERROR
 
-        # dict(subdir: repodata)
         existing_repodatas = {}
         append = append or []
         for repodata in append:
--- a/conda_build/index.py
+++ b/conda_build/index.py
@@ -789,11 +789,12 @@ class ChannelIndex:
 
         # dict(subdir: repodata)
         existing_repodatas = {}
-        append = [] if not append else append
+        append = append or []
         for repodata in append:
-            with open(repodata) as fh:
-                j = json.load(fh)
-            existing_repodatas[j.get("info", {}).get("subdir", None)] = j
+            with open(repodata) as repodata_file:
+                loaded = json.load(repodata_file)
+            subdir = loaded.get("info", {}).get("subdir", None)
+            existing_repodatas[subdir] = loaded
 
         with utils.LoggingContext(level, loggers=[__name__]):
             if not self._subdirs:
@@ -830,7 +831,7 @@ class ChannelIndex:
 
                             if existing_repodata:
                                 existing_repodata.get("packages").update(
-                                        repodata_from_packages.get("packages")
+                                    repodata_from_packages.get("packages")
                                 )
 
                             t2.set_description("Writing pre-patch repodata")
--- a/conda_build/index.py
+++ b/conda_build/index.py
@@ -258,7 +258,7 @@ def update_index(dir_path, check_md5=False, channel_name=None, patch_generator=N
         return update_index(base_path, check_md5=check_md5, channel_name=channel_name,
                             threads=threads, verbose=verbose, progress=progress,
                             hotfix_source_repo=hotfix_source_repo,
-                            current_index_versions=current_index_versions, append=appen)
+                            current_index_versions=current_index_versions, append=append)
     return ChannelIndex(dir_path, channel_name, subdirs=subdirs, threads=threads,
                         deep_integrity_check=check_md5, debug=debug).index(
                             patch_generator=patch_generator, verbose=verbose,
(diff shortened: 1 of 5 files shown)


conda_build/index.py
```diff
@@ -781,12 +781,20 @@ def __init__(self, channel_root, channel_name, subdirs=None, threads=MAX_THREADS
...
 
     def index(self, patch_generator, hotfix_source_repo=None, verbose=False, progress=False,
-              current_index_versions=None, index_file=None):
+              current_index_versions=None, index_file=None, append=None):
         if verbose:
             level = logging.DEBUG
         else:
             level = logging.ERROR
 
+        # dict(subdir: repodata)
```
jezdez, line 790: Mind elaborating a bit in the comment what this is used for?

conda_build/index.py
```diff
@@ -781,12 +781,20 @@ def __init__(self, channel_root, channel_name, subdirs=None, threads=MAX_THREADS
...
+        existing_repodatas = {}
+        append = [] if not append else append
+        for repodata in append:
+            with open(repodata) as fh:
+                j = json.load(fh)
+            existing_repodatas[j.get("info", {}).get("subdir", None)] = j
```
jezdez, line 796:
```suggestion
        append = append or []
        for repodata in append:
            with open(repodata) as repodata_file:
                loaded = json.load(repodata_file)
            repodata_subdir = loaded.get("info", {}).get("subdir", None)
            existing_repodatas[repodata_subdir] = loaded_repodata
```
Just a bit easier on the eyes.

conda_build/index.py
```diff
@@ -816,16 +825,22 @@ def index(self, patch_generator, hotfix_source_repo=None, verbose=False, progres
                                 subdir, verbose=verbose, progress=progress,
                                 index_file=index_file)
 
+                            if existing_repodata:
+                                existing_repodata.get("packages").update(
+                                        repodata_from_packages.get("packages")
+                                )
```
jezdez, line 831:
```suggestion
                                existing_repodata.get("packages").update(
                                    repodata_from_packages.get("packages")
                                )
```

jezdez (commented):
These are just minor code comments so far.

But I think I'd appreciate a bit more details on the why for this feature.
My context is basically: Adding features is cheap. But my spidey senses tingle when I see a feature being proposed that adds a whole new alternative to maintain a core functionality such as the `repodata.json`.

@chenghlee Mind to chime in on that, too?

cjmartian:
@jezdez I'd like to be able to upload packages to S3 without needing to regenerate repodata in it's entirety.

My idea on this would be to have a lambda architecture where S3 upload events would be stored in a queue. A lambda would read off the queue and would be able to take in the current repodata and update it with whatever package was just uploaded. This gives us the ability to upload builds directly to S3 and also keep the repodata up to date.

We could then also scan an S3 bucket and batch load all of the objects into the queue and use that queue to regenerate repodata from scratch.  

AFAIK `conda-index` doesn't support such a thing.

conda_build/index.py
cjmartian: Was just keeping notes, I'm happy to remove that.

conda_build/index.py
```diff
@@ -258,14 +258,14 @@ def update_index(dir_path, check_md5=False, channel_name=None, patch_generator=N
         return update_index(base_path, check_md5=check_md5, channel_name=channel_name,
                             threads=threads, verbose=verbose, progress=progress,
                             hotfix_source_repo=hotfix_source_repo,
-                            current_index_versions=current_index_versions)
+                            current_index_versions=current_index_versions, append=appen)
```
jezdez, line 261:
```suggestion
                            current_index_versions=current_index_versions, append=append)
```

jezdez (changes requested):
Just a minor thing.

jezdez:
@chenghlee Who'd be a good person to get a review from about this outside Anaconda?

dholth:
I found this `--no-remove` option in this older documentation but not in my local install. https://docs.conda.io/projects/conda-build/en/latest/resources/commands/conda-index.html?highlight=conda-build%20index#conda-index

dholth:
I'm trying to make this obsolete. A conda-index run would require the entire cache (one sqlite db file) for a channel but not the actual packages. A user could manipulate the replacement for the stat cache to index exactly a set of files, and tell index not to call `listdir/os.stat()`. This lets us index one file without adding additional cases to the main index logic.

dholth:
The standalone github.com/conda-incubator/conda-index separates "looking for packages on the filesystem" from "updating metadata cache / write repodata.json". It can be overridden to provide a similar feature by not updating "list of all packages to index" from os.listdir() each time. If the set it's comparing to is identical except for the newly to-be-indexed/removed packages then only those packages need to be available locally. https://github.com/conda-incubator/conda-index/blob/main/conda_index/index/sqlitecache.py#L424

PR #4389, opened 2022-02-26
Add `entry_points` to `outputs` in `FIELDS`

jakirkham:
This appears to be usable in recipes and [covered in the tests]( https://github.com/conda/conda-build/blob/3b99b2222a067e113a2282926871cd1e5406ee2b/tests/test-recipes/split-packages/_entry_points/meta.yaml#L19-L22 ). So add it to `FIELDS`.

### DIFF
--- a/conda_build/metadata.py
+++ b/conda_build/metadata.py
@@ -448,6 +448,7 @@ FIELDS = {
         'name': None,
         'version': None,
         'number': None,
+        'entry_points': None,
         'script': None,
         'script_interpreter': None,
         'build': None,


jakirkham:
Could someone please review? Maybe @chenghlee?

jakirkham:
@jezdez do you know if someone would be able to review this? 🙂

jakirkham:
Rebased. Could someone please review?

kenodegard (approved):

conda_build/metadata.py
```diff
@@ -448,6 +448,7 @@ def parse(data, config, path=None):
         'name': None,
         'version': None,
         'number': None,
+        'entry_points': None,
```
kenodegard, line 451: But actually shouldn't this be:
```suggestion
        'entry_points': list,
```

kenodegard (changes requested):

kenodegard:
And while we're at it this needs documenting in https://github.com/conda/conda-build/blob/master/docs/source/resources/define-metadata.rst#outputs-section

conda_build/metadata.py
jakirkham: It seems all the others here are `None`. IIUC this is because the `outputs` fields are not required since some recipes won't use `outputs`. If we make this `list`, wouldn't that imply `outputs/entry_points` is required by all recipes (even when other `outputs` fields are not)? Admittedly I could be misunderstanding things

jakirkham:
It is [documented outside of `outputs` here]( https://github.com/conda/conda-build/blob/be6b3e96e1a01a47300a513524676f283f8c799a/docs/source/resources/define-metadata.rst#python-entry-points ). While `outputs` allows many of the same keys as the top-level recipe does, it appears not all of them are defined in `outputs`. Guessing as that would be somewhat repetitive. Not sure when we decide what things to include in `outputs` as well as general recipe documentation.

conda_build/metadata.py
kenodegard: I don't think specifying it as a `list` makes it required but you may be right for other reasons, [I remember now how the schema validator we use here is incomplete and struggles with nested stuff](https://github.com/conda/conda-build/pull/4179). So I'll retract my comment and we'll leave it as `None`.

jakirkham:
Thanks Ken! 😄

Issue #4386, opened 2022-02-22
Windows conda-build tests need to move to Github Actions


CheeseMochi:
Conda-build's tests are currently running in Azure Pipelines. To be consistent across the board on where tests run, we need to move them to Github Actions

We are not making any changes to tests. However, If there are refactors that are seen to improve things, make a separate issue for them.

PR #4400, opened 2022-03-21
[WIP] Migrating Windows Tests from Azure Pipeline to GitHub Actions

beeankha:
Connecting issue https://github.com/conda/conda-build/issues/4386

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -229,7 +229,7 @@ jobs:
         shell: cmd
 
       - name: Run Serial Tests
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -111,6 +111,7 @@ jobs:
   windows:
     runs-on: windows-2019
```
beeankha, line 112: Available images are here: https://github.com/actions/virtual-environments#available-environments

Wondering if we should put `windows-latest` here instead of `windows-2019`

PR #4405, opened 2022-03-24
add menuinst JSON file validation

jaimergp:
This adds the logic to validate the new menuinst JSONs at build-time. It will only print a warning for now. If an error should be raised, we need to add a logic similar to the overlinking analysis, but I don't know if that's what people want... For now I kept it simple. Let me know!


jaimergp:
This needs https://github.com/conda/conda/pull/11882 on Windows, because `conda` has a `menuinst <2` constraint. On the other platforms we are ok because there's no such constraint (`menuinst` was never a requirement there).

jezdez (approved):

jezdez:
The test failures are concerning tho!

jaimergp:
I believe it's a caplog issue. It gets no text captured even though the exception is being raised. Need to debug.

jaimergp:
Turns out `utils.get_logger()` injects a deduplication filter by default, so the 2nd test would never raise a warning because the message (albeit with different args!) had already been 'seen'. 🤷 `dedupe=False` is enough to make it work.

jezdez (approved):

jaimergp:
Errors unrelated to this PR (to be fixed by https://github.com/conda/conda-build/pull/5059).

conda_build/post.py
```diff
@@ -1732,6 +1732,54 @@ def fix_permissions(files, prefix):
                 log.warn(str(e))
 
 
+def check_menuinst_json(files, prefix):
+    json_files = fnmatch_filter(files, "[Mm]enu[/\\]*.[Jj][Ss][Oo][Nn]")
+    if not json_files:
+        return
+
+    print("Validating Menu/*.json files")
```
travishathaway, line 1740: Would it be a better idea to make this a log.info call? I'm always a little skeptical of print statements 🤔.

jaimergp: Just following trend with the rest of the module :( Search for `print` in `post.py` and be amazed 😬

conda_build/post.py
```diff
@@ -1732,6 +1732,54 @@ def fix_permissions(files, prefix):
                 log.warn(str(e))
 
 
+def check_menuinst_json(files, prefix):
```
travishathaway, line 1735: I'm a little confused about the purpose of this function. It looks like it attempts to validate the menuinst JSON file, but it doesn't return anything. I would expect this type of function to return a boolean value or perhaps an error message. Instead it appears to just be using the logger to print this information.

This is different enough from what I would expect that it warrants a brief description as a doc string. Also consider annotating it with a return type to make that a little less ambiguous.

travishathaway: Sounds good. I know conda-build is a weird. It's appropriate to have different standards for this project.

travishathaway (approved):

PR #4407, opened 2022-03-24
Mark Two Tests with Xfail

beeankha:


tests/test_api_render.py
```diff
@@ -112,6 +112,8 @@ def test_pin_compatible_semver(testing_config):
     utils.on_win and sys.version_info < (3, 6),
     reason="Failing tests on Azure for Python 2.7"
 )
+@pytest.mark.xfail(sys.platform == "win32",
```
jezdez, line 115: You can use `conda_build.utils.on_win` here as well.

jezdez (approved):
Just a minor nit about using existing utilities. :shipit:

PR #4412, opened 2022-03-29
Migrate MacOS Tests from Azure to GitHub Actions

beeankha:
Connecting issue https://github.com/conda/conda-build/issues/4387

* * * 

From the issue linked above:

`conda-build`'s tests are currently running in Azure Pipelines. To be consistent across the board on where tests run, we need to move them to Github Actions.


.github/workflows/tests.yml
```diff
@@ -258,8 +257,124 @@ jobs:
...
+        python-version: ['3.7', '3.8']
+        conda-version: [release]
+        test-type: [serial, parallel]
+        include:
+          - python-version: '3.9'
+            conda-version: canary
+            test-type: serial
+          - python-version: '3.9'
+            conda-version: canary
+            test-type: parallel
```
jezdez, line 286: Until we have more self-hosted GitHub Actions runners via @rhoule-anaconda's work, we need to limit the amount of Python version to one [like in the conda repo](https://github.com/conda/conda/blob/7651023ca56c4bcdf200090e75ae99f62124524c/.github/workflows/ci.yml#L110) since the number of action runners is shared in the whole conda organization.

jezdez:
@beeankha The failing tests are about the `extras_require` value not being correctly discovered in skeleton. The way skeleton this has done historically is to [**patch the Python stdlib library `distutils`**](https://github.com/conda/conda-build/blob/0a6efdcfa02484954e7a3fdf5c2ec4bb3cb6682e/conda_build/skeletons/pypi.py#L91-L135) for some reason. It basically patches it to get the metadata during build-time and store it in a `pkginfo.yaml` file, which is later in the process [retrieved again from disk](https://github.com/conda/conda-build/blob/0a6efdcfa02484954e7a3fdf5c2ec4bb3cb6682e/conda_build/skeletons/pypi.py#L1195-L1196) and used as the basis in [the `get_package_metadata` function](https://github.com/conda/conda-build/blob/0a6efdcfa02484954e7a3fdf5c2ec4bb3cb6682e/conda_build/skeletons/pypi.py#L741-L752).

If `pkginfo['extras_require']` is available at that point, that means the patch for distutils didn't work correctly, which is basically a larger issue than what the error may indicate.

My guess is that something happened in distutils or setuptools recently that triggered this patching not to be successful. But I haven't checked on the actual files during the test runs. One thing to note is that distutils is being deprecated in [favor of setuptools](https://peps.python.org/pep-0632/).

wimglenn:
@jezdez  I assume you've meant to write "_If pkginfo['extras_require'] is **unavailable** at that point_`. 

The patch is successfully applied, but it's patching in the wrong place.  conda-build tries to patch stdlib `distutils/core.py` but because of the distutils deprecation, the actual file in use will be the vendored [`setuptools/_distutils/core.py`](https://github.com/pypa/setuptools/blob/main/setuptools/_distutils/core.py). 

This means the `pkginfo.yaml` isn't created at all, but it's absence is "handled" by falling back to using [`pkginfo.SDist(download_path).__dict__ `](https://github.com/conda/conda-build/blob/dfc64415614e4605d0655e910d5d7181b4b743e0/conda_build/skeletons/pypi.py#L1197-L1198). This namespace does not necessarily contain `extras_require`, even when they are present inside the sdist, so it is not a safe fallback to use in this case that the `pkginfo.yaml` file was not written out.

kenodegard (approved):
Phew, @beeankha thanks for working on this! 🚀

jezdez:
Amazing work @beeankha!

PR #4414, opened 2022-03-31
Also rewrite long prefix paths in `output` scripts $PREFIX, etc.

mariusvniekerk:
Whilst the intention initially was for ouput scripts to mostly do bundling it has emerged as a rather common pattern for recipes to not have a base recipe at all and the main build work is done in an output script.

As a result of this use case, output scripts should have the same filtering applied to them as build scripts

Example: https://github.com/conda-forge/arrow-cpp-feedstock/blob/main/recipe/meta.yaml


mariusvniekerk:
Test failed with flakiness

PR #4423, opened 2022-04-11
Allow redirecting metadata output to another directory

tobijk:


jezdez:
This needs a test :)

dholth:
This was a strangely missing feature. It is called `--output` in new conda-index. https://github.com/conda-incubator/conda-index/blob/main/conda_index/cli/__init__.py#L16

However, the metadata cache is still always under the packages.

Issue #4411, opened 2022-03-28
Remove Azure Pipeline Files


beeankha:
Depends on work being done for the issues below:

- "Migrate Windows Testing from Azure to GitHub Actions": https://github.com/conda/conda-build/issues/4386
- "Migrate MacOS Testing from Azure to GitHub Actions": https://github.com/conda/conda-build/issues/4387

Since all Azure tests will be migrated to GitHub Actions after that point, the pipeline files here can be deprecated/archived:
https://github.com/conda/conda-build/tree/master/.azure-pipelines

PR #4436, opened 2022-04-20
Remove Azure CI pipeline Files

beeankha:
Resolving: https://github.com/conda/conda-build/issues/4411

This PR is the final phase of migrating CI test infrastructure from Azure to GitHub Actions in this repo.


kenodegard:
I think I'm in favor of simply deleting all of these files (especially since this archival ends up adding back files that you had previously renamed/moved). Everything is preserved in the git history so I see little value in preserving it further.

kenodegard (changes requested):
I think I'm in favor of simply deleting all of these files (especially since this archival ends up adding back files that you had previously renamed/moved). Everything is preserved in the git history so I see little value in preserving it further.

kenodegard (approved):

PR #4443, opened 2022-04-26
Drop dependency on Setuptools

jakirkham:
Instead of relying on `setuptools` to get functionality from `pkg_resources` (which is slowly being refactored away), leverage `packaging` for the functionality used by Conda-Build.

Edit: Looked to see whether Conda needed a similar change. Appears it only indirectly uses `pkg_resources` via a vendored copy of `tqdm`. Updating the vendored copy of `tqdm` would fix that issue ( https://github.com/conda/conda/issues/11432 ).

Fixes https://github.com/conda/conda-build/issues/4733


jakirkham:
Updated with latest `master`. Please let me know if there's anything else needed here 🙂

jezdez:
One thing to note is that `pkg_resources.parse_version` falls back to the [LegacyVersion](https://github.com/pypa/setuptools/blob/565d4febd8fedf50638ea54d14f2e75bd9a4558a/pkg_resources/_vendor/packaging/version.py#L42-L51) which is a [pre-PEP440 compatible version](https://github.com/pypa/setuptools/blob/565d4febd8fedf50638ea54d14f2e75bd9a4558a/pkg_resources/_vendor/packaging/version.py#L168-L220), and as such could still be used in one or more packages out there. In other words, this has some potential to be a breaking change for a few projects and needs to be communicated as such.

FWIW I'm in favor of this move since it's way over time to standardize on PEP440 and following.

@chenghlee Have you heard anything about how compatible we are with PEP 440 nowadays?

jakirkham:
Open to other solutions if there is a preferred alternative

jezdez:
I've looked into more of the actual failing tests, and I think we shouldn't rush this for 3.22.0.

It's worth considering all the implications of the old `LegacyVersion` being used for the various non-PEP 440 conforming versions though, which happen to exist for conda packages that ship non-Python software, for example.
I do think we should support and use PEP 440 as the only standard for Python packages, but at the same time use a more relaxed version specifier for non-Python packages.

jezdez:
I've backported `LegacyVersion` to get this unblocked and get rid of the dependency on setuptools. When we'll send skeleton into the retirement, we can remove it again.

conda_build/skeletons/cpan.py
```diff
@@ -28,6 +27,7 @@
 from conda_build.config import get_or_merge_config
 from conda_build.utils import on_win, check_call_env
 from conda_build.variants import get_default_variant
+from conda_build.version import _parse as parse_version
```
jaimergp, line 30: I think we need to do this _everywhere_ where `LegacyVersion` could have been expected. So essentially we need to replicate this change in every other instance of `from pkg_resources import parse_version`.

jezdez: For the PyPI skeleton, I don't think that's needed given that PyPI will provide PEP 440 compatible versions, but the variants handling you're right.

jezdez:
@jakirkham I took the liberty to amend your PR, could you take a look if that's along the lines what you had in mind?

jakirkham:
Thanks Jannis! Doing a bit of vendoring to protect ourselves makes sense to me 🙂

conda_build/skeletons/cpan.py
jakirkham: Interesting how do they handle the long tail of packages that preceded PEP 440?

jakirkham:
Am seeing this test failure in some jobs. Not sure what the expect behavior was before, but that version number doesn't look ok.

It is worth noting that [SymPy is on GitHub]( https://github.com/sympy/sympy ) these days. So a test that was presumably for Bitbucket/Mercurial may not be as relevant (particularly [since Bitbucket dropped Mercurial support]( https://bitbucket.org/blog/sunsetting-mercurial-support-in-bitbucket )).

Maybe something using [Heptapod]( https://foss.heptapod.net/public ) would be better (if Mercurial testing is desired).

```python
___________________________ test_sympy[with version] ___________________________
[gw0] linux -- Python 3.8.16 /usr/share/miniconda/envs/test/bin/python
  File "/home/runner/work/conda-build/conda-build/src/tests/test_api_skeleton.py", line 189, in test_sympy
    api.skeletonize(
  File "/home/runner/work/conda-build/conda-build/src/conda_build/api.py", line 270, in skeletonize
    skeleton_return = module.skeletonize(packages, output_dir=output_dir, version=version,
  File "/home/runner/work/conda-build/conda-build/src/conda_build/skeletons/pypi.py", line 280, in skeletonize
    versions = sorted(pypi_data["releases"].keys(), key=Version)
  File "/usr/share/miniconda/envs/test/lib/python3.8/site-packages/packaging/version.py", line 197, in __init__
    raise InvalidVersion(f"Invalid version: '{version}'")
packaging.version.InvalidVersion: Invalid version: '0.5.13-hg'
```

jakirkham:
Any thoughts on what we should do here?

jezdez:
@jakirkham @jaimergp I've updated it to use the legacy version parser everywhere, I think that should be okay, assuming the tests pass

jezdez:
Given that 3.7 isn't maintained by conda-forge and defaults anymore, I'm ready to take the chance and remove it from the test matrix. Do you have concerns about this?

jakirkham:
None. Have also been dropping Python 3.7 from other projects I work on

jezdez (approved):

jakirkham:
Thanks Jannis! 🙏

beckermr:
xref: https://github.com/conda-forge/conda-build-feedstock/pull/201

dholth:
It looks like the new file is https://github.com/pypa/packaging/blob/main/src/packaging/version.py with underscores in front of each function.

dholth:
I'm surprised that `parse` from packaging's `version.py` has been renamed to `_parse` in our version, but it seems like there are other differences including us having `_LegacyVersion`.

Issue #4404, opened 2022-03-24
conda debug fails when called on a path without trailing slash on Windows


tobijk:
I ran

```
conda debug C:\Users\builder\tkoch\git\aggregate\python-3.7-feedstock
```

which resulted in

```
...
Traceback (most recent call last):
  File "C:\MC3_x64\Scripts\conda-debug-script.py", line 10, in <module>
    sys.exit(main())
  File "C:\MC3_x64\lib\site-packages\conda_build\cli\main_debug.py", line 93, in main
    return execute(sys.argv[1:])
  File "C:\MC3_x64\lib\site-packages\conda_build\cli\main_debug.py", line 69, in execute
    activation_string = api.debug(thing_to_debug, verbose=(not _args.activate_string_only), **_args.__dict__)
  File "C:\MC3_x64\lib\site-packages\conda_build\api.py", line 410, in debug
    os.path.isdir(recipe_or_package_path_or_metadata_tuples) else [])
  File "C:\MC3_x64\lib\genericpath.py", line 42, in isdir
    st = os.stat(s)
TypeError: stat: path should be string, bytes, os.PathLike or integer, not list
```

The expected result would have been for conda debug to realize that I was pointing it at a feedstock directory.

However, it worked when I added a trailing `\`.

PR #4448, opened 2022-05-04
Bugfix: fixes a CLI argument bug for the conda-debug command

travishathaway:
Addresses the following issue:
- https://github.com/conda/conda-build/issues/4404

Although at first glance this issue seem pretty straight forward, it was not. The bug unearthed a dormant code path that was revealed by some interesting behavior in Python.

After some debugging, it became obvious that this line wasn't doing what it intended to do:
https://github.com/conda/conda-build/blob/4d1d52218911cf08e12fda9390529c3798efa380/conda_build/cli/main_debug.py#L63

If I'm correct, the intent of the above line was to filter out everything out that was not a "conda package file", so in theory, directories would be diverted to this code path. In reality, it also allows strings like, `/tmp/some-dir/` to evaluate as `True`, which means directories have always been sent to the next `else` block (currently the working and expected behavior).

Why is that so? It all has to do with the fact that `'' in 'somestring` evaluates to `True`. Check out the example for details:

```python
>>> exts = {'.conda', '.tar.bz2'}
>>> path1 = '/tmp/some-dir-not-trailing-slash'
>>> os.path.splitext(path1)[1] in exts
False
>>> path1 = '/tmp/some-dir-with-slash/'
>>> os.path.splitext(path2)[1] in exts
True
```

With all that being said, I think it is fair to assume that the code executed under this `if` statement was never actually working in the first place because when it is run, it generates the error that we see. Therefore, I made the decision to entirely remove the execution of this code.

I'm not entirely sure why that was ever there in the first place, but it's definitely been broken for a bit. Additionally, the `conda_build.cli.main_debug` module had zero test coverage when I began, so it's very easy to imagine that this broke a while back, and we were unaware of it.

If you know more about this command, please comment, so we can find the best way to fix it!


kenodegard (commented):
I'm not thrilled about adding a new submodule but can see the value in standardizing validators. Perhaps we should identify a few other places where using validators will help simplify the code before we move ahead with this change?

conda_build/validators.py
```diff
@@ -0,0 +1,59 @@
+from __future__ import annotations
```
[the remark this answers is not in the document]
travishathaway: I understand the concern. How about moving it to the `cli` module? They are CLI option and argument validators after all.

conda_build/utils.py
```diff
@@ -2112,3 +2113,16 @@ def shutil_move_more_retrying(src, dest, debug_name):
...
+def is_conda_pkg(pkg_path: str) -> bool:
+    """
+    Determines whether string is pointing to a valid conda pkg
+    """
+    path = Path(pkg_path)
+
+    return (
+        path.is_file() and (
+            any(path.name.endswith(ext) for ext in CONDA_PACKAGE_EXTENSIONS)
+        )
```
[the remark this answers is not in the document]
travishathaway: So, I gave this a shot, but it doesn't work. Specifically for `tar.bz2` file extensions:

```python
>>> Path('mytarball.tar.bz2').suffix
'.bz2'
```

Perhaps not as performant as a simple `in` operation, but I think the `endswith` check is going to be our best bet for now.

conda_build/cli/main_debug.py
```diff
@@ -43,39 +41,42 @@ def parse_args(args):
...
 
 
-def execute(args):
-    p, _args = parse_args(args)
+ARG_VALIDATORS = (
+    ('recipe_or_package_file_path', valid.validate_is_conda_pkg_or_recipe_dir),
+)
+
+
+@valid.validate_args(ARG_VALIDATORS, get_parser())
```
kenodegard, line 61: Btw isn't this design more expensive startup-wise? Calling `get_parser` here means the `ArgumentParser` is generated upon import and not upon function call?

Further, can we rework the `validate_args` signature for a cleaner implementation?
```suggestion
@valid.validate_args(
    get_parser,  # pass in function so we can lazily evaluate
    recipe_or_package_file_path=valid.validate_is_conda_pkg_or_recipe_dir,  # define validators as kwargs
)
```

Another thought, since we use this to validate the `ArgumentParser` parameter should this validator actually be embedded in the argument parser? [Either by defining a custom type or a custom action.](https://stackoverflow.com/questions/37471636/python-argument-parsing-validation-best-practices)

```python
def conda_package(string):
    if os.path.isdir(string) or utils.is_conda_pkg(string):
        return string
    raise argparse.ArgumentTypeError


p.add_argument(
    'recipe_or_package_file_path',
    type=conda_package,
    help=...,
)
```

travishathaway: @kenodegard , really like that last thought! Much better to use the API our current argument parser provides rather come up with a custom solution.

Check out my latest revisions. I believe it addresses everything you have brought up here.

conda_build/cli/validators.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+def get_is_conda_pkg_or_recipe_error_message() -> str:
+    """Return the error displayed on the `validate_is_conda_pkg_or_recipe_dir` validator"""
+    valid_ext_str = ' or '.join(CONDA_PACKAGE_EXTENSIONS)
+    return (
+        '\nUnable to parse provided recipe directory or package file.\n\n'
+        f'Please make sure this argument is either a valid package \n'
+        f'file ({valid_ext_str}) or points to a directory containing recipe.'
+    )
```
kenodegard, line 17: Hm, since this will always be the same message can we make it a global instead?

conda_build/cli/main_debug.py
```diff
@@ -43,39 +42,37 @@ def parse_args(args):
...
-    return p, args
+    return p
 
 
-def execute(args):
-    p, _args = parse_args(args)
+def execute():
+    parser = get_parser()
+    args = parser.parse_args()
     test = True
```
kenodegard, line 59: Looks like `test` no longer serves a purpose and can be removed (see `if test` clauses on lines 70 & 76 below).

kenodegard (commented):
Looking good, just noticed that there's more we can remove since you removed the unused `if` clause.

kenodegard:
@travishathaway you may also want to rebase/squash [15f9605](https://github.com/conda/conda-build/pull/4448/commits/15f9605c96ac06a3de68b488c112ec000566edd0) to change the author/committer

conda_build/cli/main_debug.py
```diff
@@ -43,51 +42,46 @@ def parse_args(args):
...
-        if not _args.activate_string_only:
+        if not args.activate_string_only:
             print("#" * 80)
-            if test:
-                print("Test environment created for debugging.  To enter a debugging environment:\n")
-            else:
-                print("Build and/or host environments created for debugging.  To enter a debugging environment:\n")
+            print(
+                "Build and/or host environments created for debugging.  To enter a debugging environment:\n"
+            )
```
kenodegard, line 71: I think we grabbed the wrong if-clause? We've kept the else-clause from above:

```python
else:
    thing_to_debug = _args.recipe_or_package_file_path
```

Meaning `test = True` so we want to keep the first if-clause here:

```python
print("Test environment created for debugging. To enter a debugging environment:\n")
```

conda_build/cli/main_debug.py
```diff
@@ -43,51 +42,46 @@ def parse_args(args):
...
-                test_file = "conda_test_runner.bat" if on_win else "conda_test_runner.sh"
-                print(f"To run your tests, you might want to start with running the {test_file} file.")
-            else:
-                build_file = "conda_build.bat" if on_win else "conda_build.sh"
-                print(f"To run your build, you might want to start with running the {build_file} file.")
+        if not args.activate_string_only:
+            build_file = "conda_build.bat" if on_win else "conda_build.sh"
+            print(
+                f"To run your build, you might want to start with running the {build_file} file."
+            )
```
kenodegard, line 78: Ditto

kenodegard (changes requested):
Everything looks good, just want to make sure we aren't changing any user-facing messages that don't need to be changed.

travishathaway:
Putting this on hold until I can resolve a comment from another one of our projects:
https://github.com/conda/conda/issues/11463#issuecomment-1122555835

(I just need to implement `fsspec` instead of `pyfakefs`)

kenodegard (approved):

jezdez (approved):
Nice!

PR #4450, opened 2022-05-04
convert filesystem-based metadata cache to sqlite

dholth:
on large conda-build index runs, it can take a very long time to even stat each file in the cache. convert to sqlite3.

the index function should be updated to do some of the set operations in the database instead of in Python, and be updated so that the package data does not have to reside locally.

also, avoid writing any temporary files just because we want a few tiny .json from conda's info/

will vendor all unavailable imports and find a way to move improved conda reader to conda-package-handling or some other package.


dholth:
I run it like so in vscode debugger
```json
{
    // Use IntelliSense to learn about possible attributes.
    // Hover to view descriptions of existing attributes.
    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
    "version": "0.2.0",
    "configurations": [
        {
            "name": "conda-index",
            "type": "python",
            "request": "launch",
            "program": "bin/conda-index",
            "args": ["~/miniconda3/", "--verbose", "--no-progress", "--threads=1"],
            "console": "integratedTerminal",
            "cwd": "${workspaceFolder}",
        }
    ]
}
```

conda_build/index/__init__.py
```diff
@@ -921,25 +1089,305 @@ def index_subdir(self, subdir, index_file=None, verbose=False, progress=False):
...
+            cache.db.execute("DELETE FROM stat WHERE stage='fs'")
+            cache.db.executemany(
+                """
+            INSERT INTO STAT (stage, path, mtime, size)
+            VALUES ('fs', :path, :mtime, :size)
+            """,
+                listdir_stat(),
+            )
+
+        log.info("repodata into stat")
```
dholth, line 1253: probably continue doing the repodata compare in memory

conda_build/index/__init__.py
```diff
@@ -921,25 +1089,305 @@ def index_subdir(self, subdir, index_file=None, verbose=False, progress=False):
...
+            for fn in fns_in_subdir:
+                abs_fn = os.path.join(subdir_path, fn)
+                stat = os.stat(abs_fn)
+                yield {
+                    "path": cache.database_path(fn),
+                    "mtime": int(stat.st_mtime),
+                    "size": stat.st_size,
+                }
+
+        log.info("save fs state")
```
dholth, line 1241: this is the part that definitely needs to be pulled out into a different function. we also need to ensure it can compare with any stage=:parameter and that it doesn't accidentally listdir when we don't want it to, while preserving the old behavior normally.

conda_build/index/__init__.py
dholth: the part where we have to compare with fs to load the cache, and with repodata is confusing.

could we split into two functions: 1. make sure cache knows about everything in repodata 2. load from cache into repodata as necessary.

conda_build/index/__init__.py
```diff
@@ -921,25 +1089,305 @@ def index_subdir(self, subdir, index_file=None, verbose=False, progress=False):
...
+
+            # Invalidate cached files for update_set.
+            # Extract and cache update_set and add_set, then add to new_repodata_packages.
+            # This is also where we update the contents of the stat_cache for successfully
+            #   extracted packages.
+            # Sorting here prioritizes .conda files ('c') over .tar.bz2 files ('b')
+            hash_extract_set = tuple(concatv(add_set, update_set))
+
+            extract_func = functools.partial(
+                cache.extract_to_cache, self.channel_root, subdir
```
dholth, line 1142: we need a way to preview which files extract_to_cache actually has to open (those not already in the cache), maybe just a function that loads from the cache, and a second one that updates from cache, instead of "load or update as needed"

dholth:
superceded by new pull request, rebased against black-reformatted index

PR #4459, opened 2022-05-10
use set for membership queries

dholth:
speeds conda index when the number of unchanged packages is large

### DIFF
--- a/conda_build/index.py
+++ b/conda_build/index.py
@@ -926,7 +926,9 @@ class ChannelIndex:
             )
             # unchanged_set: packages in old repodata whose information can carry straight
             #     across to new repodata
-            unchanged_set = sorted(old_repodata_fns - update_set - remove_set - ignore_set)
+            unchanged_set = set(old_repodata_fns - update_set - remove_set - ignore_set)
+
+            assert isinstance(unchanged_set, set) # faster `in` queries
 
             # clean up removed files
             removed_set = (old_repodata_fns - fns_in_subdir)
@@ -937,7 +939,7 @@ class ChannelIndex:
             new_repodata_packages = {k: v for k, v in old_repodata.get('packages', {}).items() if k in unchanged_set}
             new_repodata_conda_packages = {k: v for k, v in old_repodata.get('packages.conda', {}).items() if k in unchanged_set}
 
-            for k in unchanged_set:
+            for k in sorted(unchanged_set):
                 if not (k in new_repodata_packages or k in new_repodata_conda_packages):
                     fn, rec = ChannelIndex._load_index_from_cache(self.channel_root, subdir, fn, stat_cache)
                     # this is how we pass an exception through.  When fn == rec, there's been a problem,
--- a/conda_build/index.py
+++ b/conda_build/index.py
@@ -928,7 +928,7 @@ class ChannelIndex:
             #     across to new repodata
             unchanged_set = set(old_repodata_fns - update_set - remove_set - ignore_set)
 
-            assert isinstance(unchanged_set, set) # faster `in` queries
+            assert isinstance(unchanged_set, set)   # faster `in` queries
 
             # clean up removed files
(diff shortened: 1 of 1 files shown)


conda_build/index.py
```diff
@@ -937,7 +939,7 @@ def index_subdir(self, subdir, index_file=None, verbose=False, progress=False):
             new_repodata_packages = {k: v for k, v in old_repodata.get('packages', {}).items() if k in unchanged_set}
             new_repodata_conda_packages = {k: v for k, v in old_repodata.get('packages.conda', {}).items() if k in unchanged_set}
 
-            for k in unchanged_set:
+            for k in sorted(unchanged_set):
```
dholth, line 942: unknown whether sort matters, but was sorted before

conda_build/index.py
```diff
@@ -926,7 +926,9 @@ def index_subdir(self, subdir, index_file=None, verbose=False, progress=False):
             )
             # unchanged_set: packages in old repodata whose information can carry straight
             #     across to new repodata
-            unchanged_set = sorted(old_repodata_fns - update_set - remove_set - ignore_set)
+            unchanged_set = set(old_repodata_fns - update_set - remove_set - ignore_set)
+
+            assert isinstance(unchanged_set, set) # faster `in` queries
```
dholth, line 931: prevent future problems

conda_build/index.py
barabo: Not needed - `k` is only referenced twice elsewhere (see my other comment on that above).

conda_build/index.py
```diff
@@ -937,7 +939,7 @@ def index_subdir(self, subdir, index_file=None, verbose=False, progress=False):
             new_repodata_packages = {k: v for k, v in old_repodata.get('packages', {}).items() if k in unchanged_set}
             new_repodata_conda_packages = {k: v for k, v in old_repodata.get('packages.conda', {}).items() if k in unchanged_set}
 
-            for k in unchanged_set:
+            for k in sorted(unchanged_set):
                 if not (k in new_repodata_packages or k in new_repodata_conda_packages):
```
barabo, line 943: This is still gross because both `new_repodata_packages` and `new_repodata_conda_packages` are dicts.  These should both be sets as well.

conda_build/index.py
```diff
@@ -937,7 +939,7 @@ def index_subdir(self, subdir, index_file=None, verbose=False, progress=False):
             new_repodata_packages = {k: v for k, v in old_repodata.get('packages', {}).items() if k in unchanged_set}
             new_repodata_conda_packages = {k: v for k, v in old_repodata.get('packages.conda', {}).items() if k in unchanged_set}
```
barabo, line 940: We're using these two dicts for set operations below.  It might be worth our memory to create a new set using their keys and use that for the set operations.

```python
new_repodata_packages = {k: v for k, v in old_repodata.get('packages', {}).items() if k in unchanged_set}
new_repodata_conda_packages = {k: v for k, v in old_repodata.get('packages.conda', {}).items() if k in unchanged_set}
repodata_packages = set(new_repodata_packages).union(new_repodata_conda_packages)
```

Then, later, do this:
```python
if k not in repodata_packages:
```

barabo (commented):
Such a nice find!  Just a couple more little things to tighten it more.

conda_build/index.py
barabo: Uh, nevermind - I see this is a cheap operation.

conda_build/index.py
barabo: nvm

barabo (approved):
LGTM

conda_build/index.py
dholth: it makes for a more pleasing 'we are processing package a, then package b' instead of 'we are processing packages in random order'

jezdez (approved):

Issue #3507, opened 2019-05-16
[FeatureRequest] Support PEP-518


dhirschfeld:
If `setup.py` requires dependencies not available in the conda build environment then `load_setup_py_data` will fail to parse the `setup.py` file.

***xref:*** #2458

This seems to be the exact problem that [PEP-518](https://www.python.org/dev/peps/pep-0518/#build-system-table) was designed to solve. If `conda-build` parsed the `[build-system]` table in the `pyproject.toml` it could then create an initial environment with the *`setup_requires`* dependencies such that it could then parse the `setup.py` and add the extra dependencies specified there in a second pass.

jakirkham:
I wonder what other ways there might be to solve this problem.

MosheVai:
What's the status of this? 
I am managing my code with [`poetry`](https://python-poetry.org/)  but I have a single package thats only available on `conda`.

I would like for my `pyproject.toml` to contain also the conda instructions.

dhirschfeld:
I think this has been resolved by #4480

PR #4465, opened 2022-05-12
Added load_file_data and load_str_data as functions available in meta.yaml

### LINKED ISSUE #4207: Add a load_pyproject_toml_data() template function
The Bokeh project currently uses the `install_requires` of a `setup.py` to specify the single source of truth for runtime dependencies. We use the existing `load_setup_py_data()` function that conda-build supplies to pull those dependencies into conda recipes automatically. 

We would like to move build information in to a `pyproject.toml` file, but also still keep ability to automatically pull information in to our recipes without manually duplicating it. I propose adding a `load_pyproject_toml_data()` alongside the existing `load_setup_py_data()`. 

Note: I did also find https://github.com/conda/conda-build/issues/3507 which seems possibly related, but I think the proposal here is narrower in scope, and aligned to a different use-case.

abrahammurciano:
Up until PEPs 517 and 518, the setup.py file was the standard way to declare project metatada. That made the function load_setup_py_data an excellent tool for many projects to avoid repetition of metadata in the meta.yaml file.

No longer. With the introduction of the aforementioned PEPs, metadata can be declared in many different ways. Many (if not most) of these use the pyproject.toml file to store their configurations. A notable example of this is poetry.

Instead of trying to accommodate for each and every tool, I have introduced the function load_file_data which is capable of loading (and parsing) certain common serialisation languages (specifically JSON, TOML, and YAML, but can be easily extended (though that would require changing the source code)). The rationale behind this is that the user can just parse whichever file contains their metadata and structure their meta.yaml to use whichever other tools they use without conda-build having to specifically support it.

As an added bonus, I also added a function which is capable of doing the same but reading from a string instead of from a file, so that users can pass JSON (or any other supported language) via environment variables.

Additionally, I have expounded on the bare documentation of load_setup_py_data, added documentation for the previously undocumented load_file_regex, and added documentation for the two functions I added.

Closes #4207

Related: https://github.com/python-poetry/poetry/discussions/3738#discussioncomment-434654

Possibly related? #3507

### DIFF
--- a/conda_build/skeletons/pypi.py
+++ b/conda_build/skeletons/pypi.py
@@ -1195,6 +1195,7 @@ def get_pkginfo(package, filename, pypiurl, digest, python_version, extra_specs,
             with open(join(tempdir, 'pkginfo.yaml')) as fn:
                 pkg_info = yaml.safe_load(fn)
         except OSError:
+            print("WARNING: the pkginfo.yaml file was absent, falling back to pkginfo.SDist")
             pkg_info = pkginfo.SDist(download_path).__dict__
         if new_hash_value:
             pkg_info['new_hash_value'] = ('sha256', new_hash_value)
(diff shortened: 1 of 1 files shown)


kenodegard:
@conda-bot check

abrahammurciano:
@conda-bot check

abrahammurciano:
@kenodegard to me it seems the error with that last run wasn't to do with me. Do you want to rerun that task? Or is there something I'm missing that I should fix?

abrahammurciano:
@kenodegard ready for review

bryevdv:
Big 👍  This would be great to have, and would allow us to take steps towards reducing or even getting rid of `setup.py` for Bokeh.

abrahammurciano:
@kenodegard can you please review this?

kenodegard:
@abrahammurciano Oops, I was trying to push some minor changes and managed to roll back your master branch, I still have all of your changes but due to permissions cannot push the changes back to your fork. I've opened a new PR with the changes instead.

My apologies.

PR #4467, opened 2022-05-13
Add CLI option for changing the compression settings

chrisburr:
I keep finding myself needing to build an extremely large package (many gigabytes) which causes zstandard to use an excessive amount of memory when used in combination with `level=22`. Even when it works I've seen `level=22` an hour when `level=1` takes ~10 seconds which is worthwhile when debugging and not deploying the packages. (I actually wonder if a lower level by default would be more appropriate to make installations faster but that's a separate question.)

As this is configurable behind the scenes I think it would be good to expose the option to the CLI.

Any thoughts?

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -1689,8 +1689,13 @@ def bundle_conda(output, metadata, env, stats, **kw):
         else CONDA_PACKAGE_EXTENSION_V1
     )
     with TemporaryDirectory() as tmp:
-        conda_package_handling.api.create(metadata.config.host_prefix, files,
-                                          basename + ext, out_folder=tmp)
+        conda_package_handling.api.create(
+            metadata.config.host_prefix,
+            files,
+            basename + ext,
+            out_folder=tmp,
+            compression_tuple=metadata.config.compression_tuple,
+        )
         tmp_archives = [os.path.join(tmp, basename + ext)]
 
         # we're done building, perform some checks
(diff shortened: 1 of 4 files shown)


dholth:
Sounds useful, but it might be a little too flexible and too specific to libarchive; could you wind up with a `.tar.xz` inside a `.conda` by replacing the entire tuple? What about `--zstd-compression-level` to expose just the number?

chrisburr:
> What about --zstd-compression-level to expose just the number?

That seems reasonable to me :+1: I'll adapt the PR accordingly

> One useful zstd feature that libarchive may or may not support, and that would go beyond a single number is # of compression threads. I checked the libarchive man page, it looks like its zstd filter doesn't support threads.

I came to the same conclusion as well when originally writing this PR. That said it probably won't help much as with low compression levels I find the storage itself is the bottleneck (even with very fast storage).

> It should be possible to stream tar+zstd directly into the zipfile without using a temporary file; would that be interesting to you?

Yeah I think that would be a nice improvement, especially as I often need to change `$TMPDIR` to be somewhere with more space.

chrisburr:
I go straight to `.conda`. Using `.tar.bz2` was crashing (I suspect due to RAM limits).

dholth (approved):

dholth:
```
   conda-build:
      pkg_format: 2
```

dholth:
I noticed something in boa/mamba land defaults to zstd level 15 by default, which seems like a good inflection point for time/compression.

PR #4470, opened 2022-05-16
reformat index.py with black, move into module

dholth:
to ease diffs; with a few import changes could be mergeable. Working on https://github.com/dholth/conda-build/pull/3


jezdez:
Do you intend to split `index/__init__.py` into separate submodules?

BTW https://github.com/conda/conda-build/issues/4472 may make this much easier to do.

dholth:
There are already several files in the index refactor. We do need to split it out, might as well get it over with

PR #4480, opened 2022-05-27
Added load_file_data and load_str_data as functions available in meta.yaml

### LINKED ISSUE #4207: Add a load_pyproject_toml_data() template function

kenodegard:
Continuation of #4465
Credits go to @abrahammurciano

Up until PEPs 517 and 518, the setup.py file was the standard way to declare project metatada. That made the function load_setup_py_data an excellent tool for many projects to avoid repetition of metadata in the meta.yaml file.

No longer. With the introduction of the aforementioned PEPs, metadata can be declared in many different ways. Many (if not most) of these use the pyproject.toml file to store their configurations. A notable example of this is poetry.

Instead of trying to accommodate for each and every tool, I have introduced the function load_file_data which is capable of loading (and parsing) certain common serialisation languages (specifically JSON, TOML, and YAML, but can be easily extended (though that would require changing the source code)). The rationale behind this is that the user can just parse whichever file contains their metadata and structure their meta.yaml to use whichever other tools they use without conda-build having to specifically support it.

As an added bonus, I also added a function which is capable of doing the same but reading from a string instead of from a file, so that users can pass JSON (or any other supported language) via environment variables.

Additionally, I have expounded on the bare documentation of load_setup_py_data, added documentation for the previously undocumented load_file_regex, and added documentation for the two functions I added.

Closes #4207

Related: https://github.com/python-poetry/poetry/discussions/3738#discussioncomment-434654

Possibly related? #3507


docs/source/resources/define-metadata.rst
```diff
@@ -1706,6 +1706,70 @@ retrieve a fully rendered ``meta.yaml``, use the
...
+* ``load_file_regex``: Searches a file for a regular expression and returns the
+  first match as a Python ``re.Match object``. For example::
+
+    {% set readme_heading = load_file_regex(load_file='README.rst', regex_pattern=r'^# (\S+)') %}
+    package:
+      name: {{ readme_heading.string }}
+
+* ``load_file_data``: You can also parse JSON, TOML, or YAML files and load data
+  from them. For example you can use this to load poetry configurations from
+  ``pytproject.toml``. This is especially useful as ``setup.py`` is no longer the
```
beeankha, line 1734:
```suggestion
  ``pyproject.toml``. This is especially useful as ``setup.py`` is no longer the
```

docs/source/resources/define-metadata.rst
```diff
@@ -1706,6 +1706,70 @@ retrieve a fully rendered ``meta.yaml``, use the
...
+  Then you can pass the ``EXTRA_DEPS`` environment variable to the build like so::
+
+    EXTRA_DEPS='["foo =1.0", "bar >=2.0"]' conda build path/to/recipe
+
+The functions ``load_setup_py_data``, ``load_file_regex``, and ``load_file_data``
+all take the parameters ``from_recipe_dir`` and ``recipe_dir``. If
+``from_recipe_dir`` is set to true, then ``recipe_dir`` must also be passed. In
+that case, the file in question will be searched for relative to the recipe
+directory. Otherwise the file is searched for in the source (after it is
+downloaded and extracted, if necessary). Obviously, if the given file is an
```
beeankha, line 1765:
```suggestion
downloaded and extracted, if necessary). If the given file is an
```

beeankha (approved):
Made some minor suggestions, otherwise LGTM! 👍

docs/source/resources/define-metadata.rst
```diff
@@ -1706,6 +1706,70 @@ retrieve a fully rendered ``meta.yaml``, use the
...
+  first match as a Python ``re.Match object``. For example::
+
+    {% set readme_heading = load_file_regex(load_file='README.rst', regex_pattern=r'^# (\S+)') %}
+    package:
+      name: {{ readme_heading.string }}
+
+* ``load_file_data``: You can also parse JSON, TOML, or YAML files and load data
+  from them. For example you can use this to load poetry configurations from
+  ``pytproject.toml``. This is especially useful as ``setup.py`` is no longer the
+  only standard way to define project metadata (see PEPs 517 and 518)::
```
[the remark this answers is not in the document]
kenodegard:
```suggestion
  only standard way to define project metadata (see
  `PEP 517 <https://peps.python.org/pep-0517>`_ and
  `PEP 518 <https://peps.python.org/pep-0518>`_)::
```

kenodegard:
@bryevdv We are working to define a release cadence for conda and conda-build (see conda-incubator/ceps#26) but I'd expect to see the next release in 1-2 months.

frgfm:
Hello there :wave: 

That's a wonderful feature! I stumbled across it in the documentation and tried to use. The loading of the `pyproject.toml` fails silently because this commit hasn't made its way to a release. But its presence in the documentation is confusing (I figured it out coming on this PR and checking the last available release which was right before).

I see that you intend on making regular releases, any plan to make a conda-build release soon? :) :pray:

marscher:
I'd also be glad, if this nice feature could make it to a release soon :+1: 

@frgfm I agree, that the documentation should point to "stable" instead of latest. I happened to me in the past as well. Discovered a bleeding edge feature in the main branch docs and tried it without success. But that's for another issue, right?

PR #4491, opened 2022-05-31
Fix spelling of pytest

jugmac00:
pytest dropped the dot as in `py.test` already a couple of years ago.


jugmac00:
I had a look at the commit history, and it looks like for trivial changes there is no need to include a change log entry.

jezdez (approved):

PR #4495, opened 2022-06-02
Add `patch`/`m2-patch` as a hard dependency

kenodegard:
### Description

While doing the recent conda-build 3.21.9 release (#4483) we noticed that the [`test_ensure_unix_line_endings_with_nonutf8_characters` test would fail](https://github.com/conda/conda-build/runs/6667921343?check_suite_focus=true#step:11:1160) during cleanup of it's temporary files. No good reason for why this occurred as the test would usually pass. 🤷🏻‍♂️ 

#### Rabbit hole
In trying to rework the test to be more robust we learned a couple of things:
1. The tests defined in `tests/test_patch.py` weren't actually testing what they claimed to test. They claimed to be able to successfully apply a mismatched line endings patch (e.g. an LF patch could be applied to a CRLF source and vice-versa). The tests were instead consistently testing matching line endings.
3. In trying to correct the test to actually test the expected behavior we found that the claim was in fact false. `patch` will not apply a patch with a mismatched line endings source.
4. We also discovered that the necessary `--binary` flag for `patch` while originally introduced in `patch` 2.3 does not work as we expect/need it to until it was amended in `patch` 2.6. MacOS however still ships `patch` 2.5.

#### Solution
With all of this in mind, we propose adding `patch`/`m2-patch`>=2.6 as a hard dependency of conda-build in addition to the other improvements to `conda_build.source._ensure_LF` and `conda_build.source._ensure_CRLF` and their tests.

### Checklist - did you ...

- [x] Add a file to the `news` directory ([using the template](https://github.com/conda/conda/blob/master/news/TEMPLATE)) for the next release's release notes?
     
- [X] ~Add / update outdated documentation?~

### DIFF
--- a/conda_build/source.py
+++ b/conda_build/source.py
@@ -787,6 +787,7 @@ def apply_one_patch(src_dir, recipe_dir, rel_path, config, git=None):
     attributes_output = ""
     # While --binary was first introduced in patch 2.3 it wasn't until patch 2.6 that it produced
     # consistent results across OSes. So patch/m2-patch is a hard dependency of conda-build.
+    # See conda-build#4495
     patch_exe = external.find_executable("patch")
     if not patch_exe:
         raise MissingDependency("Failed to find conda-build dependency: 'patch'")
--- a/conda_build/source.py
+++ b/conda_build/source.py
@@ -493,9 +493,7 @@ _RE_LF = re.compile(rb"(?<!\r)\n")
 _RE_CRLF = re.compile(rb"\r\n")
 
 
-def _ensure_unix_line_endings(
-    src: os.PathLike, dst: Optional[os.PathLike] = None
-) -> Path:
+def _ensure_LF(src: os.PathLike, dst: Optional[os.PathLike] = None) -> Path:
     """Replace windows line endings with Unix.  Return path to modified file."""
     src = Path(src)
     dst = Path(dst or src)  # overwrite src if dst is undefined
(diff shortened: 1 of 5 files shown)


conda_build/source.py
```diff
@@ -806,15 +803,18 @@ def try_apply_patch(patch, patch_args, cwd, stdout, stderr):
...
+            raise RuntimeError(
+                dedent(
+                    f"""
+                    Cannot use 'git' (not a git repo and/or patch) and did not find 'patch' in:
+                        {os.pathsep.join(external.dir_paths)}
+                    You can install 'patch' using apt-get, yum (Linux), Xcode (MacOSX),
+                    or conda, m2-patch (Windows),
+                    """
+                ).lstrip()
+            )
```
kenodegard, line 816: Since `patch` is now a hard dependency do we remove this check entirely or do we move this check up with `patch_exe = external.find_executable("patch")` and error with something more abrasive?

jezdez: Yeah, that makes sense.

tests/test_patch.py
```diff
@@ -28,71 +34,101 @@ def test_patch_strip_level(testing_workdir, monkeypatch):
...
+            """
+        ).lstrip()
+    )
+
+    return paths
 
 
 # TODO :: These should require a build env with patch (or m2-patch) in it.
 #         at present, only ci/github/install_conda_build_test_deps installs
 #         this.
```
kenodegard, line 75: This can be dropped.
```suggestion
```

jezdez (approved):
Thank you @kenodegard!

PR #4516, opened 2022-06-24
Replace `r-cortools` with `r-badger` for "Artistic-2.0" license testing

kenodegard:
CRAN R archived `cortools` on 2022-06-20 so this test fails.

Found a different "Artistic-2.0" licensed package (`badger`) to use instead.


beeankha (approved):
Thanks for fixing this, @kenodegard !

jezdez (approved):

PR #4569, opened 2022-08-19
Cleanup of `main_build.py`

kenodegard:
### Description

A noop PR, just cleaning up formatting and using appropriate helper functions.

### Checklist - did you ...

- [x] ~Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release notes?~
     
- [x] ~Add / update necessary tests?~
- [x] ~Add / update outdated documentation?~


conda_build/cli/main_build.py
```diff
@@ -25,21 +26,19 @@
...
-different sets of packages."""
+    p.description = dals(
+        """
+        Tool for building conda packages. A conda package is a binary tarball
+        containing system-level libraries, Python modules, executable programs, or
+        other components. conda keeps track of dependencies between packages and
+        platform specifics, making it simple to create working environments from
+        different sets of packages.
+        """
+    )
```
dholth, line 41: I don't like that dals() processes the text every single time, at least this function is not called many times.

kenodegard: Agreed, but until we decide on a policy for what to do with blocks of text (e.g. should they be triple quote strings with dedent or string concatenation with explicit `\n`), I will continue using `dals` anywhere I see paragraphs.

dholth:
When will we run black on the entire conda source code, commit that, and mark it as a revision to ignore for blame purposes?

kenodegard:
Probably not for another year or so. Looking at the changes here now I'm not sure there's value in doing the string reformatting as I did. Too much noise.

conda_build/cli/main_build.py
```diff
@@ -441,29 +441,22 @@ def execute(args):
...
+        outputs = []
         failed_recipes = []
-        recipes = [item for sublist in
-                   [glob(os.path.abspath(recipe)) if '*' in recipe
-                                                  else [recipe] for recipe in args.recipe]
-                   for item in sublist]
+        recipes = chain.from_iterable(
+            glob(abspath(recipe)) if "*" in recipe else [recipe]
+            for recipe in args.recipe
+        )
```
dholth, line 456: how does from_iterable differ from chain() here

kenodegard: `chain` expects multiple arguments, `chain.from_iterable` expects a singular argument:

```
chain(*iterable) == chain.from_iterable(iterable)
```

dholth (dismissed):

PR #4571, opened 2022-08-22
faster tests from standalone conda-index

dholth:
### Description

The conda-build tests are very slow. Although conda-index should be standalone, conda-build still has, and uses, a version of this code. Copy standalone conda-index's tests back into conda-build for a big speedup - using cached packages instead of re-downloading / building recipes from someone's git every time the tests run.

`test_index` should take < 30s to run after this change.


tests/index_hotfix_pkgs/channeldata.json
```diff
@@ -0,0 +1,129 @@
+{
```
dholth, line 1: The tests have always modified files in here. Not sure if these indexes are necessary, or if they will be regenerated as we test conda-index.

tests/test_index.py
```diff
@@ -866,38 +934,76 @@ def test_patch_from_tarball(testing_workdir):
...
-    out_files = api.build(testing_metadata)
-    for f in out_files:
+
+    archive_name = "test_index_of_removed_pkg-1.0-1.tar.bz2"
+    archive_destination = os.path.join(
+        testing_metadata.config.croot, TEST_SUBDIR, archive_name
+    )
+
+    # conda_build.api.build() calls update_index as a side effect.
+    # used to build on every test (38 seconds)
```
dholth, line 953: this is what we are avoiding

dholth:
My least favorite thing about this is that I feel like we have an odd & random choice of packages in our test index. Changing it would mean redoing all the expected output. Maybe we could use the info-only packaging trick (unpack conda package; repack only the info/ directory) to shrink these down...

msarahan (approved):
The test failures look like noise to me - a connection error on CRAN, and a Windows file SNAFU.

tests/utils.py
```diff
@@ -125,18 +135,20 @@ def put_bad_conda_on_path(testing_workdir):
     except:
         raise
```
sven6002, line 138: ```raise``` could have a message which would be cleaner if the test case failed

dholth: This re-raises the exception that triggered the except clause, with whatever message it already had

tests/utils.py
```diff
@@ -101,22 +109,24 @@ def assert_package_consistency(package_path):
...
     # it is easier to add an intentionally bad path than it is to try to scrub any existing path
-    os.environ['PATH'] = os.pathsep.join([testing_workdir, os.environ["PATH"]])
+    os.environ["PATH"] = os.pathsep.join([testing_workdir, os.environ["PATH"]])
 
-    exe_name = 'conda.bat' if on_win else 'conda'
+    exe_name = "conda.bat" if on_win else "conda"
     out_exe = os.path.join(testing_workdir, exe_name)
-    with open(out_exe, 'w') as f:
+    with open(out_exe, "w") as f:
         f.write("exit 1")
```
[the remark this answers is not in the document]
dholth: `f` is automatically closed by the `with` statement when it is left

tests/utils.py
```diff
@@ -50,20 +51,23 @@ def assert_package_consistency(package_path):
...
+
     try:
         with tarfile.open(package_path) as t:
             # Read info from tar file
             member_list = t.getnames()
-            files = t.extractfile('info/files').read().decode('utf-8')
+            files = t.extractfile("info/files").read().decode("utf-8")
             # Read info/has_prefix if present
-            if 'info/has_prefix' in member_list:
+            if "info/has_prefix" in member_list:
```
sven6002, line 63: is this filename fixed?? If so this can be an constant itself. (Following a pattern idea

tests/conftest.py
```diff
@@ -78,77 +85,97 @@ def return_to_saved_path():
...
+    d["requirements"]["build"] = []
+    d["requirements"]["run"] = []
+    d["test"]["commands"] = ['echo "A-OK"', "exit 0"]
+    d["about"]["home"] = "sweet home"
+    d["about"]["license"] = "contract in blood"
+    d["about"]["summary"] = "a test package"
+    d["about"]["tags"] = ["a", "b"]
+    d["about"]["identifiers"] = "a"
     testing_config.variant = get_default_variant(testing_config)
     testing_config.variants = [testing_config.variant]
```
sven6002, line 139: Where are the asserts?? for the test metadata

tests/utils.py
dholth: this is just unchanged test code that was reformatted

tests/conftest.py
```diff
@@ -78,77 +85,97 @@ def return_to_saved_path():
...
+            "-p",
+            env_path,
+            "python={}".format(".".join(sys.version.split(".")[:2])),
+        ]
+    )
+    monkeypatch.setenv(
+        "PATH",
+        prepend_bin_path(os.environ.copy(), env_path, prepend_prefix=True)["PATH"],
+    )
     # cleanup is done by just cleaning up the testing_workdir
```
sven6002, line 161: same thing what is the test actually testing?

tests/conftest.py
dholth: `@pytest.fixture` passes this data to any test function that has a same-named parameter `def test_fn(testing_metadata):`

tests/conftest.py
```diff
@@ -78,77 +85,97 @@ def return_to_saved_path():
...
     return env_path
 
 
 # these are functions so that they get regenerated each time we use them.
 #    They could be fixtures, I guess.
-@pytest.fixture(scope='function')
```
sven6002, line 142: these defnitions should be moved to the ```utils.py```

dholth: `pytest` requires certain setup or configuration to go into `conftest.py`

tests/conftest.py
kenodegard: `conftest.py` is a special place for defining [pytest fixtures](https://docs.pytest.org/en/6.2.x/fixture.html), fixtures are pytest's way of implementing setup/teardown logic

sven6002 (approved):

tests/test_index.py
```diff
@@ -185,41 +188,47 @@ def test_file_index_on_single_subdir_1(testing_workdir):
...
+    #     "build_number": 0,
+    #     "depends": [],
+    #     "license": "Apache",
+    #     "md5": "2e84ee54415a5021db050bd5fa5438b2",
+    #     "name": "fly",
+    #     "sha256": "79dec46aaa827ffde1bc069f740697b0d126f485ed9cb4c3db201f720601d346",
+    #     "size": 5382961,
+    #     "subdir": "osx-64",
+    #     "version": "2.5.2"
+    # }
```
travishathaway, line 224:
```suggestion

    updated_packages = expected_repodata_json.get("packages")

```

travishathaway: If we don't need it, let's get rid of it.

tests/test_index.py
```diff
@@ -262,97 +271,96 @@ def test_file_index_on_single_subdir_1(testing_workdir):
...
+            #     "run_exports": {},
+            #     "source_git_url": "/Users/scastellarin/projects/fly-package/https:/github.com/concourse/concourse.git",
+            #     "source_url": None,
+            #     "subdirs": ["osx-64"],
+            #     "summary": "A command line interface that runs a build in a container with ATC.",
+            #     "tags": None,
+            #     "text_prefix": False,
+            #     "timestamp": 0,
+            #     "version": "2.5.2",
+            # },
```
travishathaway, line 310:
```suggestion
```

Yank it out!

tests/test_index.py
```diff
@@ -582,64 +605,79 @@ def test_file_index_noarch_osx64_1(testing_workdir):
...
 
-    with open(os.path.join(workdir, subdir, 'repodata.json')) as f:
+    # build index_hotfix_pkgs: pkgs =
+    # conda_build.api.build(os.path.join(utils.metadata_dir,
+    # "_index_hotfix_pkgs"), croot=workdir) for pkg in pkgs:
+    #     conda_package_handling.api.transmute(pkg, ".conda")
+    # conda_build.api.update_index(workdir) # not testing update_index in the
+    # _build_test_index() helper
+
+    # workdir may be the same during a single test run?
```
travishathaway, line 663: I think this comment would be better phrased as a statement of uncertainty rather than a question (more direct that way).

As someone reading this code for the first time, I find questions like these to be a little confusing.

tests/test_index.py
```diff
@@ -582,64 +605,79 @@ def test_file_index_noarch_osx64_1(testing_workdir):
...
-    api.update_index(workdir)
 
-    with open(os.path.join(workdir, subdir, 'repodata.json')) as f:
+    # build index_hotfix_pkgs: pkgs =
+    # conda_build.api.build(os.path.join(utils.metadata_dir,
+    # "_index_hotfix_pkgs"), croot=workdir) for pkg in pkgs:
+    #     conda_package_handling.api.transmute(pkg, ".conda")
+    # conda_build.api.update_index(workdir) # not testing update_index in the
+    # _build_test_index() helper
+
```
travishathaway, line 662:
```suggestion
```

Nuke it! 💥

tests/test_index.py
```diff
@@ -582,64 +605,79 @@ def test_file_index_noarch_osx64_1(testing_workdir):
...
         "subdirs": [
             "noarch",
             "osx-64",
-        ]
+        ],
     }
     assert actual_channeldata_json == expected_channeldata_json
 
 
 def _build_test_index(workdir):
```
travishathaway, line 654: Doc string please 🙏

tests/test_index.py
```diff
@@ -649,11 +687,15 @@ def _build_test_index(workdir):
...
-    assert "package_has_been_revoked" not in pkg_list["revoke_test-1.0-0.tar.bz2"]["depends"]
+    assert (
+        "package_has_been_revoked"
+        not in pkg_list["revoke_test-1.0-0.tar.bz2"]["depends"]
+    )
 
     assert "remove_test-1.0-0.tar.bz2" in pkg_list
 
 
+# SLOW
```
travishathaway, line 698: How about we add a custom slow marker?

https://docs.pytest.org/en/6.2.x/mark.html

Let's use pytest's full capabilities 💪

tests/test_index.py
travishathaway: Perhaps this information would be better in a doc string? If you can't tell by now I really dislike leaving old code around that is commented out 😅

travishathaway (changes requested):
Left some suggestions and comments. Please reconsider where you could add some doc strings. I also highly recommend taking advantage of pytest's markers so that we can identify slow running tests. This would be a good thing to start doing for this repository.

Thank you for doing the work necessary to increase the speed of the tests! 🐇

dholth:
@travishathaway thank you for your comments. We will remove the entire index function from the conda-build project, addressing your test concerns. Then we can improve the tests in the standalone conda-index project.

tests/test_index.py
dholth: Fixed

travishathaway (approved):
Looks good. I would still like to see us implement pytest markers for conda-build, but that perhaps should be dealt with separately.

Issue #4601, opened 2022-10-14
Doc "Concepts » Channels and generating an index" not working on osx-arm64 


tnabtaf:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### What happened?

I am using [this doc page](https://docs.conda.io/projects/conda-build/en/latest/concepts/generating-index.html) to better understand channel creation.  

I am running on macOS 12.6 **with an Apple M1 Pro chip**

The example says to download the linux-64 and **osx-64** versions of SciPy from the anaconda channel, which I did.

It then says to run `conda index .` which I did.  And it worked!

However, when I did the next step:

`$ conda search -c file:////Users/clements/Documents/Anaconda/Conda/Conda-Index-Testing/local-channel scipy | grep local-channel`

I got nothing back.  If I drop the `grep` then `conda search` command tells me it finds it in `pkgs/main`.

If I run `conda search` on a non-existing package 

`conda search -c file:////Users/clements/Documents/Anaconda/Conda/Conda-Index-Testing/local-channel spork`

it tells me

```
Loading channels: done
No match found for: spork. Search: *spork*

PackagesNotFoundError: The following packages are not available from current channels:

  - spork

Current channels:

  - file://users/clements/Documents/Anaconda/Conda/Conda-Index-Testing/local-channel/osx-arm64
  - file://users/clements/Documents/Anaconda/Conda/Conda-Index-Testing/local-channel/noarch
  - https://repo.anaconda.com/pkgs/main/osx-arm64
  - https://repo.anaconda.com/pkgs/main/noarch
  - https://repo.anaconda.com/pkgs/r/osx-arm64
  - https://repo.anaconda.com/pkgs/r/noarch

To search for alternate channels that may provide the conda package you're
looking for, navigate to

    https://anaconda.org

and use the search bar at the top of the page.
```

And there's the rub.  Note that the two local channels it is searching are

```
file://users/clements/Documents/Anaconda/Conda/Conda-Index-Testing/local-channel/osx-arm64
file://users/clements/Documents/Anaconda/Conda/Conda-Index-Testing/local-channel/noarch
```

but I didn't create or put anything in either of those directories.  Both of those directories were created when I ran the `conda index` command.  

It looks like `conda search` figures out that I am running on an `osx-arm64` platform, not on an `osx-64` or a `linux-64` platform, so it doesn't look in either of those two directories, which is where I copied the SciPy packages to.

## Solution?

[...]

PR #4602, opened 2022-10-14
Updated the "Generated an index" example to include more platforms.

tnabtaf:
Addresses issue #4601.

### Description

As documented in issue #4601 I had issues following the 'build a channel index' example in the docs.  This PR adds support for Windows and the new Apple Silicon to the list.  Also adds a section about how to get `wget` if you don't already have it.

### Checklist - did you ...

### DIFF
--- a/docs/source/concepts/generating-index.rst
+++ b/docs/source/concepts/generating-index.rst
@@ -130,62 +130,56 @@ To build a local channel and put a package in it, follow the directions below.
 
     .. code-block:: bash
 
-      $ mkdir local-channel
-      $ cd local-channel
+        $ mkdir local-channel
+        $ cd local-channel
 
-#. Install ``wget`` if you don't already have it.
-
-    #.  Check if you have it.
-
-      #. Windows
-
-          .. code-block:: powershell
-
-            $ where wget.*
-
-      #. Linux, macOS
-
-          .. code-block:: bash
-
-            $ which wget
-
-    #. If ``wget`` is not found, then install it:
-
-        .. code-block:: bash
-
-          $ conda install wget
-
-#. Put your favorite package in the local channel.
-
-    The steps here depend on your platform.
+#. Now, download your favorite package.  We'll use SciPy in our example.  The next steps depend on your platform.
 
     #. Windows
 
         .. code-block:: bash
 
-          $ mkdir win-64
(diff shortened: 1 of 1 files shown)


travishathaway (approved):

docs/source/concepts/generating-index.rst
```diff
@@ -126,20 +126,62 @@ Example: Building a channel
...
-      $ mkdir linux-64 osx-64
 
-#. Put your favorite package in the channel.
+#. Install ``wget`` if you don't already have it.
+
+    Check if you have it.
+
+    .. code-block:: bash
+
+      $ which wget
```
kathatherine, line 142: "which" is only a unix command, so this method is only available to mac and Linux users. 

I believe `where.exe wget` would be the Windows equivalent.

docs/source/concepts/generating-index.rst
```diff
@@ -110,7 +110,7 @@ For each subdir:
 
 * Remove all packages that need to be removed.
 
-For all packages that need to be added/updated:
+* For all packages that need to be added/updated:
 
   * Extract the package to access metadata including full package name,
```
kathatherine, line 115: It'd be great to add a comma here after "metadata", too.

kathatherine (changes requested):
Couple things. Thanks for looking into this, Dave!

docs/source/concepts/generating-index.rst
tnabtaf: Thanks Katherine.  I had assumed that any Windows users will be using Windows Subsystem for Linux. I'll add this.

tnabtaf: Updated. But this is dangerous as I don't understand windows, and don't have access to a windows machine.

kathatherine: I have a vanilla Windows VM, so I attempted some testing. A few things:

- I can't get `conda install wget` to work. I had to pip install it.
- Once I did get it installed in an environment, `where wget`, `where.exe wget`, and `where wget.*` all give me the same thing: `INFO: Could not find files for the given pattern(s).`

So, seems like vanilla Windows is going to be a bit more complicated than everything Unix based.

kathatherine (approved):
Thanks so much for this, Dave! I trust you tested it as much as you were able, so approved!

tnabtaf:
@kathatherine I tested this on a clean Windows 10 install.

jezdez (approved):

Issue #4572, opened 2022-08-24
Fix `python -m` issue


kenodegard:
We currently have failing canary builds, and we are seeing two different types of errors:

-  [ModuleNotFoundError: No module named 'tlz'](https://github.com/conda/conda/runs/7731850991?check_suite_focus=true#step:10:803)
- [ModuleNotFoundError: No module named 'conda'](https://github.com/conda/conda/runs/7774814177?check_suite_focus=true#step:10:2152)

To address this, we first need to recreate these errors outside of CI and then implement a fix based on our findings. A fair amount of research has already been accomplished in this issue #4561.

One thing to keep in mind is that there are several different places where a fix could potentially be implemented:

- [conda itself](https://github.com/conda/conda)
- [conda-build](https://github.com/conda/conda-build)
- [conda-feedstock](https://github.com/AnacondaRecipes/conda-feedstock)

travishathaway:
We need to get this test working: https://github.com/conda/conda/blob/main/conda.recipe/run_test.sh

travishathaway:
Looks like the fix may be adding this line near the top of the `run_test.sh` script:

```bash
conda init --all
```

The important operation that this command performs is modifying the entry point scripts. I got the following output after testing it successfully locally:

```bash
modified      $PREFIX/condabin/conda
modified      $PREFIX/bin/conda
modified      $PREFIX/bin/conda-env
# remaining output truncated...
```

We can assume that it modified the entry point script to actually point at the Python interpreter we care about. This solution still isn't 100% satisfactory though as I really wish I knew what was causing this to be created incorrectly in the first place. But, I also fear that may be a rabbit hole not worth venturing down, so I will propose this solution in a separate pull request for our [conda feedstock](https://github.com/AnacondaRecipes/conda-feedstock).

travishathaway:
I'm going to block this issue until we do our release this week. There were some issues with the build on [conda-feedstock](https://github.com/AnacondaRecipes/conda-feedstock) that will be resolved after this upcoming release.

travishathaway:
This issue has been resolved by recent updates to the [AnacondaRecipe/conda-feedstock](https://github.com/AnacondaRecipes/conda-feedstock/pull/14)

Issue #4561, opened 2022-08-16
SPIKE: Investigate `python -m` usage


kenodegard:
### What happened?

While working on recent issues:

- https://github.com/conda/conda/pull/11589
- https://github.com/conda/conda/issues/11698
- https://github.com/conda/conda-build/pull/4556
- https://github.com/conda/conda/pull/11704
- https://github.com/conda/conda/pull/11700
- https://github.com/conda/conda/pull/11709
- https://github.com/conda/conda/pull/11666
- https://github.com/conda/conda/issues/11724
- https://github.com/conda/conda/pull/11720

we found that conda-build invokes conda via `python -m` in a few instances, causing issues when building conda itself:

- https://github.com/conda/conda/actions/runs/2820051398
- https://github.com/conda/conda/actions/runs/2835251119

Instead of invoking and using the installed conda to build the new package, conda-build manages to invoke the source code. This appears to be a result of the `python -m` side-effect.

dholth:
```[-I](https://docs.python.org/3/using/cmdline.html#cmdoption-I) option can be used to run the script in isolated mode where [sys.path](https://docs.python.org/3/library/sys.html#sys.path) contains neither the current directory nor the user’s site-packages directory. All PYTHON* environment variables are ignored, too.``` 

https://docs.python.org/3/using/cmdline.html#cmdoption-m

kenodegard:
# `python -m`

https://github.com/conda/conda-build/blob/6a1cea0111f53417aeb631aa54d7cc26d26d6628/conda_build/skeletons/pypi.py#L368

https://github.com/conda/conda-build/blob/6a1cea0111f53417aeb631aa54d7cc26d26d6628/conda_build/build.py#L1825

https://github.com/conda/conda-build/blob/6a1cea0111f53417aeb631aa54d7cc26d26d6628/conda_build/build.py#L2740

# `_CE_M`: shell activation `--dev` "feature"

https://github.com/conda/conda-build/blob/6a1cea0111f53417aeb631aa54d7cc26d26d6628/conda_build/build.py#L2731-L2738

PR #4604, opened 2022-10-19
Run Python in isolated mode unless in dev mode

jaimergp:
### Description

`conda build` generates some shell scripts wrappers to run `build.sh` and `bld.bat`. This includes setting a number of environment variables and activating the host and build environments in a stacked way. To do this, it employs `python -m conda shell.posix hook`. 

However, the `-m` semantics implies that the current working directories (`SRC_DIR` for conda-build!) and other user installation paths are considered as part of `sys.path`. This is usually ok, but there can be corner cases:

* If we are packaging `conda` and the version being packaged on `SRC_DIR` happens to change how `shell.posix hook` works, it might produce different code than what the released conda-build expects. 
* Not sure, but if we are packaging a dependency needed during the `hook` generation steps, it might also break something?

This discrepancy of which `conda` is used (the one on `SRC_DIR` vs the one installed next to `conda-build`) introduces a fragile behavior, but it can be easily fixed by just adding [`-I` (isolated mode)](https://docs.python.org/3/using/cmdline.html#cmdoption-I) to prevent leakage from non-installed locations:

> `-I`
> Run Python in isolated mode. This also implies -E and -s. In isolated mode [sys.path](https://docs.python.org/3/library/sys.html#sys.path) contains neither the script’s directory nor the user’s site-packages directory. All PYTHON* environment variables are ignored, too. Further restrictions may be imposed to prevent the user from injecting malicious code.

> `-m <module-name>`
> Search [sys.path](https://docs.python.org/3/library/sys.html#sys.path) for the named module and execute its contents as the [__main__](https://docs.python.org/3/library/__main__.html#module-__main__) module.
> ...
> [`-I`](https://docs.python.org/3/using/cmdline.html#cmdoption-I) option can be used to run the script in isolated mode where [sys.path](https://docs.python.org/3/library/sys.html#sys.path) contains neither the current directory nor the user’s site-packages directory. All PYTHON* environment variables are ignored, too.

#### Why?

This bug has been discovered at https://github.com/conda/conda/pull/11995. The sequence goes as:

* Without this PR, `conda-build` uses `python -m conda` for its shell integration. This means that it first finds `conda` the working directory (`SRC_DIR`) executes the hook as produced by the version being packaged.
* If we package https://github.com/conda/conda/pull/11995, the hook will generate shell code with the form of `python -I -m conda ...` instead of `CONDA_PREFIX/bin/conda` (the Python entry point) as the Python command to avoid shebang issues. 
* The `-I` flag will force the next `conda activate` commands to run from the `conda` installation placed next to `conda-build`, not the one from the working directory! 
* In this particular timing, we are mixing two shell code generation strategies. The former one uses `CONDA_PREFIX/bin/conda` and unsets `_CE_M` and `_CE_CONDA` (the shell variables that hold `-m` and `conda` strings, respectively). 
* This causes the hook code to be run like `python shell.posix activate`, where it should have been `python -I -m conda shell.posix activate`. Of course, `shell.posix` is interpreted as a path to a file that doesn't exist and Python errors out.

I know this is a corner case, but in my opinion, conda-build shouldn't rely on software being packaged as part of its implementation :)

### Checklist - did you ...

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -1821,7 +1821,7 @@ def _write_sh_activation_text(file_handle, m):
                             cygpath_suffix))
 
     if conda_46:
-        py_flags = '-m' if m.config.debug else '-m -I'
+        py_flags = '-m' if m.config.debug else '-I -m'
         file_handle.write(
             f"""eval "$('{sys.executable}' {py_flags} conda shell.bash hook)"\n"""
         )
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -1821,8 +1821,8 @@ def _write_sh_activation_text(file_handle, m):
                             cygpath_suffix))
 
     if conda_46:
-        file_handle.write("eval \"$('{sys_python}' -m conda shell.bash hook)\"\n".format(
-            sys_python=sys.executable,
+        file_handle.write("eval \"$('{sys_python}'{I_flag} -m conda shell.bash hook)\"\n".format(
+            sys_python=sys.executable, I_flag=('' if m.config.debug else ' -I'),
         ))
 
     if m.is_cross:
@@ -2729,15 +2729,18 @@ def write_test_scripts(metadata, env_vars, py_files, pl_files, lua_files, r_file
                 if utils.on_win:
                     tf.write('set "CONDA_SHLVL=" '
                              '&& @CALL {}\\condabin\\conda_hook.bat {}'
-                             '&& set CONDA_EXE={}'
+                             '&& set CONDA_EXE={python_exe}'
+                             '&& set CONDA_PYTHON_EXE={python_exe}'
+                             '&& set _CE_I={}'
                              '&& set _CE_M=-m'
                              '&& set _CE_CONDA=conda\n'.format(sys.prefix,
                                                                '--dev' if metadata.config.debug else '',
-                                                               sys.executable))
+                                                               '' if metadata.config.debug else '-I',
+                                                               python_exe=sys.executable))
 
                 else:
-                    tf.write("eval \"$('{sys_python}' -m conda shell.bash hook)\"\n".format(
-                        sys_python=sys.executable))
+                    tf.write("eval \"$('{sys_python}'{I_flag} -m conda shell.bash hook)\"\n".format(
+                        sys_python=sys.executable, I_flag=('' if metadata.config.debug else ' -I')))
                 tf.write(f'conda activate "{metadata.config.test_prefix}"\n')
             else:
                 tf.write('{source} "{conda_root}activate{ext}" "{test_env}"\n'.format(
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -1821,9 +1821,10 @@ def _write_sh_activation_text(file_handle, m):
                             cygpath_suffix))
 
     if conda_46:
-        file_handle.write("eval \"$('{sys_python}'{I_flag} -m conda shell.bash hook)\"\n".format(
-            sys_python=sys.executable, I_flag=('' if m.config.debug else ' -I'),
-        ))
+        py_flags = '-m' if m.config.debug else '-m -I'
+        file_handle.write(
+            f"""eval "$('{sys.executable}' {py_flags} conda shell.bash hook)"\n"""
+        )
 
     if m.is_cross:
         # HACK: we need both build and host envs "active" - i.e. on PATH,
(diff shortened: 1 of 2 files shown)


jaimergp:
The failure above was due to temporary connection issues, but I don't have permissions to issue a re-run there. Otherwise, ready for review!

jaimergp:
cc @conda/conda-core (there's no `conda-build` team, right?)

conda_build/build.py
```diff
@@ -1821,8 +1821,8 @@ def _write_sh_activation_text(file_handle, m):
                             cygpath_suffix))
 
     if conda_46:
-        file_handle.write("eval \"$('{sys_python}' -m conda shell.bash hook)\"\n".format(
-            sys_python=sys.executable,
+        file_handle.write("eval \"$('{sys_python}'{I_flag} -m conda shell.bash hook)\"\n".format(
+            sys_python=sys.executable, I_flag=('' if m.config.debug else ' -I'),
         ))
```
kenodegard, line 1826: Ah, that's why I didn't get it to work! When I had looked at the `-m` issue previously, I thought we could simply replace `-m` with `-I`.

```suggestion
        file_handle.write(
            f"""eval "$('{sys.executable}' {'-m' if m.config.debug else '-m -I'} conda shell.bash hook)"\n"""
        )
```

conda_build/build.py
```diff
@@ -2729,15 +2729,18 @@ def write_test_scripts(metadata, env_vars, py_files, pl_files, lua_files, r_file
...
                                                                '--dev' if metadata.config.debug else '',
-                                                               sys.executable))
+                                                               '' if metadata.config.debug else '-I',
+                                                               python_exe=sys.executable))
 
                 else:
-                    tf.write("eval \"$('{sys_python}' -m conda shell.bash hook)\"\n".format(
-                        sys_python=sys.executable))
+                    tf.write("eval \"$('{sys_python}'{I_flag} -m conda shell.bash hook)\"\n".format(
+                        sys_python=sys.executable, I_flag=('' if metadata.config.debug else ' -I')))
```
kenodegard, line 2743: Ditto

conda_build/build.py
```diff
@@ -2729,15 +2729,18 @@ def write_test_scripts(metadata, env_vars, py_files, pl_files, lua_files, r_file
                 if utils.on_win:
                     tf.write('set "CONDA_SHLVL=" '
                              '&& @CALL {}\\condabin\\conda_hook.bat {}'
-                             '&& set CONDA_EXE={}'
+                             '&& set CONDA_EXE={python_exe}'
+                             '&& set CONDA_PYTHON_EXE={python_exe}'
+                             '&& set _CE_I={}'
```
kenodegard, line 2734: I want to change this to `_CE_FLAGS`? Is that possible, or will that result in too many other changes?

kenodegard (changes requested):
@jaimergp hehe, you ran into the same issue we've been running into during recent releases, but unfortunately, we found workarounds at the time, so further work was postponed:
- https://github.com/conda/conda-build/issues/4561
- https://github.com/conda/conda-build/issues/4572

conda_build/build.py
```diff
@@ -1821,9 +1821,10 @@ def _write_sh_activation_text(file_handle, m):
                             cygpath_suffix))
 
     if conda_46:
-        file_handle.write("eval \"$('{sys_python}' -m conda shell.bash hook)\"\n".format(
-            sys_python=sys.executable,
-        ))
+        py_flags = '-m' if m.config.debug else '-m -I'
```
kenodegard, line 1824: Looks like it needs to be `-I -m` https://docs.python.org/3/using/cmdline.html#command-line

jaimergp: Good catch!
```suggestion
        py_flags = '-m' if m.config.debug else '-I -m'
```

dholth:
Are the CI tests run in debug mode?

conda_build/build.py
```diff
@@ -2727,17 +2728,25 @@ def write_test_scripts(metadata, env_vars, py_files, pl_files, lua_files, r_file
...
+                            sys.prefix,
+                            '--dev' if metadata.config.debug else '',
+                            '' if metadata.config.debug else '-I',
+                            python_exe=sys.executable
+                        )
+                    )
                 else:
-                    tf.write("eval \"$('{sys_python}' -m conda shell.bash hook)\"\n".format(
-                        sys_python=sys.executable))
+                    py_flags = '-m' if metadata.config.debug else '-m -I'
```
kenodegard, line 2746: And here:

```suggestion
                    py_flags = '-m' if metadata.config.debug else '-I -m'
```

kenodegard (approved):

jaimergp:
Thanks!

PR #4612, opened 2022-11-02
Use updated channel label

kenodegard:
### Description

We recently altered the label (see https://github.com/conda/conda/pull/12040#discussion_r1010173002) used when pushing builds to conda-canary, allowing us to push builds for both main and feature branches. Updating canary tests here to use the new label.

### Checklist - did you ...


kenodegard:
This is actually undesired since the main label is the same as being unlabled, the following are identical:

```
conda install -c conda-canary/label/main conda
conda install -c conda-canary conda
```

See https://github.com/conda/conda/pull/12048

PR #4619, opened 2022-11-04
Update path logic to avoid Windows path sensitivity

kenodegard:
### Description

`test_convert` in `tests/test_cli.py` has become particularly flaky on Windows recently (updates to the Windows runners perhaps?). This is an attempt to adjust the code to avoid this flakiness.

### Checklist - did you ...

### DIFF
--- a/conda_build/convert.py
+++ b/conda_build/convert.py
@@ -273,18 +273,21 @@ def update_lib_contents(lib_directory, temp_dir, target_platform, file_path):
     elif target_platform == 'unix':
         temp_dir = Path(temp_dir)
         src_dir = temp_dir / "Lib"
-        dest_dir = temp_dir / "lib"
+        dst_dir = temp_dir / "lib"
 
         # rename Lib to lib (unix is case sensitive)
-        src_dir.rename(dest_dir)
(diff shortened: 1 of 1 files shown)


chenghlee (dismissed):

dholth:
Why does it need to rename source dir to dest dir (different case?) That is the strange thing about this code.

Maybe it could rename source dir to intermediate-dir, and then rename to source dir (different case)

kenodegard:
@dholth 🤷🏼‍♂️ renaming src to dst is an atomic operation?

based on the [error message](https://github.com/conda/conda-build/actions/runs/3388870367/jobs/5633340193#step:11:1176) I don't think an intermediate dir would make a difference?

dholth:
Considering the possibility that renaming Lib to lib is not allowed because Lib already exists.

Directory or file renames are atomic.

dholth:
@kenodegard I'll walk across the room to my Windows machine and try this out.

kenodegard:
Lol the test is still flaky, see https://github.com/conda/conda-build/actions/runs/3395410223/jobs/5645381714#step:11:1193

PR #4622, opened 2022-11-04
Flaky path convert test

dholth:
### Description

Check intermediate lib rename technique

### Checklist - did you ...

### DIFF
--- a/conda_build/convert.py
+++ b/conda_build/convert.py
@@ -273,18 +273,21 @@ def update_lib_contents(lib_directory, temp_dir, target_platform, file_path):
     elif target_platform == 'unix':
         temp_dir = Path(temp_dir)
         src_dir = temp_dir / "Lib"
-        dest_dir = temp_dir / "lib"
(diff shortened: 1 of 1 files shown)


conda_build/convert.py
```diff
@@ -268,15 +268,19 @@ def update_lib_contents(lib_directory, temp_dir, target_platform, file_path):
         except IndexError:
             pass
 
+        # why not rename here?
         shutil.move(os.path.join(temp_dir, 'lib'), os.path.join(temp_dir, 'Lib'))
 
     elif target_platform == 'unix':
         temp_dir = Path(temp_dir)
         src_dir = temp_dir / "Lib"
+        intermediate_dir = temp_dir / "notLib"
```
dholth, line 277: if this works we should use the technique when we are on a case-insensitive filesystem, which is usually true on Windows an Mac, and rarely true on Linux

dholth:
Didn't help; we think the directory is busy on Windows. Can we skip the test on Windows?

PR #4621, opened 2022-11-04
Add Allure reporting, update Python version test matrix, and prune test_api_build.py

sven6002:
### Description

Miscellaneous improvements:
- Adds Allure reporting
- Updates the Python version test matrix
- Prunes tests from `test_api_build.py`

Resolves https://github.com/conda/conda-build/issues/4558

### Checklist - did you ...

- [ ] ~Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release notes?~
     
- [x] Update / Remove necessary tests?
- [x] Add / update outdated docustring?

PR #4648, opened 2022-11-23
Cleanup replay variables & use setup-miniconda

kenodegard:
### Description

While reviewing #4621 I noticed that the replay definitions were unnecessarily inconsistent so this attempts to correct that. I also ended up updating the setups to use https://github.com/conda-incubator/setup-miniconda.

A few changes of note:
- removed `mkdir` statements since these directories are created if missing
- removed custom pytest-replay arg definitions for Windows and reused implementation used in both Linux and macOS
- standardized replay names
- split pytest commands across multiple lines for easier reading
- use setup-miniconda for Linux, Windows, and macOS

Depends on https://github.com/conda/conda/pull/12184

### Checklist - did you ...


tests/test-recipes/metadata/_c_vendoring_detection/conda_build_config.yaml
```diff
@@ -33,8 +33,6 @@ rust_compiler:
   - rust
 rust_compiler_version:
   - 1.29.0
-CONDA_BUILD_SYSROOT:
-  - /opt/MacOSX10.10.sdk        # [osx]
```
kenodegard, line 37: `CONDA_BUILD_SYSROOT` as an environment variable is easier to use as we can extract the value programmatically (see "Setup variables" in the macos job of `tests.yml`)

ci/github/activate_conda
```diff
@@ -1,5 +0,0 @@
-#!/usr/bin/env bash
-
-unset CONDA_SHLVL
-eval "$(${1} -m conda shell.bash hook)"
-conda activate base
```
kenodegard, line 5: Moot with setup-conda

ci/github/install_conda_build_test_deps
```diff
@@ -1,51 +0,0 @@
...
-  else
-    if [[ $(uname) =~ .*inux.* ]]; then
-      conda install -y --show-channel-urls shellcheck
-    else
-      conda install -y --show-channel-urls conda-forge::shellcheck
-    fi
-  fi
-  if [[ ! -d ../conda_build_test_recipe ]]; then
-    pushd ..
-      git clone "https://github.com/conda/conda_build_test_recipe"
```
kenodegard, line 46: This has been extracted into a separate step in the linux/macos jobs and reuses the actions/checkout GH action

ci/github/install_conda_build_test_deps
```diff
@@ -1,51 +0,0 @@
...
-  # with their channel, but, well, yuck.
-  if [[ ${MACOS_ARM64} == yes ]]; then
-    echo "Not installing shellcheck as it is unavailable on macOS arm64 at present"
-  else
-    if [[ $(uname) =~ .*inux.* ]]; then
-      conda install -y --show-channel-urls shellcheck
-    else
-      conda install -y --show-channel-urls conda-forge::shellcheck
-    fi
-  fi
```
kenodegard, line 43: All of this logic has been moved into the `requirements*.txt` files

ci/github/run_conda_forge_build_setup_osx
```diff
@@ -1,55 +0,0 @@
...
-conda config --set add_pip_as_python_dependency false
-
-# CONDA_PREFIX might be unset
-export CONDA_PREFIX="${CONDA_PREFIX:-$(conda info --json | jq -r .root_prefix)}"
-
-mkdir -p "${CONDA_PREFIX}/etc/conda/activate.d"
-echo "export CONDA_BUILD_SYSROOT='${CONDA_BUILD_SYSROOT}'"                 > "${CONDA_PREFIX}/etc/conda/activate.d/conda-forge-ci-setup-activate.sh"
-echo "export CPU_COUNT='${CPU_COUNT}'"                                    >> "${CONDA_PREFIX}/etc/conda/activate.d/conda-forge-ci-setup-activate.sh"
-echo "export PYTHONUNBUFFERED='${PYTHONUNBUFFERED}'"                      >> "${CONDA_PREFIX}/etc/conda/activate.d/conda-forge-ci-setup-activate.sh"
-echo "export MACOSX_DEPLOYMENT_TARGET='${MACOSX_DEPLOYMENT_TARGET}'"      >> "${CONDA_PREFIX}/etc/conda/activate.d/conda-forge-ci-setup-activate.sh"
```
kenodegard, line 51: unnecessary since we export `CONDA_BUILD_SYSROOT`

ci/github/run_conda_forge_build_setup_osx
```diff
@@ -1,55 +0,0 @@
...
-if [[ ! -d ${CONDA_BUILD_SYSROOT} || "$OSX_FORCE_SDK_DOWNLOAD" == "1" ]]; then
-    echo "Downloading ${MACOSX_DEPLOYMENT_TARGET} sdk"
-    curl -L -O https://github.com/phracker/MacOSX-SDKs/releases/download/11.3/MacOSX${MACOSX_DEPLOYMENT_TARGET}.sdk.tar.xz
-    tar -xf MacOSX${MACOSX_DEPLOYMENT_TARGET}.sdk.tar.xz -C "$(dirname "$CONDA_BUILD_SYSROOT")"
-    # set minimum sdk version to our target
-    if [[ ${INSTALL_XCODE} == 1 ]]; then
-      plutil -replace MinimumSDKVersion -string ${MACOSX_DEPLOYMENT_TARGET} $(xcode-select -p)/Platforms/MacOSX.platform/Info.plist
-      plutil -replace DTSDKName -string macosx${MACOSX_DEPLOYMENT_TARGET}internal $(xcode-select -p)/Platforms/MacOSX.platform/Info.plist
-    fi
-fi
```
kenodegard, line 30: No need to fetch our own SDK, it comes with the GH runner

.github/workflows/tests.yml
```diff
@@ -277,76 +286,74 @@ jobs:
...
         run: |
-          set -e -u
-          source ci/github/activate_conda "${{ github.workspace }}/miniconda/bin/python"
-          conda remove conda-verify -y
-          echo "safety_checks: disabled" >> ~/.condarc
-          echo "local_repodata_ttl: 1800" >> ~/.condarc
-          mkdir -p ~/.conda
-          conda create -n blarg1 -yq python=2.7
-          conda create -n blarg3 -yq python=3.7
-          conda create -n blarg4 -yq python nomkl numpy pandas svn
```
jezdez, line 341: What's up with these, any idea where they came from?

ci/github/install_conda_build_test_deps
jezdez: Good riddance!

jezdez (approved):
This is a big leap forward, thanks @kenodegard!

.github/workflows/tests.yml
```diff
@@ -27,6 +27,7 @@ jobs:
     defaults:
       run:
         shell: bash -l {0}
+        working-directory: ./src/
     strategy:
       fail-fast: false
       matrix:
```
mbargull, line 33: Do we want to switch to newer Python versions while we're at it?
(Seems appropriate since this is a CI overhaul PR.)
I'd go for `['3.7', '3.10']`, i.e.,
- use the lowest version still supported
- include the latest version for which all packages are available
  (you can also try `3.11`, but I'm not sure if everything is ready yet; I'd probably wait for this until spring or so)

.github/workflows/tests.yml
```diff
@@ -277,76 +286,74 @@ jobs:
...
+      - name: Setup environment
         run: |
-          set -x -e -u
-          curl -o ${{ github.workspace }}/miniconda.sh https://repo.anaconda.com/miniconda/Miniconda3-latest-MacOSX-x86_64.sh
-          chmod +x ${{ github.workspace }}/miniconda.sh
-          ${{ github.workspace }}/miniconda.sh -b -p ${{ github.workspace }}/miniconda
-          source ci/github/activate_conda "${{ github.workspace }}/miniconda/bin/python"
+          conda install -q -y -c defaults --file tests/requirements.txt
+          conda install -q -y -c defaults --file tests/requirements-macos.txt
+          conda install -q -y ${{ env.CONDA_CHANNEL_LABEL }}::conda
```
mbargull, line 319:
```suggestion
          conda install -q -y -c defaults \
            --file tests/requirements.txt \
            --file tests/requirements-macos.txt
          conda install -q -y -c defaults ${{ env.CONDA_CHANNEL_LABEL }}::conda
```
same as above

.github/workflows/tests.yml
```diff
@@ -277,76 +286,74 @@ jobs:
...
-
-      - name: Run Parallel Tests
-        if: matrix.test-type == 'parallel'
+      - name: Run tests
         run: |
-          set -e -u
-          source ci/github/activate_conda "${{ github.workspace }}/miniconda/bin/python"
-          conda remove conda-verify -y
-          echo "safety_checks: disabled" >> ~/.condarc
-          echo "local_repodata_ttl: 1800" >> ~/.condarc
```
mbargull, line 337: A toplevel
```sh
conda config --set auto_update_conda False
conda config --set local_repodata_ttl: 7200
```
could make sense to avoid unnecessary changes during the test run in case some test runs a `conda install` in the base env and to avoid re-downloading repodata at some point during the CI run (`7200` would be "don't update repodata if it got updated in the last two hours from the former update").

mbargull (changes requested):
Thanks for removing clutter :).

I didn't look at all of the changes (actually, didn't look at the main replay and setup-miniconda bits :smile:), but a few things caught my eye.

ci/github/install_conda_build_test_deps
mbargull: :broom: :bucket: +1

.github/workflows/tests.yml
mbargull: Those were added to "pre-warm" the package cache, i.e., avoid downloading those packages during the (parallel) tests, I'd guess to mostly avoid concurrency issues. At some point those had `--download-only` I think -- either I'm misremembering or those got lost somewhere one the way (with `--download-only` those env wouldn't have been created which explains why Mike/Ray(?) chose such "meaningful" names :laughing: ...)
We could leave them out and see if anything causes trouble, I'd say. If it leads to any flakiness, we could add something like this (or rather something more precise) back again.

kenodegard: yeah, I was suspecting it was for caching, if we need it we should create a pytest fixture to do the caching

.github/workflows/tests.yml
```diff
@@ -41,69 +42,65 @@ jobs:
...
-          conda install -y -c conda-canary/label/dev conda
-
-      - name: Install default Conda if release version
-        if: matrix.conda-version == 'release'
+      - name: Setup environment
         run: |
-          conda install -y conda
+          conda install -q -y -c defaults --file tests/requirements.txt
+          conda install -q -y -c defaults --file tests/requirements-linux.txt
+          conda install -q -y ${{ env.CONDA_CHANNEL_LABEL }}::conda
```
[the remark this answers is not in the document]
sven6002: are teh requirements based on operating system.. wouldn't it be an idea to use an `.env` file that way you can just segragated the installation based on Operating System.. Intead of having multiple `.req.txt` files

.github/workflows/tests.yml
```diff
@@ -277,76 +286,74 @@ jobs:
...
+        # macOS needs to define $CONDA_BUILD_SYSROOT and $MACOSX_DEPLOYMENT_TARGET to know where to find the SDKs
+        # the Anaconda compilers only support up to macOS 10.15
         run: |
-          set -x -e -u
-          source ci/github/activate_conda "${{ github.workspace }}/miniconda/bin/python"
-          source ci/github/run_conda_forge_build_setup_osx
+          sudo xcode-select --switch /Applications/Xcode_11.7.app
+
+          echo "CONDA_BUILD_SYSROOT=$(xcrun --sdk macosx --show-sdk-path)" >> $GITHUB_ENV
+          echo "MACOSX_DEPLOYMENT_TARGET=$(xcrun --sdk macosx --show-sdk-version)" >> $GITHUB_ENV
```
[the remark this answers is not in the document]
kenodegard: Unfortunately, the GH runner doesn't offer any older SDKs, is this important enough to download additional SDKs?

.github/workflows/tests.yml
sven6002: @mbargull In regards with the newer versions there is an effort to overhaul the `conda-build` itself to support the newer version.. it just may be out of the scope for this PR

.github/workflows/tests.yml
sven6002: same suggestion.. we could combine all the necessary reqs for dependency installation

.github/workflows/tests.yml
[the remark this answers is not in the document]
kenodegard: @mbargull I didn't know we could specify multiple files like that! It also appears that we can also include individual packages at the same time:

```suggestion
          conda install -q -y -c defaults \
              --file tests/requirements.txt \
              --file tests/requirements-linux.txt \
              ${{ env.CONDA_CHANNEL_LABEL }}::conda
```

@sven6002 conda environment files do not support selectors, so we'd still have per OS environments

.github/workflows/tests.yml
kenodegard: I'll move these into `ci/github/.condarc` that we use with our setup-miniconda

.github/workflows/tests.yml
kenodegard: yeah, I'm in favor of doing this in a separate PR

.github/workflows/tests.yml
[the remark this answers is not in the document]
kenodegard: Or does the `MACOSX_DEVELOPMENT_TARGET` not need to match the SDK?

tests/test-recipes/metadata/_c_vendoring_detection/conda_build_config.yaml
kenodegard: How do we feel about dynamically generating `cbc.yaml`?

beeankha (dismissed):
I know @mbargull left some comments/change requests that might need to be addressed, but otherwise this LGTM!

tests/test-recipes/metadata/_c_vendoring_detection/conda_build_config.yaml
mbargull: Yes, that'd be exactly what I had in mind :).

.github/workflows/tests.yml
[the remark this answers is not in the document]
mbargull: Yes, they don't have to match!

.github/workflows/tests.yml
mbargull: Fine by me. It should be a small PR that we can slip on in any time!

.github/workflows/tests.yml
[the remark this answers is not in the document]
mbargull: > It also appears that we can also include individual packages at the same time:

Yes, and both ways are fine but have their respective small differences, see the part from my previous comment:
> We could put the `${{ env.CONDA_CHANNEL_LABEL }}::conda` also in the first `install` for the same reasons. A counter argument would be to try to have `${{ env.CONDA_CHANNEL_LABEL }}` out of the channel list when installing the other deps, though. (I added the `-c defaults` here to for config/solver consistency's sake.)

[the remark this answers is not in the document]
kenodegard: 🤦🏼‍♂️ lol, I got too excited to finish reading

I see what you mean, well since the conda-canary channel only contains `conda-*` packages and none of these other dependencies it shouldn't make a difference between a one-liner and splitting it?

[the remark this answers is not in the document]
mbargull: > 🤦🏼‍♂️ lol, I got too excited to finish reading

Hehe :)

> I see what you mean, well since the conda-canary channel only contains `conda-*` packages and none of these other dependencies it shouldn't make a difference between a one-liner and splitting it?

Yes! (I just didn't check what's in -canary these days ;).)

tests/test-recipes/metadata/_c_vendoring_detection/conda_build_config.yaml
kenodegard: Since all of the affected recipes are built via `conda_build.api.build` I opted to create a pytest fixture that returned a `variants` dictionary. Is this sufficiently dynamic or do we want an actual `cbc.yaml` file generated? If this is sufficient we should be able to also remove the `CONDA_BUILD_SYSROOT` and `MACOSX_DEPLOYMENT_TARGET` from the macOS setup in `tests.yaml`.

tests/conftest.py
```diff
@@ -205,3 +207,29 @@ def single_version():
...
+        "MACOSX_DEPLOYMENT_TARGET": [
+            subprocess.run(
+                ["xcrun", "--sdk", "macosx", "--show-sdk-version"],
+                check=True,
+                capture_output=True,
+                text=True,
+            ).stdout.strip(),
+            "10.9",
+        ],
+    }
```
mbargull, line 235: For the builds at conda-forge we usually have them as environment variables and then override them via cbc.yaml is required.
To be closer to that, we could also have the fixture like this:
```suggestion
def variants_conda_build_sysroot(monkeypatch):
    if not on_mac:
        return {}

    monkeypatch.setenv(
        "MACOSX_DEPLOYMENT_TARGET",
        subprocess.run(
            ["xcrun", "--sdk", "macosx", "--show-sdk-version"],
            check=True,
            capture_output=True,
            text=True,
        ).stdout.strip(),
    )
    monkeypatch.setenv(
        "CONDA_BUILD_SYSROOT",
        subprocess.run(
            ["xcrun", "--sdk", "macosx", "--show-sdk-path"],
            check=True,
            capture_output=True,
            text=True,
        ).stdout.strip(),
    )
    return {"MACOSX_DEPLOYMENT_TARGET": ["10.9"]}
```

tests/test-recipes/metadata/_c_vendoring_detection/conda_build_config.yaml
mbargull: I think it should be fine to use `variants` instead of the YAML file (close enough ;).)
I'd go for setting things via env vars and then override the deployment target as variant, see https://github.com/conda/conda-build/pull/4648/files#r1038188689 .

mbargull (approved):
LGTM, thanks!

tests/conftest.py
kenodegard: @mbargull the intention was for `{"MACOSX_DEPLOYMENT_TARGET": ["10.9"]}` to always override the `MACOSX_DEPLOYMENT_TARGET` environment variable (assuming its passed to `api.build`), right?

mbargull: Ah, whoops, my bad. No, I meant to still have it run two versions one without the override and one with. Hold on...

tests/conftest.py
```diff
@@ -205,3 +207,29 @@ def single_version():
...
+    monkeypatch.setenv(
+        "MACOSX_DEPLOYMENT_TARGET",
+        subprocess.run(
+            ["xcrun", "--sdk", "macosx", "--show-sdk-version"],
+            check=True,
+            capture_output=True,
+            text=True,
+        ).stdout.strip(),
+    )
+    return {"MACOSX_DEPLOYMENT_TARGET": ["10.9"]}
```
mbargull, line 235: Maybe this works for retaining the two cases?
(Untested; I never used `fixtures`+`params` before...)
```suggestion
@pytest.fixture(scope="function", params=[{}, {"MACOSX_DEPLOYMENT_TARGET": ["10.9"]})
def variants_conda_build_sysroot(monkeypatch, request):
    if not on_mac:
        return {}

    monkeypatch.setenv(
        "CONDA_BUILD_SYSROOT",
        subprocess.run(
            ["xcrun", "--sdk", "macosx", "--show-sdk-path"],
            check=True,
            capture_output=True,
            text=True,
        ).stdout.strip(),
    )
    monkeypatch.setenv(
        "MACOSX_DEPLOYMENT_TARGET",
        subprocess.run(
            ["xcrun", "--sdk", "macosx", "--show-sdk-version"],
            check=True,
            capture_output=True,
            text=True,
        ).stdout.strip(),
    )
    return request.param
```

kenodegard: Ah! This makes more sense! This is an improvement over the previous test cases since we're properly testing not overriding envvars and then overriding.

Since this is specific to macOS I think we need an additional tweak:

```suggestion
@pytest.fixture(
    scope="function",
    params=[{}, {"MACOSX_DEPLOYMENT_TARGET": ["10.9"]}] if on_mac else [{}],
)
def variants_conda_build_sysroot(monkeypatch, request):
    if not on_mac:
        return {}

    monkeypatch.setenv(
        "CONDA_BUILD_SYSROOT",
        subprocess.run(
            ["xcrun", "--sdk", "macosx", "--show-sdk-path"],
            check=True,
            capture_output=True,
            text=True,
        ).stdout.strip(),
    )
    monkeypatch.setenv(
        "MACOSX_DEPLOYMENT_TARGET",
        subprocess.run(
            ["xcrun", "--sdk", "macosx", "--show-sdk-version"],
            check=True,
            capture_output=True,
            text=True,
        ).stdout.strip(),
    )
    return request.param
```

.github/workflows/tests.yml
```diff
@@ -137,22 +135,15 @@ jobs:
...
         with:
           fetch-depth: 0
 
-      - name: Set temp dirs correctly
-        shell: cmd
-        # https://github.com/actions/virtual-environments/issues/712
-        run: |
-          echo "TMPDIR=$env:USERPROFILE\AppData\Local\Temp" >> $env:GITHUB_ENV
-          echo "TEMP=$env:USERPROFILE\AppData\Local\Temp" >> $env:GITHUB_ENV
-          echo "TMP=$env:USERPROFILE\AppData\Local\Temp" >> $env:GITHUB_ENV
```
dholth, line 154: is this no longer needed?

.github/workflows/tests.yml
```diff
@@ -251,7 +239,17 @@ jobs:
           set PERL=
           set LUA=
           set R=
-          pytest --color=yes -vv -n 0 --basetemp %UserProfile%\cbtmp_serial --cov conda_build --cov-report xml -m "serial" ${{ env.pytest-replay }}
+          pytest ^
```
dholth, line 242: 24-caret

What is the ^ character doing at the end of these lines?

dholth (approved):

.github/workflows/tests.yml
kenodegard: that's the line continuation character for CMD

.github/workflows/tests.yml
kenodegard: Probably not, I tried to ignore the Windows job as I intended to review that in a followup PR

dholth:
The force-push's make it more difficult to collaborate on PRs

kenodegard:
@dholth true, in this case I opted to do a rebase & force push because of the number of conflicts with main

dholth:
> @dholth true, in this case I opted to do a rebase & force push because of the number of conflicts with main

Now that we are doing a lot of squash merges, git checkout branch ; git merge origin/main works pretty well

dholth:
tar should be available. You could use "python -m tarfile". Sometimes people like to use gnu tar or a tar that comes with git, on Windows CI.

sven6002 (approved):
looks good

.github/workflows/tests.yml
```diff
@@ -73,69 +74,77 @@ jobs:
...
+            --replay-record-dir="${{ env.REPLAY_DIR }}" \
+            --replay-base-name="${{ env.REPLAY_NAME }}" \
+            --alluredir="${{ env.ALLURE_DIR }}" \
+            -m "${{ env.PYTEST_MARKER }}" \
+            ./tests
 
       - name: Tar Allure Results
         if: always()
-        run: tar -zcf allure-results.tar.gz allure-results
+        run: tar -zcf "${{ env.ALLURE_DIR }}.tar.gz" "${{ env.ALLURE_DIR }}"
```
dholth, line 147: Aren't we in `${{ github.workspace }}` already `ALLURE_DIR: ${{ github.workspace }}/allure-results`

Will have to double check the artifacts to see if the tar paths are concise.

kenodegard: AFAIK no, we are in `${{ github.workspace }}/src` since we checkout two different repos and specify the default working-directory

dholth: Can we vendor the test recipe instead of pulling it from an archived repository? Seems like a recipe for tears.

kenodegard: Possibly but this PR is already pretty large and longrunning, can we do that in a followup PR?

kenodegard: probably not, it's referred to in several tests as a cloned git repo

kenodegard: we can perhaps change it to be a submodule?

dholth (approved):
Thanks

beeankha (approved):

PR #4650, opened 2022-11-24
fix: conda-build CLI overrode condarc's zstd_compression_level with the default value

mbargull:
### Checklist - did you ...

- [x] Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [x] Add / update necessary tests?
- [ ] Add / update outdated documentation?

fixes gh-4649:
> The `conda-build` command defines a default value for `--zstd-compression-level` (`conda_build.config.zstd_compression_level_default=22`) without querying the Conda config for the `conda_build.zstd_compression_level` key. Since parameters given by the CLI take precedence over those from `.condarc`, a `zstd_compression_level` set in the latter is always overridden by the default value when `conda-build ...` is invoked.

PR #4652, opened 2022-11-25
Reduce flakiness of CI test runs

mbargull:
### Description

I started to include unrelated test fixes in gh-4650 to easier inspect failing/passing of relevant tests.
Those started to grow and it's cleaner to have them separated, anyway, so here we go.

### Checklist - did you ...


mbargull:
I didn't add a news entry because this is just chore that doesn't touch anything outside the test code.
If you'd still like to have a news entry, let me know.

mbargull:
Re: `conf=testing_config`:
Previously, tests failed because (even when only `render`ing) concurrent writes/deletions around the `{croot}/work` happened at `conda_build.config.Config.compute_build_id` (at the end in its `if old_dir != work_dir:` branch) happened. I didn't dig into it, but it might just be that another test creates the `work` (without a timestamp) folder and then `compute_build_id` wants to move it.
In any case, `testing_config` sets the `croot` to `testing_workdir` and as such the single tests use their own root dir without seeing others' `work` folders.

tests/requirements.txt
```diff
@@ -13,6 +13,6 @@ psutil
 pytest
 pytest-cov
 pytest-mock
-pytest-rerunfailures
+flaky
```
jezdez, line 16: We recently switched to pytest-rerunfailures because it's better maintained and used in the conda code base. You should be able to achieve the same as the `rerun_filter` parameter with the `condition` parameter: https://github.com/pytest-dev/pytest-rerunfailures#re-run-individual-failures

jezdez (changes requested):

tests/requirements.txt
mbargull: > We recently switched to pytest-rerunfailures because it's better maintained

Yeah, saw that too, but

> You should be able to achieve the same as the `rerun_filter` parameter with the `condition` parameter: https://github.com/pytest-dev/pytest-rerunfailures#re-run-individual-failures

this appears to be just for the "outside" conditions but doesn't get the test's report information: https://github.com/pytest-dev/pytest-rerunfailures/blob/10.3/pytest_rerunfailures.py#L323-L324
I.e., `pytest-rerunfailures` seems to lack comparable/flexible functionality w.r.t. `flaky`.
AIUI, if we want to use `pytest-rerunfailures`, we have to either be non-selective at all or add `--only-rerun=ConnectionError` to `addopts` in `setup.cfg` and thus be too selective. Since the latter could become a footgun, the former is more sensible.

Do you want me to switch to `pytest-rerunfailures` and thus remove the selectivity on `ConnectionError`?

mbargull:
Closing in favor of gh-4653

PR #4653, opened 2022-11-25
Reduce flakiness of CI test runs

mbargull:
### Description

Ignore this for now. Testing an alternative to gh-4652 ...
### Checklist - did you ...


mbargull:
Comments copied from gh-4652:
> I didn't add a news entry because this is just chore that doesn't touch anything outside the test code. If you'd still like to have a news entry, let me know.

> Re: `config=testing_config`: Previously, tests failed because (even when only `render`ing) concurrent writes/deletions around the `{croot}/work` happened at `conda_build.config.Config.compute_build_id` (at the end in its `if old_dir != work_dir:` branch) happened. I didn't dig into it, but it might just be that another test creates the `work` (without a timestamp) folder and then `compute_build_id` wants to move it. In any case, `testing_config` sets the `croot` to `testing_workdir` and as such the single tests use their own root dir without seeing others' `work` folders.

mbargull:
(Added description/motivation to the OP.)

tests/conftest.py
```diff
@@ -120,6 +123,54 @@ def boolify(v):
...
+            saved_kwdefaults[member] = kwdefaults
+            member.__kwdefaults__ = {**kwdefaults, arg_name: new_default}
+
+    def restore_defaults():
+        for func, defaults in saved_defaults.items():
+            func.__defaults__ = defaults
+        for func, kwdefaults in saved_kwdefaults.items():
+            func.__kwdefaults__ = kwdefaults
+
+    request.addfinalizer(restore_defaults)
```
jezdez, line 171: This is way too magic and would be more obviously implemented by using pytest's `monkeypatch` fixture to overwrite `conda_buid.config.get_or_merge_config` to apply the test config.

jezdez (changes requested):
Thank you for working on this @mbargull!

I don't think your approach with rewriting the function defaults for `conda_build.api` is a good pattern and could be achieved by a more mundane monkey patching of the `conda_build.config.get_or_merge_config` function, to be automatically returning the appropriate test config.

While I enjoy seeing the use of `pytestmark`, I agree that it'll likely be forgotten by future contributors when adding new tests, creating rather unpleasant implicit differences in behavior again, increasing the risk of concurrency issues.

I would suggest to indeed use `autouse=True` to automatically enable the test config, if you think the majority of test modules need it (as it seems) and instead try to find a way to disable it when needed.

mbargull:
> I don't think your approach with rewriting the function defaults for `conda_build.api` is a good pattern and could be achieved by a more mundane monkey patching of the `conda_build.config.get_or_merge_config` function, to be automatically returning the appropriate test config.

Haha, and that's what I get to only look at level `n` but not level `n+1` :) -- I didn't even see/look for that all those functions use `get_or_merge_config`.
So, yes, very much agreed, I'll change it to monkeypatch `get_or_merge_config`; much cleaner.

> I would suggest to indeed use `autouse=True` to automatically enable the test config, if you think the majority of test modules need it (as it seems) and instead try to find a way to disable it when needed.

Okay, I guess `autouse=True` is fine since we can opt out via the added `@pytest.mark.no_<...>`.
Mind you, this would mean that we create a temporary directory for every single test then (through `testing_config` -> `testing_workdir` -> `tmpdir` fixture requests).
We then probably want to set `pytestmark = pytest.mark.usefixtures("no_default_testing_config")` for some test modules like `test_api_consistency.py`.

jezdez:
> > I don't think your approach with rewriting the function defaults for `conda_build.api` is a good pattern and could be achieved by a more mundane monkey patching of the `conda_build.config.get_or_merge_config` function, to be automatically returning the appropriate test config.
> 
> Haha, and that's what I get to only look at level `n` but not level `n+1` :) -- I didn't even see/look for that all those functions use `get_or_merge_config`. So, yes, very much agreed, I'll change it to monkeypatch `get_or_merge_config`; much cleaner.

Oops, conda-build keeps on giving 🤣 

> > I would suggest to indeed use `autouse=True` to automatically enable the test config, if you think the majority of test modules need it (as it seems) and instead try to find a way to disable it when needed.
> 
> Okay, I guess `autouse=True` is fine since we can opt out via the added `@pytest.mark.no_<...>`. Mind you, this would mean that we create a temporary directory for every single test then (through `testing_config` -> `testing_workdir` -> `tmpdir` fixture requests). We then probably want to set `pytestmark = pytest.mark.no_default_testing_config` for some test modules like `test_api_consistency.py`.

Hmm, yeah, let's see what the impact on runtime is for that

jezdez:
Seems like it's only one test failing now! `test_skeleton_pypi_arguments_work`

mbargull:
> Seems like it's only one test failing now! `test_skeleton_pypi_arguments_work`

Ha! That was a sneaky PyPI update since Saturday :grin: .
I fixed the test to use the same package version as before. The the commit message from https://github.com/conda/conda-build/pull/4653/commits/c6463d9842fd034bf5b70c3bcd71c4fd586d8dfd for details.

mbargull:
This is now a much smaller and cleaner change set with which I'm content :).
The test run time is not noticeably affected since we have rather heavy test cases that dominate all other running times.

@jezdez, thanks for your thorough and thoughtful review, I appreciate it!
Let me know if there's anything else you'd want me to change.

mbargull:
Remaining failure is the permission error on Windows during `pip install` for `peppercorn` again.

kenodegard (approved):
Reran failed test, seems to be passing now

mbargull:
Looks like
```python
@pytest.fixture(scope="function", autouse=True)
def default_testing_config(testing_config, monkeypatch, request):
```
doesn't work (reliably/at all?): https://github.com/conda/conda-build/pull/4665#issuecomment-1337525913

PR #4657, opened 2022-11-29
Pin msumastro to 1.1.6

kenodegard:
### Description

Attempt to resolve newly failing test.

`msumastro` was recently updated (11/26-27/2022, v1.2.0-2), one of the changes were to switch from `setup.py` to `pyproject.toml`.

### Checklist - did you ...

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -291,7 +291,7 @@ def test_skeleton_pypi_arguments_work(testing_workdir):
 
     https://github.com/conda/conda-build/pull/1384
     """
-    args = ['pypi', 'msumastro', '--pin-numpy']
+    args = ["pypi", "msumastro", "--version=1.1.6", "--pin-numpy"]
     main_skeleton.execute(args)
     assert os.path.isdir('msumastro')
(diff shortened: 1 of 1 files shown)


mbargull:
WDYT about
> msumastro has been updated a couple of days ago and versions >=1.2 now
store the metadata in setup.cfg which is not handled by skeletons.pypi.
It is easy to fix this if we reuse the code from _load_setup_py_data but
we might as well use/advertize newer projects like grayskull instead.

I'm generally in favor of separating concerns here, e.g., moving skeleton code into a/multiple separate projects (`grayskull` for PyPI for example).
As such, I'd be fine with not updating `skeletons.pypi` anymore.
(Plus, there is also the problem that `setuptools.read_configuration` is deprecated and they didn't yet expose `setuptools.config.pyprojecttoml.read_configuration` as a public export.)

kenodegard:
@mbargull nothing has been decided wrt `conda skeleton` but none of us have had much appetite or time to update it either so we've been recommending `grayskull` when asked

PR #4660, opened 2022-11-30
[DOC] Add note about non-git local source location needing to be within the feedstock directory tree

bskinn:
### Description

Based on [recent personal experience](https://github.com/conda-forge/sphobjinv-feedstock/pull/1#issuecomment-1332372132), when building with an unpacked sdist that's *not* a git repository, in order for the source to be visible inside the docker container it appears to have to reside inside the feedstock directory tree.

That is, the build only worked when I unpacked the sdist in `{repo}/recipe/{sdist unpack}`. If I used an unpack location outside `{repo}/recipe`, the build would fail at the `cp -a ...` step.

NB: I didn't try unpacking it as `{repo}/{sdist unpack}`.

NB also, I didn't try to build using a git repo housed outside the repository.

---

I'm aiming this PR to clarify this requirement for the location of the source on the local filesystem.

If this is *not actually supposed to be a requirement*, then this PR should probably be a bug ticket, instead. Happy to put it wherever's best, and happy to do further testing if needed.

### Checklist - did you ...

- [x] Add / update necessary tests? (N/A)

### DIFF
--- a/docs/source/resources/define-metadata.rst
+++ b/docs/source/resources/define-metadata.rst
@@ -171,7 +171,9 @@ Source from a local path
 
 If the path is relative, it is taken relative to the recipe
 directory. The source is copied to the work directory before
-building.
+building. The source must be present in the `recipe/`
+directory (or a sub-directory) in order to be visible
+from inside the Docker container used for the build process.
 
 .. code-block:: yaml
 


bskinn:
Weird, I completely missed the CLA comment, even though it clearly came into my email.

beeankha:
@conda-bot check

PR #4615, opened 2022-11-03
Fix `lru_cache` issues

kenodegard:
### Description

This fixes the failing tests introduced by #4593.

1. Replaces `lru_cache(None)` with `lru_cache(maxsize=None)`
2. Deprecates `conda_build.environ.system_vars` in favor of `conda_build.environ.os_vars`
3. Create shim for `conda_build.os_utils.ldd.get_linkages` to circumvent hashing issue

### Checklist - did you ...

- [x] Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [ ] ~Add / update necessary tests?~
- [ ] ~Add / update outdated documentation?~

PR #4665, opened 2022-12-03
Fix: Build environments not activated

mbargull:
### Description

In https://github.com/conda-forge/conda-feedstock/pull/181#issuecomment-1334110585 we encountered unexpected failures for PyPy variants of the builds with
```cmd
D:/bld/conda_1669913819035/_h_env/python.exe: error while loading shared libraries: libpypy3.9-c.dll: cannot open shared object file: No such file or directory
```
It turned out, this happened because the build environments were not activated at all.
The correct Python interpreter was launched because the build script used `$PYTHON`.
The CPython variants built fine "by chance" only since our CPython builds carry library lookup patches to also function without an activated environment and other programs needed for the build were already on `PATH` from outside the build environment.

This happened due to a regression caused by https://github.com/conda/conda-build/pull/4615/commits/cecb031f8e4a37adaffbd9a84bf3195fd51e6d7a .
Previously, `conda`'s `memoized` bailed out from caching anything because of the additionally passed non-hashable `dict`.
Memoizing with a `MetaData` instance as a key is problematic since that object/its members are highly mutable during `conda-build`'s work.

The actual fix in commit 7f047c7 just removes the memoization.
I also added two guards to avoid regressions:
1. Make `MetaData` non-hashable in 9fd7165.
2. Add a test in 9fde4c1 + 060f10f to check if the prefixes are added to `PATH` as expected.

Note that the added test did not replicate the behavior from the conda-forge build entirely.
During the test runs, the prefixes were only added in a wrong order (possibly due to merged `build`/`host` at the time of memoization) but they were not completely omitted like in the `conda-feedstock` build.
These subtleties can be expected due to different setups and different times of memoization.
It is also likely that the early memoization changed other parts of the builds; changes to `PATH` were just most visible.

----
This PR also adds PREFIX/bin to PATH on Windows and removes PREFIX root from PATH on Unix.
Those were two bugs I noticed broke the added test case.
Fixing that should be non-controversial without any downsides, really (those are plain bugs anyway).

### Checklist - did you ...

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -2584,6 +2584,9 @@ def construct_metadata_for_test(recipedir_or_package, config):
 
 
 def write_build_scripts(m, script, build_file):
+    # TODO: Prepending the prefixes here should probably be guarded by
+    #         if not m.activate_build_script:
+    #       Leaving it as is, for now, since we need a quick, non-disruptive patch release.
     with utils.path_prepended(m.config.host_prefix):
         with utils.path_prepended(m.config.build_prefix):
             env = environ.get_dict(m=m)
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -2587,8 +2587,8 @@ def write_build_scripts(m, script, build_file):
     # TODO: Prepending the prefixes here should probably be guarded by
     #         if not m.activate_build_script:
     #       Leaving it as is, for now, since we need a quick, non-disruptive patch release.
-    with utils.path_prepended(m.config.host_prefix):
-        with utils.path_prepended(m.config.build_prefix):
+    with utils.path_prepended(m.config.host_prefix, False):
+        with utils.path_prepended(m.config.build_prefix, False):
             env = environ.get_dict(m=m)
     env["CONDA_BUILD_STATE"] = "BUILD"
 
--- a/conda_build/utils.py
+++ b/conda_build/utils.py
@@ -950,11 +950,10 @@ def get_build_folders(croot):
 
 
 def prepend_bin_path(env, prefix, prepend_prefix=False):
-    # bin_dirname takes care of bin on *nix, Scripts on win
-    env['PATH'] = join(prefix, bin_dirname) + os.pathsep + env['PATH']
+    env['PATH'] = join(prefix, "bin") + os.pathsep + env['PATH']
     if sys.platform == "win32":
         env['PATH'] = join(prefix, "Library", "mingw-w64", "bin") + os.pathsep + \
-                      join(prefix, "Library", "usr", "bin") + os.pathsep + os.pathsep + \
+                      join(prefix, "Library", "usr", "bin") + os.pathsep + \
                       join(prefix, "Library", "bin") + os.pathsep + \
                       join(prefix, "Scripts") + os.pathsep + \
                       env['PATH']
@@ -985,9 +984,10 @@ def sys_path_prepended(prefix):
 
 
 @contextlib.contextmanager
-def path_prepended(prefix):
+def path_prepended(prefix, prepend_prefix=True):
+    # FIXME: Unclear why prepend_prefix=True for all platforms.
     old_path = os.environ['PATH']
-    os.environ['PATH'] = prepend_bin_path(os.environ.copy(), prefix, True)['PATH']
+    os.environ['PATH'] = prepend_bin_path(os.environ.copy(), prefix, prepend_prefix)['PATH']
     try:
         yield
     finally:
--- a/tests/test_api_build.py
+++ b/tests/test_api_build.py
@@ -1657,12 +1657,13 @@ def test_script_env_warnings(testing_config, recwarn):
 @pytest.mark.slow
 def test_activated_prefixes_in_actual_path(testing_config, testing_metadata):
     file = "env-path-dump"
+    testing_metadata.config.activate = True
     meta = testing_metadata.meta
     meta["requirements"]["host"] = []
     meta["build"]["script"] = [
         f"echo %PATH%>%PREFIX%/{file}" if on_win else f"echo $PATH>$PREFIX/{file}"
     ]
-    outputs = api.build(testing_metadata, activate=True)
+    outputs = api.build(testing_metadata)
     env = {"PATH": ""}
     prepend_bin_path(env, testing_metadata.config.host_prefix)
     prepend_bin_path(env, testing_metadata.config.build_prefix)
@@ -1672,4 +1673,7 @@ def test_activated_prefixes_in_actual_path(testing_config, testing_metadata):
         for path in package_has_file(outputs[0], file).strip().split(os.pathsep)
         if path in expected_paths
     ]
-    assert actual_paths == expected_paths
+    # We get the PATH entries twice:
+    #   1. from the environment activation hooks,
+    #   2. also beforehand from utils.path_prepended at the top of build.write_build_scripts
+    assert actual_paths == expected_paths * 2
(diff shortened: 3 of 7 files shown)


mbargull:
I updated the OP with explanations for what happened and how this is mitigated.

cc @jezdez, @kenodegard 
Another quick patch release would be highly appreciated since this is actually a severe issue with many unknowns in what parts under which build configurations it impacts.

conda_build/build.py
```diff
@@ -2584,8 +2584,11 @@ def construct_metadata_for_test(recipedir_or_package, config):
 
 
 def write_build_scripts(m, script, build_file):
-    with utils.path_prepended(m.config.host_prefix):
-        with utils.path_prepended(m.config.build_prefix):
+    # TODO: Prepending the prefixes here should probably be guarded by
+    #         if not m.activate_build_script:
+    #       Leaving it as is, for now, since we need a quick, non-disruptive patch release.
+    with utils.path_prepended(m.config.host_prefix, False):
+        with utils.path_prepended(m.config.build_prefix, False):
```
mbargull, line 2591: Before removing this in a future release, we should add more test cases.
The test added in this PR only checks for the case when there is `build`+`host`, `activate=True` and no `outputs`.
Other configurations that should be tested:
- `activate=False`
- `build` and `host` merged
  - with `activate=True` 
  - with `activate=False` 
- recipes with `outputs`
  - and those for potential combinations of the aforementioned other cases.

I consider this out of scope for this PR since we should put out a patch release quickly and because those other cases touch other issues like https://github.com/conda/conda-build/issues/3993 -- that's another can of worms (one we should take care of, but not hastily in a small patch release).

conda_build/utils.py
```diff
@@ -950,11 +950,10 @@ def get_build_folders(croot):
 
 
 def prepend_bin_path(env, prefix, prepend_prefix=False):
-    # bin_dirname takes care of bin on *nix, Scripts on win
-    env['PATH'] = join(prefix, bin_dirname) + os.pathsep + env['PATH']
+    env['PATH'] = join(prefix, "bin") + os.pathsep + env['PATH']
```
mbargull, line 953: `conda activate` always adds `"bin"` to `PATH`. We have to have it here too. The addition of `"Scripts"` is taken care of a few lines below.

conda_build/utils.py
```diff
@@ -985,9 +984,10 @@ def sys_path_prepended(prefix):
 
 
 @contextlib.contextmanager
-def path_prepended(prefix):
+def path_prepended(prefix, prepend_prefix=True):
+    # FIXME: Unclear why prepend_prefix=True for all platforms.
     old_path = os.environ['PATH']
-    os.environ['PATH'] = prepend_bin_path(os.environ.copy(), prefix, True)['PATH']
+    os.environ['PATH'] = prepend_bin_path(os.environ.copy(), prefix, prepend_prefix)['PATH']
```
mbargull, line 990: No idea why this is `prepend_prefix=True` for all platforms.
For now, I left it that way and only changed it to `False` in `build.write_build_scripts` where I needed it.
I did not look at the other places where `path_prepended` is used. For me, this is also out of scope for this PR, but I'd like us to fix it, hence the added `FIXME`-note.

tests/test_api_build.py
```diff
@@ -1652,3 +1652,35 @@ def assert_keyword(keyword):
...
+    ]
+    outputs = api.build(testing_metadata)
+    env = {"PATH": ""}
+    # We get the PATH entries twice: (which we should fix at some point)
+    #   1. from the environment activation hooks,
+    #   2. also beforehand from utils.path_prepended at the top of
+    #      - build.write_build_scripts on Unix
+    #      - windows.build on Windows
+    #        And apparently here the previously added build env gets deactivated
+    #        from the activation hook, hence only host is on PATH twice.
```
mbargull, line 1674: Maybe `CONDA_SHLVL`  is not reset for Windows builds? IDK. I didn't check since debugging for Windows is rather time-consuming on my end.
This can be investigated later; not something we have to do in this PR.

mbargull:
This PR also adds PREFIX/bin to PATH on Windows and removes PREFIX root from PATH on Unix.
Those were two bugs I noticed broke the added test case.
Fixing that should be non-controversial without any downsides, really (those are plain bugs anyway).

travishathaway (changes requested):
Hi @mbargull 

Thanks for this pull request. So far everything is looking okay with this pull request and I will be happy to approve, but I would really appreciate it if you could add a descriptive doc string to your test explaining exactly why it exists and perhaps its shortcomings as you mentioned.

We are trying hard to make sure the tests in conda-build are more maintainable and these simple descriptions are very helpful for future maintainers.

Thank you! ✌️

mbargull:
@travishathaway, thanks for the review! Addressed in 1ebe72c .

> We are trying hard to make sure the tests in conda-build are more maintainable and these simple descriptions are very helpful for future maintainers.

Glad to hear that process started! :smile_cat:

dholth (approved):
LGTM

dbast (approved):

dbast:
@travishathaway I don't mind doc strings in test cases... more important for me: test cases follow a given/when/then or setup/run-subject-under-test/compare-outcome structure (which is given here), which often makes the code self explanatory beyond to what can described in english via doctsrings.

travishathaway (approved):

mbargull:
<details>
<summary>Failure in <code>Tests / linux (3.9, canary, parallel)</code> is unrelated...</summary>

... but it's still odd, since after
https://github.com/conda/conda-build/commit/1053650e8bc311546163dc8c8219d510df0ea9ea#diff-e52e4ddd58b7ef887ab03c04116e676f6280b824ab7469d5d3080e5cba4f2128R126-R144
we shouldn't see `/usr/share/miniconda/envs/test/conda-bld/work` used in these kinds of tests anymore...
Something about the fixture/its "autouse" doesn't seem to work properly *sigh*...
```python
__________________ test_no_satisfiable_variants_raises_error ___________________
[gw0] linux -- Python 3.9.15 /usr/share/miniconda/envs/test/bin/python
  File "/usr/share/miniconda/envs/test/lib/python3.9/shutil.py", line 825, in move
    os.rename(src, real_dst)
FileNotFoundError: [Errno 2] No such file or directory: '/usr/share/miniconda/envs/test/conda-bld/work' -> '/usr/share/miniconda/envs/test/conda-bld/abc_1670250124813/work'

  File "/home/runner/work/conda-build/conda-build/tests/test_variants.py", line 106, in test_no_satisfiable_variants_raises_error
    api.render(recipe, permit_unsatisfiable_variants=True)
  File "/home/runner/work/conda-build/conda-build/conda_build/api.py", line 38, in render
    metadata_tuples = render_recipe(recipe_path, bypass_env_check=bypass_env_check,
  File "/home/runner/work/conda-build/conda-build/conda_build/render.py", line 827, in render_recipe
    m.config.compute_build_id(m.name(), m.version(), reset=reset_build_id)
  File "/home/runner/work/conda-build/conda-build/conda_build/config.py", line 615, in compute_build_id
    shutil.move(old_dir, work_dir)
  File "/usr/share/miniconda/envs/test/lib/python3.9/shutil.py", line 845, in move
    copy_function(src, real_dst)
  File "/usr/share/miniconda/envs/test/lib/python3.9/shutil.py", line 444, in copy2
    copyfile(src, dst, follow_symlinks=follow_symlinks)
  File "/usr/share/miniconda/envs/test/lib/python3.9/shutil.py", line 264, in copyfile
    with open(src, 'rb') as fsrc:
FileNotFoundError: [Errno 2] No such file or directory: '/usr/share/miniconda/envs/test/conda-bld/work'
```
</details>

jakirkham:
It sounds like we would like to capture this in a patch release of conda-build (like 3.23.3). Is that something that could be started?

jezdez:
Yes

PR #4568, opened 2022-08-19
Standardize license disclaimer

kenodegard:
### Description

Adding standardized license disclaimer ~and pragmas~. See https://github.com/conda/conda/pull/11183.

Copyright year picked based on the first commit timestamp.

### Checklist - did you ...

- [x] ~Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release notes?~
     
- [x] ~Add / update necessary tests?~
- [x] ~Add / update outdated documentation?~

PR #4676, opened 2022-12-15
Update disclaimer/LICENSE to reflect conda-build originally being part of conda

kenodegard:
### Description

We were reminded (https://github.com/conda/conda-build/pull/4568#issuecomment-1353773401) that conda-build's code was originally part of conda and only later (~2014) was it moved into a separate repo.

This change uses standard SPDX syntax for disclaimers (I opted to use `©`, if there are any objections or concerns I'll switch it back to `Copyright (c)`), credits Continuum Analytics with the original copyright, and recognizes the rename to Anaconda in 2017.

Sought to otherwise standardize the disclaimer and LICENSE.txt to mirror that of conda (https://github.com/conda/conda/pull/12183).

Ref https://matija.suklje.name/how-and-why-to-properly-write-copyright-statements-in-your-code

### Checklist - did you ...

### DIFF
--- a/.github/disclaimer.txt
+++ b/.github/disclaimer.txt
@@ -1,2 +1,3 @@
-Copyright (C) 2014 Anaconda, Inc
+SPDX-FileCopyrightText: © 2012 Continuum Analytics, Inc. <http://continuum.io>
+SPDX-FileCopyrightText: © 2017 Anaconda, Inc. <https://www.anaconda.com>
 SPDX-License-Identifier: BSD-3-Clause
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,23 +1,25 @@
-Except where noted below, conda is released under the following terms:
+BSD 3-Clause License
 
-(c) 2012 Continuum Analytics, Inc. / http://continuum.io
-All Rights Reserved
+© 2012 Continuum Analytics, Inc. <http://continuum.io>
+© 2017 Anaconda, Inc. <https://www.anaconda.com>
+All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
 modification, are permitted provided that the following conditions are met:
+
     * Redistributions of source code must retain the above copyright
       notice, this list of conditions and the following disclaimer.
     * Redistributions in binary form must reproduce the above copyright
       notice, this list of conditions and the following disclaimer in the
       documentation and/or other materials provided with the distribution.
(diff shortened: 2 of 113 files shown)


.github/disclaimer.txt
```diff
@@ -1,2 +1,3 @@
-Copyright (C) 2014 Anaconda, Inc
+SPDX-FileCopyrightText: © 2012 Continuum Analytics, Inc. <http://continuum.io>
+SPDX-FileCopyrightText: © 2017 Anaconda, Inc. <https://www.anaconda.com>
```
mbargull, line 2: Let's not overcomplicate things and go with
```suggestion
SPDX-FileCopyrightText: © 2012 Anaconda, Inc. <https://www.anaconda.com>
```
The copyright remained with the same entity ([CIK: 0001538860](https://www.sec.gov/edgar/browse/?CIK=0001538860)), after all.
Let's just us the current name.

mbargull (changes requested):

jezdez:
Hmm, I wonder if https://hynek.me/til/copyright-years/ might be worth another look, a lot of corporate project have removed the years altogther!

mbargull:
> Hmm, I wonder if https://hynek.me/til/copyright-years/ might be worth another look, a lot of corporate project have removed the years altogether!

Ah, nice find! It's good to see that this is has some notable examples. That'd be my favored solution because it's more faithful, https://github.com/conda/conda-build/pull/4568#issuecomment-1353773401 :
> However, a lot of the code/source files are from 2016+, i.e., `(c) 2012`/`(c) 2014` is not representative/true for that.

Going with `SPDX-FileCopyrightText: © Anaconda, Inc. <https://www.anaconda.com>` in the file headers and either leaving `LICENSE.txt` untouched or just substituting Continuum -> Anaconda sounds reasonable.
(Though, I'd leave `2012` in `LICENSE.txt` -- doesn't really hurt at that single place, plus it is a faithful statement in there.)

.github/disclaimer.txt
mbargull:
```suggestion
SPDX-FileCopyrightText: © Anaconda, Inc. <https://www.anaconda.com>
```

jezdez:
No, this is blocked by the legal work being done in providing a trademark license to NumFOCUS and can now picked up again.

PR #4677, opened 2022-12-16
Remove logic branching for `conda_43`/`conda_44`/`conda_45`/`conda_46`/`conda_47`

kenodegard:
### Description

Since conda-build's minimum compatible conda is 4.13 we can remove all older conda logic.

### Checklist - did you ...

### DIFF
--- a/conda_build/render.py
+++ b/conda_build/render.py
@@ -67,11 +67,12 @@ def bldpkg_path(m):
 
 
 def actions_to_pins(actions):
-    specs = []
-    spec_name = lambda x: x.dist_name
-    if 'LINK' in actions:
-        specs = [' '.join(spec_name(spec).split()[0].rsplit('-', 2)) for spec in actions['LINK']]
-    return specs
+    if "LINK" not in actions:
+        return [
+            " ".join(spec.dist_name.split()[0].rsplit("-", 2))
+            for spec in actions["LINK"]
+        ]
+    return []
 
 
 def _categorize_deps(m, specs, exclude_pattern, variant):
--- a/conda_build/render.py
+++ b/conda_build/render.py
@@ -67,7 +67,7 @@ def bldpkg_path(m):
 
 
 def actions_to_pins(actions):
-    if "LINK" not in actions:
+    if "LINK" in actions:
         return [
             " ".join(spec.dist_name.split()[0].rsplit("-", 2))
             for spec in actions["LINK"]
--- a/conda_build/render.py
+++ b/conda_build/render.py
@@ -20,7 +20,6 @@ from .conda_interface import (UnsatisfiableError, ProgressiveFetchExtract,
                               TemporaryDirectory)
 from .conda_interface import execute_actions
 from .conda_interface import pkgs_dirs
-from .conda_interface import conda_43
 from .conda_interface import specs_from_url
 from .utils import CONDA_PACKAGE_EXTENSION_V1, CONDA_PACKAGE_EXTENSION_V2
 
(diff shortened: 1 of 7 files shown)


sven6002 (commented):
Can you comment on the epic as well that these pieces are removed

sven6002 (dismissed):

conda_build/render.py
```diff
@@ -69,10 +68,7 @@ def bldpkg_path(m):
 
 def actions_to_pins(actions):
     specs = []
-    if conda_43:
-        spec_name = lambda x: x.dist_name
-    else:
-        spec_name = lambda x: str(x)
+    spec_name = lambda x: x.dist_name
     if 'LINK' in actions:
         specs = [' '.join(spec_name(spec).split()[0].rsplit('-', 2)) for spec in actions['LINK']]
```
mbargull, line 73:
```suggestion
    if 'LINK' in actions:
        specs = [' '.join(spec.dist_name.split()[0].rsplit('-', 2)) for spec in actions['LINK']]
```

beeankha (dismissed):

conda_build/utils.py
```diff
@@ -1504,36 +1504,7 @@ def filter_info_files(files_list, prefix):
...
-            pass
-
-    conda_log_level = logging.WARN
-    if config and config.debug:
-        conda_log_level = logging.DEBUG
-    with LoggingContext(conda_log_level):
-        # this clears out the path from conda's cache, which otherwise thinks
-        #    that things are still installed here
-        _rm_rf(path)
+    return _rm_rf(path)
```
mbargull, line 1507: Should remove this indirection and remove the `as _rm_rf` from the `import` line.

tests/test_api_build.py
```diff
@@ -781,15 +780,6 @@ def test_noarch_python_1(testing_config):
     assert 'package_metadata_version' in extra
 
 
-@pytest.mark.sanity
-@pytest.mark.xfail(conda_47, reason="parallel verify/execute in conda 4.7 breaks legacy noarch, which depends on having the env files present before pre-link scripts are run.")
```
mbargull, line 785: Should only remove the `@pytest.mark.xfail` but not the whole test case.

mbargull (changes requested):

conda_build/render.py
kenodegard: yup, I saw a few refactors I could do, just wanted to make this PR easy to review at first, I'll take another pass

conda_build/utils.py
kenodegard: I agree, unfortunately doing so will deprecate the `config` argument

I wanted to get https://github.com/conda/conda/pull/12125 merged before starting to mark  deprecations per [CEP9](https://github.com/conda-incubator/ceps/blob/main/cep-9.md)

tests/test_api_build.py
kenodegard: so rewrite the test to check for an expected exception?

conda_build/render.py
mbargull: kk. I also don't mind keeping the cleanup simple -- just looked at the few surrounding lines.
(I'm often unsure when PRs are done/meant for review -- I like the "mark as draft" functionality to communicate that ;).

conda_build/utils.py
mbargull: Ah, right. In the moment I forgot that we have a few downstream users (including code I wrote :facepalm:) that import things from everywhere.

tests/test_api_build.py
mbargull: Nvm, I somehow read it as `not conda_47`, sorry. Okay to remove then :).

conda_build/render.py
kenodegard: @mbargull no worries, I def try to follow the draft/ready distinction, we haven't been very good at following a consistent process as a community, something to aspire for in 2023

personally, I always appreciate reviews no matter what stage the PR is in

beeankha (approved):
Thanks for doing this, @kenodegard !

PR #4679, opened 2022-12-16
Move tar-ing into separate step to avoid shadowing test results

kenodegard:
### Description

https://github.com/conda/conda-build/pull/4621 added allure reports and adds a `tar` command:

https://github.com/conda/conda-build/blob/9951c3ee0e208cf9c7e3c42a0b88b44560f04989/.github/workflows/tests.yml#L281-L282

Unfortunately, this `tar` command occurs within the same step as `pytest` resulting in the `pytest` exitcode being shadowed by the `tar` exitcode. The fix is to move the `tar` process into a separate step.

See this test that reports the test step as having passed despite there being failures: https://github.com/conda/conda-build/actions/runs/3709733101/jobs/6288635122#step:11:3184

### Checklist - did you ...

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -117,7 +117,6 @@ jobs:
               -m "serial" tests \
               ${{ env.pytest-replay }} \
               --alluredir=allure-results
-          tar -zcf allure-results.tar.gz allure-results
 
       - name: Run Parallel tests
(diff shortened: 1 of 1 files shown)


jezdez (approved):

sven6002:
This is good to go..

PR #4681, opened 2022-12-19
Remove custom temporary directories

kenodegard:
### Description

We noticed with the recent update to the GitHub Windows Runner that the `tests/test_api_build.py::test_failed_recipe_leaves_folders` became very flaky:
- https://github.com/conda/conda-build/actions/runs/3716460346/jobs/6332948506
- https://github.com/conda/conda-build/actions/runs/3711735797/jobs/6332950346
- https://github.com/conda/conda-build/actions/runs/3727885677/jobs/6332968601

This is an attempt to fix the issue by using the standard temporary directories.

### Checklist - did you ...

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -111,7 +111,7 @@ jobs:
               --color=yes \
               -vv \
               -n 0 \
-              --basetemp /tmp/cb_serial \
+              --basetemp "${{ runner.temp }}/${{ matrix.test-type }}" \
               --cov conda_build \
               --cov-report xml \
               -m "serial" tests \
@@ -125,7 +125,7 @@ jobs:
               --color=yes \
               -vv \
               -n auto \
-              --basetemp /tmp/cb \
+              --basetemp "${{ runner.temp }}/${{ matrix.test-type }}" \
               --cov conda_build \
               --cov-append \
               --cov-report xml \
@@ -171,14 +171,6 @@ jobs:
         uses: actions/checkout@v2
         with:
           fetch-depth: 0
-
-      - name: Set temp dirs correctly
-        shell: cmd
-        # https://github.com/actions/virtual-environments/issues/712
-        run: |
-          echo "TMPDIR=$env:USERPROFILE\AppData\Local\Temp" >> $env:GITHUB_ENV
-          echo "TEMP=$env:USERPROFILE\AppData\Local\Temp" >> $env:GITHUB_ENV
-          echo "TMP=$env:USERPROFILE\AppData\Local\Temp" >> $env:GITHUB_ENV
 
       - name: Setup Python ${{ matrix.python-version }}
         uses: actions/setup-python@v1
@@ -256,10 +248,6 @@ jobs:
           pip install allure-pytest
           conda-build --version
           pushd .. && git clone https://github.com/conda/conda_build_test_recipe && popd
-          mkdir %UserProfile%\cbtmp_serial
-          mkdir %UserProfile%\cbtmp
-          for /d %%F in (%UserProfile%\cbtmp_serial\*) do rd /s /q "%%F"
-          for /d %%F in (%UserProfile%\cbtmp\*) do rd /s /q "%%F"
           call conda create -n blarg -yq --download-only python=2.7||exit 1
           call conda create -n blarg -yq --download-only python=3.8||exit 1
           call conda create -n blarg -yq --download-only python cmake||exit 1
@@ -279,7 +267,7 @@ jobs:
           set PERL=
           set LUA=
           set R=
-          pytest --color=yes -vv -n 0 --basetemp %UserProfile%\cbtmp_serial --cov conda_build --cov-report xml -m "serial" ${{ env.pytest-replay }} --alluredir=allure-results
+          pytest --color=yes -vv -n 0 --basetemp "${{ runner.temp }}\${{ matrix.test-type }}" --cov conda_build --cov-report xml -m "serial" ${{ env.pytest-replay }} --alluredir=allure-results
         shell: cmd
 
       - name: Run Parallel Tests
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -172,14 +172,6 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Set temp dirs correctly
-        shell: cmd
-        # https://github.com/actions/virtual-environments/issues/712
-        run: |
-          echo "TMPDIR=$env:USERPROFILE\AppData\Local\Temp" >> $env:GITHUB_ENV
```
sven6002, line 179: we don't need to add these in env variables for windows anymore???

.github/workflows/tests.yml
```diff
@@ -393,7 +381,7 @@ jobs:
           conda create -n blarg1 -yq python=2.7
```
sven6002, line 381: do we still need that python=2.7 since upper and lower bounds are `3.7 and `3.10`

.github/workflows/tests.yml
kenodegard: `%TEMP%` and `%TMP%` are default variables on windows, this was added to address the issue here https://github.com/actions/virtual-environments/issues/712 which is hopefully a non-issue today (that's a 2 year old issue)

sven6002: ok.. then removing that should sound good

.github/workflows/tests.yml
sven6002: since you did change the upper and lower limits in the previous PR.. thought this would be redundant

kenodegard: AFAIK this environment is unused in our tests (even if it is used it should be refactored into a pytest fixture) so it can likely be removed entirely, my intention was to do a Windows-only cleanup once we get https://github.com/conda/conda-build/pull/4648 merged

sven6002 (approved):
I like the pytest mark for flaky.. works with me

tests/test_api_build.py
```diff
@@ -1180,21 +1177,13 @@ def test_failed_recipe_leaves_folders(testing_config, testing_workdir):
...
-    locale.getpreferredencoding(False)
-    for lock in locks:
-        if os.path.isfile(lock.lock_file):
-            any_locks = True
-            dest_path = base64.b64decode(os.path.basename(lock.lock_file))
-            if hasattr(dest_path, "decode"):
-                dest_path = dest_path.decode(errors="backslashreplace")
-            locks_list.add((lock.lock_file, dest_path))
-    assert not any_locks, "\n".join(
+    locks = {lock.lock_file for lock in locks if os.path.isfile(lock.lock_file)}
```
sven6002, line 1183: this makes sense.. with removing the old set up variables.. and decoding..

tests/test_api_build.py
```diff
@@ -1180,21 +1177,13 @@ def test_failed_recipe_leaves_folders(testing_config, testing_workdir):
...
+    assert not locks, "\n".join(
         "remaining locks:",
```
sven6002, line 1185: could there me more than one locked `path`

tests/test_api_build.py
kenodegard: I removed it because, as far as I could tell, the decoding didn't expose anything meaningful. I also think the decoding was gibberish, so perhaps it used to mean something in the past before conda adopted `filelock` for the locking mechanism?

tests/test_api_build.py
kenodegard: supposedly different locks are created depending on what logic is being blocked from simultaneously running but in the failing test there is only ever one lock that fails to be released

sven6002: that is true.. i think if we need to expand this we will need to see what set of locks are there..

travishathaway (approved):

beeankha (dismissed):

tests/test_api_build.py
```diff
@@ -1170,6 +1167,8 @@ def test_unknown_selectors(testing_config):
     api.build(recipe, config=testing_config)
 
 
+# the locks can be very flaky on GitHub Windows Runners
+@pytest.mark.flaky(rerun=5, reruns_delay=2)
```
mbargull, line 1171: Hmm... Given that GitHub Actions and Azure Pipelines are widely used by, e.g., conda-forge, this flakiness isn't unlikely to be encountered in production.
We should investigate this more (i.e., create issue/testing PR to track this).

kenodegard: https://github.com/conda/conda-build/issues/4685

tests/test_api_build.py
```diff
@@ -1178,20 +1177,14 @@ def test_failed_recipe_leaves_folders(testing_config, testing_workdir):
...
-            if hasattr(dest_path, 'decode'):
-                dest_path = dest_path.decode(sys.getfilesystemencoding(), errors='backslashreplace')
-            locks_list.add((lock.lock_file, dest_path))
-    assert not any_locks, "remaining locks:\n{}".format('\n'.join('->'.join((l, r))
-                                                                for (l, r) in locks_list))
+    locks = {lock.lock_file for lock in locks if os.path.isfile(lock.lock_file)}
+    assert not locks, "\n".join(
+        "remaining locks:",
+        *(f"  {lock}" for lock in locks.items()),
+    )
```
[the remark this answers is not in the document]
kenodegard: hm good point, doing `not locks` will also be sufficient:

Screen Shot 2022-12-22 at 11 51 13 AM

```suggestion
    assert not {lock.lock_file for lock in locks if os.path.isfile(lock.lock_file)}
```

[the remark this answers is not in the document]
mbargull: (This might not be terribly important in this case, but for longer lists/entries you get a proper diff if you use `==`. With `not ...` those longer entries will actually just be truncated instead.)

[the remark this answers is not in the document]
kenodegard: We run `pytest -v` in CI so doesn't that already show us the full diff?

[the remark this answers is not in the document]
mbargull: `assert not ...` will show you `AssertionError: assert not ...<truncated>`.
`assert [] == ...` will show you `AssertionError: assert [] == ...<truncated>` and additionally a diff (also truncated).
(A non-truncated "full diff" gets output if you run `pytest -vv`.)

sven6002 (approved):
Already approved before holidays

PR #4690, opened 2023-01-04
Replace bundled conda-index with standalone conda-index package

### LINKED ISSUE #4645: Replace conda_build/index.py with standalone conda-index
### Checklist

### What is the idea?

conda-build generates repodata.json during its run, using get_build_index https://github.com/conda/conda-build/blob/main/conda_build/index.py#L112

This confusing function both downloads repodata.json, and generates new repodata.json for the local package cache / the packages that were just built.

It also provides the "conda-build index" command line interface, or "conda index".

Now that we have a much faster standalone conda-index descended from conda_build/index.py, conda-build should add a dependency on conda-index, and call that instead. (The downloading parts of `get_build_index` will probably have to stay)

- [ ] package conda-index
- [ ] figure out what to do about the new CLI "conda index" (provided by standalone conda-index) versus the old CLI "conda index" provided by conda-build
- [ ] depend on conda-index
- [ ] rewrite get_build_index

### Why is this needed?

This will ensure that we have one excellent repodata.json generator.

### What should happen?

_No response_

### Additional Context

_No response_

dholth:
### Description

Delegate conda-build's conda-index while keeping code for deprecation cycle reasons.

Could we delete the hotfix-source-repo argument as it has been unused for longer than a deprecation cycle?

Fix #4645 

We should probably delete the entire `test_index.py` which is now part of conda-index. Is there any part of index testing that needs to remain here, perhaps the legacy API? Assuming `get_build_index` is tested as part of all the conda-build invocations done by us.

### Checklist - did you ...


conda_build/build_index.py
```diff
@@ -0,0 +1,272 @@
...
+        "lua": "lua",
+        "nodejs": "js",
+        "perl": "perl",
+        "php": "php",
+        "ruby": "ruby",
+        "m2-base": "m2",
+        "msys2-conda-epoch": "m2w64",
+    }
+    NAMESPACE_PACKAGE_NAMES = frozenset(NAMESPACES_MAP)
+    NAMESPACES = frozenset(NAMESPACES_MAP.values())
```
dholth, line 64: Is this used?

conda_build/build_index.py
```diff
@@ -0,0 +1,272 @@
+# Copyright (C) 2014 Anaconda, Inc
+# SPDX-License-Identifier: BSD-3-Clause
+"""
+conda-build's use of conda-index, delegated to now-separate conda-index package.
```
dholth, line 4: Copied to make keeping track of all the changes easier, and to discourage future `conda_build.index` imports

conda_build/cli/main_index.py
```diff
@@ -90,8 +90,8 @@ def execute(args):
 
     api.update_index(args.dir, check_md5=args.check_md5, channel_name=args.channel_name,
```
dholth, line 91: Would love to delete this whole file.

`conda-index` has a nice click CLI, but doesn't take the same arguments.

conda_build/environ.py
```diff
@@ -776,9 +775,10 @@ def get_install_actions(prefix, specs, env, retries=0, subdir=None,
 
     bldpkgs_dirs = ensure_list(bldpkgs_dirs)
 
-    index, index_ts, _ = get_build_index(subdir, list(bldpkgs_dirs)[0], output_folder=output_folder,
-                                      channel_urls=channel_urls, debug=debug, verbose=verbose,
-                                      locking=locking, timeout=timeout)
+    bldpkgs_dir = list(bldpkgs_dirs)[0]
+    index, index_ts, _ = build_index.get_build_index(subdir, bldpkgs_dir, output_folder, False,
```
dholth, line 779: inline function courtesy of pycharm refactor feature

conda_build/index.py
```diff
@@ -12,10 +12,9 @@
 import json
 from numbers import Number
 import os
-from os.path import abspath, basename, getmtime, getsize, isdir, isfile, join, splitext, dirname
+from os.path import abspath, basename, getmtime, getsize, isdir, isfile, join, splitext
```
dholth, line 15: "remove unused imports"

conda_build/index.py
```diff
@@ -237,33 +126,9 @@ def _ensure_valid_channel(local_folder, subdir):
...
-                        deep_integrity_check=check_md5, debug=debug).index(
-                            patch_generator=patch_generator, verbose=verbose,
-                            progress=progress,
-                            hotfix_source_repo=hotfix_source_repo,
-                            current_index_versions=current_index_versions,
-                            index_file=index_file)
-
+    return build_index.update_index(dir_path, check_md5, channel_name, patch_generator, threads,
+                 verbose, progress, hotfix_source_repo, subdirs, warn,
+                 current_index_versions, debug, index_file)
```
dholth, line 131: As of today an unused copy of the full index implementation remains in this file, for deprecation cycle reasons.

tests/test_index.py
```diff
@@ -353,7 +353,7 @@ def test_index_noarch_osx64_1(testing_workdir):
                 "home": "https://anaconda.org/conda-test/conda-index-pkg-a",
```
dholth, line 353: can we delete this entire file, as a nearly identical one exists in conda-index?

jezdez: Yes we can

conda_build/build_index.py
jezdez: To clarify, we couldn't continue to have index.py as the place for this code to live? Or asked differently, the get_build_index function, is that something that we have in conda-index? I'm a little worried we're introducing another place where users will end up importing from :)

jezdez (changes requested):
I have a few questions about the deprecation strategy of this code, and generally think we need to use deprecation warnings here and not only shims around the old code paths.

conda_build/build_index.py
dholth: `get_build_index` is only in conda-build. It is mostly about fetching repodata.json and not about generating repodata.json.

conda_build/build_index.py
dholth: it wouldn't be surprising if conda.base.constants exists in all conda > 4.7 or the oldest one we can depend on.

conda_build/build_index.py
dholth: However we might want to raise deprecationwarning when conda_build.index is imported.

dholth:
I'm still figuring out the best way to do deprecation. The current strategy is to leave the unused methods in, and come back a year later to remove the code.

.github/workflows/tests.yml
```diff
@@ -161,14 +161,14 @@ jobs:
...
+        python-version: ['3.7', '3.10']
         conda-version: [release]
         test-type: [serial, parallel]
         include:
-          - python-version: '3.9'
+          - python-version: '3.10'
             conda-version: canary
             test-type: serial
-          - python-version: '3.9'
+          - python-version: '3.10'
```
dholth, line 171: Maybe deal with a version we are testing under? What is this include directive for?

.github/workflows/tests.yml
```diff
@@ -161,14 +161,14 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.7', '3.8', '3.9', '3.10']
+        python-version: ['3.7', '3.10']
```
dholth, line 164: Attempt to test faster.

.github/workflows/tests.yml
```diff
@@ -182,22 +182,24 @@ jobs:
           fetch-depth: 0
 
       - name: Setup Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v1
+        uses: actions/setup-python@v4
```
dholth, line 185: Might as well use the newest one.

.github/workflows/tests.yml
```diff
@@ -182,22 +182,24 @@ jobs:
...
           Start-Process "VCForPython27.msi" /qn -Wait
         shell: pwsh
 
+      # would be great to cache this
       - name: Install miniconda
         run: |
           rmdir C:\Strawberry /s /q
           choco install wget visualstudio2017-workload-vctools
-          wget https://repo.anaconda.com/miniconda/Miniconda3-latest-Windows-x86_64.exe
+          wget -q https://repo.anaconda.com/miniconda/Miniconda3-latest-Windows-x86_64.exe
```
dholth, line 202: Quiet. This download is pretty slow. Why not use New-Object net.webclient as above?

.github/workflows/tests.yml
```diff
@@ -242,11 +244,9 @@ jobs:
           call conda config --set always_yes yes
           call conda config --set auto_update_conda no
           call conda info
-          python -c "import sys; print(sys.version)"
-          python -c "import sys; print(sys.executable)"
-          python -c "import sys; print(sys.prefix)"
+          python -c "import sys; print(sys.version); print(sys.executable); print(sys.prefix)"
           call conda update -q --all||exit 1
-          call conda install -q pip python-libarchive-c pytest git pytest-cov jinja2 m2-patch flake8 mock requests contextlib2 chardet glob2 perl pyflakes pycrypto posix m2-git anaconda-client numpy beautifulsoup4 pytest-xdist pytest-mock filelock pkginfo psutil pytz tqdm conda-package-handling||exit 1
+          call conda install -q pip pytest git pytest-cov jinja2 m2-patch flake8 mock requests contextlib2 chardet glob2 perl pyflakes pycrypto posix m2-git anaconda-client numpy beautifulsoup4 pytest-xdist pytest-mock filelock pkginfo psutil pytz tqdm conda-package-handling conda-index||exit 1
```
dholth, line 249: It would be nice if we could pull these out of the recipe or setuptools dependencies.

.github/workflows/tests.yml
```diff
@@ -258,7 +258,7 @@ jobs:
           conda-build --version
           pushd .. && git clone https://github.com/conda/conda_build_test_recipe && popd
           call conda create -n blarg -yq --download-only python=2.7||exit 1
-          call conda create -n blarg -yq --download-only python=3.8||exit 1
+          call conda create -n blarg -yq --download-only python=3.10||exit 1
```
dholth, line 261: Why not download a relevant version. Not sure why we need to test conda create here.

conda_build/cli/main_index.py
dholth: this might be able to use the non-api update_index, which is basically "call update_index for every dir in a list"

.github/workflows/tests.yml
kenodegard: `include`s are appended to the `matrix`

.github/workflows/tests.yml
kenodegard: need to rebase

.github/workflows/tests.yml
kenodegard: rebase will remove much of this

dholth:
@kenodegard I updated, and tried to keep the main branch's version of the test configs.

conda_build/build_index.py
kenodegard: I don't think this submodule should exist.

Plus, all of `conda.index` should be deprecated with warnings to import from `conda-index` instead.

dholth: Do you mean `conda.index` or `conda_build.index`

Reminder that conda-build's `get_build_index` doesn't belong in conda-index.

dholth: @kenodegard @jezdez I moved this module back to conda_build.index

dholth:
I've updated this to remove the new build_index.py and copy it over to index.py instead (since the redundant fuctionality in index is in standalone conda-index).

news/4645-use-conda-index
```diff
@@ -0,0 +1,20 @@
...
+* Depend on standalone conda-index instead of bundled indexing code.
+
+### Bug fixes
+
+* <news item>
+
+### Deprecations
+
+* `conda-build index --file` is no longer supported.
+* `conda-build index --hotfix-source-repo` is removed.
```
kenodegard, line 12:
```suggestion
* `conda-build index --file` is no longer supported and is pending deprecation. (#4690)
* `conda-build index --hotfix-source-repo` is no longer supported and is pending deprecation. (#4690)
```

kenodegard (changes requested):
Minor edits but is otherwise good to go

tests/test_api_skeleton_cran.py
```diff
@@ -17,7 +17,7 @@
 @pytest.mark.parametrize(
     "package,license_id,license_family,license_files",
     [
-        ("r-rmarkdown", "GPL-3", "GPL3", {"GPL-3"}),
+        # ("r-rmarkdown", "GPL-3", "GPL3", {"GPL-3"}),  # fails with connection error
```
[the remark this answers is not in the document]
dholth: Seems like a pointless test IMO

news/4645-use-conda-index
dholth: Trying to communicate that `--file` no longer does anything, and `--hotfix-source-repo` produces an "unknown command line argument" error.

kenodegard: Per CEP9 we shouldn't be removing things outright. We should update the code to allow these arguments but raise deprecation warnings with instructions on what to do instead.

tests/test_api_skeleton_cran.py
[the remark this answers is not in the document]
kenodegard: That's fine but let's have that conversation/effort elsewhere, we want to keep PRs as focused as possible

kenodegard (commented):
My biggest question is how do we know if the new conda-index works like the old conda-index? Is it possible to get `tests/test_index.py` to verify the new codes? Or add `tests/cli/test_main_index.py`?

dholth:
> My biggest question is how do we know if the new conda-index works like the old conda-index? Is it possible to get `tests/test_index.py` to verify the new codes? Or add `tests/cli/test_main_index.py`?

conda-index has its own pretty good test suite that is based on the old conda-build suite, in the same way that the rest of the conda-index code is based on its counterpart from conda-build.

news/4645-use-conda-index
dholth: Doesn't CEP9 only apply to conda?

dholth: This script is probably run in a non-interactive cron job. We could print a warning and fail to do what the user expects, since they are not actually there to observe the deprecation warning. Instead the current code exits.

tests/test_api_skeleton_cran.py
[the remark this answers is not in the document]
dholth: We constantly face a simple choice between a focused PR and a PR with passing tests.

PR #4691, opened 2023-01-04
Adjust testing matrix

kenodegard:
### Description

Only run tests with all supported Python versions on Linux where tests are faster, and runners are plentiful. On Windows (where tests are slower) and macOS (where runners are limited), only run tests on the lower and upper Python bounds with the stable and canary conda packages, respectively.

### Checklist - did you ...


.github/workflows/tests.yml
```diff
@@ -310,11 +312,17 @@ jobs:
...
+        conda-version: [release]
         test-type: [serial, parallel]
-      max-parallel: 10
+        include:
+          - python-version: '3.10'
+            conda-version: canary
+            test-type: serial
+          - python-version: '3.10'
+            conda-version: canary
+            test-type: parallel
```
jezdez, line 325: Do we want to allow these to fail?

kenodegard: no? we should always be ready to release?

beeankha (approved):

PR #4695, opened 2023-01-06
Speedup `test_subpackages.py::test_intradependencies`

kenodegard:
### Description

`test_intradependencies` would previously take ~1min to run locally on macOS and Linux but 4+min to run on Windows. On shared machines (e.g., GitHub Actions) this could take 10+min on Windows, see [this recent CI run on Windows](https://github.com/conda/conda-build/actions/runs/3851180271/jobs/6563654598#step:10:1471).

Changes include:
- replaced `git` dependency with cheaper `ca-certificates` package (kB vs. MB),
- simplified dependency logic by dropping `m2-base` dependency for Windows,
- replaced bad dummy names (`abc` and `foo`),
- and shortened verbose naming

With these changes, these tests still complete locally in ~1min for macOS/Linux and ~1.5min for Windows.

### Checklist - did you ...


jezdez (approved):

dholth:
Have we tried turning parallelism off on the parallel test, to see if it runs at the same speed

jezdez:
> Have we tried turning parallelism off on the parallel test, to see if it runs at the same speed

I don't think we have!

PR #4696, opened 2023-01-06
Speedup `test_variants.py::test_get_package_variants*`

kenodegard:
### Description

Both `test_get_package_variants_from_file` and `test_get_package_variants_from_dictionary_of_lists` tests would take 3+min to run locally on all OSes. On shared machines (e.g., GitHub Actions) this could take 7+min, see [this recent CI run on Windows](https://github.com/conda/conda-build/actions/runs/3851180271/jobs/6563654598#step:10:1472).

Changes include:
- remove the unused `build_config.yaml`,
- drop the unnecessary `numpy` dependency, and
- standardize the directory definition to follow what is used elsewhere in the test suite.

With these changes, these tests complete locally in less than 30s each.

### Checklist - did you ...


jezdez (dismissed):

tests/test_variants.py
```diff
@@ -12,8 +13,7 @@
 from conda_build import api, exceptions, variants
 from conda_build.utils import package_has_file
 
-thisdir = os.path.dirname(__file__)
-recipe_dir = os.path.join(thisdir, 'test-recipes', 'variants')
+from .utils import variants_dir
 
 
 def test_later_spec_priority(single_version, no_numpy_version):
```
sven6002, line 19: we could move this into the utils folder as well..

tests/test_variants.py
```diff
@@ -35,38 +35,76 @@ def test_later_spec_priority(single_version, no_numpy_version):
...
+
     # one for each Python version.  Numpy is not strictly pinned and should present only 1 dimension
     assert len(metadata) == 2
-    assert sum('python >=2.7,<2.8' in req for (m, _, _) in metadata
-               for req in m.meta['requirements']['run']) == 1
-    assert sum('python >=3.5,<3.6' in req for (m, _, _) in metadata
-               for req in m.meta['requirements']['run']) == 1
+    assert (
+        sum(
+            "python >=2.7,<2.8" in req
```
sven6002, line 57: do we update the python version here??

tests/test_variants.py
```diff
@@ -35,38 +35,76 @@ def test_later_spec_priority(single_version, no_numpy_version):
...
+            for (m, _, _) in metadata
+            for req in m.meta["requirements"]["run"]
+        )
+        == 1
+    )
+    assert (
+        sum(
+            "python >=3.5,<3.6" in req
```
sven6002, line 65: same here

tests/test_variants.py
kenodegard: move what?

tests/test_variants.py
kenodegard: not necessary, see the `no_numpy_version` fixture, which defines the variants we are building against

sven6002: sorry that was wrong comment.. I'm deleting it

tests/test_variants.py
sven6002: sorry wrong comment

tests/test_variants.py
kenodegard: The versions of Python utilized here are insignificant, but I went ahead and refactored this test (and other similar tests) to make this less ambiguous.

kenodegard:
Whoa, this last round of refactors made this PR very noisy...perhaps this is better split into smaller PRs?

kenodegard:
Discussed with @sven6002 and realized that it would be better to do the refactor in a follow-up PR, cherry-picked today's commits into a new branch, and reverted back to the original edits from last week.

sven6002 (approved):

Issue #4718, opened 2023-01-17
Update versions listed in `conftest.py`


beeankha:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### What happened?

`/conda-build/tests/conftest.py` needs the following to have updated versions:
- Python
- Numpy
- Conda
- Operating systems

The fixtures should also default to not returning hard-coded values, if possible.

Along with this, any tests that are using the fixtures in this file need to be updated accordingly.

beeankha:
Closing this ticket as this issue is no longer relevant!

Issue #4717, opened 2023-01-17
`numpy` version for conda-build is pointing to 1.16 on MacOS (M1 only)


beeankha:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### What happened?

When running `tests/test_variants.py::test_numpy_used_variable_looping`, the following error pops up:

```
    raise DependencyNeedsBuildingError(exc, subdir=subdir)
conda_build.exceptions.DependencyNeedsBuildingError: Unsatisfiable dependencies for platform osx-arm64: {'numpy=1.16'}
```

`conda-build/tests/test-recipes/variants/numpy_used/conda_build_config.yaml` contains the following:

```
[snippet]
numpy:
    - 1.16
    - 1.19
```

We should either upgrade the `numpy` version for M1 Mac systems or we need to skip this test for now; after the fix, we'll remove the `skipif` decorator.

### Additional Context

Release notes from `numpy`'s latest version for M1 support: https://numpy.org/doc/stable/release/1.21.4-notes.html

skupr-anaconda:
@beeankha`numpy` should be **1.19** for `osx-arm64` platform, see https://github.com/AnacondaRecipes/aggregate/blob/b655086f99139bf951d7a59ce1315249d969c58d/conda_build_config.yaml#L158-L162

Issue #4590, opened 2022-09-29
Pruning the `conda-build` test cases


sven6002:
### Summary

The testing strategy for the `conda-build` repo is currently outdated and was last maintained thoroughly 6 years ago. So just as (a reference to Star Trek) USS Enterprise underwent multiple refits, redesigns, and upgrade-based missions it was tasked with and the captains that led those missions.

As the orbital shipyard is used for such a purpose URSA team is also updating and re-formatting / redesigning the test cases for `conda-build` right now.

### Tasks
- [x] test_api_build_conda_v2.py **(no pruning required)** @sven6002
- [x] https://github.com/conda/conda-build/pull/4722
- [x] https://github.com/conda/conda-build/pull/4723
- [x] https://github.com/conda/conda-build/issues/4558
- [x] test_api_consistency.py **(no pruning required)** @sven6002
- [x] test_api_convert.py **(no pruning required)** @sven6002
- [x] test_api_debug.py
    - [x] https://github.com/conda/conda-build/pull/4701
    - [x] https://github.com/conda/conda-build/pull/4757
- [x] https://github.com/conda/conda-build/pull/4724
- [x] https://github.com/conda/conda-build/pull/4747
- [x] https://github.com/conda/conda-build/pull/4736
- [x] https://github.com/conda/conda-build/pull/4748
- [x] test_api_test.py **(no pruning required)** @kenodegard
- [x] https://github.com/conda/conda-build/pull/4721
- [x] https://github.com/conda/conda-build/pull/4746
- [x] test_check.py **(no pruning required)** @kenodegard
- [x] https://github.com/conda/conda-build/pull/4749
- [x] test_conda_interface.py **(no pruning required)** @kenodegard
- [x] test_config.py **(no pruning required)** @kenodegard
- [x] https://github.com/conda/conda-build/pull/4729
- [x] https://github.com/conda/conda-build/pull/4741
- [x] https://github.com/conda/conda-build/pull/4739
- [x] test_environ.py **(no pruning required)** @kenodegard
- [x] test_index.py **(no pruning required, to be deprecated in favor of conda-index)** @kenodegard
- [x] test_inspect.py **(no pruning required)** @kenodegard
- [x] https://github.com/conda/conda-build/pull/4744
- [x] https://github.com/conda/conda-build/pull/4683
- [x] https://github.com/conda/conda-build/pull/4745
- [x] https://github.com/conda/conda-build/pull/4728
[...]

PR #4720, opened 2023-01-17
Prune/update `test_variant.py` file

beeankha:
Related to:
- https://github.com/conda/conda-build/issues/4590

While updating `conda-build/tests/test_variants.py`, the following issues were discovered and filed:
- https://github.com/conda/conda-build/issues/4717
- https://github.com/conda/conda-build/issues/4718

### DIFF
--- a/tests/test_variants.py
+++ b/tests/test_variants.py
@@ -10,6 +10,7 @@ import sys
 import pytest
 import yaml
 
+from conda.common.compat import on_mac
 from conda_build import api, exceptions
 from conda_build.variants import (
     combine_specs,
@@ -150,7 +151,7 @@ def test_no_satisfiable_variants_raises_error():
 def test_zip_fields():
     """Zipping keys together allows people to tie different versions as sets of combinations."""
     v = {'python': ['3.7', '3.10'], 'vc': ['9', '14'], 'zip_keys': [('python', 'vc')]}
-    ld = variants.dict_of_lists_to_list_of_dicts(v)
+    ld = dict_of_lists_to_list_of_dicts(v)
     assert len(ld) == 2
     assert ld[0]['python'] == '3.7'
     assert ld[0]['vc'] == '9'
@@ -159,7 +160,7 @@ def test_zip_fields():
 
     # allow duplication of values, but lengths of lists must always match
     v = {'python': ['3.7', '3.7'], 'vc': ['9', '14'], 'zip_keys': [('python', 'vc')]}
-    ld = variants.dict_of_lists_to_list_of_dicts(v)
+    ld = dict_of_lists_to_list_of_dicts(v)
     assert len(ld) == 2
     assert ld[0]['python'] == '3.7'
     assert ld[0]['vc'] == '9'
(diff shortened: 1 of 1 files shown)


tests/test_variants.py
```diff
@@ -151,21 +151,21 @@ def test_no_satisfiable_variants_raises_error():
...
-    v = {'python': ['2.7', '3.5'], 'vc': ['9', '14'], 'zip_keys': [('python', 'vc')]}
+    v = {'python': ['3.7', '3.10'], 'vc': ['9', '14'], 'zip_keys': [('python', 'vc')]}
     ld = variants.dict_of_lists_to_list_of_dicts(v)
     assert len(ld) == 2
-    assert ld[0]['python'] == '2.7'
+    assert ld[0]['python'] == '3.7'
     assert ld[0]['vc'] == '9'
-    assert ld[1]['python'] == '3.5'
+    assert ld[1]['python'] == '3.10'
     assert ld[1]['vc'] == '14'
```
kenodegard, line 160: The actual package used here isn't important so let's abstract it. I also suggest changing the imports to explicitly import stuff from the `conda_build.variants` (instead of importing it as `variants`):

```suggestion
    variants = {'packageA': ['1.2', '3.4'], 'packageB': ['5', '6'], 'zip_keys': [('pacakgeA', 'packageB')]}
    zipped = dict_of_lists_to_list_of_dicts(v)
    assert len(zipped) == 2
    assert zipped[0]['packageA'] == '1.2'
    assert zipped[0]['packageB'] == '5'
    assert zipped[1]['packageA'] == '3.4'
    assert zipped[1]['packageB'] == '6'
```

tests/test_variants.py
```diff
@@ -281,6 +281,7 @@ def test_ensure_valid_spec_on_run_and_test(testing_workdir, testing_config, capl
     assert "Adding .* to spec 'pytest-mock  1.6'" not in text
 
 
+@pytest.mark.skipif(sys.platform == "darwin", reason="Unsatisfiable dependencies for M1 MacOS: {'bzip2=1.0.6'}")
```
kenodegard, line 284: We don't want to skip these because they still work on Intel macOS, either explicitly test for M1 machines or mark them as `xfail`:

```suggestion
@pytest.mark.skipif(on_mac and platform.machine() == "arm64", reason="Unsatisfiable dependencies for M1 MacOS: {'bzip2=1.0.6'}")
```

tests/test_variants.py
```diff
@@ -176,7 +176,7 @@ def test_validate_spec():
     """
     spec = {
         # normal expansions
-        "foo": [2.7, 3.7, 3.8],
+        "foo": [3.7, 3.8],
```
kenodegard, line 179: Let's abstract this:

```suggestion
        "foo": [1.2, 3.4],
```

I also think this test should be reworked into a small test using parameterization

tests/test_variants.py
beeankha: It looks like this is being done in PR #4714 , so I'll leave this bit alone

sven6002 (approved):
this is a good idea

tests/test_variants.py
kenodegard: this wasn't actually done in the other PR

kenodegard (approved):

PR #4725, opened 2023-01-18
Refactor `test_published_examples.py` to use test parameterization

kenodegard:
### Description

Use test parameterization and avoid unnecessary string splitting

Xref https://github.com/conda/conda-build/issues/4590

### Checklist - did you ...


tests/test_published_examples.py
```diff
@@ -1,37 +1,36 @@
...
-    return os.path.join(published_examples, request.param)
+    check_call_env([conda_path, "skeleton", "pypi", "click"])
+    check_call_env([conda_path, "build", "click"])
 
 
-# This tests any of the folders in the test-recipes/published_code folder that don't start with _
 @pytest.mark.sanity
+@pytest.mark.parametrize(
+    "recipe",
+    [
```
travishathaway, line 26: Really not a big fan of these chonky inline list comprehensions. Could we at least assign it to a module level variable so we can better express the semantic meaning of it? I know you added a comment, which is nice, but maybe instead of have the comment appear in the middle of the list comprehension, it could appear above the variable assignment.

I'm not going to be nit-picky and block on this though. Please let me know what you think.

kenodegard: hehe and I am not a fan of assigning temporary variables that clutter the namespace, especially those that are only used once

I agree that the comment should be moved so I'll take another look

beeankha (dismissed):

beeankha (approved):

PR #4737, opened 2023-01-20
Updating `test_subpackages.py`

beeankha:
Related to:
- https://github.com/conda/conda-build/issues/4590

### Checklist - did you ...


tests/test_subpackages.py
```diff
@@ -91,7 +91,7 @@ def test_subpackage_variant_override(testing_config):
     assert len(outputs) == 3
 
 
-def test_intradependencies(testing_workdir, testing_config):
```
kenodegard, line 94: the `testing_workdir` creates a temporary directory **and** `cd`s into that directory so I don't think it should be removed here

kenodegard (changes requested):

tests/test_subpackages.py
kenodegard: on a second review, it doesn't look like this test puts anything into the cwd anyhow so this won't be necessary after all

kenodegard (approved):

PR #4771, opened 2023-02-13
Added missing activate_in_script  in allowed FIELDS

npanpaliya:
### Description
`activate_in_script` is missing from `FIELDS` because of which `conda_build.api.check` was failing for recipes having `activate_in_script` in the build section.

### Checklist - did you ...

### DIFF
--- a/conda_build/metadata.py
+++ b/conda_build/metadata.py
@@ -494,6 +494,7 @@ FIELDS = {
         "error_overdepending": None,
         "error_overlinking": None,
         "overlinking_ignore_patterns": [],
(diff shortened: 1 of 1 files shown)


npanpaliya:
I've signed the CLA.

kenodegard (approved):

kenodegard:
The tests are failing because setting the field type to `bool` changes the default value for `build/activate_in_script` to `False`. Without setting the field type the default value is `None` which translates to `True` in `MetaData.activate_in_script`.

PR #4798, opened 2023-03-09
Delegate `get_build_index` to external conda-index package

dholth:
### Description

This is a less ambitious "depend on conda-index" PR that should be able to follow deprecation.

In our large production repository, new conda-index is on the order of 10 times faster than the code that is currently inside conda-build. It is mostly a rewrite of the older code. The new code is faster and simpler, removing complications intended to work around performance problems in the version that is currently embedded inside conda-build.

Since conda-index removes a couple of features, we can't replace the end user script "conda-build index" a.k.a. "conda index". Instead, only modify `get_build_index` and retain the implementation of those scripts. Then we can mark everything else in `conda_build.index`, besides `get_build_index` as deprecated and remove it next year.

We don't have deprecation helpers for conda-build yet?

### Checklist - did you ...


dholth:
https://github.com/conda/conda-build/pull/4828 instead.

PR #4763, opened 2023-01-31
Install downstream packages in correct `subdir`

kenodegard:
### Description

Xref https://github.com/conda/conda-build/issues/4750

### Checklist - did you ...

- [ ] Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [ ] Add / update necessary tests?
- [ ] Add / update outdated documentation?

Issue #4750, opened 2023-01-24
Downstream tests download to wrong subdir (breaks boa+libsolv>=0.7.23)


mbargull:
### What happened?

See https://github.com/conda/conda-index/issues/89 and linked issues.

> This comes from https://github.com/openSUSE/libsolv/commit/cc2da2e789f651b2d0d55fe31c258426bf9e984d#diff-2f0af950e6d2249e0340147b80f5b8194de5362f23026ff65cdb65dfa57696aaR314-R320 , i.e., a newly added strict check in `libsolv=0.7.23` update.
That check is probably reasonable/good -- but `conda-build` is not so strict and dumps the to-be-downstream-tested packages in the current subdir.

In https://github.com/conda-forge/conda-feedstock/pull/197#issuecomment-1400851029 the `noarch` package `conda-smithy` gets downloaded to `conda-build/linux-64` which is then indexed and thus results in a `repodata.json` for `linux-64` with a `noarch` package.

----

I didn't dig deep enough, but I guess somewhere around
https://github.com/conda/conda-build/blob/3.23.3/conda_build/build.py#L3144
-> https://github.com/conda/conda-build/blob/3.23.3/conda_build/render.py#L311
is the culprit.

mbargull:
Until this bug is fixed, https://github.com/conda-forge/boa-feedstock/pull/61 should act as a workaround.

dholth:
conda-build intentionally combines noarch and native arch packages in an index here https://github.com/conda/conda-build/blob/main/conda_build/index.py#L164-L165

kenodegard:
It looks like this is unrelated to `libsolv` but is rather some other change in `mamba` that started to check whether the package's `subdir` matches that of the `repodata.json`'s `subdir` (I haven't pinpointed the exact change).

Conda-build rearchives/copies the downstream test's package from the package chache into the `$CONDA_PREFIX/conda-bld` index and uses the host `subdir`. I believe it happens in this hack (see line 220):

https://github.com/conda/conda-build/blob/44defeda1c4cf252a42c07dbd813a03bcc500a51/conda_build/render.py#L202-L223

See #4763 for a potential fix

kenodegard:
I now understand that the `subdir` checks on `repodata.json` was added to `libsolv` (not `mamba`). Either way, the investigations done above and the [potential fix](https://github.com/conda/conda-build/pull/4763) is valid.

jezdez:
[...]

PR #4803, opened 2023-03-13
Add tests for #4763

kenodegard:
### Description

Adds missing tests for #4750, #4763.

This adds tests for the 5 different use cases for `conda_build.render.find_pkg_dir_or_file_in_pkgs_dirs`:
1. package not found
1. package tarball found
1. pacakge directory found
1. package directory found, generate on demand tarball
1. package directory found, generate on demand tarball, place in different subdir from host

Took the opportunity to also refactor `conda_build.render.find_pkg_dir_or_file_in_pkgs_dirs` to use `pathlib`.

### Checklist - did you ...

### DIFF
--- a/conda_build/render.py
+++ b/conda_build/render.py
@@ -1,5 +1,7 @@
 # Copyright (C) 2014 Anaconda, Inc
 # SPDX-License-Identifier: BSD-3-Clause
+from __future__ import annotations
+
 from collections import OrderedDict, defaultdict
 from functools import lru_cache
 import json
@@ -8,7 +10,6 @@ from os.path import isdir, isfile, abspath
 from pathlib import Path
 import random
 import re
-import shutil
 import string
 import subprocess
 import sys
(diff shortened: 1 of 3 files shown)


conda_build/render.py
```diff
@@ -200,37 +201,36 @@ def _filter_run_exports(specs, ignore_list):
...
-            pkg_loc = pkg_file
-            # create the tarball on demand.  This is so that testing on archives works.
-            with tarfile.open(pkg_file, 'w:bz2') as archive:
-                for entry in os.listdir(pkg_dir):
-                    archive.add(os.path.join(pkg_dir, entry), arcname=entry)
-
-            # use the package's subdir
+def find_pkg_dir_or_file_in_pkgs_dirs(
+    distribution: str, m: MetaData, files_only: bool = False
+) -> str | None:
```
jezdez, line 206: I don't think this can be used in the return type, but only in argument declarations. 

```suggestion
) -> Optional[str]:
```

kenodegard: per the docs the union type is allowed anywhere we used to say `Optional`:
- https://docs.python.org/3/library/typing.html#typing.Optional
- https://docs.python.org/3/library/stdtypes.html#types-union

jezdez (approved):

PR #4804, opened 2023-03-14
Update setup.py

johnnynunez:
up to 3.8 instead of 3.6


johnnynunez:
@kenodegard python <3.7 support is deleted

PR #4816, opened 2023-03-21
Add support for defining `_CONDA_BUILD_ISOLATED_ACTIVATION` in `meta.yaml`

kenodegard:
### Description

The `_CONDA_BUILD_ISOLATED_ACTIVATION` environment variable is special and can currently only be set prior to running `conda build`, (e.g. `_CONDA_BUILD_ISOLATED_ACTIVATION=1 conda build ...`). This makes it hard to enable for packaging teams that do not normally expose this kinds of customizations. So instead of exclusively relying on a pre-set environment variable, let's also allow the value to be set within `meta.yml`'s `build/script_env`.

### Checklist - did you ...

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -1809,11 +1809,11 @@ bundlers = {
 }
 
 
-def _write_sh_activation_text(file_handle, m):
+def _write_sh_activation_text(file_handle, m: MetaData):
     cygpath_prefix = "$(cygpath -u " if utils.on_win else ""
     cygpath_suffix = " )" if utils.on_win else ""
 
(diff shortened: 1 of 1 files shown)


kenodegard:
To indicate that its private?

jezdez:
I'm not sure this should be "private" if it becomes part of the recipe format. Are there other ways to add this as a first-level recipe option?

kenodegard:
@jezdez I don't want this environment variable to become an official `meta.yaml` setting. IMO isolated mode should be the only mode conda is ever invoked as. We also need to work to eliminate the need to run conda via subprocess in the first place. Furthermore, there needs to be proper API end-points that conda-build should call instead.

kenodegard:
we are collectively displeased with this solution and should consider something like https://github.com/conda-forge/conda-feedstock/pull/204 instead

PR #4817, opened 2023-03-21
Require the source when rendering if load_file_data is used

### LINKED ISSUE #4807: `load_file_data` depends on textual occurrence of certain strings in meta.yaml
### Checklist

### What happened?

I have just spent an hour or two trying to figure out why conda build couldn't find my pyproject.toml that I was trying to load with [`load_file_data`](https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#loading-data-from-other-files).

I believe the culprit is [this](https://github.com/conda/conda-build/blob/7e9d1231057a6cae01d3116d399d9d52717cff21/conda_build/metadata.py#L1917) and the following few methods.

It seems that metadata.py has logic that looks for the occurrence of the literal string "load_setup_py_data" in the meta.yaml, and if it finds it, triggers some kind of special behavior related to making files in the source directory available later so that things like `load_file_data` can have access to them.  There is similar logic for "load_setuptools" and "load_file_regex" . . . but not for "load_file_data".

So if you try to use `load_file_data` on its own in your meta.yaml, it will give a "no such file error".  Of course, it should be possible to do this, since nowadays some projects are moving to using only pyproject.toml and no setup.py.

A quick fix would be to add a similar check for "load_file_data" to the strings that are already being checked.  However, the whole idea of triggering different behavior based on the occurrence of specific seems a bit fragile.  It leads to highly surprising behavior, since it means things will magically "work" regardless of where the trigger string occurs, even if it's in a comment or surrounded by other text (and hence magically break if the trigger string is removed).  So if your meta.yaml looks like this:

```
{% set pyproject = load_file_data('pyproject.toml') %}

package:
  name: testpack
  version: {{ pyproject['project']['version'] }}

source:
  path: ..

requirements:
  build:
    - python
    - setuptools
  run:
    - python

# whatever_you_do_don't_load_setup_py_dataxxx!!!
```

. . . then it will work, but removing the comment at the end will break it.

I don't know enough about how conda-build works to know what behavior exactly is being triggered by these string checks, or whether there's another way to do it, but it would be nice if a more robust solution could be found to avoid this kind of thing.

### Conda Info

```shell
active environment : None
            shell level : 0
       user config file : /home/brenbarn/.condarc
 populated config files : /home/brenbarn/.condarc
          conda version : 23.1.0
    conda-build version : 3.23.3
         python version : 3.9.16.final.0
       virtual packages : __archspec=1=x86_64
                          __glibc=2.35=0
                          __linux=5.15.0=0
                          __unix=0=0
       base environment : /home/brenbarn/miniconda3  (writable)
      conda av data dir : /home/brenbarn/miniconda3/etc/conda
  conda av metadata url : None
          package cache : /home/brenbarn/miniconda3/pkgs
                          /home/brenbarn/.conda/pkgs
       envs directories : /home/brenbarn/miniconda3/envs
                          /home/brenbarn/.conda/envs
               platform : linux-64
             user-agent : conda/23.1.0 requests/2.28.2 CPython/3.9.16 Linux/5.15.0-60-generic ubuntu/22.04.1 glibc/2.35
                UID:GID : 1000:1000
             netrc file : None
           offline mode : False
```

### Conda Config

```
==> /home/brenbarn/.condarc <==
auto_activate_base: False
```

jjhelmus:
### Description

Require the source to be downloaded and available if the `load_file_data` Jinja2 function is used in the recipe's meta.yaml.

closes #4807 

### Checklist - did you ...

### DIFF
--- a/conda_build/metadata.py
+++ b/conda_build/metadata.py
@@ -1930,11 +1930,20 @@ class MetaData:
         return "load_file_regex" in meta_text
 
     @property
+    def uses_load_file_data_in_meta(self):
+        meta_text = ""
+        if self.meta_path:
+            with open(self.meta_path, "rb") as f:
+                meta_text = UnicodeDammit(f.read()).unicode_markup
+        return "load_file_data" in meta_text
+
+    @property
     def needs_source_for_render(self):
         return (
             self.uses_vcs_in_meta
             or self.uses_setup_py_in_meta
             or self.uses_regex_in_meta
+            or self.uses_load_file_data_in_meta
         )
 
     @property
--- a/news/4817-require-source-when-load-file-data-used
+++ b/news/4817-require-source-when-load-file-data-used
@@ -1,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Require the source when rendering a recipe that uses the load_file_data function (#4817, fixes #4807)
+
+### Deprecations
+
+* <news item>
+
+### Docs
+
+* <news item>
+
+### Other
+
+* <news item>
--- a/tests/test-recipes/metadata/jinja_load_toml_from_source/meta.yaml
+++ b/tests/test-recipes/metadata/jinja_load_toml_from_source/meta.yaml
@@ -1,0 +1,8 @@
+{% set pyproject = load_file_data("pyproject.toml") %}
+
+package:
+  name: example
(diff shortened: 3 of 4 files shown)


kenodegard (approved):
Looks good, I'll keeping an eye on the tests and rerunning the flaky ones

jezdez (approved):

PR #4826, opened 2023-03-24
`create_test.py`: Add type hints and refactor to use pathlib

kenodegard:
### Description

`create_test.py` is where a package's test scripts are generated. Here we add type hints and start to replace `os.path` logic with `pathlib`. Also update docs to reflect what has been happening in the code.

### Checklist - did you ...


beeankha (dismissed):

conda_build/create_test.py
```diff
@@ -3,44 +3,42 @@
...
+from pathlib import Path
 
-from conda_build.utils import copy_into, ensure_list, glob, on_win, rm_rf
+from .metadata import MetaData
+from .utils import copy_into, ensure_list, on_win, rm_rf
 
 
-def create_files(m, test_dir=None):
-    """
-    Create the test files for pkg in the directory given.  The resulting
```
dholth, line 17: Where are the test files created now?

kenodegard: not sure what you mean? the docstring was updated and now reads:

```
Copy all test files from recipe over into testing directory.
```

dholth: Doesn't this call some other functions that create e.g. a Python script that tries to import things the recipe asks for, and then copies them over?

beeankha (approved):

conda_build/create_test.py
kenodegard: No, `create_files` doesn't _create_ any new files, it only copies existing ones. `create_python_files` (and similarly named functions for other supported languages) are where the test scripts you're referencing are created.

PR #4828, opened 2023-03-27
Use conda-index dependency instead of obsolete built-in index

### LINKED ISSUE #4645: Replace conda_build/index.py with standalone conda-index

dholth:
### Description

This replaces conda-build's index with conda-index. Fix #4645 .

Standalone conda-index caches package metadata to sqlite and fetches everything necessary for repodata.json in a single query, instead of slowly opening thousands or hundreds of thousands of individual small files. It is simpler and on the order of 10 times faster in production compared to the older `conda-build index`. Unlike the older code, it no longer reads from the previous repodata.json trying to only replace metadata related to changed packages. Instead, the data only flows in one direction from packages -> metadata cache -> repodata.json.

We have tried several strategies. This code simply keeps the old implementation for users of `conda-build index` aka `conda index` while printing many warnings to use the new package. Instead of worrying about a deprecation schedule for the missing `file` flag this code anticipates completely removing the old command and API.

Most users get the benefit automatically when conda-build uses new-conda-index internally as a side effect of doing builds.

I think it is likely that very few users need to run conda-index themselves rather than relying on hosting.

### Checklist - did you ...


conda_build/cli/main_index.py
```diff
@@ -90,8 +90,8 @@ def execute(args):
 
     api.update_index(args.dir, check_md5=args.check_md5, channel_name=args.channel_name,
                      threads=args.threads, subdir=args.subdir, patch_generator=args.patch_generator,
-                     verbose=args.verbose, progress=args.progress, hotfix_source_repo=args.hotfix_source_repo,
-                     current_index_versions=args.current_index_versions_file, index_file=args.file)
```
dholth, line 94: it might be nice to print a warning if you pass file

conda_build/index.py
```diff
@@ -107,9 +57,18 @@ def map(self, func, *iterables):
...
+    bldpkgs_dir,
+    output_folder=None,
+    clear_cache=False,
+    omit_defaults=False,
+    channel_urls=None,
+    debug=False,
+    verbose=True,
+    locking=None,
+    timeout=None,
+):
```
dholth, line 71: add comment "this is confusing" may exist in other PR's

conda_build/index.py
```diff
@@ -226,1195 +200,46 @@ def get_build_index(subdir, bldpkgs_dir, output_folder=None, clear_cache=False,
...
+    progress=False,
+    hotfix_source_repo=None,
+    subdirs=None,
+    warn=True,
+    current_index_versions=None,
+    debug=False,
+    index_file=None,
+):
+    # conda-build calls update_index on a single subdir internally, but
+    # conda-index expects to index every subdir under dir_path
```
dholth, line 225: this is possible used as an api by conda_build's index cli (although I would rather you use conda-index's new CLI with added features)

conda_build/index.py
```diff
@@ -61,31 +31,6 @@ def map(self, func, *iterables):
...
-        "lua": "lua",
-        "nodejs": "js",
-        "perl": "perl",
-        "php": "php",
-        "ruby": "ruby",
-        "m2-base": "m2",
-        "msys2-conda-epoch": "m2w64",
-    }
-    NAMESPACE_PACKAGE_NAMES = frozenset(NAMESPACES_MAP)
-    NAMESPACES = frozenset(NAMESPACES_MAP.values())
```
dholth, line 87: we could keep these maybe

conda_build/render.py
```diff
@@ -2,36 +2,43 @@
 # SPDX-License-Identifier: BSD-3-Clause
```
dholth, line 2: we might be able to pull these changes into a separate PR or rebase with the reformat

conda_build/utils.py
```diff
@@ -1404,6 +1404,8 @@ class LoggingContext:
         "conda_build.index",
         "conda_build.noarch_python",
         "urllib3.connectionpool",
+        "conda_index",
+        "conda_index.index",
```
dholth, line 1408: this affects logging for those module

tests/test_index.py
```diff
@@ -1,1198 +0,0 @@
-# Copyright (C) 2014 Anaconda, Inc
```
dholth, line 1: 10% less test time

dholth: The other conda-build tests that build packages, wind up calling get_build index and indirectly test the indexing code. Standalone conda-index has its own faster version of these tests that we delete here.

jezdez (dismissed):
I like this PR also best.

jezdez:
@dholth Could you rebase on top of main please?

dholth:
> I like this PR also best.

What is best

jezdez:
> 

I mean compared to the other options to deprecate the current index subcommand, #4798 and #4841.

recipe/meta.yaml
```diff
@@ -28,7 +28,12 @@ requirements:
   run:
     - beautifulsoup4
     - chardet
-    - conda >=4.13
+    - conda  >=4.13
+    - conda-index
+    # - conda-verify  >=3.0.2  # optional as of CB 3.12.0
+    - contextlib2   # [py<34]
+    - enum34        # [py<34]
+    - pathlib2      # [py<3]
```
travishathaway, line 36: Are these lines really necessary? 😅 

I didn't know we had to support Python `<3.4`...

conda_build/cli/main_index.py
```diff
@@ -85,7 +87,7 @@ def parse_args(args):
     p.add_argument(
         "-f",
         "--file",
-        help="A file that contains a new line separated list of packages to add to repodata.",
+        help="A file that contains a new line separated list of packages to add to repodata.  Deprecated, will be removed in a future version of conda build.",
```
travishathaway, line 90: I know that conda-build hasn't officially adopted the [Conda Release Schedule](https://github.com/conda-incubator/ceps/blob/main/cep-8.md) but I still think it would be a good idea to provide a version number when this will be removed. Otherwise, it probably never will 🤷.

travishathaway (commented):
I like the direction that this pull request is going in, but I think we need to be more specific about when conda-index will be officially removed from conda-build.

Eventually I would like to see the this command removed from the CLI of conda-build and replaced by adding conda-index as a proper subcommand plugin (https://github.com/conda/conda-index/issues/81).

Is there a way that we can add our intention to do this either via the release notes or in the documentation? It doesn't have to be completed in this pull request, and I would be more than happy to help out.

conda_build/cli/main_index.py
```diff
@@ -96,6 +98,11 @@ def parse_args(args):
 def execute(args):
     _, args = parse_args(args)
 
+    if args.file:
+        warnings.warn("--file is no longer implemented.")
+    if args.hotfix_source_repo:
+        warnings.warn("--hotfix-source-repo is no longer implemented.")
+
```
jaimergp, line 105: These aren't available in the new conda-index, right? Are there equivalent functionalities? Should people be warned with a `DeprecationWarning` before we do remove these?

jaimergp (commented):
Thanks Daniel!

I have some questions:

- What's the story for users of the `conda index` subcommand in this PR? Are they silently being upgraded to the `conda-index` rewrite? If that's the case, what backwards incompatible changes should they expect?
- Same but for users of `conda_build.api`.

conda_build/index.py
```diff
@@ -40,6 +39,7 @@
 #  BAD BAD BAD - conda internals
 from conda.core.subdir_data import SubdirData
 from conda.models.channel import Channel
+from conda_index.index import update_index as _update_index
```
jaimergp, line 42: Users have been silently updated to the new `conda-index` via this import. Have we considered deprecating this module first with a warning? What's the story in the future? Will `conda_build.index` always exist? If not, should we add a module-level warning (bis)?

dholth: How do we add a module warning without triggering it when conda-build imports

conda_build/cli/main_index.py
dholth: `--hotfix` was removed before I started working on conda-build and `--file` isn't implemented in new conda-index. The replacement is that new conda-index is fast.

recipe/meta.yaml
dholth: No I didn't touch the recipe as part of this PR.

conda_build/cli/main_index.py
dholth: > Thanks Daniel!
> 
> I have some questions:
> 
> * What's the story for users of the `conda index` subcommand in this PR? Are they silently being upgraded to the `conda-index` rewrite? If that's the case, what backwards incompatible changes should they expect?
> * Same but for users of `conda_build.api`.

Yes, everyone is upgraded to the new conda-index with the possible exception of people who import parts of this file that we have left in for backwards compatibility to follow a deprecation policy that this project has not officially adopted.

Users should expect equivalent indexing with a sqlite cache instead of a many-small-files cache.

In this PR we try to map all the API calls to forwarded calls to new-conda-index.

conda_build/index.py
jezdez: I think the goal would be eventually to replace this code completely with the implementation in conda-index via an own subcommand. There are two groups of users here:

- End-users of the index subcommand, they can continue to use it, and they shouldn't see any difference to the built-in index feature.
- Makers who import from `conda_build.index` in their scripts/tools and rely on the Python API. They'll need to be provided with enough information and a rollout plan to make the necessary changes.

I only worry about the second group, and how to reach them. They'll unlikely read the conda-build changelog, so we'll need to let them know via `DeprecationWarning` instead. I don't think it has to be a module-level warning, maybe a function wrapper like `conda.deprecations` would suffice.

jezdez:
> I like the direction that this pull request is going in, but I think we need to be more specific about when conda-index will be officially removed from conda-build.
> 
> Eventually I would like to see the this command removed from the CLI of conda-build and replaced by adding conda-index as a proper subcommand plugin ([conda/conda-index#81](https://github.com/conda/conda-index/issues/81)).

💯 

> Is there a way that we can add our intention to do this either via the release notes or in the documentation? It doesn't have to be completed in this pull request, and I would be more than happy to help out.

I think there is a way to do this, providing a deprecation warning for importers of the `conda_build.index` module (or its functions etc) and an additional note about the removal in N numbers of releases.

I'd suggest in 2 releases (a fast version of CEP 9), e.g. in 3.27.0 by current nomenclature, which would coincide with the conda deprecation release 23.9.0.

jezdez:
Bonus points if we could do the same for conda skeleton 😬 But maybe that's asking for too much!

dholth:
We need to finish this PR. Somehow it has been delayed at least since January over 

> Bonus points if we could do the same for conda skeleton 😬 But maybe that's asking for too much!

@travishathaway if you want to make conda-index a cli plugin, PR's are welcome against https://github.com/conda/conda-index although I'm not sure how to resolve the redesigned click-CLI differences (we thought about giving it a different name but haven't found the right synonym for 'index' ('index2'?)

dholth:
I've updated this PR to leave the `conda-build index` command aka `conda index` alone; it will use the old implementation. conda's deprecation handler is copied in to avoid depending on very new conda.

Only conda-build's own indexing should use the new implementation. That way normal use gets the benefit and API or CLI use gets compatible behavior in case the `--file` feature is needed.

Users without those unusual requirements should use standalone `conda-index` for fast indexing, repodata.json.zst support etc.

conda_build/index.py
dholth: Updated to expose all the old API's with the old implementation, while printing warnings.

conda_build/build.py
```diff
@@ -37,7 +37,7 @@
 from conda_build.config import Config
 from conda_build.create_test import create_all_test_files
 from conda_build.exceptions import CondaBuildException, DependencyNeedsBuildingError
-from conda_build.index import get_build_index, update_index
+from conda_build.index import _delegated_update_index, get_build_index
```
dholth, line 40: better name welcome

conda_build/cli/main_index.py
```diff
@@ -96,6 +99,10 @@ def parse_args(args):
 def execute(args):
     _, args = parse_args(args)
 
+    warnings.warn(
+        "conda-build index is deprecated. Use standalone conda-index package."
```
jezdez, line 103:
```suggestion
        "conda-build index is deprecated. Use the standalone conda-index package instead."
```

news/4645-use-conda-index
```diff
@@ -0,0 +1,22 @@
...
+### Bug fixes
+
+* <news item>
+
+### Deprecations
+
+* `conda-build index --file` is no longer supported.
+* `conda-build index --hotfix-source-repo` is removed.
+* Prefer `python -m conda_index`, instead of `conda-build index` or `conda
+  index`, to use faster indexing code.
```
jezdez, line 14: I think the users should prefer the use of the index subcommand long-term, when conda-index has implemented a plugin based subcommand. In the meantime let's not send them down the `-m conda_index` rabbit hole here in the changelog, and instead point them to the conda-index docs. That'll allow us to have one source of truth about how to interact with conda-index long-term, its own docs.

```suggestion
* Prefer the [standalone conda-index package](https://conda.github.io/conda-index/),
  instead of `conda-build index` or `conda index`, to use faster indexing code.
```

recipe/meta.yaml
jezdez, line 36:
```suggestion
```

recipe/meta.yaml
```diff
@@ -28,7 +28,12 @@ requirements:
   run:
     - beautifulsoup4
     - chardet
-    - conda >=4.13
+    - conda  >=4.13
```
jezdez, line 31:
```suggestion
    - conda >=4.13
```

tests/requirements.txt
```diff
@@ -3,7 +3,8 @@
 anaconda-client
 beautifulsoup4
 chardet
-conda >=4.13
+conda >= 4.13
```
jezdez, line 6:
```suggestion
conda >=4.13
```
IIRC matchspec parsing can be brittle, was this change intentional?

jezdez (approved):
I like the idea of us setting a clear target on what comes next when we actually remove the index code from conda-build: a conda-build 4.0.0 release. It's a backward-incompatible change to remove the index code, no matter the amount of warning we do ahead of time.

jaimergp:
> I'm not sure how to resolve the redesigned click-CLI differences

@dholth, is this about potential conflicts with `conda-build`'s `conda index` command? I guess we can go the packaging route and use a `run_constrained` rule to make sure there are now CLI collisions. AKA a subcommand-providing `conda-index` version will only be compatible with a conda-build that has dropped the `conda index` subcommand.

tests/requirements.txt
dholth: No

dholth:
> > I'm not sure how to resolve the redesigned click-CLI differences
> 
> @dholth, is this about potential conflicts with `conda-build`'s `conda index` command? I guess we can go the packaging route and use a `run_constrained` rule to make sure there are now CLI collisions. AKA a subcommand-providing `conda-index` version will only be compatible with a conda-build that has dropped the `conda index` subcommand.

This is about updating shell scripts that run conda index.

news/4645-use-conda-index
dholth: A reminder that `conda index` is currently an alias for `conda-build index`

news/4645-use-conda-index
```diff
@@ -0,0 +1,22 @@
...
+* Depend on standalone conda-index instead of bundled indexing code. (#4690)
+
+### Bug fixes
+
+* <news item>
+
+### Deprecations
+
+* `conda-build index --file` is no longer supported.
+* `conda-build index --hotfix-source-repo` is removed.
```
dholth: Now we anticipate removing the entire `conda index` instead of removing the two features not implemented in standalone `conda-index`

tests/test_api_consistency.py
```diff
@@ -214,4 +218,5 @@ def test_api_update_index():
         False,
         None,
         None,
+        inspect._empty,
```
dholth, line 221: It looks like we aren't too far off from the `getfullargspec` version, `parameters` appears to include the positional and `**kwargs` parameters. `getfullargspec()` doesn't return anything for a `@wraps` function...

conda_build/deprecations.py
```diff
@@ -0,0 +1,275 @@
+# Copyright (C) 2014 Anaconda, Inc
```
travishathaway, line 1: Is there a reason we're not importing this directly from `conda`?

travishathaway: Thanks for clarifying 👌.

travishathaway (approved):
Seems reasonable enough. My only reservation about the prior implementation was not having a clear deprecation path (i.e. which version we remove it in), but it looks like we've addressed that with the inclusion of our handy deprecation decorator.

kenodegard (approved):
I like this and think it strikes a good enough balance between deprecating/providing backports and switching to the new conda-index codes

My only worry is that conda-build 4.0.0 feels too much like the fabled conda 5.0.0 which never happened but I don't think that should block this

jezdez (approved):

PR #4834, opened 2023-03-29
Drop extra Jinja w/package names from variants.rst

jakirkham:
### Description

AIUI including the package name in Jinja was something that use to be needed, but at some point this was added automatically when matching against a package name. So suggesting we drop this from the doc to simplify things for users learning about this functionality.

### Checklist - did you ...


pentschev:
Thanks @jakirkham !

beeankha:
Hi @jakirkham , thanks for this pull request! Please add a [news file](https://github.com/conda/conda-build/blob/main/news/TEMPLATE) and include information about which version of conda-build we started to be able to drop these variables from (in the news file blurb), if you know that info!

jakirkham:
AFAIK this was never needed in conda-build 3 (when the variants were introduced). Though I could be wrong, in which case it is some very old version of conda-build 3 (and I don't know what version that would be)

sven6002 (dismissed):
Looks good to me..

beeankha (approved):

jakirkham:
Thanks all! 🙏

Anything else needed here?

jezdez (approved):

jakirkham:
Thanks again! 🙏

PR #4840, opened 2023-04-06
Convert `setup.py` to `pyproject.toml` using hatch build system

kenodegard:
### Description

This converts the `setup.py` to `pyproject.toml` using `hatchling` and `hatch-vcs`. Since we no longer rely on `versioneer.py` we remove all related codes (no deprecation since these are build time only codes).

Xref https://github.com/conda/conda/pull/12509

### Checklist - did you ...


recipe/meta.yaml
```diff
@@ -103,10 +101,8 @@ test:
   source_files:
     - tests
 
-outputs:
-  - type: wheel
-  - name: {{ PKG_NAME }}
```
jezdez, line 108: What is this removal?

jezdez (approved):
Looks good to me, even though I feel a little uncertain of not being able to install conda from PyPI, making the dependencies in `pyproject.toml` not resolvable

recipe/meta.yaml
kenodegard: I don't believe this makes any difference.

We don't produce more than one output from this recipe (this counter intuitively suggests we should be?). And the `conda-forge`/`AnacondaRecipes` recipes don't do this.

PR #4841, opened 2023-04-06
Conda index helpful changes from other PR

dholth:
### Description

Some changes from the other attempt. May need to adjust where/whether warnings are printed.

### Checklist - did you ...

### DIFF
--- a/conda_build/utils.py
+++ b/conda_build/utils.py
@@ -1406,6 +1406,7 @@ class LoggingContext:
         "urllib3.connectionpool",
         "conda_index",
         "conda_index.index",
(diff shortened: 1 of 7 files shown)


conda_build/utils.py
```diff
@@ -1406,6 +1406,7 @@ class LoggingContext:
         "urllib3.connectionpool",
         "conda_index",
         "conda_index.index",
+        "conda_index.index.convert_cache",
```
dholth, line 1409: If an existing old-style "many small files" cache exists, it's converted to sqlite. Prints logs, fails tests that want no logs, unless we add it here.

PR #4845, opened 2023-04-07
Enable `xattr` test for macOS

kenodegard:
### Description

Xattr test previously only ran on Linux (if `setfattr` is present), made necessary changes to allow the test to also run on macOS.

Updated `testing_homedir` fixture to use a simpler `tempfile.TemporaryDirectory` implementation.

### Checklist - did you ...

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -5,8 +5,8 @@ import subprocess
 import sys
 import tempfile
 from collections import defaultdict
+from collections.abc import Iterator
 from pathlib import Path
-from typing import Generator
 
 import pytest
 from conda.common.compat import on_mac, on_win
@@ -60,7 +60,7 @@ def testing_workdir(tmpdir, request):
 
 
 @pytest.fixture(scope="function")
-def testing_homedir() -> Generator[Path, None, None]:
+def testing_homedir() -> Iterator[Path]:
     """Create a temporary testing directory in the users home directory; cd into dir before test, cd out after."""
     saved = Path.cwd()
     try:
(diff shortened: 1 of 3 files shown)


tests/conftest.py
```diff
@@ -3,11 +3,13 @@
 import os
 import subprocess
 import sys
+import tempfile
 from collections import defaultdict
 from pathlib import Path
+from typing import Generator
```
travishathaway, line 9: Deprecated since version 3.9: [collections.abc.Generator](https://docs.python.org/3/library/collections.abc.html#collections.abc.Generator) now supports subscripting ([]). See [PEP 585](https://peps.python.org/pep-0585/) and [Generic Alias Type](https://docs.python.org/3/library/stdtypes.html#types-genericalias)  (source: https://docs.python.org/3/library/typing.html?highlight=typing#typing.Generator)

You can also just use `Iterator` too if you don't need to specify the extra types in a `Generator`:

```python
from collections.abc import Iterator

def oh_it_iterates() -> Iterator[str]:
    yield "iter"
    yield "ator"
    yield "!!!"
```

travishathaway (changes requested):
Looks good. I just want to make sure we are not using any of the deprecated types in the `typing` module.

tests/conftest.py
kenodegard: Oh cool! I didn't know this!

kenodegard: hm since this was a change introduced in Python 3.9 I think we should wait until we drop Python 3.8 support before making the switch from `typing` to `collections.abc`

jezdez (approved):

sven6002 (approved):
Looks good to me

Issue #4370, opened 2022-01-27
Python 3.11 support


kenodegard:
### Description

Need to investigate changes necessary to support Python 3.11.

```[tasklist]
- [x] distutils is deprecated in stdlib
- [x] Get Python 3.11 build (even if alpha) to test conda and conda-build
- [x] Review 3.11 release notes to see if other (packaging) related changes are expected
- [ ] https://github.com/conda/conda/issues/11170
- [ ] https://github.com/conda/conda-build/pull/4852
```

jezdez:
@chenghlee Can you ask the packaging team to look into whether we can/should build a Python 3.11 early release package to a separate channel (conda-canary?) to try this out? We could set up the CI system to use the channel while it hasn't officially been released yet to be part of defaults.

kenodegard:
if the package is released as a beta version, it won't be selected for installation unless explicitly requested, so per that understanding, it shouldn't be problematic to release to `main` either? making the beta releases generally available should help other packages with their testing too

kenodegard:
Depends on conda/conda#11170

PR #4852, opened 2023-04-17
Add Python 3.11 support

kenodegard:
### Description

Add Python 3.11 to test matrix and minor test updates.

Xref https://github.com/conda/conda-build/issues/4370

### Checklist - did you ...


jezdez:
pre-commit.ci autofix

.github/workflows/tests.yml
```diff
@@ -63,14 +63,14 @@ jobs:
       fail-fast: false
       matrix:
         # test all lower versions (w/ stable conda) and upper version (w/ canary conda)
-        python-version: ['3.8', '3.9']
+        python-version: ['3.8', '3.9', '3.10']
```
sven6002, line 66: do we still need 3.8.. now..

tests/requirements.txt
```diff
@@ -20,7 +20,6 @@ pkginfo
 psutil
 py  # https://github.com/ContinuumIO/anaconda-issues/issues/13198
 py-lief
-pycrypto
```
sven6002, line 23: does need to change to https://pypi.org/project/pycryptodome/3.11.0 instead?

.github/workflows/tests.yml
kenodegard: yes, Python 3.8 support hasn't been dropped yet so it still needs testing

sven6002: is the plan to drop it for next release round??

kenodegard: at this time we only drop Python 3.8 once it reaches EOL/when support is dropped from the defaults channel

sven6002: makes sense

PR #4873, opened 2023-05-05
check return code when selecting patch type

jjhelmus:
### Description

When selecting between the .native, .lf, and .crlf patches, do not mark the result as successful if a non-zero (failure) return value was returned.

This also allows platforms whose patch commands do not support the `--binary` argument, such as FreeBSD, to use patches in recipes.  

### Checklist - did you ...

### DIFF
--- a/conda_build/source.py
+++ b/conda_build/source.py
@@ -879,6 +879,9 @@ def _get_patch_attributes(
                         result[check_name] = False
                         pass
                     else:
+                        if process.returncode:
+                            result[check_name] = False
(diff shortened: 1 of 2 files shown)


jezdez (changes requested):
Thanks for the patch @jjhelmus, could you add a news item?

jjhelmus:
> Thanks for the patch @jjhelmus, could you add a news item?

Done

jezdez (approved):

PR #4878, opened 2023-05-10
add type annotations created by monkeytype

jjhelmus:
### Description

This PR is for demonstration and discussion, it is unlikely that these changes should be merged as is.

Add type annotations to the conda_build library created by running the test suite with [MonkeyType](https://monkeytype.readthedocs.io/en/stable/index.html).

This steps used to accomplished this were:
```
make setup
conda activate conda-build
python -m pip install monkeytype
monkeytype run -m pytest
monkeytype list-modules | grep conda_build > modules
cat modules | while read mod
do
    echo "Adding types to $mod"
    monkeytype apply --pep_563 $mod
done
```  

### Checklist - did you ...

### DIFF
--- a/conda_build/_load_setup_py_data.py
+++ b/conda_build/_load_setup_py_data.py
@@ -1,17 +1,19 @@
 # Copyright (C) 2014 Anaconda, Inc
 # SPDX-License-Identifier: BSD-3-Clause
+from __future__ import annotations
 import logging
 import os
 import sys
+from typing import TYPE_CHECKING, Dict, List, Optional, Union
 
 
 def load_setup_py_data(
-    setup_file,
-    from_recipe_dir=False,
-    recipe_dir=None,
-    work_dir=None,
-    permit_undefined_jinja=True,
-):
+    setup_file: str,
+    from_recipe_dir: bool=False,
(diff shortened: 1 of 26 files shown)


travishathaway:
Hi @jjhelmus,

Thanks for submitting this. I just wanted to chime in and give my two cents on this 😄. TLDR: I don't think we should worry about adding type hinting to conda-build. Below are my reasons:

1. conda-build has maintainability problems that type hints won't solve. For example there are various examples of terse variable names in the code: https://github.com/conda/conda-build/blob/07a9d43dc24e2bf12b1c61b7d7298582755c5a01/conda_build/build.py#L156
I think we should attempt to fix these first and perhaps add type hints manually at the same time.
2. I am of the opinion that for type hints to truly be useful, we should also be using a type checker like `mypy`. I don't think that's feasible for a project like conda-build given its size that the refactor effort involved to reach a sufficient level of coverage.
3. Adding type hints puts an extra burden on new contributions, and we will need to train reviewers to account for this in their pull request reviews. Not everyone is currently aligned on the need for type hints.

Hope that makes sense. Anyways, thanks for submitting this pull request and starting the conversation. Looking forward to a healthy debate.

PR #4901, opened 2023-05-26
Log the extra-meta data we burn into the package for better testability

dbast:
### Description

This an improvement to #4303, which enabled to burn-in extra meta data into packages. The feature is used to e.g. burn-in CI_RUN_IDs, GIT_URL, GIT_SHA1 into packages to be able to trace them back to specific build runs and commits in the recipe feedstock.

Its currently pretty cumbersome to check if a build run embeds the right extra meta data or not (requires inspecting the resulting package). 

This PR:
* makes it much easier by logging the extra-meta data that is burn-in into the build log.
* makes it much easier to co-relate a package and its build log as both contain the same extra meta data

This particularly helps to test https://github.com/conda-forge/conda-smithy/pull/1577 and also constantly check via future build logs that the right meta data is in place.

pinging also @jaimergp :) 

### Checklist - did you ...

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -1447,7 +1447,8 @@ def write_about_json(m):
         if m.config.extra_meta:
             log = utils.get_logger(__name__)
             log.info(
-                "Adding the following extra-meta data to about.json: %s", m.config.extra_meta
+                "Adding the following extra-meta data to about.json: %s",
+                m.config.extra_meta,
             )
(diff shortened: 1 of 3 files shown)


jaimergp (approved):
LGTM! The Windows failure looks like a race condition unrelated to this.

jezdez (approved):

PR #4921, opened 2023-06-28
Convert conda-build entrypoints into plugins

kenodegard:
### Description

Might as well make conda-build into a proper plugin too!

This introduces a new submodule, `conda.cli.plugins`, where the pluggy hookimpls live. Otherwise minimal changes were made.

This also marks the legacy style plugin interfaces as deprecated, e.g.:

```
conda-build ...
python -m conda_build.cli.main_build ...
```

### Checklist - did you ...


conda_build/cli/plugins.py
jezdez, line 1: I would not name this **plugins** in plural, since in pluggy terminology, conda-build itself is “a plugin” (~a collection of implementations of plugin hooks).

It's better to use `conda_build/plugin.py` (or `conda_build/plugin/subcommands.py` if we have many) to define all hook implementations, that's also what [we went with for CLS](https://github.com/conda/conda-libmamba-solver/blob/main/conda_libmamba_solver/plugin.py).

conda_build/cli/main_build.py
```diff
@@ -601,3 +597,7 @@ def main():
         )
         sys.exit(1)
     return
+
+
+if __name__ == "__main__":
+    main()
```
jezdez, line 603: Do we need this?

jezdez (changes requested):

conda_build/cli/main_build.py
kenodegard: We don't _need_ this, I added it solely to be consistent (only some of the subcommands were previously callable via `python -m`)

but since it's a deprecated interface it probably doesn't make sense to enable the ability now...I'll remove these

jezdez (approved):

jezdez (changes requested):
Wait, this'll require a run constraint for a newer version of conda!

kenodegard:
@jezdez added but technically not necessary? the old interface is still there and the plugin system is only imported in `conda.plugin` which is only invoked via the plugin system itself?

travishathaway:
@kenodegard, I was testing this out locally today, and one thing I noticed is that the commands show up twice on the `conda --help` page. Once in the `conda commands available from other packages:` and another time in the `... (legacy)`.

Is there anything we can do to prevent that? I think that would require a change on the conda side.

jezdez (approved):

kenodegard:
conda-build's various subcommands will continue to be listed twice until conda is modified to exclude duplicates (probably the correct route?) or the executables are deprecated/removed

while the subcommand displaying twice is noisy its not incorrect (the legacy plugins are still there) and IMO isn't enough to block this PR

Issue #4944, opened 2023-07-27
lief crashing parsing .so file on windows


tl-hbk:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### What happened?

Should the call to `_collect_needed_dsos` still take `files` instead of `files_to_inspect`?

https://github.com/conda/conda-build/blob/655dbe1e6c760c95c329dc044ba94f753499b531/conda_build/post.py#L1455-L1462

In this [commit](https://github.com/conda/conda-build/commit/6deb528dd0d5ff76df9709f8a6081fe4f8ed189b) code was added 
to only check against platform specific filetypes but the call to `_collect_needed_dsos` didn't change.

https://github.com/conda/conda-build/blob/6deb528dd0d5ff76df9709f8a6081fe4f8ed189b/conda_build/post.py#L863-L871

https://github.com/conda/conda-build/blob/6deb528dd0d5ff76df9709f8a6081fe4f8ed189b/conda_build/post.py#L930-L940

I've run into an issue where a package I'm working on with a vendored 3rd party dependency has both .pyd and .so files 
and the process crashes when trying to inspect the .so file on windows using py-lief. 

https://github.com/conda/conda-build/blob/655dbe1e6c760c95c329dc044ba94f753499b531/conda_build/os_utils/liefldd.py#L357-L359

I've created a minimal reproduction of the issue with just one of the .pyd and .so files 

### Conda Info

```shell
active environment : None
       user config file : C:\Users\runneradmin\.condarc
 populated config files : C:\Users\runneradmin\.condarc
          conda version : 23.5.2
    conda-build version : 3.26.0
         python version : 3.11.4.final.0
       virtual packages : __archspec=1=x86_64
                          __win=0=0
       base environment : C:\Miniconda3  (writable)
      conda av data dir : C:\Miniconda3\etc\conda
  conda av metadata url : None
           channel URLs : https://repo.anaconda.com/pkgs/main/win-64
                          https://repo.anaconda.com/pkgs/main/noarch
                          https://repo.anaconda.com/pkgs/r/win-64
                          https://repo.anaconda.com/pkgs/r/noarch
                          https://repo.anaconda.com/pkgs/msys2/win-64
                          https://repo.anaconda.com/pkgs/msys2/noarch
          package cache : C:\Users\runneradmin\conda_pkgs_dir
       envs directories : C:\Miniconda3\envs
                          C:\Users\runneradmin\.conda\envs
[...]

PR #4951, opened 2023-07-28
Only collect dsos for files that will be inspected

tl-hbk:
### Description

This PR fixes the issue mentioned in https://github.com/conda/conda-build/issues/4944

After going through `post.py` and looking at where `all_needed_dsos` and `needed_dsos_for_file` are used, I think that it isn't necessary to collect dsos for non platform specific types.  

As I mentioned in the github issue, at the beginning of `check_overlinking_impl`, there is a check to filter for only files for whatever platform is being built

```
    files_to_inspect = []
    filesu = []
    for f in files:
        path = join(run_prefix, f)
        filetype = codefile_type(path)
        if filetype and filetype in filetypes_for_platform[subdir.split("-")[0]]:
            files_to_inspect.append(f)
        filesu.append(f.replace("\\", "/"))

    if not files_to_inspect:
        return dict()
```

Later in the code after `all_needed_dsos, needed_dsos_for_file = _collect_needed_dsos(...)` is called, `needed_dsos_for_file ` is only used in a loop that checks against the filtered files.

```
    for f in files_to_inspect:
        needed = needed_dsos_for_file[f]
```

### Checklist - did you ...


tl-hbk:
@kenodegard sorry if you're not the right person to ping for this but I was wondering if someone could take a look at this PR and the issue associated.

beckermr (changes requested):
This appears to be safe to merge, but it does need a test.

beckermr:
@tl-hbk Can you point me to original package build where you found this error?I see your github with the reproducer, but I'd like to look at the upstream package directly.

tl-hbk:
There wasn't a public feedstock where this happened.  It occured in a build for an internal package at my employer.

tl-hbk:
Sure I'll try to add that for a test.

PR #5033, opened 2023-10-10
Update `conda inspect channels`

kenodegard:
### Description

Update `conda inspect channels` to use latest `solver`/`transactions` logic instead of legacy `get_index`/`actions`.

Xref https://github.com/conda/conda/issues/13192

### Checklist - did you ...


conda_build/inspect_pkg.py
```diff
@@ -155,61 +159,42 @@ def replace_path(binary, path, prefix):
...
-        log.info("######## Testing platform %s ########", platform)
-        channels = [channel]
-        index = get_index(channel_urls=channels, prepend=False, platform=platform)
-        for _, rec in index.items():
-            # If we give channels at the command line, only look at
-            # packages from those channels (not defaults).
-            if channel != "defaults" and rec.get("schannel", "defaults") == "defaults":
-                continue
-            name = rec["name"]
+    for subdir in ["osx-64", "linux-32", "linux-64", "win-32", "win-64"]:
```
jaimergp, line 164: Shouldn't this use `KNOWN_SUBDIRS` in `conda.somewhere.constants`?

conda_build/inspect_pkg.py
```diff
@@ -155,61 +159,42 @@ def replace_path(binary, path, prefix):
...
                     prepend=False,
-                    platform=platform,
+                    subdir=subdir,
                 )
-                success &= bool(install_steps)
             except KeyboardInterrupt:
                 raise
-            # sys.exit raises an exception that doesn't subclass from Exception
-            except BaseException as e:
+            except BaseException as err:
```
jaimergp, line 188: Why is `check_install` raising `SystemExit`? AFAIK it's only calling the solver 🤔 , which doesn't `sys.exit`, does it?

conda_build/inspect_pkg.py
```diff
@@ -155,61 +159,42 @@ def replace_path(binary, path, prefix):
...
                 # and should be ignored here
                 continue
-            # Don't fail just because the package is a different version of Python
-            # than the default.  We should probably check depends rather than the
-            # build string.
-            build = rec["build"]
-            match = has_py.search(build)
-            assert match if "py" in build else True, build
-            if match:
-                additional_packages = [f"python={match.group(1)}.{match.group(2)}"]
```
jaimergp, line 186: Where did `additional_packages` go? Can you explain why is it being removed?

conda_build/inspect_pkg.py
```diff
@@ -87,23 +93,21 @@ def __str__(self):
...
-    return None
+    packages: Iterable[str],
+    subdir: str | None = None,
+    channel_urls: Iterable[str] = (),
+) -> None:
+    with TemporaryDirectory() as prefix:
+        Solver(
+            prefix,
+            channel_urls,
+            [subdir],
```
jaimergp, line 108: Will this fail if `subdir` is `None`? I'd make sure the default value is compatible with the `Solver` class.

jaimergp: Also, I remember that conda-build would merge `native_subdir, "noarch"` in a single made-up subdir internally, so I don't know if this assumption was here too (in case we need to pass `"noarch"` here). I think it doesn't, but I'd double check just in case.

conda_build/inspect_pkg.py
```diff
@@ -87,23 +93,21 @@ def __str__(self):
...
+            specs_from_args(packages),
+        ).solve_for_transaction().print_transaction_summary()
```
jaimergp, line 110: Looks like this output should be very similar to original one, so ok for me, but just mentioning that it might not be, in case that's important.

jaimergp, line 110:
```suggestion
        ).solve_for_transaction(ignore_pinned=True).print_transaction_summary()
```

Do we need this here due to the `pinned=False` in the original block?

jaimergp (dismissed):
LGTM but I've added some comments of things to double check because the conda-build<>conda interface for the solver is full of hacks (e.g. how subdirs are merged when fetching the index) and I don't know if those are assumed to apply here or not.

conda_build/inspect_pkg.py
kenodegard: As I understood (and from my testing) this doesn't make a difference. The package already defines its Python dependency so theres no need for us to also define it here.

This was actually prompted by the removal of the regex search on 183. Packages that do not follow the `pyNN` build string pattern fails (e.g. https://anaconda.org/conda-test/itsdangerous/files).

See `conda inpsect channels conda-test --test-installable`.

conda_build/inspect_pkg.py
kenodegard: No idea what it may have done at some point in the past. Happy to switch this to a plain `Exception` clause and to remove the misleading comment.

conda_build/inspect_pkg.py
kenodegard: Good catch

conda_build/inspect_pkg.py
kenodegard: subdir fallback: https://github.com/conda/conda-build/pull/5033/commits/9c1100f918643817f39656a1f6b82c64c83e51a9

conda_build/inspect_pkg.py
kenodegard: tightened exception handling: https://github.com/conda/conda-build/pull/5033/commits/e5ba600f640cff21c3ea6bee7521b0fce9582729

conda_build/inspect_pkg.py
```diff
@@ -186,8 +186,7 @@ def test_installable(channel: str = "defaults") -> bool:
                 )
             except KeyboardInterrupt:
                 raise
-            except BaseException as err:
-                # sys.exit raises an exception that doesn't subclass from Exception
+            except Exception as err:
```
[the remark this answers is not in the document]
kenodegard: 🤦🏼‍♂️ I checked `KeyError.__mro__` not `KeyboardInterrupt.__mro__`...

conda_build/inspect_pkg.py
```diff
@@ -87,23 +94,21 @@ def __str__(self):
 untracked_package = _untracked_package()
 
 
+@deprecated.argument("3.28.0", "4.0.0", "package", rename="subdir")
```
kenodegard, line 97:
```suggestion
@deprecated.argument("3.28.0", "4.0.0", "platform", rename="subdir")
```

conda_build/inspect_pkg.py
```diff
@@ -110,23 +110,21 @@ def __str__(self):
 untracked_package = _untracked_package()
 
 
+@deprecated.argument("3.28.0", "4.0.0", "platform", rename="subdir")
+@deprecated.argument("3.28.0", "4.0.0", "prepend")
+@deprecated.argument("3.28.0", "4.0.0", "minimal_hint")
```
jezdez, line 115: four dot oh, four dot oh!

kenodegard: not for long https://github.com/conda/conda-build/issues/5064

jezdez (dismissed):

conda_build/inspect_pkg.py
kenodegard: oh no! i forgot to update the deprecation versions!

```suggestion
@deprecated.argument("24.1.0", "24.3.0", "platform", rename="subdir")
@deprecated.argument("24.1.0", "24.3.0", "prepend")
@deprecated.argument("24.1.0", "24.3.0", "minimal_hint")
```

PR #5066, opened 2023-11-10
Fix conda_index.index verbose DEBUG/INFO logging

mbargull:
### Description

Log messages from `conda_index.index` get put out with high verbosity.
As an example, I built https://github.com/conda-forge/zlib-feedstock/tree/main
which yielded a build log with 936 lines -- 260 of those lines (about 28%) were `INFO`/`DEBUG` messages from `conda_index.index` like these:
<details>

```
DEBUG:conda_index.index:found subdirs {'linux-64', 'noarch'}
DEBUG:conda_index.index.sqlitecache:CondaIndexCache channel_root=/opt/conda/conda-bld, subdir=linux-64 db_filename=/opt/conda/conda-bld/linux-64/.cache/cache.db cache_is_brand_new=False
DEBUG:conda_index.index.sqlitecache:linux-64 listdir
DEBUG:conda_index.index.sqlitecache:linux-64 save fs state
DEBUG:conda_index.index:linux-64 find packages to extract
DEBUG:conda_index.index:linux-64 extract 1 packages
DEBUG:conda_index.index.sqlitecache:cache linux-64/libzlib-1.2.13-hd590300_5.tar.bz2
DEBUG:conda_index.index.sqlitecache:libzlib-1.2.13-hd590300_5.tar.bz2 early close
INFO:conda_index.index:linux-64 cached 64 KB from 1 packages at 1.0 MB/second
DEBUG:conda_index.index.sqlitecache:CondaIndexCache channel_root=/opt/conda/conda-bld, subdir=noarch db_filename=/opt/conda/conda-bld/noarch/.cache/cache.db cache_is_brand_new=False
DEBUG:conda_index.index.sqlitecache:noarch listdir
DEBUG:conda_index.index.sqlitecache:noarch save fs state
DEBUG:conda_index.index:noarch find packages to extract
DEBUG:conda_index.index:noarch extract 0 packages
INFO:conda_index.index:noarch cached 0 B from 0 packages at 0 B/second
INFO:conda_index.index:Subdir: linux-64 Gathering repodata
DEBUG:conda_index.index.sqlitecache:CondaIndexCache channel_root=/opt/conda/conda-bld, subdir=linux-64 db_filename=/opt/conda/conda-bld/linux-64/.cache/cache.db cache_is_brand_new=False
INFO:conda_index.index:linux-64 Writing pre-patch repodata
[...]

PR #5059, opened 2023-11-07
CI adjustments for conda-libmamba-solver as default

jaimergp:
### Description

conda 23.10.0 requires conda-libmamba-solver and has `solver=libmamba` as default, so conda-build's CI is going to be using it now.

### Checklist - did you ...

- [x] Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [X] Add / update necessary tests?
- [ ] Add / update outdated documentation?

PR #5068, opened 2023-11-10
Fix testing config monkeypatch for concurrent test flakiness

mbargull:
The default_testing_config monkeypatching fixture was added in gh-4653 but did not consider "from .config import get_or_merge_config" cases in which get_or_merge_config is already bound and thus not patched.

### Description

I noticed flaky tests again in gh-5066.
This is essentially a follow-up to gh-4653 (... about a year later....).

### Checklist - did you ...

### DIFF
--- a/conda_build/config.py
+++ b/conda_build/config.py
@@ -914,6 +914,8 @@ class Config:
 
 
 def _get_or_merge_config(config, variant=None, **kwargs):
+    # This function should only ever be called via get_or_merge_config.
+    # It only exists for us to monkeypatch a default config when running tests.
     if not config:
         config = Config(variant=variant)
     else:
--- a/conda_build/config.py
+++ b/conda_build/config.py
@@ -913,8 +913,7 @@ class Config:
             self.clean(remove_folders=False)
 
 
-def get_or_merge_config(config, variant=None, **kwargs):
-    """Always returns a new object - never changes the config that might be passed in."""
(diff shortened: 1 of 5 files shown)


mbargull:
pre-commit.ci autofix

mbargull:
N.B.: There are still some failing tests. Most of those are most likely either
- fixed by gh-5059,
- happening because we get concurrent accesses to the package cache.
    (We should at some point discuss `conda-build`s `pkgs-dirs` handling in general.)

conda_build/config.py
```diff
@@ -913,8 +913,7 @@ def __exit__(self, e_type, e_value, traceback):
             self.clean(remove_folders=False)
 
 
-def get_or_merge_config(config, variant=None, **kwargs):
-    """Always returns a new object - never changes the config that might be passed in."""
+def _get_or_merge_config(config, variant=None, **kwargs):
```
jaimergp, line 916: I'm confused about this. Why do we need the private indirection? Just to monkey patch it? If that's the case, it might be good to drop a comment here so future folks are not tempted to undo it :D

jaimergp (approved):
LGTM. Just one question about the new `_function`.

conda_build/config.py
mbargull: Yes, good idea, thanks.

jezdez (approved):

PR #5072, opened 2023-11-15
flatten requirements lists

wolfv:
This is a precursor to flatten requirement lists to be able to implement a Jinja function that returns multiple packages.

The idea is that we could use `{{ compiler('c') }}` to return a compiler & stdlib combination as two packages.


wolfv:
Note that this doesn't add the `{lang}_stdlib*` variables to the used variables and therefore doesn't compute the correct variants (same problem as current upstream implementation).

PR #5076, opened 2023-11-19
Tests: Fix get_or_merge_testing_config for cli tests

mbargull:
### Description

`cli.main_build.execute` (et al.) call the function with `**args.__dict__` with args having default values (e.g., `croot=None`) via `parse_args`.

Follow-up to gh-5068.
See https://github.com/conda/conda-build/pull/5066#issuecomment-1817033741

### Checklist - did you ...

### DIFF
--- a/tests/test-recipes/metadata/entry_points/meta.yaml
+++ b/tests/test-recipes/metadata/entry_points/meta.yaml
@@ -20,3 +20,7 @@ requirements:
     - setuptools
   run:
     - python
+
+# Ensure we get different build strings for concurrently tested packages.
+extra:
+  dummy: '{{ pytest_name is defined }}'
--- a/tests/cli/test_main_build.py
+++ b/tests/cli/test_main_build.py
@@ -2,7 +2,6 @@
 # SPDX-License-Identifier: BSD-3-Clause
 import os
 import re
-import sys
 from pathlib import Path
 
 import pytest
@@ -104,8 +103,7 @@ def test_build_output_build_path(
     args = ["--output", testing_workdir]
     main_build.execute(args)
     test_path = os.path.join(
-        sys.prefix,
-        "conda-bld",
+        testing_config.croot,
         testing_config.host_subdir,
         "test_build_output_build_path-1.0-1.tar.bz2",
     )
@@ -125,7 +123,7 @@ def test_build_output_build_path_multiple_recipes(
     main_build.execute(args)
 
     test_path = lambda pkg: os.path.join(
-        sys.prefix, "conda-bld", testing_config.host_subdir, pkg
+        testing_config.croot, testing_config.host_subdir, pkg
     )
     test_paths = [
         test_path("test_build_output_build_path_multiple_recipes-1.0-1.tar.bz2"),
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -118,11 +118,17 @@ def default_testing_config(testing_config, monkeypatch, request):
         return
 
     def get_or_merge_testing_config(config, variant=None, **kwargs):
-        merged_kwargs = {}
         if not config:
-            merged_kwargs.update(testing_config._testing_config_kwargs)
-        merged_kwargs.update(kwargs)
-        return _get_or_merge_config(config, variant, **merged_kwargs)
+            # If no existing config, override kwargs that are None with testing config defaults.
+            # (E.g., "croot" is None if called via "(..., *args.__dict__)" in cli.main_build.)
+            kwargs.update(
+                {
+                    key: value
+                    for key, value in testing_config._testing_config_kwargs.items()
+                    if kwargs.get(key) is None
+                }
+            )
+        return _get_or_merge_config(config, variant, **kwargs)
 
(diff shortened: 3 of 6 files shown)


mbargull:
pre-commit.ci autofix

mbargull:
pre-commit.ci autofix

mbargull:
I took the opportunity to let this run multiple times (on Sunday so I don't cause large CI queues for others).
Overall, things look quite better than before but there are still some failures occasionally:
1. [` macos (3.8, release, parallel) `](https://github.com/conda/conda-build/actions/runs/6924808470/job/18834562766#step:8:1427)
```python
FAILED tests/test_variants.py::test_build_run_exports_act_on_host - conda.exceptions.PackagesNotFoundError: The following packages are not available from current channels:

  - openssl
  - vc=9*

Current channels:

  - file:///Users/runner/work/_temp/parallel/popen-gw0/test_build_run_exports_act_on_0/win-32

To search for alternate channels that may provide the conda package you're
looking for, navigate to

    https://anaconda.org/

and use the search bar at the top of the page.
```
2. [` linux (3.11, canary, serial) `](https://github.com/conda/conda-build/actions/runs/6917611524/job/18829288175#step:8:234)
```python
FAILED tests/cli/test_main_build.py::test_build_add_channel - KeyError: PackageRecord(_hash=3447543960105723200, name='conda_build_test_requirement', version='1.0', build='h35b8010_0', build_number=0, channel=Channel("conda_build_test/linux-64"), subdir='linux-64', fn='conda_build_test_requirement-1.0-h35b8010_0.tar.bz2', md5='b643e7c94252b812aaff56ca1c83f002', url='https://conda.anaconda.org/conda_build_test/linux-64/conda_build_test_requirement-1.0-h35b8010_0.tar.bz2', arch='x86_64', platform='linux', depends=(), timestamp=1497374008.185, size=1978)
```
3. [` linux (3.11, canary, parallel) `](https://github.com/conda/conda-build/actions/runs/6917611524/job/18826528537#step:8:1410)
```python
FAILED tests/cli/test_main_build.py::test_test_extra_dep - KeyError: PackageRecord(_hash=-5768541241801201921, name='test_test_extra_dep', version='1.0', build='1', build_number=1, channel=Channel("test_test_extra_dep0/linux-64"), subdir='linux-64', fn='test_test_extra_dep-1.0-1.tar.bz2', md5='2001052b4eac59e5e1c5808642dcf21a', url='file:///home/runner/work/_temp/parallel/popen-gw1/test_test_extra_dep0/linux-64/test_test_extra_dep-1.0-1.tar.bz2', sha256='22f404ef55a1a1c360402893d07cd136846a7012ee7fed0daad9fcf24524d89d', arch=None, platform=None, depends=(), license='contract in blood', timestamp=1700406669.469, size=3037)
```

----
Note:
- All of these are *sporadic* failures and thus could be difficult to reproduce locally.
- Ignore the `serial`/`parallel` test distinction for now; `serial` marked tests are actually run in parallel currently.
- There *could* be interactions with `conda-libmamba-solver`, but (esp. for 2. and 3.) it is likely caused by package (index) cache corruptions (the tests re-use the same `croot`/`pkgs-dirs` and will occasionally trample on each other).

Re. 1: "Current channels" shows `.../win-32` which is odd because the test uses `platform="win", arch="64"`.
Re. 2: This is for the non-local https://anaconda.org/conda_build_test/conda_build_test_requirement/files .
Re. 3: This is for the local `test_test_extra_dep` which is in the correct `Channel("test_test_extra_dep0/linux-64")`.

tests/test-recipes/metadata/entry_points/meta.yaml
```diff
@@ -20,3 +20,7 @@ requirements:
     - setuptools
   run:
     - python
+
+# Ensure we get different build strings for concurrently tested packages.
+extra:
+  dummy: '{{ pytest_name is defined }}'
```
jaimergp, line 26: I don't get this part 🤔 Isn't this going to always be true? And how are these `extra` key-value pairs affecting build strings? I thought only `conda_build_config` keys got there, as well as some host deps?

mbargull: `conda-build` computes the hash from the set of used variant variables
To determine which variables are used, it searches the recipes for occurrences.
For `meta.yaml` this means if it finds a `{{ this_variable<whatever> }}`, it adds `this_variable` to the set of variables.
The hash in the build string will thus change depending on the value of `this_variable` (totally ignoring how it is otherwise used in the recipe).

mbargull:
I've opened gh-5078 for the general notes on serial tests and collisions in the package cache.

tests/test-recipes/metadata/entry_points/meta.yaml
mbargull: > I thought only `conda_build_config` keys got there, as well as some host deps?

Those are only the inputs (which I provide in the test functions via `variants={"pytest": [request.node.name]}`) but what actually makes it into the hash is determined as written above.

kenodegard (approved):

Issue #5064, opened 2023-11-09
Adopt expedited CEP-9 deprecation policy


kenodegard:
### Summary

While we have the deprecation helpers available we haven't formalized when and how deprecations occur in conda-build. This has resulted in deprecations being marked for eventual removal in the fictitious `4.0.0` release.

```[tasklist]
- [ ] https://github.com/conda/conda-build/issues/4975
- [x] Adopt 2 releases deprecation schedule (i.e., the next release marks code as deprecations (not pending deprecation), the subsequent release removes the code) https://github.com/conda/conda-build/pull/5080
- [x] Convert ongoing deprecations to conform with expedited release schedule https://github.com/conda/conda-build/pull/5080
```

PR #5080, opened 2023-11-21
Officially adopt CEP-8 and an expedited CEP-9

kenodegard:
### Description

With this PR we are officially adopting the bi-monthly release cycle used by conda as outlined in CEP-8. This will be evident with the switch from SemVer to CalVer starting with the January 2024 release (`24.1.0`).

In addition, we will also be adopting an expedited version of the deprecation policy (CEP-9). In this modified version, conda-build will skip the pending deprecation phase and only require one release before removing deprecated codes. To this end, all `4.0.0` deprecations have been modified to be deprecated in `24.1.0`.

Xref #5064

### Checklist - did you ...


conda_build/deprecations.py
```diff
@@ -127,6 +128,50 @@ def inner(*args, **kwargs):
...
+    def action(
+        self,
+        deprecate_in: str,
+        remove_in: str,
+        action: Action,
+        *,
+        addendum: str | None = None,
+        stack: int = 0,
+    ):
+        class DeprecationMixin:
```
jaimergp, line 140: Do we need to define the Mixin inside the function or can it live outside?

conda_build/deprecations.py
```diff
@@ -236,8 +281,8 @@ def _get_module(self, stack: int) -> tuple[ModuleType, str]:
         import inspect  # expensive
 
         try:
-            frame = inspect.stack()[2 + stack]
-            module = inspect.getmodule(frame[0])
+            frame = sys._getframe(2 + stack)
```
jaimergp, line 284: Why this change? Is `sys._getframe` guaranteed to be stable? According to [the docs](https://docs.python.org/3/library/sys.html#sys._getframe), it's a CPython implementation detail (e.g. it might be absent in PyPy?). So if we are doing this for performance, we should fall back to `inspect.stack` if unavailable.

```
getframe = getattr(sys, "_getframe", None)
frame = frame(2+stack) if getframe else inspect.stack()[2+stack]
```

kenodegard: Ditto, also see https://github.com/conda/conda/pull/12919

jaimergp: Ok, since we are catching AttributeError it's ok I guess.

beeankha (approved):

travishathaway (approved):

conda_build/index.py
```diff
@@ -332,7 +332,7 @@ def _delegated_update_index(
 # Everything below is deprecated to maintain API/feature compatibility.
 
 
-@deprecated("3.25.0", "4.0.0", addendum="Use standalone conda-index.")
+@deprecated("3.25.0", "24.1.0", addendum="Use standalone conda-index.")
```
dholth, line 335: This entire file needs to be deleted ASAP

dholth: Have we finished rewriting `conda index` so that it calls new conda index, or can we remove the command entirely and require the nicer python3 -m conda_index click CLI

kenodegard: the subcommand is marked for deprecation in 24.1.0:

https://github.com/conda/conda-build/blob/328f7e1f9939125efcc934eb5f2c44343b303208/conda_build/cli/main_index.py#L100-L121

it looks like you've marked all of the old index codes for deprecation, we can focus on removing all of this once we finish the November release

PR #5099, opened 2023-12-06
add pip check to recipe tests

dholth:
### Description

lief was added to the pip-style dependencies, causing pip check to fail since py-lief's conda package has no `dist-info` (pip metadata)

Xref https://github.com/conda/conda-build/issues/5101

### Checklist - did you ...

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,7 +36,6 @@ dependencies = [
   "jinja2",
   "jsonschema >=4.19",
   "libarchive-c",
-  "lief",
   "menuinst",
   "packaging",
   "pkginfo",
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -52,8 +52,6 @@ requirements:
     - requests
     - tomli                        # [py<311]
     - tqdm
-  test:
-    - pip
   run_constrained:
     - conda-verify  >=3.1.0
 
@@ -63,6 +61,9 @@ test:
     - conda_build
     # new/updated submodules (can be dropped after 1-2 releases)
     - conda_build.index
+  requires:
+    - setuptools
+    - pip
   files:
     - test_bdist_conda_setup.py
   commands:
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -52,6 +52,8 @@ requirements:
     - requests
     - tomli                        # [py<311]
     - tqdm
+  test:
+    - pip
   run_constrained:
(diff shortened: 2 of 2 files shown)


recipe/meta.yaml
```diff
@@ -78,6 +78,7 @@ test:
     - conda debug --help
     # bdist_conda
     - python test_bdist_conda_setup.py bdist_conda --help
+    - python -m pip check
```
kenodegard, line 81: Move this to the top of the commands list (otherwise the comments don't make sense)

kenodegard: Also need to add pip to `test/requires`

recipe/meta.yaml
```diff
@@ -51,6 +51,8 @@ requirements:
     - six
     - tomli                        # [py<311]
     - tqdm
+  test:
+    - pip
```
kenodegard, line 55: this belongs in the `test` section below

kenodegard (dismissed):

pyproject.toml
```diff
@@ -36,7 +36,6 @@ dependencies = [
   "jinja2",
   "jsonschema >=4.19",
   "libarchive-c",
-  "lief",
```
kenodegard, line 39:
```
# "lief",  # see https://github.com/conda/conda-build/issues/5101
```

dholth: I expect the canary section of this CI build to run the pip check test.

kenodegard: canaries are only built for PRs if the https://github.com/conda/conda-build/labels/build%3A%3Areview label is applied

kenodegard (approved):

recipe/meta.yaml
```diff
@@ -61,9 +61,13 @@ test:
     - conda_build
     # new/updated submodules (can be dropped after 1-2 releases)
     - conda_build.index
+  requires:
+    - setuptools
```
JeanChristopheMorinPerso, line 65: This feels weird to me. Is it because pip check complains that setuptools isn't present?

dholth: setuptools was in the other version of the recipe/feedstock. Maybe for `- test_bdist_conda_setup.py`

kenodegard: this is copied over from the AnacondaRecipes conda-build-feedstock where it was originally added in https://github.com/AnacondaRecipes/conda-build-feedstock/pull/26, not sure why

Issue #5106, opened 2023-12-11
How to interpet linking WARNINGs from conda-build 3.28.x?


jdblischak:
I am working on fixing the linking issues in the [tiledb-feestock recipe](https://github.com/conda-forge/tiledb-feedstock/blob/e512aec8b00b62d2c7bfaf06e66b41406ce96f1c/recipe/meta.yaml) (specifically PR https://github.com/conda-forge/tiledb-feedstock/pull/225)

However, I just started getting some never before seen WARNING messages. I couldn't find them documented, nor could I find similar conda-forge recipes with these WARNINGs. They look like the following:

```
[...]

PR #5041, opened 2023-10-18
Refactor which_package & which_prefix

kenodegard:
### Description

Consolidate `which_package` implementations and replaces legacy `conda.models.dist.Dist` usage with modern `conda.models.records.PrefixRecords`.

Discovered while working to remove usage of the following:
- `linked_data_no_multichannels`
- `linked_data`
- `linked`
- `is_linked`

Depends on #5055 

### Checklist - did you ...

- [x] Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [ ] ~Add / update necessary tests?~
- [ ] ~Add / update outdated documentation?~

PR #5108, opened 2023-12-12
`which_package` to yield unique collection of package records

kenodegard:
### Description

The refactor of `which_package` accidentally resulted in the same package record being returned multiple times (once for each _file_ collision). This results in misleading build warnings.

Instead we only want to return a unique set of package records.

Resolves https://github.com/conda/conda-build/issues/5106

### Checklist - did you ...

### DIFF
--- a/conda_build/inspect_pkg.py
+++ b/conda_build/inspect_pkg.py
@@ -73,9 +73,8 @@ def which_package(
     path = prefix / path
 
     for prec in PrefixData(str(prefix)).iter_records():
-        for file in prec["files"]:
-            if samefile(prefix / file, path):
-                yield prec
+        if any(samefile(prefix / file, path) for file in prec["files"]):
+            yield prec
 
 
 def print_object_info(info, key):
--- a/conda_build/post.py
+++ b/conda_build/post.py
@@ -1187,9 +1187,7 @@ def _lookup_in_prefix_packages(
     if len(precs_in_reqs) == 1:
         _print_msg(
             errors,
-            "{}: {} found in {}{}".format(
-                info_prelude, n_dso_p, precs_in_reqs[0], and_also
-            ),
+            f"{info_prelude}: {n_dso_p} found in {precs_in_reqs[0]}{and_also}",
             verbose=verbose,
         )
     elif in_whitelist:
@@ -1201,25 +1199,20 @@ def _lookup_in_prefix_packages(
     elif len(precs_in_reqs) == 0 and len(precs) > 0:
         _print_msg(
             errors,
-            "{}: {} found in {}{}".format(
-                msg_prelude, n_dso_p, [prec.name for prec in precs], and_also
-            ),
+            f"{msg_prelude}: {n_dso_p} found in {[str(prec) for prec in precs]}{and_also}",
             verbose=verbose,
         )
         _print_msg(
             errors,
-            "{}: .. but {} not in reqs/run, (i.e. it is overlinking)"
-            " (likely) or a missing dependency (less likely)".format(
-                msg_prelude, [prec.name for prec in precs]
-            ),
+            f"{msg_prelude}: .. but {[str(prec) for prec in precs]} not in reqs/run, "
+            "(i.e. it is overlinking) (likely) or a missing dependency (less likely)",
             verbose=verbose,
         )
     elif len(precs_in_reqs) > 1:
         _print_msg(
             errors,
-            "{}: {} found in multiple packages in run/reqs: {}{}".format(
-                warn_prelude, in_prefix_dso, precs_in_reqs, and_also
-            ),
+            f"{warn_prelude}: {in_prefix_dso} found in multiple packages in run/reqs: "
+            f"{[str(prec) for prec in precs_in_reqs]}{and_also}",
             verbose=verbose,
         )
     else:
--- a/news/5108-fix-which_package
+++ b/news/5108-fix-which_package
@@ -1,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Fix `conda_build.inspect_pkg.which_package` so it does not return duplicate package records. (#5108)
+
+### Deprecations
+
+* <news item>
+
+### Docs
+
+* <news item>
+
+### Other
+
+* <news item>
--- a/news/5108-fix-which_package
+++ b/news/5108-fix-which_package
@@ -5,6 +5,7 @@
 ### Bug fixes
 
 * Fix `conda_build.inspect_pkg.which_package` so it does not return duplicate package records. (#5108)
+* Fix `conda_build.post._lookup_in_prefix_packages` to display `str(PackageRecord)` instead of `repr(PackageRecord)`. (#5108)
 
 ### Deprecations
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -31,8 +31,6 @@ from conda_build.metadata import MetaData
 from conda_build.utils import check_call_env, copy_into, prepend_bin_path
 from conda_build.variants import get_default_variant
 
-pytest_plugins = ["conda.testing"]
-
 
 @pytest.fixture(scope="function")
 def testing_workdir(monkeypatch: MonkeyPatch, tmp_path: Path) -> Iterator[str]:
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -31,6 +31,8 @@ from conda_build.metadata import MetaData
 from conda_build.utils import check_call_env, copy_into, prepend_bin_path
 from conda_build.variants import get_default_variant
 
+pytest_plugins = ["conda.testing"]
+
 
 @pytest.fixture(scope="function")
 def testing_workdir(monkeypatch: MonkeyPatch, tmp_path: Path) -> Iterator[str]:
--- a/tests/test_inspect_pkg.py
+++ b/tests/test_inspect_pkg.py
@@ -2,6 +2,9 @@
 # SPDX-License-Identifier: BSD-3-Clause
 from __future__ import annotations
 
+import json
+from pathlib import Path
+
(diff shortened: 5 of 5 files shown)


jezdez (dismissed):

tests/test_inspect_pkg.py
```diff
@@ -0,0 +1,26 @@
...
+
+@pytest.mark.skipif(
+    parse(conda_version) < Version("23.5.0"),
+    reason="tmp_env fixture first available in conda 23.5.0",
+)
+def test_which_package(tmp_env: TmpEnvFixture):
+    with tmp_env("ca-certificates") as prefix:
+        pd = PrefixData(prefix)
+        prec = pd.get("ca-certificates")
+        precs = list(which_package(prefix / prec.files[0], prefix))
```
mbargull, line 24: To ensure that the test works in the future, we should either pin this to a certain build of `ca-certificates` or just compose a self-contained test environment here à la
```python
conda_meta = Path(tmp_dir, "conda-meta")
conda_meta.mkdir()
(conda_meta / "history").touch()
(conda_meta / "test-which_package-1-0.json").write_text(prefix_record_json)
```
with `prefix_record_json` being
```json
{
  "build": "0",
  "build_number": 0,
  "channel": "test-which_package-channel",
  "files": [
    "hardlink",
    "softlink"
  ],
  "name": "test-which_package",
  "paths_data": {
    "paths": [
      {
        "_path": "hardlink",
        "path_type": "hardlink",
        "size_in_bytes": 0
      },
      {
        "_path": "softlink",
        "path_type": "softlink",
        "size_in_bytes": 0
      }
    ],
    "paths_version": 1
  },
  "version": "1"
}
```
(or just a copy of a `conda-meta/ca-certificates-*.json`).

kenodegard: good idea! reworked this in https://github.com/conda/conda-build/pull/5108/commits/2ce0e20ecf6603d9104d166a4c6ea4e26647449d

news/5108-fix-which_package
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Fix `conda_build.inspect_pkg.which_package` so it does not return duplicate package records. (#5108)
```
kenodegard, line 7:
```suggestion
* Fix `conda_build.inspect_pkg.which_package` so it does not return duplicate package records. (#5108)
* Fix `conda_build.post._lookup_in_prefix_packages` to display `str(PackageRecord)` instead of `repr(PackageRecord)`. (#5108)
```

tests/test_inspect_pkg.py
```diff
@@ -0,0 +1,100 @@
...
+
+@pytest.mark.skipif(
+    parse(conda_version) < Version("23.5.0"),
+    reason="tmp_env fixture first available in conda 23.5.0",
+)
+def test_which_package(tmp_path: Path):
+    # create a dummy environment
+    (tmp_path / "conda-meta").mkdir()
+    (tmp_path / "conda-meta" / "history").touch()
+    # a dummy package with a unique file (softlink) and a shared file (shared)
```
mbargull, line 24:
```suggestion
    # a dummy package with a unique file (softlink) and a shared file (shared)
    (tmp_path / "shared").touch()
    (tmp_path / "softlink").symlink_to(tmp_path / "shared")
```

tests/test_inspect_pkg.py
```diff
@@ -0,0 +1,100 @@
...
+                            "size_in_bytes": 0,
+                        },
+                    ],
+                    "paths_version": 1,
+                },
+                "version": "1",
+            }
+        )
+    )
+    # a dummy package with a unique file (hardlink) and a shared file (shared)
```
mbargull, line 52:
```suggestion
    # a dummy package with a unique file (hardlink) and a shared file (shared)
    (tmp_path / "hardlink").touch()
```

tests/test_inspect_pkg.py
kenodegard: we don't actually need the files to exist for `which_package` to work

jezdez (approved):

tests/test_inspect_pkg.py
mbargull: If we want to properly test for the symlink cases, we do.
Yes, `conda_build.utils.samefile` has a fallback so it works even if the files don't exist.
But here, to cover the case that one package reports multiple paths for the same file, we need the symlink so that `Path.samefile` will be true for hard- as well as symlink.

kenodegard: you're right, I've expanded the softlink test to be more comprehensive

conda_build/inspect_pkg.py
```diff
@@ -73,9 +73,8 @@ def which_package(
     path = prefix / path
 
     for prec in PrefixData(str(prefix)).iter_records():
-        for file in prec["files"]:
-            if samefile(prefix / file, path):
-                yield prec
+        if any(samefile(prefix / file, path) for file in prec["files"]):
```
dholth, line 76: `(n * m)` ?

kenodegard: worst case yes, same as before

`any` returns on the first `True`: https://docs.python.org/3/library/functions.html#any

conda_build/post.py
```diff
@@ -1201,25 +1199,20 @@ def _lookup_in_prefix_packages(
     elif len(precs_in_reqs) == 0 and len(precs) > 0:
         _print_msg(
             errors,
-            "{}: {} found in {}{}".format(
-                msg_prelude, n_dso_p, [prec.name for prec in precs], and_also
-            ),
+            f"{msg_prelude}: {n_dso_p} found in {[str(prec) for prec in precs]}{and_also}",
```
jaimergp, line 1202: This upgrades from name to the full str representation. i.e. from `python` to `defaults::python-3.12.0-h123abc_0`, I think? Is that intentional?

conda_build/inspect_pkg.py
jaimergp: This is what avoids the duplicated records to begin with, right?

jaimergp: In a subsequent PR, we could see if implementing a cached `path -> artifact` map would speed this up. It can be a veeery slow part of the post processing stage in conda-build.

jaimergp:
Do we need to de-dup in line 1180 in `post.py` or are we confident there will be no dups?

```python
    precs = list(which_package(in_prefix_dso, run_prefix))
```

conda_build/inspect_pkg.py
kenodegard: > This is what avoids the duplicated records to begin with, right?

correct, this is essentially what was happening pre https://github.com/conda/conda-build/pull/5041: 

https://github.com/conda/conda-build/blob/c71c4abee1c85f5a36733c461f224941ab3ebbd1/conda_build/inspect_pkg.py#L57-L62

> cached `path -> artifact` map

definitely something to consider

conda_build/post.py
kenodegard: yes, this is now more detailed (but still less verbose than `str(precs)`) than before

the change was inspired by the fix below on line 1214-1215, where `precs_in_reqs: List[PackageRecord]` was cast to string resulting in a very verbose output, see https://github.com/conda/conda-build/pull/5108#pullrequestreview-1782145301

kenodegard:
@jaimergp AFAIK no deduping is needed once we merge this PR, with the modifications in `which_package` it is only be possible to return a specific `PackageRecord` once

tests/test_inspect_pkg.py
```diff
@@ -0,0 +1,126 @@
...
+from conda.core.prefix_data import PrefixData
+from packaging.version import Version, parse
+
+from conda_build.inspect_pkg import which_package
+
+
+@pytest.mark.skipif(
+    parse(conda_version) < Version("23.5.0"),
+    reason="tmp_env fixture first available in conda 23.5.0",
+)
```
mbargull, line 19:
```suggestion
from conda.core.prefix_data import PrefixData

from conda_build.inspect_pkg import which_package


```

tests/conftest.py
```diff
@@ -31,6 +31,8 @@
 from conda_build.utils import check_call_env, copy_into, prepend_bin_path
 from conda_build.variants import get_default_variant
 
+pytest_plugins = ["conda.testing"]
+
```
mbargull, line 35: Probably not needed anymore?
```suggestion
```

conda_build/post.py
mbargull: I think it's good that this is now more consistent with the full spec which was already output at https://github.com/conda/conda-build/pull/5108/files#diff-5073339a2e88f80714f82a001af334869aa5df9627b2f00ffd18d7a2035f537eR1190 .

conda_build/post.py
```diff
@@ -1201,25 +1199,20 @@ def _lookup_in_prefix_packages(
...
             verbose=verbose,
         )
         _print_msg(
             errors,
-            "{}: .. but {} not in reqs/run, (i.e. it is overlinking)"
-            " (likely) or a missing dependency (less likely)".format(
-                msg_prelude, [prec.name for prec in precs]
-            ),
+            f"{msg_prelude}: .. but {[str(prec) for prec in precs]} not in reqs/run, "
```
mbargull, line 1207: Although I find the change to use `str(prec)` in the `len(precs_in_reqs) == 0 and len(precs) > 0` case above a welcome change, one could argue that we might only want to output the name here so users get a more succinct "you might've missed `package` in run requirements" message.

But I don't have a strong opinion on this one.

mbargull (approved):
Added some minor nit comments, but overall this LGTM, thanks!

mbargull (approved):

beeankha (approved):

Issue #5126, opened 2023-12-22
Package checks extremely slow with large build environments


mdekstrand:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### What happened?

When the build environment is very large, the post-packaging checks take an extremely long time. The culprit seems to be https://github.com/conda/conda-build/blob/main/conda_build/post.py#L988 (or somewhere in its call chain), where it it seems to wind up calling `stat` on *every* file in the build environment. This becomes a problem when building a Rust package, because the current Rust compiler toolchain installs documentation (35K files)  and — after `cargo build` has completed — a copy of the cargo index (37K files). On Linux, this takes several minutes; on Windows, it takes hours. For an example of a public build that is affected, see conda-forge/deno-feedstock#79.

I determined this by two things:

1. running `strace` on the conda-build while it appeared stuck, saw many `stat` calls
2. <kbd>^C</kdb> the stuck conda-build, found this call chain in the backtrace

I just started noticing this recently; I don't know if it is a change in conda-build or a change in how the rust compiler is packaged in conda-forge.

travishathaway:
Hi @mdekstrand,

Thanks for reporting this. I've looked into this, and it does appear to be a performance regression from version `3.27.x` to `3.28.x`. 

My hunch right now is that it is coming from this function:

- https://github.com/conda/conda-build/blob/main/conda_build/utils.py#L2225

Tagging @kenodegard as he works much more in the conda-build base than I do.

kenodegard:
Resolved in #5130

PR #5130, opened 2024-01-02
Deprecate `conda_build.utils.samefile` in favor of path to package mapping

kenodegard:
### Description

The `which_package` refactor https://github.com/conda/conda-build/pull/5041 switched from using `os.path.normcase` to compare paths to calling `Path.samefile` which calls `os.stat`. This became especially expensive for certain builds so we need to ~revert~ revisit this change.

The proposal here is to implement a cached mapping of files to packages making `which_package` into `O(n)` for the first lookup and `O(1)` for all subsequent lookups.

We add a benchmark/battery test to better compare implementations. Here are the results of running the benchmark test with a few different versions/ideas:

| Version/Idea | Time |
|---|---|
| `3.27.0` | ~9s |
| `3.28.2` | ~16m48s |
| `normpath` | ~1m6s |
| `normcase` | ~35s |
| mapping | ~2s |

Resolves https://github.com/conda/conda-build/issues/5126

### Checklist - did you ...

### DIFF
--- a/news/5126-samefile-regression
+++ b/news/5126-samefile-regression
@@ -4,7 +4,7 @@
 
 ### Bug fixes
 
-* Use `os.path.normpath` instead of `Path.samefile` to avoid `os.stat` calls. (#5126 via #)
+* Use `os.path.normpath` instead of `Path.samefile` to avoid `os.stat` calls. (#5126 via #5130)
 
 ### Deprecations
 
--- a/news/5126-samefile-regression
+++ b/news/5126-samefile-regression
@@ -4,7 +4,7 @@
 
 ### Bug fixes
 
-* Use `os.path.normpath` instead of `Path.samefile` to avoid `os.stat` calls. (#5126 via #5130)
+* Update `which_package` to use a cached mapping of files to packages (`O(1)`) instead of relying on `Path.samefile` comparisons (`O(n * m)`). (#5126 via #5130)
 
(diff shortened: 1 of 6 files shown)


news/5126-samefile-regression
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Use `os.path.normpath` instead of `Path.samefile` to avoid `os.stat` calls. (#5126 via #)
```
kenodegard, line 7:
```suggestion
* Use `os.path.normpath` instead of `Path.samefile` to avoid `os.stat` calls. (#5126 via #5130)
```

jezdez (approved):

news/5126-samefile-regression
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Use `os.path.normpath` instead of `Path.samefile` to avoid `os.stat` calls. (#5126 via #5130)
```
kenodegard, line 7:
```suggestion
* Update `which_package` to use a cached mapping of files to packages (`O(1)`) instead of relying on `Path.samefile` comparisons (`O(n * m)`). (#5126 via #5130)
```

jezdez (approved):

Issue #5136, opened 2024-01-11
Link checker cannot find dependencies for multi-output recipe in 3.28.3


carterbox:
### What happened?

I believe there is an error with the link checker in conda-build version 3.28.3.

When building the libsecp256k1 recipe [from this pull request](https://github.com/conda-forge/staged-recipes/pull/23741#issuecomment-1887863659), we get a missing dependency error as follows:

```
['  ERROR (libsecp256k1,lib/libsecp256k1.so): Needed DSO x86_64-conda-linux-gnu/sysroot/lib64/libc.so.6 not found in any CDT/compiler package, nor the whitelist?!']
```
It seems that the link checker cannot find libc.so.6 for the libsecp256k1 output even though the compiler jinja template is included in the build requirements section for that output.

The same requirements are used for the libsecp256k1-2 output, but the link checker can find libc there.

I was able to recproduce the issue locally, but only with conda-build 3.28.3.
```
conda build -m .ci_support/linux64.yaml recipes/libsecp256k1 --error-overlinking
```
Earlier versions of conda-build including 3.26, 3.27 did not check the links of libsecp256k1.so (because it is a symlink?), and so do not raise the error. 3.28.2 checks the links but is able to find libc.

### Conda Details

<details>
<summary><code>conda info</code></summary>

```shell
active environment : base
    active env location : /home/DCHING/miniforge3
            shell level : 1
       user config file : /home/DCHING/.condarc
 populated config files : /home/DCHING/miniforge3/.condarc
                          /home/DCHING/.condarc
          conda version : 23.10.0
    conda-build version : 3.28.3
         python version : 3.10.13.final.0
       virtual packages : __archspec=1=skylake
                          __glibc=2.34=0
                          __linux=6.5.2=0
                          __unix=0=0
       base environment : /home/DCHING/miniforge3  (writable)
      conda av data dir : /home/DCHING/miniforge3/etc/conda
  conda av metadata url : None
           channel URLs : https://conda.anaconda.org/conda-forge/linux-64
                          https://conda.anaconda.org/conda-forge/noarch
          package cache : /home/DCHING/miniforge3/pkgs
                          /home/DCHING/.conda/pkgs
       envs directories : /home/DCHING/miniforge3/envs
                          /home/DCHING/.conda/envs
               platform : linux-64
[...]

PR #5141, opened 2024-01-16
Restore pre-`3.28` behavior for `which_package`

mbargull:
### Description

Fix linking check regressions by restoring pre-`3.28` behavior for `which_package`.

Fixes gh-5136.

### Checklist - did you ...

- [x] ~~Add / update outdated documentation?~~

### DIFF
--- a/conda_build/inspect_pkg.py
+++ b/conda_build/inspect_pkg.py
@@ -76,7 +76,7 @@ def which_package(
     # On Windows, be lenient and allow case-insensitive path comparisons.
     # NOTE: On macOS, although case-insensitive filesystem is default, still
     #       require case-sensitive matches (i.e., normcase on macOS is a no-op).
-    normcase_path = normcase(str(path).replace(os.sep, "/"))
+    normcase_path = normcase(path)
 
     for prec in PrefixData(str(prefix)).iter_records():
         files = prec["files"]
--- a/conda_build/inspect_pkg.py
+++ b/conda_build/inspect_pkg.py
@@ -11,7 +11,7 @@ from collections import defaultdict
 from functools import lru_cache
 from itertools import groupby
 from operator import itemgetter
-from os.path import abspath, basename, dirname, exists, join
+from os.path import abspath, basename, dirname, exists, join, normcase
 from pathlib import Path
 from typing import Iterable, Literal
 
@@ -67,43 +67,21 @@ def which_package(
     Given the path (of a (presumably) conda installed file) iterate over
     the conda packages the file came from.  Usually the iteration yields
     only one package.
-
-    We use lstat since a symlink doesn't clobber the file it points to.
     """
-    prefix = Path(prefix)
-
-    # historically, path was relative to prefix, just to be safe we append to prefix
-    # get lstat before calling _file_package_mapping in case path doesn't exist
     try:
-        lstat = (prefix / path).lstat()
-    except FileNotFoundError:
-        # FileNotFoundError: path doesn't exist
-        return
-    else:
-        yield from _file_package_mapping(prefix).get(lstat, ())
+        path = Path(path).relative_to(prefix)
+    except ValueError:
+        # If path is already relative to prefix, we get a ValueError.
+        pass
+    # On Windows, be lenient and allow case-insensitive path comparisons.
+    # NOTE: On macOS, although case-insensitive filesystem is default, still
+    #       require case-sensitive matches (i.e., normcase on macOS is a no-op).
+    normcase_path = normcase(str(path).replace(os.sep, "/"))
 
-
-@lru_cache(maxsize=None)
-def _file_package_mapping(prefix: Path) -> dict[os.stat_result, set[PrefixRecord]]:
-    """Map paths to package records.
-
-    We use lstat since a symlink doesn't clobber the file it points to.
-    """
-    mapping: dict[os.stat_result, set[PrefixRecord]] = {}
     for prec in PrefixData(str(prefix)).iter_records():
-        for file in prec["files"]:
-            # packages are capable of removing files installed by other dependencies from
-            # the build prefix, in those cases lstat will fail, while which_package wont
-            # return the correct package(s) in such a condition we choose to not worry about
-            # it since this file to package lookup exists primarily to detect clobbering
-            try:
-                lstat = (prefix / file).lstat()
-            except FileNotFoundError:
-                # FileNotFoundError: path doesn't exist
-                continue
-            else:
-                mapping.setdefault(lstat, set()).add(prec)
-    return mapping
+        files = prec["files"]
+        if normcase_path in (normcase(file) for file in files):
+            yield prec
 
 
 def print_object_info(info, key):
--- a/conda_build/inspect_pkg.py
+++ b/conda_build/inspect_pkg.py
@@ -71,7 +71,7 @@ def which_package(
     try:
         path = Path(path).relative_to(prefix)
     except ValueError:
-        # If path is already relative to prefix, we get a ValueError.
+        # ValueError: path is already relative to prefix
         pass
     # On Windows, be lenient and allow case-insensitive path comparisons.
     # NOTE: On macOS, although case-insensitive filesystem is default, still
@@ -79,8 +79,7 @@ def which_package(
     normcase_path = normcase(path)
 
     for prec in PrefixData(str(prefix)).iter_records():
-        files = prec["files"]
-        if normcase_path in (normcase(file) for file in files):
+        if normcase_path in (normcase(file) for file in prec["files"]):
             yield prec
 
 
--- a/news/5141-fix-which_package
+++ b/news/5141-fix-which_package
@@ -4,7 +4,7 @@
 
 ### Bug fixes
 
-* Fix linking check regressions by restoring pre-3.28 behavior for which_package. (#5141)
+* Fix linking check regressions by restoring pre-3.28 behavior for `conda_build.inspect_pkg.which_package`. (#5141)
 
 ### Deprecations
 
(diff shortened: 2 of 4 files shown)


conda_build/inspect_pkg.py
```diff
@@ -67,43 +67,21 @@ def which_package(
...
-    except FileNotFoundError:
-        # FileNotFoundError: path doesn't exist
-        return
-    else:
-        yield from _file_package_mapping(prefix).get(lstat, ())
-
+        path = Path(path).relative_to(prefix)
+    except ValueError:
+        # If path is already relative to prefix, we get a ValueError.
+        pass
```
mbargull, line 75: I don't know why this was changed to take paths not relative to the prefix in 3.28 .
To keep the change set small, I just added handling of non-relatives paths here instead of at `which_packages` call sites.

conda_build/inspect_pkg.py
```diff
@@ -67,43 +67,21 @@ def which_package(
...
+    # On Windows, be lenient and allow case-insensitive path comparisons.
+    # NOTE: On macOS, although case-insensitive filesystem is default, still
+    #       require case-sensitive matches (i.e., normcase on macOS is a no-op).
+    normcase_path = normcase(str(path).replace(os.sep, "/"))
```
mbargull, line 79: In the current 3.28 versions, there was a behavioral change in that builds on a case-insensitive filesystem on Unix systems (i.e., macOS with default file system) would ignore case mismatches.
Pre-3.28 behavior was to check for case-sensitivity on all Unixes (since `normcase`'s implementation only ignores cases on Windows).

tests/conftest.py
```diff
@@ -227,7 +227,13 @@ def conda_build_test_recipe_path(tmp_path_factory: pytest.TempPathFactory) -> Pa
     # clone conda_build_test_recipe locally
     repo = tmp_path_factory.mktemp("conda_build_test_recipe", numbered=False)
     subprocess.run(
-        ["git", "clone", "https://github.com/conda/conda_build_test_recipe", str(repo)],
+        [
+            "git",
+            "clone",
+            "https://github.com/conda/conda_build_test_recipe",
+            "--branch=1.21.11",
```
mbargull, line 234: This is an unrelated change to make the tests pass.
It should probably not be picked verbatim for the `main` branch.

tests/test_inspect_pkg.py
```diff
@@ -101,7 +129,11 @@ def test_which_package(tmp_path: Path):
     precs_missing = list(which_package(tmp_path / "missing", tmp_path))
     assert not precs_missing
 
-    precs_hardlinkA = list(which_package(tmp_path / "hardlinkA", tmp_path))
+    if on_win:
+        # On Windows, be lenient and allow case-insensitive path comparisons.
+        precs_hardlinkA = list(which_package(tmp_path / "Hardlinka", tmp_path))
+    else:
+        precs_hardlinkA = list(which_package(tmp_path / "hardlinkA", tmp_path))
```
mbargull, line 136: See https://github.com/conda/conda-build/pull/5141/files#r1454195421 above:
`conda-build<3.28` would only allow case-insensitive matches on Windows.

tests/test_inspect_pkg.py
```diff
@@ -172,10 +212,14 @@ def test_which_package_battery(tmp_path: Path):
...
-    # this occurs when, e.g., a package removes files installed by another package
+    # removed files should still return a package
+    # this occurs when, e.g., a build script removes files installed by another package
+    # (The opposite case with files really missing from the run environment,
+    # e.g., due to a post-install script removing them, is less likely and not
+    # covered.)
+    #  covered here.)
     for file in removed:
-        assert not len(list(which_package(tmp_path / file, tmp_path)))
+        assert len(list(which_package(tmp_path / file, tmp_path))) == 1
```
mbargull, line 222: This change is a direct consequence of https://github.com/conda/conda-build/pull/5141/files#r1454205167 above, i.e., files may be removed during build from the `host` prefix.

mbargull:
@kenodegard, I reviewed the code for `which_package` for <=3.27` and found some more behavioral changes that `3.28.x` introduced.
See the code as well as review comments in this PR for details.

I think we should revert the function's behavior to (mostly) what was done in `<3.27` and put out a new `3.28` patch release for it.
We should discuss further changes (behavioral and/or performance-wise) for future versions rather than the `3.28.x` line.

conda_build/inspect_pkg.py
kenodegard: yeah this change came about because I saw how before every call to `which_package` we would strip the prefix, so inlining the relative logic here makes sense

conda_build/inspect_pkg.py
kenodegard: from Python docs on `os.path.normcase`:
> On Windows, convert all characters in the pathname to lowercase, and also convert forward slashes to backward slashes.

```suggestion
    # On Windows, be lenient and allow case-insensitive path comparisons.
    # NOTE: On macOS, although case-insensitive filesystem is default, still
    #       require case-sensitive matches (i.e., normcase on macOS is a no-op).
    normcase_path = normcase(path)
```

tests/test_inspect_pkg.py
```diff
@@ -24,15 +25,27 @@ def test_which_package(tmp_path: Path):
     (tmp_path / "internal").symlink_to(tmp_path / "hardlinkA")  # packageA
     (tmp_path / "external").symlink_to(tmp_path / "hardlinkB")  # packageA
     (tmp_path / "hardlinkB").touch()  # packageB
+    # Files might be deleted from the prefix during the build, but they should
+    # still be recognized since they will be present in the run environment.
+    (tmp_path / "deleted").unlink(missing_ok=True)  # packageA
+    (tmp_path / "deleted_shared").unlink(missing_ok=True)  # packageA & packageB
```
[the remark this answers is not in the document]
kenodegard: ah I see, a mistake on my part in understanding where overlinking detection happened

tests/conftest.py
kenodegard: since we're about to start the January releases I'm thinking we will want to change the target branch and revert this change

mbargull: Do you mean this commit in particular or the whole PR?
I'm totally fine to drop the commit -- it was just a means to get "all green".
For the PR itself, I'd like us to have another `3.28` patch release if possible.
The thing is that we don't have any working `3.28` package on conda-forge (some are already marked `broken`) and I'd like us to get one such release before we introduce further changes from the upcoming version.
Meaning, as a maintainer of infrastructure-related packages on conda-forge, I'd like to do one of the following:
1. Put out a build for a new `3.28` patch release once tagged upstream.
2. Add this PR as a patch to the recipe for the current version to get a working downstream build.
3. Mark all `3.28` builds as `broken` and thus continue to use `3.27` for now on conda-forge.

Option 1 being the most preferred.
Option 2 is not ideal since we'd have divergent behavior up- and downstream.
Option 3 would be the last resort to get things working again.
(For conda-forge we'd rather want to increase not decrease the use of linking checks and omitting the linking checks via config would only create maintenance burden and is thus not an option I'm willing to go with.)

tests/test_inspect_pkg.py
```diff
@@ -172,10 +212,14 @@ def test_which_package_battery(tmp_path: Path):
...
             assert len(list(which_package(path, tmp_path))) == 1
 
-    # removed files should return no packages
-    # this occurs when, e.g., a package removes files installed by another package
+    # removed files should still return a package
+    # this occurs when, e.g., a build script removes files installed by another package
+    # (The opposite case with files really missing from the run environment,
+    # e.g., due to a post-install script removing them, is less likely and not
+    # covered.)
+    #  covered here.)
```
kenodegard, line 220:
```suggestion
    # (post-install scripts removing files from the run environment is less
    # likely and not covered)
```

news/5141-fix-which_package
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Fix linking check regressions by restoring pre-3.28 behavior for which_package. (#5141)
```
kenodegard, line 7:
```suggestion
* Fix linking check regressions by restoring pre-3.28 behavior for `conda_build.inspect_pkg.which_package`. (#5141)
```

conda_build/inspect_pkg.py
```diff
@@ -67,43 +67,21 @@ def which_package(
...
-            try:
-                lstat = (prefix / file).lstat()
-            except FileNotFoundError:
-                # FileNotFoundError: path doesn't exist
-                continue
-            else:
-                mapping.setdefault(lstat, set()).add(prec)
-    return mapping
+        files = prec["files"]
+        if normcase_path in (normcase(file) for file in files):
```
kenodegard, line 83:
```suggestion
        if normcase_path in (normcase(file) for file in prec["files"]):
```

conda_build/inspect_pkg.py
```diff
@@ -67,43 +67,21 @@ def which_package(
...
-        lstat = (prefix / path).lstat()
-    except FileNotFoundError:
-        # FileNotFoundError: path doesn't exist
-        return
-    else:
-        yield from _file_package_mapping(prefix).get(lstat, ())
-
+        path = Path(path).relative_to(prefix)
+    except ValueError:
+        # If path is already relative to prefix, we get a ValueError.
```
kenodegard, line 74:
```suggestion
        # ValueError: path is already relative to prefix
```

kenodegard (approved):
minor suggestions

tests/conftest.py
kenodegard: working with the team to figure out a release plan

added to 24.1.x milestone only because this is something that will need to be resolved and merged before the release can proceed

kenodegard (approved):
I added the regression test I wrote in #5140, otherwise LGTM

minrk:
Thanks for the fix!

PR #5148, opened 2024-01-23
only warn about build.script_env when splitting with --no-test

minrk:
rather than unconditionally warning about a possibly irrelevant scenario.

### Description

Any time build.script_env is used produces a warning (one for each variable) that if you happen to split build and test with `--no-test` you should make sure the test env matches. The only way to avoid this warning is to never use build.script_env.

This moves the warning to a _single_ warning, displayed only if `--no-test` is given. So in the vast majority of cases, no warning is produced.

### Checklist - did you ...


minrk:
Finally came back to this to fix the tests! Ready for review now.

These warnings cause a huge amount of noise in e.g. conda-forge builds that roughly amount to "warning: I hope you are doing what you are doing," which seems unnecessary when the information about whether that's the case (separate test or not) is available at the time of the warning.

conda_build/build.py
```diff
@@ -3730,6 +3730,21 @@ def build_tree(
...
                 else:
+                    # warn about script_env if splitting with --no-test
+                    script_env = utils.ensure_list(
+                        metadata.get_value("build/script_env", [])
+                    )
+                    if script_env:
+                        env_name_list = ", ".join(
+                            [item.partition("=")[0] for item in script_env]
+                        )
+                        warnings.warn(
```
jaimergp, line 3741: Shouldn't this be `logging.Logger.warning` instead? 🤔

minrk: I'm afraid I can't answer that. I didn't pick where the warning goes, I just moved the condition. If it should change, let me know or feel free to take over the PR with the change you would prefer.

PR #5151, opened 2024-01-25
Remove deprecations

kenodegard:
### Description

Removes all deprecations marked for removal in 24.1.0.

Entrypoints removal is postponed given ongoing discussion (https://github.com/conda/conda-build/issues/5081).

### Checklist - did you ...

- [x] Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [x] Add / update necessary tests?
- [ ] ~Add / update outdated documentation?~

PR #5074, opened 2023-11-15
Dist removal

mbargull:
### Description

Work in progress to remove legacy `conda.models.dist.Dist` usage.
Excerpt from gh-5069 :
> \[...\] `conda-build`'s slow (and memory intensive) ramp up before it gets to its actual first environment solving.
> \[...\] work on purging legacy behavior/code such that we could at least prepare to make use of the deferred creation of package records.
> A first step in that direction would be to finally remove `conda-build`'s usage of the old `Dist` objects (removed from `conda`'s core workings in 2018).
> 
> ----
> #### Impact
> - Prepare usage of newer/more performant package index handling in `conda-build`.
> - Removal and possible removal tech debt not only in `conda-build` but also `conda`.
> - **Breaking** changes for library uses of `conda_build`:
>     - not only might we remove `Dist`-related functions which might be used downstream,
>     - but also not using a `Dist`-based package index anymore could break downstream expectations.

Closes gh-5069
### Checklist - did you ...

- [x] Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [ ] Add / update necessary tests?
- [ ] Add / update outdated documentation?

PR #5152, opened 2024-01-26
 Prepare deprecation of full index and action usage 

mbargull:
### Checklist - did you ...


mbargull:
@kenodegard, @jezdez, I noticed we need to deprecate some more things when we replace the `conda.core.get_index` and `install_actions` usage.
This is based on gh-5151 and gh-5074, hence the large overall changeset.
The actual changes are only in commit 7ea64c61326af2a5c5879e9ff57763c2f4e0530e .

What this does overall is:
1. Mark channel data functionality in `conda_build.index` (now in `conda-index`) as deprecated since it requires iterating over the full index.
2. Mark `actions`-related arguments as deprecated since `actions` is now always empty or a `{PREFIX_ACTION: str, LINK_ACTION: list[PackageRecord]}` dictionary for which the prefix is redundant because the call sites always have the prefix already which in turn means `actions` is really now equivalent to a list of `PackageRecord`.
    We want to remove the `install_actions` functionality which manually overrides the solver's index with the full index.
    As such, this PR prepares the functions and their arguments to just handle lists of `PackageRecord` so we can do away with all the legacy `actions` stuff in the next version.

conda_build/index.py
```diff
@@ -474,6 +421,9 @@ def _clear_newline_chars(record, field_name):
             record[field_name] = record[field_name][0].strip().replace("\n", " ")
 
 
+@deprecated(
+    "24.1.0", "24.5.0", addendum="Use `conda_index._apply_instructions` instead."
+)
 def _apply_instructions(subdir, repodata, instructions):
```
mbargull, line 427: I marked this explicitly as deprecated since I know there are some conda-forge scripts that use `_apply_instructions`.
(We probably shouldn't to continue to use the underscore-prefixed function from `conda-idex` either. But this way we can at least mark it for removal here.)

I assume we don't have to mark every single underscore-prefixed symbol here separately, right?!

conda_build/environ.py
```diff
@@ -976,18 +995,53 @@ def get_install_actions(
...
+    debug=False,
+    locking=True,
+    bldpkgs_dirs=None,
+    timeout=900,
+    disable_pip=False,
+    max_env_retry=3,
+    output_folder=None,
+    channel_urls=None,
+):
+    return get_install_actions(
```
mbargull, line 1023: Having the direction from "new function"->"old function" here lets us avoid too many internal changes for now.
(In the next version, this function will of course just hold the actual code.)

jezdez: This will trigger the deprecation warning during runtime though, even though it shouldn't when being called as `get_package_records`.

Please extract the bulk of the `get_install_actions` function into a separate “private” function, which can be called from both `get_install_actions` (decorated for deprecation) and `get_package_records` (without deprecation).

conda_build/index.py
jezdez: Every function that we know is used and plan to remove, independent whether it's prefixed with an underscore or not, is going to need the decorator.

Remember [Hyrum's Law](https://www.hyrumslaw.com/), everything is API in the eye of the user, we can only manage the way it's being used and informing the users about our intentions, is good engineering practice.

jezdez (changes requested):

conda_build/environ.py
mbargull: Ah, good catch!
I've put the function definition now into `get_package_records` with just unwrapping/wrapping the `actions` dictionary to keep changes minimal for now.

conda_build/index.py
mbargull: > Every function that we know is used \[...\]

The problem is, we of course don't know which are used in the wild...
I've now just plonked the decorator on everything in `conda_build.index` :shrug:.

(If anyone knows a convenient way to really handle API<->internals definition in Python which we could use for future modules/packages, I'm all ears. Python, since it's one of those "everything is public and dynamic" languages, is just awful when Mr. Wright et al. point out these issues...)

mbargull:
Needs `news`/`CHANGELOG.md` entries.
I'm afk for a bit, so feel free to add things at will.

kenodegard (approved):

jezdez (approved):

mbargull:
Thanks for the updates and reviews!
I'm out for today -- GL HF on releasing! :crossed_fingers:

PR #5158, opened 2024-01-30
Fix entrypoint definition in pyproject.toml 

mbargull:
### Description

Just a small amendment for the equivalent changes to `recipe/meta.yaml` from
https://github.com/conda/conda-build/pull/5151/files#diff-f3725a55bf339595bf865fec73bda8ac99f283b0810c205442021f29c06eea9aR12 .

### Checklist - did you ...

- ~~[ ] Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release notes?~~
     

kenodegard (approved):

pyproject.toml
```diff
@@ -55,15 +55,15 @@ repository = "https://github.com/conda/conda-build"
...
-conda-index = "conda_build.cli.main_index:main"
-conda-inspect = "conda_build.cli.main_inspect:main"
-conda-metapackage = "conda_build.cli.main_metapackage:main"
-conda-render = "conda_build.cli.main_render:main"
-conda-skeleton = "conda_build.cli.main_skeleton:main"
-conda-debug = "conda_build.cli.main_debug:main"
+conda-build = "conda_build.cli.main_build:execute"
+conda-convert = "conda_build.cli.main_convert:execute"
+conda-develop = "conda_build.cli.main_develop:execute"
+conda-index = "conda_build.cli.main_index:execute"
```
mbargull, line 61: Oh, how do we want to go about `conda-index`?
(The same line is also still in `recipe/meta.yaml` and thus needs adjusting there too.)

The cleanest way would be to
1. add the `conda-index` script to the `conda-index` package,
2. add a `run_constrains` to `conda-index` with `conda-build>=24.1`,
3. release a new version of `conda-index` with those changes,
4. add a that version as the lower bound of the `conda-index` dependency in `conda-build`.

... But IDK, if it's feasible to require a new `conda-index` release before the impending `conda-build=24.1` release....

kenodegard: we are already planning on a `conda-index 0.4.0` release to resolve this (and a circular dependency issue for Python 3.12 support)

`conda-build` should no longer be making any `conda-index` entrypoint scripts

mbargull: If it's not possible to do a new `conda-index` release beforehand, we may want to set this as `conda-index = "conda_index.cli.cli"`?

kenodegard: as for it still being in `recipe/meta.yaml`, that's a mistake we need to rectify

kenodegard (approved):

PR #5160, opened 2024-01-31
use actual platform-specific `perl` path during TEST stage

jvolkening:
### Description

In troubleshooting some failed win_64 CI builds for PRs submitted to conda-forge *staged-recipes* (https://github.com/conda-forge/staged-recipes/pull/25137), I believe I narrowed the failed tests to the fact that the path determined for `perl` in the code modified here was setting the wrong path. Specifically, although these are `noarch: generic` recipes, they are being tested by the `staged-recipes` CI on Windows in addition to Linux and OSX. Since the value of `metadata.config.host_platform` here is set to `noarch` but the [code that sets the path to `perl`](https://github.com/conda/conda-build/blob/465d386318cca048d660c3a5aa7088b076473ab5/conda_build/config.py#L567C3-L572C19) is testing `if platform.startswith("win"):`, the resulting path is incorrect.

I saw another issue and commit relating to the use of `host_platform` instead of `sys.platform` in another part of the codebase in order to enable cross-compiling. In this instance, I can't foresee a situation in which the command string being built up in `write_test_scripts` would be run by `test` on a platform different from that on which it was called.

I haven't performed the points below yet, but I will if the change proposed seems valid. I'd like to get feedback on the change first.

### Checklist - did you ...

### DIFF
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -3235,7 +3235,7 @@ def _write_test_run_script(
             tf.write(
                 '"{perl}" "{test_file}"\n'.format(
                     perl=metadata.config.perl_bin(
-                        metadata.config.test_prefix, metadata.config.host_platform
+                        metadata.config.test_prefix, sys.platform
                     ),
                     test_file=join(metadata.config.test_dir, "run_test.pl"),
                 )


jvolkening:
I signed the CLA

jvolkening:
@conda-bot please refresh

conda_build/build.py
```diff
@@ -3235,7 +3235,7 @@ def _write_test_run_script(
             tf.write(
                 '"{perl}" "{test_file}"\n'.format(
                     perl=metadata.config.perl_bin(
-                        metadata.config.test_prefix, metadata.config.host_platform
+                        metadata.config.test_prefix, sys.platform
```
jaimergp, line 3238: Shouldn't this be `metadata.config.build_platform`?

jvolkening: Not entirely sure, but I can't see where that is defined or used in the codebase. `sys.platform` seems to be used in a number of places.

PR #5175, opened 2024-02-09
Avoid encoding error when warning about error in ensure_binary

minrk:
### Description

openmpi builds on conda-forge are failing at https://github.com/conda-forge/openmpi-feedstock/pull/142

The source of the issue (still unclear) is an invalid utf-8 path, but the intended error handling doesn't safely catch errors.

The problem is in the _logs_ of the failure to parse then raise because the warning itself fails to be formatted:

```
Traceback (most recent call last):
  File "/opt/conda/lib/python3.10/site-packages/conda_build/os_utils/liefldd.py", line 54, in ensure_binary
    return lief.parse(str(file))
TypeError: '/home/conda/feedstock_root/build_artifacts/openmpi-mpi_1707395567538/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold/lib/\x01\udce4\x05'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/conda/bin/conda-mambabuild", line 10, in <module>
    sys.exit(main())
  File "/opt/conda/lib/python3.10/site-packages/boa/cli/mambabuild.py", line 256, in main
    call_conda_build(action, config)
  File "/opt/conda/lib/python3.10/site-packages/boa/cli/mambabuild.py", line 228, in call_conda_build
    result = api.build(
  File "/opt/conda/lib/python3.10/site-packages/conda_build/api.py", line 254, in build
    return build_tree(
  File "/opt/conda/lib/python3.10/site-packages/conda_build/build.py", line 3789, in build_tree
    packages_from_this = build(
  File "/opt/conda/lib/python3.10/site-packages/conda_build/build.py", line 2877, in build
    newly_built_packages = bundlers[pkg_type](output_d, m, env, stats)
  File "/opt/conda/lib/python3.10/site-packages/conda_build/build.py", line 2004, in bundle_conda
    files = post_process_files(metadata, initial_files)
  File "/opt/conda/lib/python3.10/site-packages/conda_build/build.py", line 1815, in post_process_files
    post_build(m, new_files, build_python=python)
  File "/opt/conda/lib/python3.10/site-packages/conda_build/post.py", line 1818, in post_build
    post_process_shared_lib(m, f, prefix_files, host_prefix)
  File "/opt/conda/lib/python3.10/site-packages/conda_build/post.py", line 1680, in post_process_shared_lib
    mk_relative_linux(
[...]

PR #5176, opened 2024-02-09
allow LIEF to fail when using patchelf

minrk:
avoids unconditional failure when LIEF raises, even when not used

### Description

related to #5175, this allows LIEF to fail just like patchelf.

Previously, even when specifying `method=patchelf`, LIEF was unconditionally run. This adds error handling matching patchelf, so errors in an unused method don't prevent proceeding with patchelf.


beeankha (changes requested):
Hi @minrk , thank you so much for this PR! A few things are requested (or are open for discussion) prior to merge:

1. Is there any way that this code could be re-worked to use `log.warn` / `log.error` (with `log = utils.get_logger(__name__)`) for the error handling? I understand that you most likely wanted to remain consistent with the rest of the `mk_relative_linux` function, but we can take this opportunity to improve error handling patterns (if you feel inclined to do so!)

2. It would be great to see a test or two for this change; addtionally it might help to have more details in the PR description for steps on how to reproduce LIEF being unconditionally run even when `method=patchelf` was specified

3. Please add a news file with a brief description of this change

minrk:
I switched it to log and added news. I'm afraid I don't know how to exercise this failure in a test.

news/5176-lief-fail
```diff
@@ -0,0 +1,3 @@
+### Bug fixes
+
+* Error handling when LIEF fails is now consistent with patchelf
```
beeankha, line 3:
```suggestion
* Error handling when `LIEF` fails is now consistent with `patchelf`. (#5176)
```

beeankha (dismissed):

jaimergp:
Hm, the modified function is now using two different methods to report errors and warnings to the user. This implies slightly different formats, and also different streams (logging will print to stderr and print to stdout, as currently configured). I'd rather stick to one of the methods so:

- Either change everything to logging (and move `logger` to the top-level module namespace, under the imports)
- Revert to print with the same format as the other messages. We would need to create an issue to tackle a general move to logging, if one doesn't exist yet

beeankha:
Per Jaime's [comment,](https://github.com/conda/conda-build/pull/5176#issuecomment-2224985028) I reverted the logging changes; I also went ahead and [filed an issue](https://github.com/conda/conda-build/issues/5415) to change all error-reporting to logging.

dholth (approved):

PR #5216, opened 2024-03-06
NEW: Better package splitting for multi-output recipes with negative glob in outputs/files

### LINKED ISSUE #4196: Negative matching 'files' for outputs

carterbox:
### Description
Ressurects #4197. Closes #4196.

The purpose of this PR is to make it easier to split packages into multiple outputs using glob expressions. This is accomplished in two ways:

1. A negative pattern match is provided. This makes it easier to use fewer glob expressions if you need to include an entire directory tree except for a single file (type). For example, you might want to glob `lib/**/libfoo*`, but not `lib/**/*.a`
2. Only the files installed to the PREFIX in the top-level build are considered. This removes the need to craft your glob expressions to avoid the artifacts installed to the PREFIX by host dependencies. For example, you can now glob `include/**/*` without pulling in the headers of other packages. It is already the behavior of a single-output recipe to ignore files added to the prefix by host dependencies, so I'm not sure why this feature didn't make it into multi-output recipes.

These new behaviors are only used if the include/exclude keywords are used under the files key. The previous behavior is retained.

```yaml

outputs:
  - name: foo
    files:
      - bin/*  # old behavior; includes artifacts from other packages
  - name: bar
    files:
      include:
        - bin/*  # new behavior; only matches artifacts from this recipe
      exclude:  # optional
        - bin/*.exe # new behavior
  - name: zee
    script: install.py  # old behavior
```

### Checklist - did you ...


docs/source/resources/define-metadata.rst
```diff
@@ -1308,6 +1308,22 @@ contents.
          - *.some-extension
          - somefolder/*.some-extension
 
+Files can be excluded by specifying `files` as a dictionary separating
+files to `include` from those to `exclude`:
```
carterbox, line 1312:
```suggestion
Files can be excluded by specifying `files` as a dictionary separating
files to `include` from those to `exclude`. Files from other packages
are automatically excluded:
```

carterbox:
I'm not sure that the failure for linux 3.8 23.5.0 serial is related to this PR.

```
----------------------------- Captured stderr call -----------------------------
  File "/usr/share/miniconda/envs/test/lib/python3.8/site-packages/conda/exception_handler.py", line 16, in __call__
  File "/usr/share/miniconda/envs/test/lib/python3.8/site-packages/conda/cli/main.py", line 66, in main_subshell
  File "/usr/share/miniconda/envs/test/lib/python3.8/site-packages/conda/cli/conda_argparse.py", line 31, in <module>
  File "/usr/share/miniconda/envs/test/lib/python3.8/site-packages/conda/base/context.py", line 24, in <module>
ModuleNotFoundError: No module named 'conda._vendor.appdirs'
```

conda_build/build.py
```diff
@@ -1802,7 +1804,14 @@ def post_process_files(m: MetaData, initial_prefix_files):
     return new_files
 
 
-def bundle_conda(output, metadata: MetaData, env, stats, **kw):
+def bundle_conda(
+    output,
+    metadata: MetaData,
+    env,
+    stats,
+    new_prefix_files: set[str],
```
jaimergp, line 1812: Since this is adding a new positional argument to a (let's assume) public signature, what about making it a `None` kwarg. This way downstream users won't run into number of arguments errors:

```suggestion
    new_prefix_files: set[str] = None,
```

conda_build/build.py
```diff
@@ -1915,10 +1924,26 @@ def bundle_conda(output, metadata: MetaData, env, stats, **kw):
...
+            exclude = files.get("exclude", [])
+            exclude_files = {
+                os.path.normpath(pth)
+                for pth in utils.expand_globs(exclude, metadata.config.host_prefix)
+            }
+            keep_files = {
+                os.path.normpath(pth)
+                for pth in utils.expand_globs(include, metadata.config.host_prefix)
+            }
+            keep_files = new_prefix_files.intersection(keep_files) - exclude_files
```
jaimergp, line 1941: Just to make sure, the precedence of operators here is:

* From all new files added to PREFIX
* Keep the ones seen in `include`
* And then remove those in `exclude`

My set theory is rusty so, again, just to double check, is it the same if we diff `keep_files` and `exclude_files` first, and then intersect the result with `new_prefix_files`?

conda_build/build.py
```diff
@@ -2093,7 +2118,13 @@ def bundle_conda(output, metadata: MetaData, env, stats, **kw):
     return final_outputs
 
 
-def bundle_wheel(output, metadata: MetaData, env, stats):
+def bundle_wheel(
+    output,
+    metadata: MetaData,
+    env,
+    stats,
+    new_prefix_files: set[str],
```
jaimergp, line 2126: Same comment here about the API breakage. Not sure how prominent the usage of wheel outputs is, though.

docs/source/resources/define-metadata.rst
```diff
@@ -1299,9 +1299,10 @@ You can specify files to be included in the package in 1 of
 Explicit file lists are relative paths from the root of the
 build prefix. Explicit file lists support glob expressions.
 Directory names are also supported, and they recursively include
-contents.
+contents. Files installed to the prefix by host dependencies will
+be matched by glob expressions.
```
jaimergp, line 1303: I would maybe even highlight this further with an admonition or something similar. I wasn't aware of that design issue!

news/files-exclude.rst
```diff
@@ -0,0 +1,24 @@
+Enhancements:
+-------------
+
+* implemented new include/exclude sections for glob expressions in multi-output `outputs/files`
```
jaimergp, line 4:
```suggestion
* implemented new include/exclude sections for glob expressions in multi-output `outputs/files`. (#4196 via #5216)
```

tests/test-recipes/split-packages/copying_files/meta.yaml
```diff
@@ -9,9 +9,12 @@ requirements:
...
     files:
-      - subpackage_file1
-      - somedir
-      - "*.ext"
+      include:
+        - subpackage_file1
+        - somedir
+        - "*.ext"
+      exclude:
+        - "*3.ext"
```
jaimergp, line 17: Should we replace the old test with a new one instead of adding a new case? Maybe I would add a new output in this meta.yaml and then test the different behaviors differently, if possible (e.g. how the list-of-str behaviour does copy _anything_ in PREFIX, but include/exclude doesn't?).

jaimergp (commented):
I like this a lot. The PR is super clear and the documentation is on-point. I only added a couple comments to clarify my own doubts. The new positional argument could be considered API breaking, so let's see how the rest of the team feels about it. I don't have strong feelings, just applying an abundance of caution Just In Case ✨ .

Once we have discussed that, I'll be super happy to approve. 👍

conda_build/build.py
mbargull: The set difference would be the intersection with the negated set, so
(A ∩ B ) \ C = (A ∩ B ) ∩ C' = A ∩ B ∩ C' = A ∩ (B ∩ C') = A ∩ (B \ C)
. Or to put it another words: We end up with intersection the same sets (set intersection is associative).

docs/source/resources/define-metadata.rst
```diff
@@ -1311,6 +1312,24 @@ contents.
          - *.some-extension
          - somefolder/*.some-extension
 
+Files can be excluded by specifying `files` as a dictionary separating
+files to `include` from those to `exclude`. Files installed to the prefix
+by host dependencies are automatically excluded when the include/exclude
+syntax is used:
```
mbargull, line 1318: We need to mention the precedence here, i.e., `exclude` entries have higher priority as per the implementation.

mbargull: This mentions the new behavior, i.e., "Files [...] excluded when the include/exclude syntax is used".
But since this is a deviation from the previous "include" logic (which would be more of a `force-include` since it allows packaging pre-existing files.), I think we should make the difference more prevalent/explain the behavior for the old non-dict case more explicitly.

conda_build/build.py
carterbox: Yes; it's the same. I could have also implemented it as you have described:

```suggestion
            keep_files = new_prefix_files.intersection(keep_files - exclude_files)
```

conda_build/build.py
carterbox: https://github.com/conda/conda-build/pull/5216#discussion_r1534003328

docs/source/resources/define-metadata.rst
carterbox: > with an admonition or something similar.

I don't know what you mean. You want stronger language? For example: "WARNING! Files installed to the prefix by host dependencies also will be matched by glob expressions."

> I wasn't aware of that design issue!

Yeah, I'm not sure whether this is a bug/oversight or a feature. It's definitely unexpected to me. The most active contributor in this section is @msarahan; maybe they know?

jaimergp: I meant an [admonition in the RST sense](https://documatt.com/restructuredtext-reference/admonitions.html). Triple-backtick + `warning` kind of syntax. Equivalent to this:

> [!WARNING]
> Files installed to the prefix by host dependencies also will be matched by glob expressions.

docs/source/resources/define-metadata.rst
carterbox: > We need to mention the precedence here, i.e., `exclude` entries have higher priority as per the implementation.

`exclude` has a higher priority than which operation? It's unclear to me what ambiguity we would be removing by mentioning priority.

> This mentions the new behavior, i.e., "Files [...] excluded when the include/exclude syntax is used". But since this is a deviation from the previous "include" logic (which would be more of a `force-include` since it allows packaging pre-existing files.), I think we should make the difference more prevalent/explain the behavior for the old non-dict case more explicitly.

```suggestion
When using `outputs/files`, any file in the prefix (including those
installed by host dependencies) matching one of the glob expressions is
included in the output. Greater control over file matching may be
achieved by using `outputs/files/include` and `outputs/files/exclude`
glob expressions. When using include/exclude, only files installed by
the current recipe are considered. i.e. files in the prefix installed
by host dependencies are not matched.
```

conda_build/build.py
jaimergp: Thanks for the explanations! Marking this as resolved then.

conda_build/build.py
```diff
@@ -2817,7 +2848,7 @@ def build(
                     # This is wrong, files has not been expanded at this time and could contain
                     # wildcards.  Also well, I just do not understand this, because when this
                     # does contain wildcards, the files in to_remove will slip back in.
-                    if "files" in output_d:
+                    if "files" in output_d and not isinstance(output_d["files"], dict):
```
jaimergp, line 2851:
```suggestion
                    if not isinstance(output_d.get("files") or (), dict):
```

conda_build/build.py
```diff
@@ -1915,10 +1924,26 @@ def bundle_conda(output, metadata: MetaData, env, stats, **kw):
...
-        keep_files = {
-            os.path.normpath(pth)
-            for pth in utils.expand_globs(files, metadata.config.host_prefix)
-        }
+        if isinstance(files, dict):
+            # When file matching with include/exclude lists, only
+            # new_prefix_files are considered. Files in the PREFIX from other
+            # recipes (dependencies) are ignored
+            include = files.get("include", [])
+            exclude = files.get("exclude", [])
```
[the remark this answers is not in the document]
jaimergp: See line 1818:

```python
files = output.get("files", [])
```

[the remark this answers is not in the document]
carterbox: My understanding is that `files.get("include")` may return `None`, but `include` *must* be a list. So we use the falsiness of `None` to replace it with `[]`. Equivalent to:

```
include = files['include'] if ('include' in files and files['include'] != None) else []
```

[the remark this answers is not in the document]
jaimergp: You might have

```
- output: something
  files:
    include:
      - path
    exclude:
      - not-this-path  # [linux]
```

and then you'll have `files["exclude"] = None`. Not sure at what point we are ensuring that `include` is a list.

Under that circumstance `files.get("exclude", [])` will be `None` instead of the expected `[]`. Is that what you mean? I think we are saying the same 😬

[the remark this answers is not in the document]
carterbox: Yes. We are saying the same thing. I'm just trying to figure out if we should be doing something stronger than converting Falsy values into the empty list. Perhaps convert any non-list into a list? Something like:

```
include = files.get("include") if isinstance(files.get("include"), list) else []
```

docs/source/resources/define-metadata.rst
h-vetinari: > > We need to mention the precedence here, i.e., `exclude` entries have higher priority as per the implementation.
> 
> `exclude` has a higher priority than which operation?

The kind of precedence where if you do
```
  files:
    include:
      - lib/libfoo.so
    exclude:
      - lib/libfoo.so
```
the exclusion wins.

conda_build/build.py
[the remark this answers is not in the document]
jaimergp: That might silence errors when users populate it with a dict or something accidentally. I think we can just play it simple with falsy promotions.

tests/test-recipes/split-packages/copying_files/meta.yaml
carterbox: I have updated the test recipe to include multiple outputs: one uses the old file matching and one uses the new file matching.

docs/source/resources/define-metadata.rst
carterbox: Done in 642906b404b72aa3dc6b2d3342e295520585d65f

docs/source/resources/define-metadata.rst
carterbox: It is now explained that when files match both exclude and include, they are excluded. d1b12a16537cb7507bcf9ba8e5984c3c819bdac2

conda_build/build.py
carterbox: I have updated this condition to:
```python
                    if (
                        "files" in output_d
                        and output_d["files"] is not None
                        and not isinstance(output_d["files"], dict)
                    ):
```

carterbox (commented):
I believe I have addressed all outstanding comments.

jaimergp (approved):

docs/source/resources/define-metadata.rst
```diff
@@ -1301,7 +1301,11 @@ build prefix. Explicit file lists support glob expressions.
 Directory names are also supported, and they recursively include
 contents.
 
-.. code-block:: none
+.. warning:: When defining `outputs/files` as a list, any file in the prefix
```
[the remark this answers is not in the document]
carterbox: Completed

docs/source/resources/define-metadata.rst
```diff
@@ -1323,6 +1326,29 @@ contents.
          - *.some-extension
          - somefolder/*.some-extension
 
+Greater control over file matching may be
+achieved by defining `files` as a dictionary separating files to
+`include` from those to `exclude`.
```
jaimergp, line 1331:
```suggestion
achieved by defining ``files`` as a dictionary separating files to
``include`` from those to ``exclude``.
```

RST 😬

docs/source/resources/define-metadata.rst
```diff
@@ -1323,6 +1326,29 @@ contents.
...
+When using include/exclude, only files installed by
+the current recipe are considered. i.e. files in the prefix installed
+by host dependencies are excluded. include/exclude may not be used
+simultaneously with glob expressions listed directly in `outputs/files`.
```
jaimergp, line 1335:
```suggestion
simultaneously with glob expressions listed directly in ``outputs/files``.
```

news/files-exclude.rst
```diff
@@ -0,0 +1,24 @@
+Enhancements:
+-------------
+
+* implemented new include/exclude sections for glob expressions in multi-output `outputs/files`. (#4196 via #5216)
```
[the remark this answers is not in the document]
jaimergp: e.g. https://github.com/conda/conda-build/blob/main/news/5222-deprecating-conda_interface

The headers should have ###, and the filename is extension-less.

jaimergp (approved):
LGTM after fixing the little MD/RST nits.

jaimergp (approved):

carterbox:
What is the difference between the parallel and serial tests? Only the macos parallel tests are failing. It looks like it only changes parameters related to pytest and codecov, not conda itself?

jaimergp:
Serial jobs run tests that can't be run concurrently. Parallel jobs take rhe rest and run them with two processes wt the same time. The tests need to be marked as serial with a decorator to be considered as such, so the default is parallel.

jaimergp:
`tests/test_subpackages.py::test_subpackage_recipes[copying_files]` is failing on macOS and Windows @carterbox

docs/source/resources/define-metadata.rst
```diff
@@ -1313,7 +1313,10 @@ build prefix. Explicit file lists support glob expressions.
 Directory names are also supported, and they recursively include
 contents.
 
-.. code-block:: none
+.. warning::
+   When defining `outputs/files` as a list, any file in the prefix (including those installed by host dependencies) matching one of the glob expressions is included in the output.
```
isuruf, line 1317:
```suggestion
   When defining `outputs/files` as a list without specifying `outputs/script`, any file in the prefix (including those installed by host dependencies) matching one of the glob expressions is included in the output.
```

wolfv:
@carterbox I've started to implement this idea in `rattler-build`, too: https://github.com/prefix-dev/rattler-build/pull/819 

Would love your feedback once it's in a testable state. This is also in preparation of a "top-level" cache build (that can be split up into multiple packages).

carterbox:
Thanks, @isuruf for fixing my platform detection logic in the tests.

@jaimergp, Looks like the [failed test on Linux](https://github.com/conda/conda-build/actions/runs/9320567770/job/25657645218?pr=5216#step:10:447) is in a test unrelated to this PR. Maybe someone can restart the tests to see if it is ephemeral or not?

isuruf (approved):

jaimergp (dismissed):

conda_build/build.py
```diff
@@ -2706,7 +2735,11 @@ def build(
                     # This is wrong, files has not been expanded at this time and could contain
                     # wildcards.  Also well, I just do not understand this, because when this
                     # does contain wildcards, the files in to_remove will slip back in.
-                    if "files" in output_d:
+                    if (
+                        "files" in output_d
+                        and output_d["files"] is not None
+                        and not isinstance(output_d["files"], dict)
+                    ):
                         output_d["files"] = set(output_d["files"]) - to_remove
```
kenodegard, line 2743:
```suggestion
                    if (files := output_d.get("files")) and not isinstance(files, dict):
                        output_d["files"] = set(files) - to_remove
```

conda_build/build.py
```diff
@@ -1802,10 +1809,26 @@ def bundle_conda(output, metadata: MetaData, env, stats, **kw):
...
+            keep_files = {
+                os.path.normpath(pth)
+                for pth in utils.expand_globs(include, metadata.config.host_prefix)
+            }
+            keep_files = new_prefix_files.intersection(keep_files) - exclude_files
+        else:
+            keep_files = {
+                os.path.normpath(pth)
+                for pth in utils.expand_globs(files, metadata.config.host_prefix)
+            }
```
kenodegard, line 1831: Given the change in behavior above with only allowing new files to be included should this behavior here be deprecated?

```suggestion
        else:
            keep_files = {
                os.path.normpath(pth)
                for pth in utils.expand_globs(files, metadata.config.host_prefix)
            }
            
            if keep_files - new_prefix_files:
                deprecated.topic(...)
```

kenodegard (approved):
this looks good, just a minor suggestion and a question about deprecating the old behavior

conda_build/build.py
isuruf: It's still useful for eg when building `gcc_bootstrap` packages. Deprecating the implicit behaviour and adding a new flag would be useful.

docs/source/resources/define-metadata.rst
```diff
@@ -1323,6 +1326,29 @@ contents.
          - *.some-extension
          - somefolder/*.some-extension
 
+Greater control over file matching may be
+achieved by defining ``files`` as a dictionary separating files to
+``include`` from those to ``exclude``.
+When using include/exclude, only files installed by
+the current recipe are considered. i.e. files in the prefix installed
+by host dependencies are excluded. include/exclude may not be used
```
JeanChristopheMorinPerso, line 1334:
```suggestion
by host dependencies are excluded. include/exclude must cannot be used
```
`cannot` instead of `may` since the code is strict about the fact that both can't be used together?

beeankha: Just a small grammatical edit
```suggestion
by host dependencies are excluded. include/exclude must not be used
```

conda_build/build.py
carterbox: I feel that the current behavior of including all files from the prefix for multi-output and only installed files from the prefix for single-output recipes is inconsistent and thus non-intuitive. [rattler-build is adding a section called "always_include_files"](https://github.com/prefix-dev/rattler-build/pull/819#issuecomment-2148525698) for recipes that want to repackage artifacts from their host dependencies.

I'm favor of deprecating the old behavior, but I'm uncertain about any removal or replacement schedule.

h-vetinari: > [rattler-build is adding a section called "always_include_files"](https://github.com/prefix-dev/rattler-build/pull/819#issuecomment-2148525698) for recipes that want to repackage artifacts from their host dependencies.

This exists in conda already, and is [in use](https://github.com/search?q=org%3Aconda-forge%20%2Falways_include_files%3A%2F%20path%3Arecipe%2Fmeta.yaml&type=code) in conda-forge. Main use-case for that from my POV is to be explicit about overwriting some CMake metadata if an incremental build adds more targets to an already-existing CMake file ([example](https://github.com/conda-forge/abseil-cpp-feedstock/issues/62)).

jaimergp: We can deprecate that behaviour in a separate PR too. This PR has been open for too long and I wouldn't want to test your patience further @carterbox <3

jaimergp (approved):

kenodegard (approved):

conda_build/build.py
jaimergp: I'll open an issue to capture this feedback, though I think with the documentation this PR added we are in a better place already.

jaimergp:
Thanks @carterbox and everyone involved! Three months and 64 comments later, this one is in!

PR #5222, opened 2024-03-08
Deprecating `conda_build.conda_interface`

beeankha:
### Description

Miscellaneous deprecations that were pending in `conda_build.conda_interface` are being done here, which encourages conda-build and other downstream projects to import from conda directly. 

Note: `cc_conda_build` deprecation will be handled in a separate PR as it will involve some refactoring work.

### Checklist - did you ...


beeankha:
pre-commit.ci autofix

conda_build/__init__.py
```diff
@@ -15,3 +16,16 @@
     "render",
     "skeleton",
 ]
+
+# Skip context logic for docs since we don't install all dependencies
```
ryanskeith, line 20: Could you expand on this comment? I am unsure what it means.

ryanskeith (dismissed):
Other than a small ask to expand a comment.  This all looks great.

conda_build/__init__.py
beeankha, line 20:
```suggestion
# Skip context logic for doc generation since we don't install all dependencies in the CI doc build environment,
# see .readthedocs.yml file
```

kenodegard (approved):

conda_build/__init__.py
```diff
@@ -15,3 +16,17 @@
     "render",
     "skeleton",
 ]
+
+# Skip context logic for doc generation since we don't install all dependencies in the CI doc build environment,
+# see .readthedocs.yml file
```
jezdez, line 21: I would argue this is a bug and needs to be resolved by installing all dependencies in the Ci doc build environment. Maybe add caching there? Switch to setup-miniconda?

conda_build/__init__.py
```diff
@@ -15,3 +16,17 @@
...
+    import os
+
+    from conda.base.context import reset_context
+
+    # Disallow softlinks. This avoids a lot of dumb issues, at the potential cost of disk space.
+    os.environ["CONDA_ALLOW_SOFTLINKS"] = "false"
+    reset_context()
+
+except ImportError:
+    pass
```
jezdez, line 32: This code will be run on every import of `conda_build` and seems like an overreach to fix a documentation rendering issue.

I would suggest to either
- make this conditional [on the default Read The Doc environment variables](https://docs.readthedocs.io/en/stable/reference/environment-variables.html) (messy), or
- move this into a custom Sphinx extension ([how-to](https://www.sphinx-doc.org/en/master/development/tutorials/helloworld.html), basically a module with a `setup(app)` function) (clean).

kenodegard: This is supposed to run for ALL imports of `conda_build`, it's only for docs rendering that we don't need this

conda_build/__init__.py
kenodegard: Sure but that will require that we stop using the RTD pipeline and switch to a custom workflow which is out of scope for this PR

conda_build/__init__.py
jezdez: Gotcha, I'd appreciate if we'd not have to handle the special case for the docs here though. Ultimately it's not a huge deal though.

jezdez (approved):

Issue #5224, opened 2024-03-11
Recipe rendering very slow in certain cases


mbargull:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### What happened?

Recipe rendering can be very slow in certain cases.
Recently, I was confronted with the case of https://github.com/conda-forge/arrow-cpp-feedstock for which a `conda-smithy rerender` apparently takes between 10 and 30 minutes.
To put things into context:
The recipe
1. is rendered 6 times by `conda-smithy` (for `linux-64`/`linux-aarch64`/`linux-ppc64le`/`osx-64`/`osx-arm64`/`win-64`),
2. has outputs for 12 separate packages,
4. produces 200 package builds (due to variants defined in `conda-forge-pinning`+`.ci_support/migrations`),
5. includes 68 `pin_subpackage` calls via jinja.

Which resulted in roughly
- 60'000 calls to `.metadata.MetaData.get_recipe_text`,
- 50'000 calls to `.jinja_context.pin_subpackage`,
- 120'000 calls to `.metadata.select_lines`,
- 10'000'000 calls to `.metadata.eval_selector`.

(I stopped the tracing profiler at somepoint, so those numbers are extrapolated; the order of magnitude should match actual numbers, though.)

Higher call counts are to be expected due to `conda-build`'s repeated partial parsing of the recipe.
They are additionally unproportionally increased due to some oddities like `.metadata.MetaData.build_id` not being cached for `.jinja_context.pin_subpackage` at
https://github.com/conda/conda-build/blob/24.1.2/conda_build/jinja_context.py#L406
which means each `pin_subpackage_against_outputs` call will re-extract the recipe text over and over again due to
https://github.com/conda/conda-build/blob/24.1.2/conda_build/metadata.py#L1660 .

mbargull:
cc @xhochy, @h-vetinari since you have the "pleasure" dealing with this issue when maintaining `arrow-cpp-feedstock`.

I've proposed gh-5225 to treat some of the symptoms of this issue.
It's not doing doing "much" (as in, the changes do not make the process >100 times faster as would actually fixing it), but at least we'd get around 4 times speedup for the `arrow-cpp-feedstock` case.

mbargull:
This issue is currently being investigated.
Quick update on the recent progress:
- CodSpeed benchmarks got enabled in gh-5233 so we can better track these performance issues.
[...]

PR #5225, opened 2024-03-11
Miscellaneous render speedups

mbargull:
### Description

Addresses parts of gh-5224 .

The proposed changes do not fix the actual underlying issues with redundancy in the recipe re-processing happening while rendering.
They do treat symptoms of it, though.
This is done by avoiding a couple of obvious redundant function calls, substituting `copy.deepcopy` by `pickle.dumps/loads`, and foremost memoizing the lower level `select_lines` and regular expression evaluations.

It is intentionally narrowly scoped and does not alter the overall rendering process to avoid potential breaking changes (i.e., this PR is meant to be able to be "merged whenever" with no further dependencies/conflicts).

For the `conda-forge/arrow-cpp-feedstock` case from gh-5224 it cut the the overall processing time to about a quarter.

### Checklist - did you ...

### DIFF
--- a/conda_build/config.py
+++ b/conda_build/config.py
@@ -8,6 +8,7 @@ from __future__ import annotations
 import copy
 import math
 import os
+import pickle
 import re
 import shutil
 import time
@@ -822,9 +823,10 @@ class Config:
 
     def copy(self):
         new = copy.copy(self)
-        new.variant = copy.deepcopy(self.variant)
+        # Use picke.loads(pickle.dumps(...) as a faster copy.deepcopy alternative.
+        new.variant = pickle.loads(pickle.dumps(self.variant, -1))
         if hasattr(self, "variants"):
-            new.variants = copy.deepcopy(self.variants)
+            new.variants = pickle.loads(pickle.dumps(self.variants, -1))
         return new
 
     # context management - automatic cleanup if self.dirty or self.keep_old_work is not True
--- a/conda_build/config.py
+++ b/conda_build/config.py
@@ -823,10 +823,16 @@ class Config:
 
     def copy(self):
         new = copy.copy(self)
-        # Use picke.loads(pickle.dumps(...) as a faster copy.deepcopy alternative.
-        new.variant = pickle.loads(pickle.dumps(self.variant, -1))
+        # Use picke.loads(pickle.dumps(...) as a faster deepcopy alternative, if possible.
+        try:
+            new.variant = pickle.loads(pickle.dumps(self.variant, -1))
+        except:
+            new.variant = copy.deepcopy(self.variant)
         if hasattr(self, "variants"):
-            new.variants = pickle.loads(pickle.dumps(self.variants, -1))
+            try:
+                new.variants = pickle.loads(pickle.dumps(self.variants, -1))
+            except:
+                new.variants = copy.deepcopy(self.variants)
         return new
 
     # context management - automatic cleanup if self.dirty or self.keep_old_work is not True
--- a/conda_build/metadata.py
+++ b/conda_build/metadata.py
@@ -318,8 +318,7 @@ def select_lines(data, namespace, variants_in_place):
             # back to the non-memoized version to avoid wrong/lossy conversions.
             if isinstance(v, (list, set, tuple)):
                 namespace_copy[k] = tuple(
-                    tuple(e) if isinstance(e, (list, set)) else e
-                    for e in v
+                    tuple(e) if isinstance(e, (list, set)) else e for e in v
                 )
         namespace_tuple = tuple(namespace_copy.items())
         # Raise TypeError if anything in namespace_tuple is not hashable.
@@ -340,8 +339,7 @@ def _select_lines_memoized(data, namespace_tuple, variants_in_place):
         namespace["environ"] = os.environ
     if "pin_run_as_build" in namespace:
         namespace["pin_run_as_build"] = {
-            key: dict(value)
-            for key, value in namespace["pin_run_as_build"]
+            key: dict(value) for key, value in namespace["pin_run_as_build"]
         }
     return _select_lines(data, namespace, variants_in_place)
 
@@ -2290,7 +2288,8 @@ class MetaData:
             if "name" in out:
                 name_re = re.compile(r"^{}(\s|\Z|$)".format(out["name"]))
                 in_reqs = any(
-                    re_match(name_re, req) for req in self.get_depends_top_and_out("run")
+                    re_match(name_re, req)
+                    for req in self.get_depends_top_and_out("run")
                 )
                 if in_reqs:
                     break
@@ -2449,11 +2448,15 @@ class MetaData:
                 )
                 if build_reqs:
                     build_reqs = [
-                        req for req in build_reqs if not re_match(subpackage_pattern, req)
+                        req
+                        for req in build_reqs
+                        if not re_match(subpackage_pattern, req)
                     ]
                 if host_reqs:
                     host_reqs = [
-                        req for req in host_reqs if not re_match(subpackage_pattern, req)
+                        req
+                        for req in host_reqs
+                        if not re_match(subpackage_pattern, req)
                     ]
                 if run_reqs:
                     run_reqs = [
--- a/conda_build/metadata.py
+++ b/conda_build/metadata.py
@@ -280,8 +280,7 @@ def select_lines(data, namespace, variants_in_place):
         if line.lstrip().startswith("#"):
             # Don't bother with comment only lines
             continue
-        m = sel_pat.match(line)
-        if m:
+        if "[" in line and "]" in line and (m := sel_pat.match(line)):
             cond = m.group(3)
             try:
                 if eval_selector(cond, namespace, variants_in_place):
--- a/conda_build/metadata.py
+++ b/conda_build/metadata.py
@@ -966,8 +966,8 @@ def get_output_dicts_from_metadata(metadata, outputs=None):
             outputs.append(OrderedDict(name=metadata.name()))
     for out in outputs:
         if (
-            "package:" in metadata.get_recipe_text()
-            and out.get("name") == metadata.name()
+            out.get("name") == metadata.name()
+            and "package:" in metadata.get_recipe_text()
         ):
             combine_top_level_metadata_with_output(metadata, out)
     return outputs
--- a/conda_build/metadata.py
+++ b/conda_build/metadata.py
@@ -109,6 +109,10 @@ numpy_compatible_re = re.compile(r"pin_\w+\([\'\"]numpy[\'\"]")
 # used to avoid recomputing/rescanning recipe contents for used variables
 used_vars_cache = {}
 
+# Placeholders singletons for os/os.environ used to memoize select_lines().
+_selector_placeholder_os = object()
+_selector_placeholder_os_environ = object()
+
 
 @lru_cache(maxsize=200)
 def re_findall(pattern, text, flags=0):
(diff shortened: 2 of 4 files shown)


conda_build/metadata.py
```diff
@@ -109,6 +109,31 @@ def remove_constructor(cls, tag):
 # used to avoid recomputing/rescanning recipe contents for used variables
 used_vars_cache = {}
 
+# Placeholders singletons for os/os.environ used to memoize select_lines().
+_selector_placeholder_os = object()
+_selector_placeholder_os_environ = object()
+
+
+@lru_cache(maxsize=200)
```
mbargull, line 117: These `maxsize=200` are more or less arbitrarily chosen.
The default is `128` IIRC, but I just wanted to specify some (reasonable) number such that the temptation to add `maxsize=None` later is lessened.
We may not want to use `maxsize=None` deliberately to avoid unbound resource usage downstream, e.,g. in case multiple recipes are rendered in the same process.

beckermr:
Oh this is fabulous! Thank you!

conda_build/config.py
```diff
@@ -822,9 +823,10 @@ def clean_pkgs(self):
 
     def copy(self):
         new = copy.copy(self)
-        new.variant = copy.deepcopy(self.variant)
+        # Use picke.loads(pickle.dumps(...) as a faster copy.deepcopy alternative.
+        new.variant = pickle.loads(pickle.dumps(self.variant, -1))
         if hasattr(self, "variants"):
-            new.variants = copy.deepcopy(self.variants)
+            new.variants = pickle.loads(pickle.dumps(self.variants, -1))
```
beckermr, line 829: Do you want a try except here that reverts back to deepcopy n case something cannot be pickled?

mbargull: Thanks for taking a look!
I don't think we should get anything in there that `pickle` won't be able to handle; but a try/except also wouldn't hurt, of course.

conda_build/config.py
```diff
@@ -822,9 +823,16 @@ def clean_pkgs(self):
...
+        try:
+            new.variant = pickle.loads(pickle.dumps(self.variant, -1))
+        except:
+            new.variant = copy.deepcopy(self.variant)
         if hasattr(self, "variants"):
-            new.variants = copy.deepcopy(self.variants)
+            try:
+                new.variants = pickle.loads(pickle.dumps(self.variants, -1))
+            except:
+                new.variants = copy.deepcopy(self.variants)
```
jezdez, line 835: Let's do this without adding pickle and a bare except

conda_build/metadata.py
jezdez: Memoizing compiled patterns might give a slight advantage here

conda_build/metadata.py
```diff
@@ -109,6 +109,31 @@ def remove_constructor(cls, tag):
...
+    if isinstance(pattern, re.Pattern):
+        return pattern.match(text, flags)
+    return re.match(pattern, text, flags)
+
+
+@lru_cache(maxsize=200)
+def re_search(pattern, text, flags=0):
+    if isinstance(pattern, re.Pattern):
+        return pattern.search(text, flags)
+    return re.search(pattern, text, flags)
```
jezdez, line 135: Not a fan of this repetition and the pattern of having to cache regex

conda_build/metadata.py
```diff
@@ -268,6 +293,58 @@ def eval_selector(selector_string, namespace, variants_in_place):
...
+            if isinstance(v, (list, set, tuple)):
+                namespace_copy[k] = tuple(
+                    tuple(e) if isinstance(e, (list, set)) else e for e in v
+                )
+        namespace_tuple = tuple(namespace_copy.items())
+        # Raise TypeError if anything in namespace_tuple is not hashable.
+        hash(namespace_tuple)
+    except TypeError:
+        return _select_lines(data, namespace, variants_in_place)
+    return _select_lines_memoized(data, namespace_tuple, variants_in_place)
```
jezdez, line 328: This needs tests to verify that it's doing the right thing

mbargull:
> Another case of a hyper optimization by adding a bunch of hacks, notably adding pickle into the system where there is no need before. 

Please adjust you attitude here. I will not continue to work on this given such preface.

These optimizations are hacks around the actual problem, yes (see top comment).
The actual problem has not been addressed because of hard to reason about code flow with many places touching global or semi-global state; which is why this is scoped to functions that do not interact with these (semi-)global states.

> notably adding pickle into the system where there is no need before.

There would be no need for this if `.config.Config.copy` would not be called an excessive amount of time (same code flow/state problem as above). Unless the number of calls is reduced, one needs to replace `deepcopy` with something else like `pickle`. I'm not fan of using `pickle` here (still imposes high resource usage) but also not a fan of having `deepcopy` here at all; if you have better or at least comparable solutions to this, please do apply them.

Analyzing the root causes and untangling the code flow paths is a much more elaborate work that I am not willing to do (as it enters the realm of questionable effort given that resources could be directed to the newer recipe format instead).

----
These changes have not been made out of fun, but because it poses an non-negligible strain on resources (compute and maintainer "wait" times) for conda-forge (that I have not been aware of before), see the case laid out in https://github.com/conda/conda-build/issues/5224 .

If you do want alternative implementations to the proposed ones, please profile this yourself or get someone to do so (I've used `py-instrument` for sampling and `yappi` for tracing -- advice: only use `yappi` with changes from this PR or on a small test case than `conda-smithy rerender` of `conda-forge/arrow-cpp-feedstock`); it's otherwise hard to reason about the impact on such changes.

beckermr:
Hey @jezdez! Sorry I have been MIA in the conda world for a while. 

It may be useful to detail the pain points on render times throughout conda-forge. This is important context behind this PR. @mbargull mentioned a few of the locations, but I can detail more.

 - As @mbargull mentioned, long rerenders are becoming more common as we expand our build matrices and for certain things like compilers. This stresses our automated rerendering service and causes usability issues.
 - The conda-forge autotick bot spends a significant amount of time rerendering feedstocks. The long rerenders rate limit and stress the bot's infrastructure.
 - Every conda-forge build of an actual package also renders the recipe. Summed over all of the packages we build, even small changes in performance will have a big effect on our CI usage.
 
So I strongly disagree with the characterization of this PR as hyper-optimization. I do understand that code like this brings in additional maintenance burdens. @mbargull has been a long-term contributor to conda-build and is active in the community. I think he'll be around to help on maintenance here, but he should speak for himself.

dholth:
For reference, Python's built-in [re cache](https://github.com/python/cpython/blob/3.12/Lib/re/__init__.py#L271-L278) appears to have 512 and 256 entries.

mbargull:
@beckermr, thanks for giving more details!
You are far more involved with this day-to-day workings at conda-forge and knowledgeable about what this impacts.
The motivation for this PR is mostly an empathy thing realizing that you and the others have to deal with it...
To me personally, this is actually very low priority since I'm at most tangential (if at all) affected by this.
Which is also to say: IMO, the decision on when to include these or similar changes (`24.5` already? this year? next year?) is somewhat dependent how the affected parties push for it.

> @mbargull has been a long-term contributor to conda-build and is active in the community. I think he'll be around to help on maintenance here, but he should speak for himself.

Yes, happy to do so -- but even happier to rip such workarounds out again whenever they cease to be relevant!

jezdez:
> > Another case of a hyper optimization by adding a bunch of hacks, notably adding pickle into the system where there is no need before.
> 
> Please adjust you attitude here. I will not continue to work on this given such preface.

As context for everyone, while preparing conda 24.3.0 yesterday, I found https://github.com/conda/conda/pull/13568 having been merged with little to no discussion which introduced a custom version parser into the recently added conda deprecations system, diverging from a standard parser. This is what I referred to above as "another case" related to inconvenient tradeoffs.

Generally speaking, I'm trying to encourage to follow engineering practices that solve the core problems rather than pile hacks onto hacks. I'll try harder to elaborate why I think some tradeoffs are not worth making in that context.

> These optimizations are hacks around the actual problem, yes (see top comment). The actual problem has not been addressed because of hard to reason about code flow with many places touching global or semi-global state; which is why this is scoped to functions that do not interact with these (semi-)global states.
> 
> > notably adding pickle into the system where there is no need before.
> 
> There would be no need for this if `.config.Config.copy` would not be called an excessive amount of time (same code flow/state problem as above). Unless the number of calls is reduced, one needs to replace `deepcopy` with something else like `pickle`. I'm not fan of using `pickle` here (still imposes high resource usage) but also not a fan of having `deepcopy` here at all; if you have better or at least comparable solutions to this, please do apply them.

To be clear, giving feedback of requiring changes to this PR does not imply that I have to provide a better solution instead. I agree that it is worth documenting the performance bottleneck that you're seeing (which you started in #5224, thank you!), so we can make the appropriate fixes, without having to compromise on code quality. You know that's how we gotten into this mess in the first place.

> Analyzing the root causes and untangling the code flow paths is a much more elaborate work that I am not willing to do (as it enters the realm of questionable effort given that resources could be directed to the newer recipe format instead).

I totally agree! This is significant work and it's not helped by patching over it again.

> These changes have not been made out of fun, but because it poses an non-negligible strain on resources (compute and maintainer "wait" times) for conda-forge (that I have not been aware of before), see the case laid out in #5224 .

I understand the context as I've read the ticket before making the code review, and take my responsibilities seriously, just like your contribution. I'm sorry if my review didn't live up to that.

> If you do want alternative implementations to the proposed ones, please profile this yourself or get someone to do so (I've used `py-instrument` for sampling and `yappi` for tracing -- advice: only use `yappi` with changes from this PR or on a small test case than `conda-smithy rerender` of `conda-forge/arrow-cpp-feedstock`); it's otherwise hard to reason about the impact on such changes.

Thanks for the advise, I'll add this to #5224, if you don't mind since it would enable others to pick this up as well.

jezdez:
> Hey @jezdez! Sorry I have been MIA in the conda world for a while.

Hello! No apologies needed, as always :)

> It may be useful to detail the pain points on render times throughout conda-forge. This is important context behind this PR. @mbargull mentioned a few of the locations, but I can detail more.
> 
> * As @mbargull mentioned, long rerenders are becoming more common as we expand our build matrices and for certain things like compilers. This stresses our automated rerendering service and causes usability issues.
> * The conda-forge autotick bot spends a significant amount of time rerendering feedstocks. The long rerenders rate limit and stress the bot's infrastructure.
> * Every conda-forge build of an actual package also renders the recipe. Summed over all of the packages we build, even small changes in performance will have a big effect on our CI usage.

This is good context, which was missing in #5224, thank's for elaborating here!

> So I strongly disagree with the characterization of this PR as hyper-optimization. I do understand that code like this brings in additional maintenance burdens. @mbargull has been a long-term contributor to conda-build and is active in the community. I think he'll be around to help on maintenance here, but he should speak for himself.

I have to disagree that it's been effective for conda and conda-build to move solving fundamental problems to a later time, it's made both increasingly hard to maintain, maybe in some areas even impossible as @mbargull alluded in relation to the flow of the application logic.

dholth:
[The deepcopy code](https://github.com/python/cpython/blob/main/Lib/copy.py#L135-L158) is a slow way to copy Python objects, json loads/dumps or pickle are known optimizations. We used the [technique in conda-index](https://github.com/conda/conda-index/blob/42b55cba0b50fe3ad7f54543e6ba2e000971954b/conda_index/index/__init__.py#L346-L347) to save time. In an old stackoverflow post e.g. the pickle loop is 2.7x faster than copy.deepcopy.

mbargull:
@jezdez, thanks for elaborating more.
To explain my averse reaction to your initial review: The main review comment read more as a blanket statement which to me gave the overall issue (and how we have to tackle it) less recognition than warranted and bode a likely less productive and strenuous discussion (which I'm not eager to invest my time in).

To be clear from my side: None of these memoization approaches (in this PR and already elsewhere in the code base) are what I *want* us to have.
The problem we are facing is that we *would want* actual fixes, but coming up with those requires an unknown amount of work (that I can't provide at the moment and I don't know who else would do/sponsor such work).
Since `conda-build` is continuously used, we'd have to decide whether we are OK with leaving things as is until someone provides actual fixes or whether we want to provide some aid in the meantime that can be easily reverted once it's fixed.
(This is similar to what I tell people who voice it would be better to more or less only focus on `rattler-build` and not not invest in `conda-build` anymore; until we have a solution ready for direct deployment, we need pragmatic approaches.)

Do note that the changes herein are intentionally structured such that they can easily be reverted:
1. Different changes are put into separate commits so can be reverted (and also cherry-picked if we want to split this),
2. Even if single commits can't be automatically reverted, one only has to rename function names/calls (`_select_lines` --> `select_lines`; `re_*` --> `re.*`) or re-indent code (in `Config.copy`) and remove the then extraneous code.
3. The actual workings are unchanged without any hard to follow restructuring (which is what makes point 2 possible).

These workarounds, of course, do unfortunately add maintenance burden.
In my opinion, due to the above points and the narrowed scope of the changes, it is an acceptable burden to take.
That is, not in the general sense, but only in light of the strain on the (conda-forge's) ecosystem this is trying to alleviate.

----
That being said, what this PR needs is definitely more explanatory comments (if we add workarounds, we should mark them as such and explain, e.g., why the heck we'd add 50 lines only to be able to memoize `select_lines`) and ideally someone reproduce/confirm the impact of these changes (since we have to gauge impact vs added maintenance).

mbargull:
> [The deepcopy code](https://github.com/python/cpython/blob/main/Lib/copy.py#L135-L158) is a slow way to copy Python objects, json loads/dumps or pickle are known optimizations.

I opt'ed to not use `json` for this since, depending on your data, it is a destructive process.
E.g., not only would it not capture instances of custom classes but also `loads(dumps({1:2})) == {"1": 2} != {1: 2}`.

conda_build/metadata.py
mbargull: As mentioned by @dholth below, the `re` module already caches compiled patterns.
In contrast to `re`'s functions, the wrappers here not only cache for the pattern but also the matched string.
This is because we have redundant `re.*(same_pattern, same_string)` calls due to the repeated re-processing mentioned in https://github.com/conda/conda-build/issues/5224#issue-2179165566 .
As such, the memoizations here not try to sidestep the recompilation of patterns but the actual regex search/matches.

conda_build/metadata.py
mbargull: Personally, I'm also triggered by unnecessary repetitions, but I'm fine with it in this case since it should be a set-and-forget thing.

That said, I would make sense to profile this (for majorly affected cases) again to see which of these memoization wrappers have individually significant impact.
Meaning, we possibly only need a subset of them to get roughly the same results.

conda_build/metadata.py
mbargull: +1 Makes sense to add/extend tests for `select_lines` which ensure repeated calls to `select_lines` and the non-memoized `_select_lines` always yield the same results.

kenodegard: Enabling CodSpeed benchmarks and adding a new `test_select_lines_battery` test in https://github.com/conda/conda-build/pull/5233 so we can better measure improvements

kenodegard:
Thanks everyone for their input here. In an effort to get this discussion moving again (and to help us make data driven decisions with benchmarks) this PR is superseded by several other PRs.

Smaller PRs with isolated changes for easier review and clearer understanding of which are syntax improvements versus speed improvements:
- https://github.com/conda/conda-build/pull/5253
- https://github.com/conda/conda-build/pull/5278
- https://github.com/conda/conda-build/pull/5279
- https://github.com/conda/conda-build/pull/5280
- https://github.com/conda/conda-build/pull/5281
- https://github.com/conda/conda-build/pull/5283

And an alternative solution to the `select_lines` caching proposed here:
- https://github.com/conda/conda-build/pull/5237

PR #5234, opened 2024-03-13
Typing `conda_build.api`

kenodegard:
### Description

Adding much needed type annotations to `conda_build.api`.

### Checklist - did you ...


conda_build/api.py
```diff
@@ -15,13 +15,16 @@
 #    to conda-build's functionality.
 import os
 import sys
+from collections.abc import Iterable
```
schuylermartin45, line 18: For anyone else who might wonder why this isn't under `typing`: https://stackoverflow.com/questions/52827463/collections-iterable-vs-typing-iterable-in-type-annotation-and-checking-for-iter

conda_build/api.py
```diff
@@ -108,20 +114,28 @@ def render(
     return list(output_metas.values())
 
 
-def output_yaml(metadata, file_path=None, suppress_outputs=False):
+def output_yaml(
+    metadata: MetaData,
+    file_path: str | os.PathLike | Path | None = None,
```
schuylermartin45, line 119: If you prefer, there's:
```suggestion
    file_path: Optional[str | os.PathLike | Path] = None,
```
I've been using `Optional` for readability as that's the terminology many other languages use for `| None`.

conda_build/api.py
```diff
@@ -154,29 +155,47 @@ def get_output_file_paths(
...
-            recipe_path_or_metadata, "config"
-        ), f"Expecting metadata object - got {recipe_path_or_metadata}"
+
+    elif hasattr(recipe_path_or_metadata, "config"):
         metadata = [(recipe_path_or_metadata, None, None)]
-    #    Next, loop over outputs that each metadata defines
+
+    elif isinstance(recipe_path_or_metadata, Iterable) and all(
+        isinstance(recipe, tuple)
+        and len(recipe) == 3
```
schuylermartin45, line 164: Should we consider removing the magic number for a `Final[int]`?

conda_build/api.py
```diff
@@ -262,12 +286,12 @@ def build(
...
 
 def test(
-    recipedir_or_package_or_metadata,
-    move_broken=True,
-    config=None,
-    stats=None,
+    recipedir_or_package_or_metadata: str | os.PathLike | Path | MetaData,
+    move_broken: bool = True,
+    config: Config | None = None,
+    stats: dict | None = None,
```
schuylermartin45, line 292: Can we add type annotations to `dict[]`? Even if it is something like `dict[str, Any]`, that'd be an improvement.

conda_build/build.py
```diff
@@ -4,6 +4,8 @@
 Module that does most of the heavy lifting for the ``conda build`` command.
 """
 
+from __future__ import annotations
```
schuylermartin45, line 7: You might want to consider adding this to every file you've touched. It gets very difficult to track which annotations are available in what version. It's icky, but might prevent future headaches.

conda_build/api.py
schuylermartin45: I see that `stats` comes up in a few places. We could use a type alias so we don't have to repeat ourselves so much.

conda_build/skeletons/pypi.py
```diff
@@ -254,30 +260,26 @@ def _formating_value(attribute_name, attribute_value):
...
+    pypi_url: str = "https://pypi.io/pypi/",
+    noprompt: bool = True,
+    version_compare: bool = False,
+    python_version: str | None = None,
+    manual_url: bool = False,
+    all_extras: bool = False,
+    noarch_python: bool = False,
+    config: Config | None = None,
+    setup_options: str | Iterable[str] | None = None,
+    extra_specs: list[str] = [],
```
schuylermartin45, line 277: Defaulting to a list in Python is not advised. There is a [pylint](https://pylint.pycqa.org/en/latest/user_guide/messages/warning/dangerous-default-value.html) rule if we would like to check for this.

https://stackoverflow.com/questions/366422/how-can-i-avoid-issues-caused-by-pythons-early-bound-default-parameters-e-g-m

conda_build/skeletons/rpm.py
```diff
@@ -706,15 +712,15 @@ def write_conda_recipe(
...
-    dependency_add=[],
-    config=None,
-    distro=default_distro,
+    packages: list[str],
+    output_dir: str = ".",
+    version: str | None = None,
+    recursive: bool = False,
+    architecture: str = default_architecture,
+    override_arch: bool = True,
+    dependency_add: list[str] = [],
```
schuylermartin45, line 721: Another defaulted list.

conda_build/api.py
```diff
@@ -154,29 +155,47 @@ def get_output_file_paths(
...
         metadata = [(recipe_path_or_metadata, None, None)]
-    #    Next, loop over outputs that each metadata defines
+
+    elif isinstance(recipe_path_or_metadata, Iterable) and all(
+        isinstance(recipe, tuple)
+        and len(recipe) == 3
+        and isinstance(recipe[0], MetaData)
+        for recipe in recipe_path_or_metadata
+    ):
+        pass
```
kenodegard, line 168:
```suggestion
        metadata = recipe_path_or_metadata
```

conda_build/api.py
kenodegard: Doing so is counter to [UP007](https://docs.astral.sh/ruff/rules/non-pep604-annotation/) and will raise a linting error

conda_build/build.py
kenodegard: All files included here either already have this or it was added

conda_build/skeletons/pypi.py
kenodegard: I completely agree when we seek to modify the argument, in this case `extra_specs` is not modified, it simply extends another sequence

But either way this suggests we should say the following instead:

```suggestion
    extra_specs: Iterable[str] = (),
```

kenodegard: or just follow the `setup_options` pattern:

```suggestion
    extra_specs: str | Iterable[str] | None = None,
```

conda_build/api.py
beeankha: 📚 `:the_more_you_know:`

Thanks for linking to that!

conda_build/build.py
schuylermartin45: Gah, some of the tops of files got truncated in the preview. Carry on!

conda_build/api.py
kenodegard: Refactored to use a `NamedTuple` (`MetaDataTuple`) instead.

conda_build/api.py
kenodegard: Added a basic `StatsDict` type for now. Will need to continue iterating on it as we work our way through the code.

tests/cli/test_main_skeleton.py
```diff
@@ -54,6 +54,6 @@ def test_skeleton_pypi_arguments_work(testing_workdir):
     assert os.path.isdir("photutils")
     # Check that the setup option occurs in bld.bat and build.sh.
 
-    m = api.render("photutils")[0][0]
-    assert "--offline" in m.meta["build"]["script"]
-    assert m.version() == "1.10.0"
+    metadata = api.render("photutils")[0][0]
```
jezdez, line 57: Thank you!

jezdez (approved):

PR #5233, opened 2024-03-13
Enable CodSpeed and add select lines benchmark

kenodegard:
### Description

Enables CodSpeed benchmarks and adds a new `test_select_lines_battery` test.

### Checklist - did you ...

- [x] Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [x] Add / update necessary tests?
- [ ] Add / update outdated documentation?

Issue #5154, opened 2024-01-26
Remove old full-index-in-memory loading and `conda.plan`-related solver handling


mbargull:
In gh-5074 we removed the `conda.models.dist.Dist` usage from `conda-build`.

This included copying over the `install_actions`/`execute_actions`/`display_actions` functions from `conda.plan`
(which only continued to exist to serve these deprecated functions to `conda-build`).

These functions, which are as of gh-5074 now part of `conda_build.environ`, are reduced to have only the code  actually used in `conda-build`.
But their behavior still remained the same, i.e., they still use the same old solver invocations that loads the whole repodata index into memory, not benefiting from the reduced resource usage via https://github.com/conda/conda/pull/12050 .

To be able to work with the reduced/only partially loaded index, we need to deprecate/remove some functionality that relies on iterating over the full index.
gh-5152 marks parts related to that as deprecated.

The major part to look out for when we'll replace the current `install_actions` code is that the index has to stay in a consistent state throughout a recipe build.

---

Closes https://github.com/conda/conda-build/issues/4961

mbargull:
I'll look into this during the next couple of weeks.
In preparation of it, to get a sense on what memory/CPU consumption is currently happening, I stumbled upon the fact that peak memory usage (with `libmamba` solver) is actually dominated by a bug in `conda>=4.12`: https://github.com/conda/conda/issues/13541 .
So, when doing comparisons of resources used before/after changes in `conda-build`, one should factor out the surplus coming from that bug (patching `conda` when testing -- or, if lucky, maybe we get a patch release in the meantime).

jaimergp:
xref https://github.com/conda/conda-build/issues/4961

jaimergp:
@zklaus, summarizing our conversation earlier today:

- The easy way out: detect `context.solver == libmamba` and only populate the local part of the `_index` (no need to build PackageRecords for the whole thing)
- Same but a bit more elegant, add a class attribute to `Solver` to announce whether a full index is needed in memory or not.
- Better even: change the solver API to have a standardized, companion Index API that handles the interfacing better (with and without conda-build) instead of relying on `Solver._index` manipulation. This API should mention which channels need to be reloaded after an artifact is built and how.

dholth:
It would also be a great idea to make PackageRecord instantiate more quickly.

jaimergp:
> It would also be a great idea to make PackageRecord instantiate more quickly.

I'll create an issue for this if it doesn't exist already.

[...]

PR #5237, opened 2024-03-14
Add cached `_split_line_selector` to avoid redundant parsing in `select_lines`

kenodegard:
### Description

A continuation of https://github.com/conda/conda-build/pull/5233 with inspiration from https://github.com/conda/conda-build/pull/5225.

When rendering a recipe we end up invoking `conda_build.metadata.select_lines` a number of times.

Within `select_lines` we first iterate over every line and split the line into the line content versus the selector and second we evaluate the selector to determine whether the line should be kept.

The first process of iterating over every line and splitting it should (and can) be a one time operation. This is gained by introducing a new `_split_line_selector` function which caches the parsed result for a given text.

The second process of evaluating the selectors can also be cached but needs to be cached differently. Instead of caching this globally (since selectors can change between rendering passes) we cache the selectors locally, so a given selector will only be evaluated once for a single pass of the text in question (it's common for a given selector to be used multiple times within a single file, e.g., [conda-forge's `conda_buid_config.yaml`](https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/main/recipe/conda_build_config.yaml)).

### Checklist - did you ...


conda_build/metadata.py
```diff
@@ -1664,7 +1690,6 @@ def build_id(self):
             raise RuntimeError(
                 f"Couldn't extract raw recipe text for {self.name()} output"
             )
-        raw_recipe_text = self.extract_package_and_build_text()
```
kenodegard, line 1667: This is a duplicate invocation, look 5 lines above here

conda_build/metadata.py
```diff
@@ -267,38 +267,64 @@ def eval_selector(selector_string, namespace, variants_in_place):
...
+                    if value:
+                        lines.append(line)
+                except Exception as e:
+                    sys.exit(
+                        f"Error: Invalid selector in meta.yaml line {i + 1}:\n"
+                        f"offending line:\n"
+                        f"{line}\n"
+                        f"exception:\n"
+                        f"{e.__class__.__name__}: {e}\n"
+                    )
```
beckermr, line 327: These sys.exit calls in conda-build are very painful for downstream users of the code base as a library (i.e., conda-forge). The reason is that in the python exception hierarchy has the `SystemExit` exception raised by `sys.exit` and the standard `Exception` at the same level (https://docs.python.org/3/library/exceptions.html#exception-hierarchy). This means when one calls `sys.exit`, the invoking code cannot catch that exception without resorting to explicitly catching `SystemExit` as opposed to a normal `Exception`. The slight change in exception handling API breaks (at least my) expectations as a user of the code and has created quite a few mystery code crashes in the bot over the years. 

Can we use a standard exception here and then have the conda-build CLI code handle that exception and call `sys.exit`?

kenodegard: I completely agree that we need to get away from these `exit` calls, just created https://github.com/conda/conda-build/issues/5242

beckermr:
Have you tested this with the recipe @mbargull was working on? I'm wondering if the hotspots measured here are the same as those when big rerendering happens.

beckermr:
Here is a timing test on rerendering

```zsh
(cf-dev) beckermr@finnegan ctng-compilers-feedstock % time conda-smithy rerender 
INFO:conda_smithy.configure_feedstock:Downloading conda-forge-pinning-2024.03.19.15.37.47
INFO:conda_smithy.configure_feedstock:Extracting conda-forge-pinning to /Users/beckermr/.cache/conda-smithy
INFO:conda_smithy.configure_feedstock:__pycache__ rendering is skipped
INFO:conda_smithy.configure_feedstock:README rendering is skipped
WARNING: Setting build platform. This is only useful when pretending to be on another platform, such as for rendering necessary dependencies on a non-native platform. I trust that you know what you're doing.
WARNING: Setting build arch. This is only useful when pretending to be on another arch, such as for rendering necessary dependencies on a non-native arch. I trust that you know what you're doing.
WARNING: No numpy version specified in conda_build_config.yaml.  Falling back to default numpy value of 1.23
Adding in variants from /Users/beckermr/.cache/conda-smithy/conda_build_config.yaml
Adding in variants from /Users/beckermr/Desktop/conda-forge/ctng-compilers-feedstock/recipe/conda_build_config.yaml
Adding in variants from argument_variants
INFO:conda_smithy.configure_feedstock:Re-rendered with conda-build 24.1.3.dev39, conda-smithy 3.32.0, and conda-forge-pinning 2024.03.19.15.37.47
INFO:conda_smithy.configure_feedstock:You can commit the changes with:

    git commit -m "MNT: Re-rendered with conda-build 24.1.3.dev39, conda-smithy 3.32.0, and conda-forge-pinning 2024.03.19.15.37.47"

INFO:conda_smithy.configure_feedstock:These changes need to be pushed to github!

conda-smithy rerender  475.86s user 79.66s system 88% cpu 10:27.78 total
(cf-dev) beckermr@finnegan ctng-compilers-feedstock % time conda-smithy rerender 
INFO:conda_smithy.configure_feedstock:__pycache__ rendering is skipped
INFO:conda_smithy.configure_feedstock:README rendering is skipped
WARNING: Setting build platform. This is only useful when pretending to be on another platform, such as for rendering necessary dependencies on a non-native platform. I trust that you know what you're doing.
WARNING: Setting build arch. This is only useful when pretending to be on another arch, such as for rendering necessary dependencies on a non-native arch. I trust that you know what you're doing.
WARNING: No numpy version specified in conda_build_config.yaml.  Falling back to default numpy value of 1.23
Adding in variants from /Users/beckermr/.cache/conda-smithy/conda_build_config.yaml
Adding in variants from /Users/beckermr/Desktop/conda-forge/ctng-compilers-feedstock/recipe/conda_build_config.yaml
Adding in variants from argument_variants
INFO:conda_smithy.configure_feedstock:Re-rendered with conda-build 24.1.2, conda-smithy 3.32.0, and conda-forge-pinning 2024.03.19.15.37.47
INFO:conda_smithy.configure_feedstock:You can commit the changes with:

    git commit -m "MNT: Re-rendered with conda-build 24.1.2, conda-smithy 3.32.0, and conda-forge-pinning 2024.03.19.15.37.47"

INFO:conda_smithy.configure_feedstock:These changes need to be pushed to github!

conda-smithy rerender  642.32s user 95.69s system 89% cpu 13:42.06 total
```

So the code is about 35% faster with this change on one of our slowest rerendering tasks. That is a lot less than the 10x in the performance tests.

mbargull:
@kenodegard, thanks for taking a stab on this!
I haven't yet gotten to review this.
I could imagine this giving similar speed ups as my bolted-on approach for `select_lines` in gh-5225.
Having the line selection and per-variant-evaluation split (with the former in a global cache) could very well offset the ugly and costly "try to make `variant` hashable" dance I did in gh-5225.

----
@beckermr, thanks for trying this out already!
35% actually sounds good!
We have other parts that contribute much to the runtime overall.
For one, we have a big offset due to the unnecessary full-index creation (see gh-5154).
And then we do have the issue with conda-forge's large `conda_build_config.yaml` which is severely affected by the `.config.Config.copy`/`copy.deepcopy` issue mentioned in gh-5225.
Meaning, for `select_lines` alone, the timings look promising!

----
@kenodegard, I'll later try to strip the `arrow-cpp` example down to a test case we can use with your recent CodSpeed setup and update the tracking issue gh-5224 with the recent progress.

kenodegard:
@beckermr @mbargull the new benchmarks results are in: https://github.com/conda/conda-build/pull/5237#issuecomment-1998627131

conda_build/metadata.py
beeankha: Are we keeping [this `sys.exit` call](https://github.com/conda/conda-build/pull/5237/files#diff-a3f5613bda9366e31149c65731327047d4e29204c9e0508fd55416115b39a6bdR322) for this PR but then changing it (along with the other 80+ instances) in a separate PR? Or should this particular instance of `sys.exit` be altered before we merge this work?

kenodegard: I would like to leave this as is here and work on removing `sys.exit` separately

beeankha (approved):

conda_build/metadata.py
kenodegard: Split off into https://github.com/conda/conda-build/pull/5253

beeankha (dismissed):

jezdez (approved):

PR #5244, opened 2024-03-18
Treat deprecation warnings as errors

kenodegard:
### Description

We need to do a better job of catching deprecations occurring upstream.

Xref https://github.com/conda/conda/pull/12771

### Checklist - did you ...

### DIFF
--- a/conda_build/deprecations.py
+++ b/conda_build/deprecations.py
@@ -18,6 +18,8 @@ if TYPE_CHECKING:
 
 from . import __version__
 
+warnings.filterwarnings("default", category=DeprecationWarning, module="conda_build")
+
 
 class DeprecatedError(RuntimeError):
     pass
--- a/conda_build/deprecations.py
+++ b/conda_build/deprecations.py
@@ -18,7 +18,12 @@ if TYPE_CHECKING:
 
 from . import __version__
 
-warnings.filterwarnings("default", category=DeprecationWarning, module="conda_build")
+if not sys.warnoptions:
+    warnings.filterwarnings(
+        "default",
+        category=DeprecationWarning,
+        module="conda_build",
+    )
 
 
 class DeprecatedError(RuntimeError):
--- a/conda_build/deprecations.py
+++ b/conda_build/deprecations.py
@@ -18,13 +18,6 @@ if TYPE_CHECKING:
 
 from . import __version__
 
-if not sys.warnoptions:
-    warnings.filterwarnings(
-        "default",
-        category=DeprecationWarning,
-        module="conda_build",
-    )
-
 
 class DeprecatedError(RuntimeError):
     pass
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -158,5 +158,5 @@ filterwarnings = [
   "error::PendingDeprecationWarning:conda_build",
   "error::DeprecationWarning:conda_build",
   # ignore numpy.distutils error
-  "ignore:\\s+`numpy.distutils` is deprecated:DeprecationWarning:conda_build._load_setup_py_data",
+  'ignore:\s+`numpy.distutils` is deprecated:DeprecationWarning:conda_build._load_setup_py_data',
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -152,7 +152,9 @@ markers = [
 ]
 filterwarnings = [
   # elevate conda's deprecated warning to an error
+  "error::PendingDeprecationWarning:conda",
   "error::DeprecationWarning:conda",
   # elevate conda-build's deprecated warning to an error
+  "error::PendingDeprecationWarning:conda_build",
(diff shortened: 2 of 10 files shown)


conda_build/deprecations.py
```diff
@@ -18,6 +18,8 @@
 
 from . import __version__
 
+warnings.filterwarnings("default", category=DeprecationWarning, module="conda_build")
+
```
mbargull, line 22: We should let the decision on when to display `DeprecationWarning` up to downstream users, i.e., only opt-in to this for our tests and at most at the CLI level (though I can't think of an immediate reason why we'd want this on the CLI).

tests/test_environ.py
```diff
@@ -22,7 +24,12 @@ def test_environment_creation_preserves_PATH(testing_workdir, testing_config):
 
 def test_environment():
     """Asserting PrefixData can accomplish the same thing as Environment."""
-    assert (specs := environ.Environment(sys.prefix).package_specs())
+    with pytest.warns(
+        PendingDeprecationWarning
+        if deprecated._version_less_than("24.3")
```
mbargull, line 29: We should just use `packaging.version` as we do in other places and consider the version parsing in `DeprecationHandler` as purely internal, i.e., not use it outside of its scope.

mbargull (changes requested):
Thanks, looks good overall; just 2 small change requests.

conda_build/deprecations.py
kenodegard: While I agree that we shouldn't override what the downstream user wants, the current configuration means no users of `conda`/`conda-build` get warnings when they use deprecated features (e.g. https://github.com/conda/conda/issues/13704) which is not good

We could wrap this in an `if` block as suggested in the docs: https://docs.python.org/3/library/warnings.html#overriding-the-default-filter

tests/test_environ.py
kenodegard: Fair enough

conda_build/deprecations.py
kenodegard: We need this for the CLI when CLI arguments have been deprecated

tests/test_environ.py
kenodegard: See https://github.com/conda/conda-build/pull/5244/commits/6b026a1750207422de7eb56c2d6c49b36831c922

conda_build/deprecations.py
kenodegard: See https://github.com/conda/conda-build/pull/5244/commits/f78331dcd88f0b50251044db033230c048cf4005

mbargull: Oh, I didn't realize `DeprecationWarning` was used for CLI args.
IMO, those cases should use `UserWarning` and `DeprecationWarning`/`PendingDeprecationWarning` should only be used when `conda`/`conda_build` is used as a library -- but that'd be something to discuss somewhere else (e.g., the `conda` issue linked above).

----
Having this guarded with `if not sys.warnoptions:` would still mean that downstream (i.e., non-CLI) users don't get full control over it.
E.g., a `conda-smithy` invocation would display the warnings which would not be something one might expect in a usual `DeprecationWarning` workflow.
Meaning, I'd rather put this under `conda_build.cli` rather than general `conda_build` code.
(That all is more of a general design/convention thing; but TBH, it's not as important for `conda-build` since the downstream uses are few/niche and one might argue that it is good to have this verbose for `conda-smithy` since we currently don't have tests for deprecation warnings in its CI.)

kenodegard: Reverted in https://github.com/conda/conda-build/pull/5244/commits/29fb7bf588ad55050b34138d410ba7ac21475003

Working on https://github.com/conda/conda/pull/13716 which will need to be ported over to `conda-build`

pyproject.toml
```diff
@@ -150,3 +150,13 @@ markers = [
...
 ]
+filterwarnings = [
+  # elevate conda's deprecated warning to an error
+  "error::PendingDeprecationWarning:conda",
+  "error::DeprecationWarning:conda",
+  # elevate conda-build's deprecated warning to an error
+  "error::PendingDeprecationWarning:conda_build",
+  "error::DeprecationWarning:conda_build",
+  # ignore numpy.distutils error
+  "ignore:\\s+`numpy.distutils` is deprecated:DeprecationWarning:conda_build._load_setup_py_data",
```
kenodegard, line 161: Use string literal and avoid escaping

```suggestion
  'ignore:\s+`numpy.distutils` is deprecated:DeprecationWarning:conda_build._load_setup_py_data',
```

schuylermartin45 (approved):
I've followed this conversation on/off the last week, but the final result looks good.

beeankha (approved):

dholth:
Does this affect only the test suite or also normal usage?

mbargull (approved):

PR #5246, opened 2024-03-20
Add benchmark for high `pin_subpackage` count recipe

mbargull:
### Description

This is a performance regression benchmark for gh-5224.

### Checklist - did you ...

### DIFF
--- a/tests/test_api_render.py
+++ b/tests/test_api_render.py
@@ -7,6 +7,7 @@ should go in test_render.py
 
 import os
 import re
+from itertools import count, islice
 
 import pytest
 import yaml
@@ -15,6 +16,7 @@ from conda.common.compat import on_win
 
 from conda_build import api, render
 from conda_build.conda_interface import cc_conda_build
+from conda_build.variants import validate_spec
 
 from .utils import metadata_dir, variants_dir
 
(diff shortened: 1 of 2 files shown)


kenodegard:
@mbargull this looks promising! is this ready for review?

mbargull:
> @mbargull this looks promising! is this ready for review?

In principle yes.
IDK if we want to merge this yet, though, since it takes about 15 minutes CI time even in the further cut down version.
We could postpone merging this until we flesh out PRs addressing gh-5224 and base those PRs on this one in the mean time if we don't want to run this without improvements all the time.

tests/test_api_render.py
```diff
@@ -299,3 +301,35 @@ def test_pin_expression_works_with_python_prereleases(testing_config):
...
+        for version_count, package_count in [(1, 4), (4, 3), (4, 3)]:
+            zipped = []
+            for package, versions in islice(packages, package_count):
+                zipped.append(package)
+                variant[package] = list(islice(versions, version_count))
+            variant["zip_keys"].append(zipped)
+        # for version_count, package_count in [(3, 1), (2, 4), (1, 327)]:
+        for version_count, package_count in [(3, 1), (2, 4), (1, 33)]:
+            for package, versions in islice(packages, package_count):
+                variant[package] = list(islice(versions, version_count))
```
mbargull, line 328: To simplify things, it would probably suffice only add the single-entry keys since their number is far greater than the other entries (and `zip_keys`).
As is, these numbers just closely recreate those from conda-forge's cbc.yaml.
(Overall, the impact on the `variant` size mostly affects the `.config.Config.copy` impact on the overall run time, I'd assume.)

kenodegard: I say we leave this for now, we can always simplify it later once we've achieved some speedup.

tests/test_api_render.py
```diff
@@ -299,3 +301,35 @@ def test_pin_expression_works_with_python_prereleases(testing_config):
...
+        validate_spec("<generated>", variant)
+        return variant
+
+    ms = api.render(
+        recipe, config=testing_config, channels=[], variant=create_variant()
```
kenodegard, line 333: Wow `variant` versus `variants`, what a deep rabbit hole...

kenodegard (approved):
As long as the entire benchmark job completes under 30 min I'm ok with proceeding

I think there's value in getting to see the speed delta for this benchmark in the other PRs (we can always disable this once we've made the improvements)

tests/test_api_render.py
mbargull: Oh, yeah, maybe we do `variants=[create_variant()]` instead? I think `variant` might be some legacy thing (haven't checked though).

kenodegard: `variants` is also a dict, using `variants=[create_variant()]` will error

mbargull: I think `Config.variants` is supposed to be a list, equal to `[config.variant]` in case there is only one.
(But I'd have to check; and also check whether we'd rather want `variants` or not to avoid re-using old stuff.)

mbargull: Oh, now I get what you meant.
`render` also takes a `variants` argument, so that one wouldn't end up as `config.variants` then.
Plus, I also think that just using `api.render(..., variants=create_variants())` would be the correct thing to do since I think `Config.variant`/`Config.variants` would be the flattened/"exploded" variants (i.e., every key is supposed to have a single value; using `render(..., variant=...)` probably only worked because the test don't actually use the entries).

mbargull: Changed it to `variants`; let's see how much it affects the run time (since that way we should get multiple entries in `Config.variants`).

mbargull: > Changed it to `variants`; let's see how much it affects the run time (since that way we should get multiple entries in `Config.variants`).

Didn't change a thing, which is good to know!

mbargull:
Thanks for the review/merge :bow: !

mbargull:
Oh, one thing I kind of forgot was that by stripping the recipe down this much, I got rid of all of the selectors :hand_over_mouth: .
Meaning, for the `split_lines` optimizations, this would, of course, only evaluate the changes for passing over all lines, but without any `eval_selector(...)` calls at all...

PR #5248, opened 2024-03-21
Speed up rendering by avoiding costly/redundant function calls

mbargull:
### Description

Address some points of gh-5224.
This cherry-picks small (hopefully) non-controversial parts from gh-5225.

### Checklist - did you ...

### DIFF
--- a/conda_build/metadata.py
+++ b/conda_build/metadata.py
@@ -2641,14 +2641,11 @@ class MetaData:
         return variants.get_vars(_variants, loop_only=True)
 
     def get_used_loop_vars(self, force_top_level=False, force_global=False):
-        loop_vars = set(self.get_loop_vars())
-        return {
-            var
-            for var in self.get_used_vars(
-                force_top_level=force_top_level, force_global=force_global
-            )
-            if var in loop_vars
-        }
+        loop_vars = self.get_loop_vars()
+        used_vars = self.get_used_vars(
+            force_top_level=force_top_level, force_global=force_global
+        )
+        return set(loop_vars).intersection(used_vars)
 
     def get_rendered_recipe_text(
         self, permit_undefined_jinja=False, extract_pattern=None
--- a/conda_build/metadata.py
+++ b/conda_build/metadata.py
@@ -281,8 +281,8 @@ def select_lines(data, namespace, variants_in_place):
         if line.lstrip().startswith("#"):
             # Don't bother with comment only lines
             continue
-        m = sel_pat.match(line)
-        if m:
+        # Checking for "[" and "]" before regex matching every line is a bit faster.
+        if "[" in line and "]" in line and (m := sel_pat.match(line)):
             cond = m.group(3)
             try:
                 if eval_selector(cond, namespace, variants_in_place):
--- a/conda_build/metadata.py
+++ b/conda_build/metadata.py
@@ -1015,6 +1015,7 @@ def get_updated_output_dict_from_reparsed_metadata(original_dict, new_outputs):
     return output_d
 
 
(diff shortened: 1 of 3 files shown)


news/5248-render-speedups-func-calls
```diff
@@ -0,0 +1,19 @@
...
+
+* <news item>
+
+### Docs
+
+* <news item>
+
+### Other
+
+* <news item>
```
mbargull, line 19: I'm impartial on whether we should or should not include this news entry (meaning, I don't necessarily consider this a "significant" change).
We can include it if we want to track such changes more obviously or exclude it to reduce noise in the changelog :shrug: .

conda_build/metadata.py
```diff
@@ -2641,12 +2641,13 @@ def get_loop_vars(self):
...
     def get_used_loop_vars(self, force_top_level=False, force_global=False):
+        loop_vars = set(self.get_loop_vars())
         return {
             var
             for var in self.get_used_vars(
                 force_top_level=force_top_level, force_global=force_global
             )
-            if var in self.get_loop_vars()
+            if var in loop_vars
         }
```
kenodegard, line 2651: Isn't this just a set operation?

```python
used_vars = set(self.get_used_vars(...))
loop_vars = set(self.get_loop_vars())
return used_vars & loop_vars
```

conda_build/variants.py
```diff
@@ -697,16 +697,14 @@ def get_package_variants(recipedir_or_metadata, config=None, variants=None):
 def get_vars(variants, loop_only=False):
     """For purposes of naming/identifying, provide a way of identifying which variables contribute
     to the matrix dimensionality"""
+    first_variant, *other_variants = variants
     special_keys = {"pin_run_as_build", "zip_keys", "ignore_version"}
-    special_keys.update(set(ensure_list(variants[0].get("extend_keys"))))
+    special_keys.update(set(ensure_list(first_variant.get("extend_keys"))))
```
kenodegard, line 702:
```python
special_keys = {
    "pin_run_as_build",
    "zip_keys",
    "ignore_version",
    *ensure_list(first_variant.get("extend_keys")),
}
```

conda_build/metadata.py
```diff
@@ -1665,7 +1666,6 @@ def build_id(self):
             raise RuntimeError(
                 f"Couldn't extract raw recipe text for {self.name()} output"
             )
-        raw_recipe_text = self.extract_package_and_build_text()
```
kenodegard, line 1668: I wonder how much of a speedup removing this gives us ([since both of us did this](https://github.com/conda/conda-build/pull/5237#discussion_r1526780334))

conda_build/metadata.py
mbargull: Certainly! The changes from gh-5225 (and thus the once cherry-picked here) were the minimal/"just-bolted-on" changes with no drive-by cleanups.
But we can, of course, apply these cleanups here too if we touch these lines anyway.

mbargull: I put this now in as `return set(loop_vars).intersection(used_vars)` -- in this case here it probably doesn't make much of a difference, but if the second iterable (`used_vars`) is large, then `set.intersection` can be a bit faster since one avoids the creation of the intermediate `set` (IIRC, `a & b` also runs `a.intersection(b)` internally, but just with additional type checks -- but `set.intersection` takes iterables as inputs anyway).

kenodegard:
Per https://github.com/conda/conda-build/pull/5253#issuecomment-2052253495 I've split this into even smaller PRs

Closing in favor of #5278, #5279, #5280, #5281, #5283

PR #5255, opened 2024-03-25
Replace `sys.exit` with `CondaBuildUserError` exceptions

### LINKED ISSUE #4209: don't use sys.exit
##### `Output of conda info`

```

```

We here at conda-forge use some of the conda build APIs for our work. The code has been getting an increasing number of `sys.exit` calls for errors, which are causing our jobs to die. 

Instead, maybe the code should raise an error and then catch it at the highest level and call sys.exit? This preserves the internals for others while maintaining the same behavior.

### LINKED ISSUE #5242: Remove `sys.exit` calls
### Checklist

### What is the idea?

> These sys.exit calls in conda-build are very painful for downstream users of the code base as a library (i.e., conda-forge). The reason is that in the python exception hierarchy has the `SystemExit` exception raised by `sys.exit` and the standard `Exception` at the same level (https://docs.python.org/3/library/exceptions.html#exception-hierarchy). This means when one calls `sys.exit`, the invoking code cannot catch that exception without resorting to explicitly catching `SystemExit` as opposed to a normal `Exception`. The slight change in exception handling API breaks (at least my) expectations as a user of the code and has created quite a few mystery code crashes in the bot over the years. 
>
> Can we use a standard exception here and then have the conda-build CLI code handle that exception and call `sys.exit`?

_Originally posted by @beckermr in https://github.com/conda/conda-build/pull/5237#discussion_r1528736514_

### Why is this needed?

`sys.exit` calls are harder to catch/debug when using conda_build as an API.

### What should happen?

Review and ideally remove all `sys.exit` calls.

### Additional Context

_No response_

kenodegard:
### Description

Resolves https://github.com/conda/conda-build/issues/4209
~Resolves https://github.com/conda/conda-build/issues/5242~

### Checklist - did you ...


conda_build/exceptions.py
```diff
@@ -126,3 +128,8 @@ def __init__(self, error, *args):
         self.error = error
         self.msg = "runpaths check failed \n%s" % (error)
         super().__init__(self.msg)
+
+
+class CondaBuildUserError(CondaError):
```
dholth, line 133: Is this like ValueError?

tests/test_build.py
```diff
@@ -5,17 +5,25 @@
...
 
+import pytest
 from conda.common.compat import on_win
 
 from conda_build import api, build
+from conda_build.exceptions import CondaBuildUserError
 
 from .utils import get_noarch_python_meta, metadata_dir
 
+if TYPE_CHECKING:
```
dholth, line 24: Does this conditional speed up the tests

kenodegard: no, it makes the linter happy since we have `TCH` enabled, I don't see a strong argument for disabling this lint rule for tests

conda_build/exceptions.py
kenodegard: for our exception handling purposes it's more like `SystemExit`, so a clean exit printing the error message without the error reporting/stacktrace

PR #5260, opened 2024-03-27
Add `test_prefix_files`

kenodegard:
### Description

While working on #5255 found that `conda_build.utils.prefix_files` makes an unnecessary call to `expand_globs`. We only call `expand_globs` for symlink directories and `expand_globs` returns the path as is for symlinks.

### Checklist - did you ...


conda_build/utils.py
```diff
@@ -1780,22 +1780,25 @@ def merge_dicts_of_lists(
...
-            path = join(root, dn)
-            if islink(path):
-                res.add(path.replace(prefix_rep, "", 1))
-                res.update(expand_globs((path,), prefix))
-    return res
+    prefix = f"{prefix}{os.path.sep}"
+
+    def relpath(*path: str) -> str:
+        # this is os.path.relpath, just hacked to be faster
+        return join(*path).replace(prefix, "", 1).lstrip(os.path.sep)
```
jezdez, line 1791: Should we move the function out of `prefix_files` to not define on every call?

jezdez (approved):
Just a minor note about function definition location

conda_build/utils.py
kenodegard: Updated the algorithm slightly and eliminated the need for the helper function entirely, see https://github.com/conda/conda-build/pull/5260/commits/89e2edc26cf03a17daa9dccd381180ba5e4c6d0f

jezdez (approved):

PR #5262, opened 2024-03-28
Change hatch-vcs `version-file` to support cleaner pip editable installs

jaimergp:
### Description

In https://github.com/conda/conda-build/pull/5261 we are `pip install -e`'ing conda-build to support development better. The current configuration overwrites `__version__.py`, which then appears as modified and can be accidentally committed. We could gitignore it after removing it from the index, but I feel it's better to refactor this a tiny bit so:

- The fallback code is in `__init__.py`
- `hatch-vcs` writes to `_version.py`, which is gitignored from the get-go.

If you want to keep `__version__.py` around we can do that but I don't think it's necessary.

### Checklist - did you ...


kenodegard (approved):
Great idea! Dunno how many times I've reverted `__version__.py` after doing an editable install!

We will want to make the same change in conda.

Issue #4762, opened 2023-01-31
Better hashing of sources


chrisburr:
### Checklist

- [X] I added a descriptive title
- [X] I searched open requests and couldn't find a duplicate

### What is the idea?

Support a content dependent hash of sources rather than only hashing the source file itself. This is how [nixpkgs's `fetchzip`](https://nixos.org/manual/nixpkgs/stable/#fetchurl) works and could also be used to have a stable hash for recipes using `git_url`/`git_rev`.

### Why is this needed?

I occasionally see builds failing due to the source coming from an autogenerated git tarball/zip being different (most commonly on github, but the same applies to most git hosts). This is because the git archives are dependent on the git version used and are not intended to be stable or reproducible.

It would also be nice to have more resistance to changing tags if it was also applied to `git_url`/`git_rev`.

### What should happen?

My first guess would be to have an key like `sha256` in the sources section like `content_sha256` which would trigger the content-based hash to be checked.

### Additional Context

Mixpkgs solved this quite a long time ago with their [`fetchzip`](https://nixos.org/manual/nixpkgs/stable/#fetchurl) funcion so that would probably be a valuable source of ideas on the implementation.

jaimergp:
This would be handy if we ever start storing sources along the built artifacts. One thing to take care of is the local conda-build cache. `sha256` is used to see if the artifact is already available, so I guess with `content_sha256` we wouldn't be able to provide a cache. This would be a good option for github dynamic archives and others, but we don't need it for stable artifacts, which could still use `sha256`.

wolfv:
We could also record a content hash in the "finalized sources" section of the "rendered recipe" for the new recipe format: https://github.com/conda-incubator/ceps/pull/74

PR #4793, opened 2023-03-05
Support SHA224, SHA384 and SHA512 source hashes.

tttc3:
### Description
Provides support for validating source checksums via SHA224, SHA384 and SHA512, in addition to the existing MD5, SHA1 and SHA256 options. A similar request has been made for `boa` (see [here](https://github.com/mamba-org/boa/issues/288)).

In addition, when multiple checksums are provided in the recipe, a `RuntimeError` will now be raised if any of the checksums are wrong. Previously, this error would not necessarily be raised if one of the provided checksums was valid (if this was an intentional design choice then this part can be reverted/excluded from the pull). 

Documentation and appropriate test-recipes have been updated to include examples of using the new checksums. 

### Checklist - did you ...
- [x] Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release notes?
- [x] Add / update necessary tests?
- [x] Add / update outdated documentation?

PR #5277, opened 2024-04-12
Add new source hashing methods: `content_sha256`, `content_sha384`, `content_sha512`

jaimergp:
### Description

- Closes https://github.com/conda/conda-build/issues/4762.
- Relevant CEP submitted at https://github.com/conda/ceps/pull/100.

### Checklist - did you ...


wolfv:
I think this is cool. It would also work nicely with the new proposal for "rendered recipes" (https://github.com/conda-incubator/ceps/pull/74).

On that note - should we continue adding features to conda-build without any standardization (e.g. CEP) process?

jaimergp:
> should we continue adding features to conda-build without any standardization (e.g. CEP) process?

I'm planning to submit a CEP. I opened this draft to explore what kind of things are needed for a stable yet robust logic, cross platform. Things like permissions and so on don't translate well to Windows.

wolfv:
Awesome. Yeah, I also recently looked at a few content hash implementations in Rust but didn't find anything super convincing yet. There are a bunch though (https://crates.io/search?q=content%20hash)

jaimergp:
So far the scheme I followed looks a lot like https://github.com/DrSLDR/dasher?tab=readme-ov-file#hashing-scheme. Things to standardize would be how the tree is sorted, the normalization of the path, the separators ([to prevent this](https://flawed.net.nz/2018/02/21/attacking-merkle-trees-with-a-second-preimage-attack/)), and the allowed algorithms.

I've seen a few [merkle tree based](https://github.com/Tierion/pymerkletools) packages but we don't need all the proof stuff, or leaf querying; just comparing the root hash.

Maybe it could be implemented in a recursive way that doesn't involve obtaining the whole file tree beforehand if that increases performance or simplifies implementation elsewhere. IMO this feels like one of those CEPs that does require prototyping first to see which things have to be standardized.

conda_build/utils.py
```diff
@@ -2045,10 +2045,20 @@ def compute_content_hash(
             hasher.update(b"D")
         elif path.is_file():
             hasher.update(b"F")
+            # We need to normalize line endings for Windows-Unix compat
```
dholth, line 2048: !

jaimergp:
[`16260bf` (#5277)](https://github.com/conda/conda-build/pull/5277/commits/16260bf433e4b8110916c29df0b868873ac0f4c3) was a tricky find. Turns out that `pathlib.Path` is sorted by components (i.e. like a tuple of str), not by their str representation (see https://github.com/python/cpython/issues/76221). This somehow causes differences in sorts across platforms due to case sensitivity: https://www.diffchecker.com/69s9br4y/.

image

Anyway, the fix is easy: use `key=str` when sorting sequences of `pathlib.Path`, but phew, took that long...

conda_build/source.py
```diff
@@ -347,7 +352,7 @@ def git_mirror_checkout_recursive(
             print(f"checkout: {checkout!r}")
         if checkout:
             check_call_env(
-                [git, "checkout", checkout],
+                [git, "checkout", "-f", checkout],  # UNDO
```
jaimergp, line 355: This is needed to avoid a weird `git checkout` error with the `conda-build=1.8.1` clone test. Not sure why it detects that the repo has changes, possibly introduced by another test somewhere else. I did try to change the git clone cache but still failed. I'll try to take another look but UUUUGH conda-build.

jaimergp: Gave up and switched to constructor for this test.

jaimergp:
@conda/builds-tools this is ready for review. Please take a look to https://github.com/conda/ceps/pull/100 too, which standardizes the basics of the algorithm followed here.

conda_build/source.py
```diff
@@ -1109,6 +1115,25 @@ def provide(metadata):
...
+                        )
+                    algorithm = hash_type[len("content_") :]
+                    obtained_content_hash = compute_content_hash(
+                        src_dir,
+                        algorithm,
+                        skip=ensure_list(source_dict.get("content_hash_skip") or ()),
+                    )
+                    if expected_content_hash != obtained_content_hash:
+                        raise RuntimeError(
+                            f"{hash_type} mismatch in source item #{idx}: "
```
beckermr, line 1133: Is it not possible to list the source url here?

conda_build/utils.py
```diff
@@ -1987,6 +1987,93 @@ def sha256_checksum(filename, buffersize=65536):
...
+          encoding.
+    - UTF-8 encoded bytes of the string `-`, as separator.
+
+    Parameters
+    ----------
+    directory: The path whose contents will be hashed
+    algorithm: Name of the algorithm to be used, as expected by `hashlib.new()`
+    skip: iterable of paths that should not be checked. If a path ends with a slash, it's
+          interpreted as a directory that won't be traversed. It matches the relative paths
+          already slashed-normalized.
```
beckermr, line 2015: We should specify we mean backwards to forwards normalization.

conda_build/utils.py
```diff
@@ -1987,6 +1987,93 @@ def sha256_checksum(filename, buffersize=65536):
...
+                        hasher.update(line.encode("utf-8"))
+                except UnicodeDecodeError:
+                    # file must be binary, read the bytes directly
+                    with open(path, "rb") as fh:
+                        for chunk in iter(partial(fh.read, 8192), b""):
+                            hasher.update(chunk)
+            except OSError as exc:
+                log.debug("Skipping %s for hashing", path.name, exc_info=exc)
+        else:
+            log.debug("Can't detect type for path %s. Skipping...", path)
```
beckermr, line 2072: For these two cases is it really optimal to simply skip? Maybe we could hash the pathname instead?

I can imagine nasty edge cases where skipping yields the same hash for two things when it should not.

beckermr (changes requested):
Some comments.

conda_build/source.py
jaimergp: There are too many possibilities: `git_url`, `url`, `path`, `svn`... And we would have to adjust that code to cover for the potential URL origin. I think it's one of those rare cases where it pays to be a bit lazy 😬

conda_build/utils.py
jaimergp: Hm, good point, let me give it a try with new separators too.

wolfv:
Actually - do we need anything more than `content_sha256`? IIRC MD5 is not safe anyways. And I don't see many people use SHA1 (does it have advantages over sha256?).

beckermr:
SHA1 is faster for sure, but yeah, no reason to have MD5 at all. Its support is for legacy purposes AFAIK. I also wonder if SHA1 is secure, but IDK either way.

schuylermartin45:
> Actually - do we need anything more than `content_sha256`? IIRC MD5 is not safe anyways. And I don't see many people use SHA1 (does it have advantages over sha256?).

MD5 has had known collision vulnerabilities for a long time now: https://en.wikipedia.org/wiki/MD5
SHA-1 had a collision vulnerability discovered in 2017: https://security.googleblog.com/2017/02/announcing-first-sha1-collision.html

conda_build/source.py
beckermr: I wonder how often we have ever added url types, but ok fair enough.

beckermr:
Thanks @schuylermartin45. I am happy to leave MD5 and SHA1 in the code, but yeah it looks like we should only use SHA256 anyways. 

We probably should start deprecation cycles for the other two as well?

schuylermartin45:
Reading further, as of 2020, it takes government-level resources to compute the SHA-1 collision. But probably still not worth it.

conda_build/utils.py
jaimergp: Actually we were already hashing the path, but I didn't add a separator for unknown file types. Used `?` now.

wolfv:
As we are the ones hashing (and validating the hash) as opposed to take published hash values from somewhere, I don't see a reason to have any hash other than SHA256.

In fact, I would be fine to _only_ have a `content_hash` field that does not include the hash type at all.

conda_build/utils.py
beckermr: ah great. sorry I missed the path was already in there. Thanks!

jaimergp:
I'd rather keep the algorithm suffix. See also https://github.com/conda/conda-build/pull/4793 for other algos we might want to support. It takes very little code to add (or remove) algos to the scheme. As long as it's accepted by [`hashlib.new()`](https://docs.python.org/3/library/hashlib.html#hashlib.new), that is.

beckermr:
@wolfv one day far into the future we will regret not specifying the hash type in the name of the hash field. I cannot imagine the pain we'd be in right now if we had done that with MD5. 😨

wolfv:
Well I still don't see a point in adding additional algorithms beyond "it's easy" - nobody should use them.

beckermr:
@jaimergp Is it possible to adjust the CEP to allow any hash supported by the python stdlib?

Having to vote over and over again on adding stuff seems rather annoying?

jaimergp:
> Is it possible to adjust the CEP to allow any hash supported by the python stdlib?

The CEP doesn't even mention which hash should be used. It's a scheme for hashing directories with an algorithm of your choice.

wolfv:
Also how do we envision this to be used?

I have two use cases in mind for rattler-build:

- After the fact, record the content-hash for future re-builds as extra-assurance in case the SHA256 of the artifact becomes outdated.
- We could use it as an input alternative to `sha256` in the URL field. However, to create the content hash, I would add a CLI option in rattler-build that would look something like:

`rattler-build prepare-source --version 0.1.2 --recipe ./recipe.yaml` and then the content hash gets injected or printed to the CLI.

jaimergp:
> Well I still don't see a point in adding additional algorithms beyond "it's easy" - nobody should use them.

My point is that it's trivial to change this at any point during the review process; I'm more concerned with the actual hashing scheme proposed. The keys being added to the `meta.yaml` are secondary to that. For clarity, I'm fine with just `content_sha256` and nothing else. Is there a reason an old system wouldn't have access to sha256? They'd have bigger problems, right?

jaimergp:
> I would add a CLI option in rattler-build that would look something like [...]

Hm, true, we could add a little subcommand here to make it easier. Although honestly, I usually run `conda-build`, wait for the hashing mismatch error, and then copy the correct one 😬  (That's why I amended the text in the errors hah).

beckermr (dismissed):

jaimergp:
https://github.com/conda/conda-build/pull/4793 will probably land soon. I'll update this branch with it once it reaches main and then remove the weak hashes. If we are that concerned about `md5` and `sha1` being used, we should also study the possibility of deprecating them or at least warning about them in the logs.

beckermr:
We'll need a lot of advanced notice to deprecate them on conda-forge. For now we should probably add a lint + minimigrator to move to sha256.

wolfv:
I am not concerned, I just don't understand why we want to add them? IMO it doesn't add value. Having the MD5 hash available for the regular hash makes sense because some pacakges might publish the known-good value (and that can be used in the recipe), but for something that we have invented ourselves I don't see a use-case where it is justified to use anything other than the best available hash.

beckermr (dismissed):
LGTM! We should probably vote on the CEP before merging.

jezdez:
Setting as blocked on the CEP vote

conda_build/utils.py
```diff
@@ -1987,6 +1987,100 @@ def sha256_checksum(filename, buffersize=65536):
...
+        elif path.is_dir():
+            hasher.update(b"D")
+        elif path.is_file():
+            hasher.update(b"F")
+            # We need to normalize line endings for Windows-Unix compat
+            # Attempt normalized line-by-line hashing (text mode). If
+            # Python fails to open in text mode, then it's binary and we hash
+            # the raw bytes directly.
+            try:
+                try:
```
jezdez, line 2057: That second try block isn't needed, since we can catch multiple exception types in the same block

conda_build/utils.py
```diff
@@ -1987,6 +1987,100 @@ def sha256_checksum(filename, buffersize=65536):
...
+    skip: iterable of paths that should not be checked. If a path ends with a slash, it's
+          interpreted as a directory that won't be traversed. It matches the relative paths
+          already slashed-normalized (i.e. backwards slashes replaced with forward slashes).
+
+    Returns
+    -------
+    str
+        The hexdigest of the computed hash, as described above.
+    """
+    log = get_logger(__name__)
```
jezdez, line 2026: Let's move that to the top of the module per best practice

conda_build/utils.py
```diff
@@ -1987,6 +1987,100 @@ def sha256_checksum(filename, buffersize=65536):
...
+            try:
+                try:
+                    lines = []
+                    with open(path) as fh:
+                        for line in fh:
+                            # Accumulate all line-ending normalized lines first
+                            # to make sure the whole file is read. This prevents
+                            # partial updates to the hash with hybrid text/binary
+                            # files (e.g. like the constructor shell installers).
+                            lines.append(line.replace("\r\n", "\n"))
```
jezdez, line 2065: Hmm, that might be a memory hog, depending on how big the files are you're normalizing, it might be best to write this to a temp file

conda_build/utils.py
```diff
@@ -1987,6 +1987,100 @@ def sha256_checksum(filename, buffersize=65536):
...
+                    with open(path, "rb") as fh:
+                        for chunk in iter(partial(fh.read, 8192), b""):
+                            hasher.update(chunk)
+            except OSError as exc:
+                log.warning(
+                    "Can't open file %s. Hashing path only...", path.name, exc_info=exc
+                )
+        else:
+            log.warning("Can't detect type for path %s. Hashing path only...", path)
+            hasher.update(b"?")
```
jezdez, line 2079: I'm not completely following this error state handling, why doesn't this stop the process since it can't read the file? Wouldn't that indicate that the recipe is faulty?

jezdez (changes requested):

conda_build/utils.py
jaimergp: We don't know what kind of files a user will have in that directory. They might point `path` to something containing a device file or who knows what. Not really common practice, but that doesn't mean that their source is invalid or that we can't verify that the other stuff is actually the same.

conda_build/utils.py
jaimergp: Hm, good point. Didn't the stdlib have a temporary file object that only writes to disk after a certain size? 🤔

conda_build/utils.py
jaimergp: I was following the practice followed in the other functions, fwiw.

jaimergp: Ah, now I see why. The `get_logger` utility is defined in that module, so there's no top-level function to use. If anything, it would go at the bottom of the module? Do you prefer that or shall we leave it in-function?

conda_build/utils.py
jaimergp: Sure, but I want to catch the potential OSError in the `except UnicodeDecodeError` arm. Will that raised exception be caught in the try/except block? IOW, will this print "Hello"? I don't think it does:

```python
try:
  raise ValueError
except ValueError:
  raise RuntimeError
except RuntimeError:
  print("Hellow!")
```

conda_build/utils.py
jaimergp: Yea, `SpooledTemporaryFile`. Added it in [`c192799` (#5277)](https://github.com/conda/conda-build/pull/5277/commits/c192799513f83d5715656c8e5030bdaf9234939d)

conda_build/utils.py
jaimergp: I made this error out because it's essentially a file we can't verify, and we don't know what it might be hiding. If it causes errors, users can deliberately skip it via the `skip` parameter.

jaimergp:
@jezdez, the CEP passed, so I guess [this review](https://github.com/conda/conda-build/pull/5277#issuecomment-2504790547) can be dismissed now.

beckermr (approved):
Some comments.

conda_build/utils.py
```diff
@@ -2139,3 +2237,6 @@ def is_conda_pkg(pkg_path: str) -> bool:
 
 def package_record_to_requirement(prec: PackageRecord) -> str:
     return f"{prec.name} {prec.version} {prec.build}"
+
+
+log = get_logger(__name__)
```
jezdez, line 2242: Why is this at the bottom?

jezdez (dismissed):

conda_build/utils.py
jaimergp: Because the function `get_logger` is only defined mid-module, but I think we don't need this `log` object after all. Every function is initializing their own logger anyway (yes, I think that has a non-negligible cost, but that's a subject for other PR). I'll remove it.

jaimergp:
CI errors are unrelated. Investigating in https://github.com/conda/conda-build/pull/5577.

PR #5253, opened 2024-03-25
Remove duplicate `extract_package_and_build_text` call

kenodegard:
### Description

Looking to see just how much of a speedup we get by removing this redundant line.

Extracted from:
- https://github.com/conda/conda-build/pull/5225
- https://github.com/conda/conda-build/pull/5237
- https://github.com/conda/conda-build/pull/5248

### Checklist - did you ...

- [ ] ~Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release notes?~
     
- [ ] ~Add / update necessary tests?~
- [ ] ~Add / update outdated documentation?~

PR #5281, opened 2024-04-12
Use pickle's loads/dumps as a faster deepcopy

kenodegard:
### Description

Attempting to use `pickle` instead of `deepcopy` for a faster copy.

Split from: 
- https://github.com/conda/conda-build/pull/5225
- https://github.com/conda/conda-build/pull/5248
- https://github.com/conda/conda-build/pull/5253

### Checklist - did you ...


beckermr:
If there is a chance that anything here does not pickle, then you'd want a try except with fallback to deepcopy. IDK how likely that is though.

beckermr (commented):
My only comment here is as last time. If we encounter a data structure that does not pickle, we will get an error. It might be worth making a separate helper function with a try except block falling back to deepcopy.

conda_build/config.py
```diff
@@ -817,9 +818,10 @@ def clean_pkgs(self):
 
     def copy(self):
         new = copy.copy(self)
-        new.variant = copy.deepcopy(self.variant)
+        # Use picke.loads(pickle.dumps(...) as a faster copy.deepcopy alternative.
+        new.variant = pickle.loads(pickle.dumps(self.variant, -1))
```
dholth, line 822: Could also be spelled as `pickle.HIGHEST_PROTOCOL`

"If a negative number is specified, [HIGHEST_PROTOCOL](https://docs.python.org/3/library/pickle.html#pickle.HIGHEST_PROTOCOL) is selected."

conda_build/config.py
```diff
@@ -820,9 +821,10 @@ def clean_pkgs(self):
 
     def copy(self) -> Config:
         new = copy.copy(self)
-        new.variant = copy.deepcopy(self.variant)
+        # Use picke.loads(pickle.dumps(...) as a faster copy.deepcopy alternative.
+        new.variant = pickle.loads(pickle.dumps(self.variant, -1))
```
beckermr, line 825:
```suggestion
        new.variant = pickle.loads(pickle.dumps(self.variant, pickle.HIGHEST_PROTOCOL))
```

conda_build/config.py
```diff
@@ -820,9 +821,10 @@ def clean_pkgs(self):
...
         if hasattr(self, "variants"):
-            new.variants = copy.deepcopy(self.variants)
+            new.variants = pickle.loads(pickle.dumps(self.variants, -1))
```
beckermr, line 827:
```suggestion
            new.variants = pickle.loads(pickle.dumps(self.variants, pickle.HIGHEST_PROTOCOL))
```

beckermr:
pre-commit.ci autofix

beckermr:
@kenodegard @dholth Are we worried about errors where something cannot be pickled or shall we merge this one? The test failures appear to be unrelated warnings appearing w/ python 3.12.

minrk:
> Are we worried about errors where something cannot be pickled

Since this is currently only applied to `variant` objects and I _think_ variants are necessarily dicts of lists, strings and numbers, I don't think this is likely to come up.

beckermr (approved):
Let's add a news item and then merge!

beckermr (approved):

jaimergp:
> > DeprecationWarning: Python 3.14 will, by default, filter extracted tar archives and reject files or modify their metadata. Use the filter argument to control this behavior.
>
> I don't follow it totally but it appears to happen in extractions of packages.

Yep, this is the `tarfile` usage in `conda-package-streaming`. Added https://github.com/conda/conda-package-streaming/issues/87 to track. For now we can add a filterwarnings setting to `pyproject.toml`, I'd say. I think @zklaus was looking into it.

beckermr:
> filterwarnings setting to pyproject.toml

do you have an example of this or a link to the docs?

I am not following.

pyproject.toml
```diff
@@ -129,6 +129,7 @@ filterwarnings = [
   # ignore conda-index error
   "ignore::PendingDeprecationWarning:conda_index",
   "ignore::DeprecationWarning:conda_index",
+  "ignore::DeprecationWarning:tarfile",
```
beckermr, line 132:
```suggestion
  "ignore::DeprecationWarning: Python 3.14",
```

beckermr:
pre-commit.ci autofix

pyproject.toml
```diff
@@ -129,6 +129,7 @@ filterwarnings = [
   # ignore conda-index error
   "ignore::PendingDeprecationWarning:conda_index",
   "ignore::DeprecationWarning:conda_index",
+  "ignore:Python 3\.14 will, by default, filter extracted tar archives:DeprecationWarning:tarfile",
```
beckermr, line 132:
```suggestion
  "ignore:Python 3\\.14 will, by default, filter extracted tar archives:DeprecationWarning:tarfile",
```

pyproject.toml
```diff
@@ -129,6 +129,7 @@ filterwarnings = [
   # ignore conda-index error
   "ignore::PendingDeprecationWarning:conda_index",
   "ignore::DeprecationWarning:conda_index",
+  'ignore:\s+Python 3\.14 will, by default, filter extracted tar archives:DeprecationWarning:tarfile',
```
beckermr, line 132:
```suggestion
  'ignore:\s+Python 3.14 will, by default, filter extracted tar archives:DeprecationWarning:tarfile',
```

pyproject.toml
```diff
@@ -129,6 +129,7 @@ filterwarnings = [
   # ignore conda-index error
   "ignore::PendingDeprecationWarning:conda_index",
   "ignore::DeprecationWarning:conda_index",
+  'ignore:\s+Python 3.14 will, by default, filter extracted tar archives:DeprecationWarning:tarfile',
```
beckermr, line 132:
```suggestion
  'ignore:\s+Python:DeprecationWarning:tarfile',
```

pyproject.toml
```diff
@@ -129,6 +129,7 @@ filterwarnings = [
   # ignore conda-index error
   "ignore::PendingDeprecationWarning:conda_index",
   "ignore::DeprecationWarning:conda_index",
+  'ignore:\s+Python:DeprecationWarning',
```
beckermr, line 132:
```suggestion
  "ignore:Python 3.14 will, by default, filter extracted tar archives and reject files or modify their metadata:DeprecationWarning:tarfile",
```

pyproject.toml
```diff
@@ -129,6 +129,7 @@ filterwarnings = [
   # ignore conda-index error
   "ignore::PendingDeprecationWarning:conda_index",
   "ignore::DeprecationWarning:conda_index",
+  "ignore::DeprecationWarning",
```
beckermr, line 132:
```suggestion
  "ignore::DeprecationWarning:tarfile",
```

pyproject.toml
beckermr, line 132:
```suggestion
  "ignore::DeprecationWarning",
```

pyproject.toml
beckermr, line 132:
```suggestion
  "ignore::DeprecationWarning:tarfile.py",
```

pyproject.toml
```diff
@@ -129,6 +129,7 @@ filterwarnings = [
   # ignore conda-index error
   "ignore::PendingDeprecationWarning:conda_index",
   "ignore::DeprecationWarning:conda_index",
+  "ignore::DeprecationWarning:tarfile.py",
```
beckermr, line 132:
```suggestion
  "ignore:Python:DeprecationWarning",
```

beckermr (approved):

minrk:
Thanks for this! I've been testing it out, and with this and #5384 rerender time is reduced by ~90%, from over 30 minutes to about 4 on some recipes with a large variant matrix (petsc4py).

FWIW, I spent a little time looking into making `config.variant` itself an immutable dict via `deepfreeze` since that's already computed for the hash anyway and would eliminate the need to copy variants altogether. But there are a few places where the variant is modified in place that make it a little bit tricky. Still doable and perhaps right thing to do, but not so simple. I think it also comes at an increased cost of computing `list_of_dicts_to_dict_of_lists` because there would end up being a lot of converting tuples to lists and back.

I also noticed in https://github.com/conda-forge/conda-smithy/pull/1967 that deepfreeze is quite a lot more expensive than using HashableDict (deepfreeze takes ~10x the time of `hash(HashableDict(dict))`

PR #5284, opened 2024-04-13
Deprecate `HashableDict` in favor of `frozendict`

kenodegard:
### Description

Instead of using a customized `HashableDict` implementation we can just use the `frozendict` library (which `conda` already depends on). Using `frozendict` also means we can avoid calling `deepcopy` (`frozendict` takes care of this for us).

### Checklist - did you ...

### DIFF
--- a/conda_build/jinja_context.py
+++ b/conda_build/jinja_context.py
@@ -10,24 +10,18 @@ import re
 import time
 from functools import partial
 from io import StringIO, TextIOBase
+from typing import TYPE_CHECKING
 from warnings import warn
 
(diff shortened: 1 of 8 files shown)


kenodegard:
The current version of `frozendict.deepfreeze` mutates the original object. This is undesired behavior and needs to be corrected. Hopefully it can be addressed upstream, see https://github.com/Marco-Sulla/python-frozendict/issues/96 & https://github.com/Marco-Sulla/python-frozendict/issues/97.

beckermr (approved):

PR #5333, opened 2024-05-10
[WIP] Remove deprecated code for conda-build 24.7

beeankha:
🚧 _this work is not yet ready for review_ 🚧

### Description

This PR removes everything marked as deprecated in conda-build `24.7.x`

### Checklist - did you ...


news/5333-remove-24.7.x-deprecations
kenodegard, line 1: Confirmed this contains all of the removals, now that it's been double checked can we sort them alphabetically?

conda_build/index.py
```diff
@@ -29,7 +28,6 @@
 local_output_folder = ""
 cached_channels = []
 _channel_data = {}
-deprecated.constant("24.1", "24.7", "channel_data", _channel_data)
```
kenodegard, line 32: `channel_data` was previously postponed since the global variable (`_channel_data`) was still in use. What would it take for us to remove this global variable and instead only use a local variable?

kenodegard (changes requested):

news/5333-remove-24.7.x-deprecations
```diff
@@ -0,0 +1,95 @@
...
+
+* Remove `conda_build.conda_interface.Completer`. (#5333)
+* Remove `conda_build.conda_interface.CondaSession`. Use `conda.gateways.connection.session.CondaSession` instead. (#5333)
+* Remove `conda_build.conda_interface.InstalledPackages`. (#5333)
+* Remove `conda_build.conda_interface.NoPackagesFound`. Use `conda.exceptions.ResolvePackageNotFound` instead. (#5333)
+* Remove `conda_build.conda_interface.Unsatisfiable`. Use `conda.exceptions.UnsatisfiableError` instead. (#5333)
+* Remove `conda_build.conda_interface.symlink_conda`. (#5333)
+* Remove `conda_build.conda_interface.ArgumentParser`. Use `conda.cli.conda_argparse.ArgumentParser` instead. (#5333)
+* Remove `conda_build.conda_interface.add_parser_channels`. Use `conda.cli.helpers.add_parser_channels` instead. (#5333)
+* Remove `conda_build.conda_interface.add_parser_prefix`. Use `conda.cli.helpers.add_parser_prefix` instead. (#5333)
```
jakirkham, line 19: Should we link out to a doc, summarize or streamline the list a bit?

An example of streamlining:

```markdown
### Deprecations

* Remove `conda_build.conda_interface.*` component. (#5333):
  * `Completer`
  * ...
```

beeankha: That's a great idea!

news/5333-remove-24.7.x-deprecations
beeankha: Sounds good. I'll also try to make the news file easier to read like @jakirkham [suggested](https://github.com/conda/conda-build/pull/5333/#discussion_r1599342527)

news/5333-remove-24.7.x-deprecations
```diff
@@ -0,0 +1,95 @@
...
+  * `conda_build.conda_interface.url_path` (use `conda.utils.url_path` instead)
+  * `conda_build.conda_interface.VersionOrder` (use `conda.models.version.VersionOrder` instead)
+  * `conda_build.conda_interface.walk_prefix` (use `conda.misc.walk_prefix` instead)
+  * `conda_build.conda_interface.win_path_to_unix` (use `conda.common.path.win_path_to_unix` instead)
+  * `conda_build.index.channel_data`
+  * `conda_build.utils._convert_lists_to_sets` (use `frozendict.deepfreeze` instead)
+  * `conda_build.utils.HashableDict` (use `frozendict.deepfreeze` instead)
+  * `conda_build.utils.represent_hashabledict` (use `frozendict.deepfreeze` instead)
+  * `conda_build.utils.rm_rf(config)`
+  * `conda_build.variants.get_vars(loop_only)`
```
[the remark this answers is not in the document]
kenodegard: Personally I don't see much of a difference, it's still hard to read but that's simply a product of a lot of deprecations. We could truncate all of the `conda_interface` deprecations into ```Removed `conda_build.conda_interface`.``` but we'd loose the more detailed information about what users should be doing instead.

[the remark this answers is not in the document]
jakirkham: This seems reasonable

Was starting to wonder if we should capture this sort of thing in a linter tool (like `ruff` or similar). That would help us apply these changes throughout the different Conda* codebases

Maybe we would also be less reliant on having this in the release notes

conda_build/index.py
```diff
@@ -56,9 +53,10 @@ def get_build_index(
     global local_output_folder
     global cached_index
     global cached_channels
-    global _channel_data
     mtime = 0
 
+    _channel_data = {}
```
beeankha, line 58: Hi @mbargull ! Wanted to ping you directly per your work in https://github.com/conda/conda-build/pull/5152 marking the `channel_data` constant for deprecation; does this update to `_channel_data` (removing the global variable and making this a local variable instead) look OK to you?

mbargull: No quite -- meaning, making it local doesn't take it far enough ;).
Sorry, I should've added a code comment there:
The idea was to remove it altogether since we populate it, but don't actually use it as of now.
Meaning, we should
1. remove everything from `expanded_channels = ...` (currently line 132) until `_channel_data["defaults"] = ...` (currently line 175), and
2. either omit the third return value or just return `None` instead.

As for point 2: I did a GitHub code search for orgs `conda`, `conda-forge`, `conda-incubator`, `bioconda`, `mamba-org` and could not find any usage of `get_build_index` outside of `conda-build` itself, so removing the return value is probably fine (returning `None` instead is a more defensive variant which wouldn't break uses of `index, ts, _ = get_build_index(...)`.

kenodegard: oooh! that makes much more sense!

yeah I'd just return an empty dict then so we don't modify the return structure/type

mbargull: This would, for one, of course remove unused code, but it would also avoid iterating over the index in `expanded_channels = {rec.channel for rec in cached_index}`.
On its own this doesn't yet change much really, since the index returned by `conda.core.index.get_index`/`conda.core.index.fetch_index` already has been iterated over (`fetch_index` does `index.update((rec, rec) for rec in f.iter_records())`).
But if/when we fix that initial iteration whilst addressing gh-5154, we can finally make use of deferred package record instantiation and shave off a good 30s and a whole lot of memory consumption.
Leaving the unused `channel_data` population in place would hinder that work, of course.

mbargull: > yeah I'd just return an empty dict then so we don't modify the return structure/type

I thought about doing this too -- but I think a `None` would be a bit better since it would at least intentionally break uses of the then invalid `channel_data` (i.e., avoid silent errors in case this is really used somewhere).
N.B.: All uses in `conda-build` itself just ignore that return value anyway.

news/5333-remove-24.7.x-deprecations
```diff
@@ -0,0 +1,95 @@
...
+  * `conda_build.conda_interface._toposort` (use `conda.common.toposort._toposort` instead)
+  * `conda_build.conda_interface.unix_path_to_win` (use `conda.utils.unix_path_to_win` instead)
+  * `conda_build.conda_interface.untracked` (use `conda.misc.untracked` instead)
+  * `conda_build.conda_interface.Unsatisfiable` (use `conda.exceptions.UnsatisfiableError` instead)
+  * `conda_build.conda_interface.UnsatisfiableError` (use `conda.exceptions.UnsatisfiableError` instead)
+  * `conda_build.conda_interface.url_path` (use `conda.utils.url_path` instead)
+  * `conda_build.conda_interface.VersionOrder` (use `conda.models.version.VersionOrder` instead)
+  * `conda_build.conda_interface.walk_prefix` (use `conda.misc.walk_prefix` instead)
+  * `conda_build.conda_interface.win_path_to_unix` (use `conda.common.path.win_path_to_unix` instead)
+  * `conda_build.index.channel_data`
```
mbargull, line 82:
```suggestion
  * `conda_build.index.channel_data`
  * `conda_build.index.get_build_index` return value for `channel_data` is now always `None`
```
or something along those lines maybe?

mbargull (approved):

news/5333-remove-24.7.x-deprecations
beeankha: Oh good call, I'll update the news file with this info!

Issue #5267, opened 2024-04-02
Regression: CondaMemoryError when building


mxschmitt:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### What happened?

Looks like this broke in https://github.com/conda/conda-build/pull/522.

```
git clone https://github.com/microsoft/playwright-python
conda config --add channels conda-forge
conda install anaconda-client conda-build=24.3.0 conda-verify py-lief=0.14.1
conda install cross-python_linux-aarch64
conda build . -m conda_build_config.yaml -m conda_build_config_linux_aarch64.yaml
```

Expected: Works
Actual: CondaMemoryError: The conda process ran out of memory. Increase system memory and/or try again.

This machine has 64GB of free memory and works with:

`conda install anaconda-client conda-build=24.1.2 conda-verify py-lief=0.12.3`

so I'm fairly confident, that this is a regression.

cc @mbargul

### Conda Info

```shell
active environment : None
       user config file : /home/codespace/.condarc
 populated config files : /home/codespace/.condarc
          conda version : 24.3.0
    conda-build version : 24.3.0
         python version : 3.12.2.final.0
                 solver : libmamba (default)
       virtual packages : __archspec=1=zen2
                          __conda=24.3.0=0
                          __glibc=2.31=0
                          __linux=6.2.0=0
                          __unix=0=0
       base environment : /opt/conda  (writable)
      conda av data dir : /opt/conda/etc/conda
  conda av metadata url : None
           channel URLs : https://conda.anaconda.org/conda-forge/linux-64
                          https://conda.anaconda.org/conda-forge/noarch
                          https://repo.anaconda.com/pkgs/main/linux-64
                          https://repo.anaconda.com/pkgs/main/noarch
                          https://repo.anaconda.com/pkgs/r/linux-64
                          https://repo.anaconda.com/pkgs/r/noarch
          package cache : /opt/conda/pkgs
                          /home/codespace/.conda/pkgs
       envs directories : /opt/conda/envs
                          /home/codespace/.conda/envs
               platform : linux-64
             user-agent : conda/24.3.0 requests/2.31.0 CPython/3.12.2 Linux/6.2.0-1019-azure ubuntu/20.04.6 glibc/2.31 solver/libmamba conda-libmamba-solver/23.12.0 libmambapy/1.5.3 aau/0.4.4 c/. s/. e/.
                UID:GID : 1000:1000
             netrc file : None
           offline mode : False
```

[...]

PR #5228, opened 2024-03-11
Add compatibility for LIEF=0.14

mbargull:
### Description

Add compatibility for LIEF=0.14
Fixes gh-5227.

### Checklist - did you ...

- [x] Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [ ] Add / update necessary tests?
- [ ] Add / update outdated documentation?

PR #5345, opened 2024-05-18
Fix `frozendict` usage

kenodegard:
### Description

A very specific use case results in the code trying to modify a `frozendict` version of the output dict.

Fix #5342 

### Checklist - did you ...

- [x] Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [x] Add / update necessary tests?
- [ ] ~Add / update outdated documentation?~

PR #5348, opened 2024-05-21
Fix excessive memory use in `inspect_linkages_lief`

mbargull:
### Description

introduced in gh-5228 which replaced `binary.name` with `str(file)`.
The former returns the path to the binary whereas the latter returns the
contents of the parsed binary, naturally causing high memory usage.
That semantic change was unintended (misread upstream changes).
Since binary's path is no longer part of the object, pass it separately.

Fixes gh-5267

### Checklist - did you ...

### DIFF
--- a/conda_build/os_utils/liefldd.py
+++ b/conda_build/os_utils/liefldd.py
@@ -596,7 +596,7 @@ def inspect_linkages_lief(
                     if recurse:
                         if os.path.exists(resolved[0]):
                             todo.append([resolved[0], lief.parse(resolved[0])])
-                already_seen.add(get_uniqueness_key(filename2, binary2))
+                already_seen.add(uniqueness_key)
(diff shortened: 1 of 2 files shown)


kenodegard (approved):

kenodegard:
@beeankha @jezdez since this was originally a 24.3.x regression, how do y'all want to proceed with this?

beeankha:
> @beeankha @jezdez since this was originally a 24.3.x regression, how do y'all want to proceed with this?

I'm currently out of office but if @jezdez is also on board, I can work on a patch release starting on Wednesday May 21st, especially if https://github.com/conda/conda-build/pull/5345 looks good to go and can also be included (otherwise I can do a separate patch release to include that change)

beeankha (approved):

PR #5363, opened 2024-06-05
Use lazy `Index` in conda-build

zklaus:
### Checklist - did you ...

### DIFF
--- a/conda_build/index.py
+++ b/conda_build/index.py
@@ -6,7 +6,7 @@ from functools import partial
 from os.path import dirname
 
 from conda.base.context import context
-from conda.core.index import get_index
+from conda.core.index import Index
 from conda.exceptions import CondaHTTPError
 from conda.utils import url_path
 from conda_index.index import update_index as _update_index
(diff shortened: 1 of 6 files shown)


pyproject.toml
```diff
@@ -124,19 +124,23 @@ filterwarnings = [
...
   "error::DeprecationWarning:conda_build",
+  "ignore:Python 3.14 will, by default, filter extracted tar archives:DeprecationWarning:conda_package_streaming",
   # ignore numpy.distutils error
   'ignore:\s+`numpy.distutils` is deprecated:DeprecationWarning:conda_build._load_setup_py_data',
   # ignore conda-index error
   "ignore::PendingDeprecationWarning:conda_index",
   "ignore::DeprecationWarning:conda_index",
   "ignore:Python 3.14 will, by default, filter extracted tar archives and reject files or modify their metadata:DeprecationWarning",
+  "ignore:conda.core.index._supplement_index_with_system is pending deprecation:PendingDeprecationWarning:conda",
+  "ignore:conda.core.index._make_virtual_package is pending deprecation:PendingDeprecationWarning:conda",
```
jaimergp, line 135: We should fix these warnings instead of ignoring them, I'd say.

conda_build/index.py
```diff
@@ -6,7 +6,7 @@
 from os.path import dirname
 
 from conda.base.context import context
-from conda.core.index import get_index
+from conda.core.index import Index
```
jaimergp, line 9: In principle, we won't need this during the deprecation period, right? So we can release a new conda without this PR and it _should_ work, correct?

zklaus: Yes.

pyproject.toml
zklaus: Well, yes, but these come from conda-libmamba-solver, so I judged that fix outside the scope of this PR.
Of course we can make the fix in a branch there and use that development version here as well.

zklaus:
Hm. Seems I cannot reopen this PR. Too many times? I'll try if commenting fixes that and otherwise open a new PR.

PR #5371, opened 2024-06-12
Replace `sys.exit` calls in `conda_build/metadata.py`

beeankha:
### Description

Replacing `sys.exit` call in a few functions in `conda_build/metadata.py` with the new `CondaBuildUserError` exception for better error handling, along with corresponding unit tests. 

The changes in this PR are separated out from work previously done by @kenodegard in #5255
Xref #4209 

### Checklist - did you ...

- [ ] ~~Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release notes?~~
     
- [ ] ~~Add / update outdated documentation?~~

### DIFF
--- a/conda_build/metadata.py
+++ b/conda_build/metadata.py
@@ -1408,12 +1408,11 @@ class MetaData:
     ):
         """variant contains key-value mapping for additional functions and values
         for jinja2 variables"""
-        # undefined_jinja_vars is refreshed by self.parse again
-        undefined_jinja_vars = ()
         # store the "final" state that we think we're in.  reloading the meta.yaml file
         #   can reset it (to True)
         final = self.final
-        # always parse again at least once.
+
+        # always parse again at least once
         self.parse_again(
             permit_undefined_jinja=True,
             allow_no_other_outputs=allow_no_other_outputs,
@@ -1421,6 +1420,8 @@ class MetaData:
         )
         self.final = final
 
+        # recursively parse again so long as each iteration has fewer undefined jinja variables
+        undefined_jinja_vars = ()
         while set(undefined_jinja_vars) != set(self.undefined_jinja_vars):
             undefined_jinja_vars = self.undefined_jinja_vars
             self.parse_again(
@@ -1429,18 +1430,8 @@ class MetaData:
                 bypass_env_check=bypass_env_check,
             )
             self.final = final
-        if undefined_jinja_vars:
-            self.parse_again(
-                permit_undefined_jinja=False,
-                allow_no_other_outputs=allow_no_other_outputs,
-                bypass_env_check=bypass_env_check,
-            )
-            sys.exit(
-                f"Undefined Jinja2 variables remain ({self.undefined_jinja_vars}).  Please enable "
-                "source downloading and try again."
-            )
 
-        # always parse again at the end, too.
+        # always parse again at the end without permit_undefined_jinja
         self.parse_again(
             permit_undefined_jinja=False,
             allow_no_other_outputs=allow_no_other_outputs,
--- a/conda_build/metadata.py
+++ b/conda_build/metadata.py
@@ -12,12 +12,12 @@ import time
 import warnings
 from collections import OrderedDict
 from functools import lru_cache
-from os.path import isfile, join
+from os.path import isdir, isfile, join
 from typing import TYPE_CHECKING, NamedTuple, overload
 
 import yaml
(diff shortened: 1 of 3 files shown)


travishathaway (approved):

conda_build/metadata.py
```diff
@@ -46,18 +55,12 @@
...
     OutputDict = dict[str, Any]
     OutputTuple = tuple[OutputDict, "MetaData"]
 
-try:
-    import yaml
-except ImportError:
-    sys.exit(
-        "Error: could not import yaml (required to read meta.yaml "
-        "files of conda recipes)"
-    )
```
jezdez, line 60: Was that basically an implicit dependency? Do we require it now in meta.yml?

conda_build/metadata.py
```diff
@@ -350,16 +353,9 @@ def yamlize(data):
     try:
         return yaml.load(data, Loader=StringifyNumbersLoader)
     except yaml.error.YAMLError as e:
-        if "{{" in data:
-            try:
-                import jinja2
-
-                jinja2  # Avoid pyflakes failure: 'jinja2' imported but unused
-            except ImportError:
-                raise exceptions.UnableToParseMissingJinja2(original=e)
```
jezdez, line 359: Just to understand, why are we removing this here? This looks like something 3rd party users might be relying on to identify why this is failing to parse

conda_build/metadata.py
```diff
@@ -767,28 +761,18 @@ def _git_clean(source_meta):
...
 
     # make a copy of the input so we have no side-effects
     ret_meta = source_meta.copy()
-    # loop over the old versions
-    for key, has in zip(git_rev_tags[1:], has_rev_tags[1:]):
-        # update if needed
-        if has:
-            ret_meta[git_rev_tags[0]] = ret_meta[key]
-        # and remove
-        ret_meta.pop(key, None)
```
jezdez, line 791: That seems like an unrelated change to this PR?

conda_build/metadata.py
```diff
@@ -2024,17 +1997,6 @@ def _get_contents(
...
-            import jinja2
-        except ImportError:
-            print("There was an error importing jinja2.", file=sys.stderr)
-            print(
-                "Please run `conda install jinja2` to enable jinja template support",
-                file=sys.stderr,
-            )  # noqa
-            with open(self.meta_path) as fd:
-                return fd.read()
-
```
jezdez, line 2037: Is jinja2 a dependency now?

conda_build/metadata.py
dholth: I'm surprised that it doesn't say "ruamel.yaml" but that package is certainly a regular dependency now.

beeankha: It was previously an optional dependency but is currently [listed as a requirement in `meta.yml`](https://github.com/conda/conda-build/commit/f60aa9a3bee04e0d6d39166dba231786c3b14939#diff-0286c1d624d881984aef773a82433d955692c2fcbc773e7a6230fdb9602a23a3R31).

This particular code change is more "cleanup" vs "remove `sys.exit`" (per work originally done in [this PR](https://github.com/conda/conda-build/pull/5255/files#diff-a3f5613bda9366e31149c65731327047d4e29204c9e0508fd55416115b39a6bdL50-L57)), so I will revert this change; once the broken-out parts of the "remove `sys.exit` calls" work are merged, the [original PR](https://github.com/conda/conda-build/pull/5255) will get rebased and will mostly focus on cleanup.

conda_build/metadata.py
beeankha: Similar to `yaml`/`pyyaml`, `jinja2` used to be an optional dependency but is now [listed as a requirement in `meta.yml`](https://github.com/conda/conda-build/commit/c538eeaf4c23d0c3959a44f881997865f32d7707). However, since this is more "cleanup" vs "remove `sys.exit`", I will revert this change and leave it for the [original PR](https://github.com/conda/conda-build/pull/5255).

conda_build/metadata.py
beeankha: Yep, I'm going to revert this change and leave it for the [original "remove `sys.exit` calls" PR](https://github.com/conda/conda-build/pull/5255) (which got very unwieldy and is thus getting broken up into smaller chunks like this PR.

travishathaway (approved):

Issue #4251, opened 2021-07-20
Deprecate or remove conda develop


jjpr-mit:
The `conda develop` command has been informally deprecated multiple times:  

- https://github.com/conda/conda-build/issues/1992#issuecomment-525777055
- https://github.com/conda/conda-build/issues/2633#issuecomment-359572215

But it is still present [in the code](https://github.com/conda/conda-build/blob/3.21.4/conda_build/cli/main_develop.py) and [in the documentation](https://github.com/conda/conda-build/blob/3.21.4/docs/source/resources/commands/conda-develop.rst).  That is very likely misleading to anyone trying to learn best practices for scientific computing.  

Example scenario:  As soon as a new scientific programmer gets comfortable enough with Python to start working on multiple projects, the next logical step is to learn how to manage environments and dependencies.  They'll start by adding a `setup.py` with dependencies to each project, and most likely they'll start using conda to manage environments.  (That's what I recommend to users of the servers I manage.) They'll create a conda environment, activate it, and then run `pip install -e .`, and the code from their project and all its dependencies will be available to anything they run in that environment.  If they use Jupyter, they'll install that environment as a kernel.  

And then, as development progresses, they'll add more stuff.  They'll `conda install` more packages.  And now they have a problem:  as described in [the current documentation](https://github.com/conda/conda/blob/4.10.2/docs/source/user-guide/tasks/manage-environments.rst#using-pip-in-an-environment), it's dangerous to interleave `conda install` and `pip install`.  Inevitably they'll end up with an environment that's inconsistent, un-upgradeable and unusable.  

On their next project, they'll try to avoid that problem.  The logical solution is to use whatever is the equivalent of `pip install -e` in conda.  Which will lead them to `conda develop`, which looks like the solution to their problem.  They will then waste a day or more trying to get `conda develop` to work, eventually stumbling across #1992 or #2633.  And then they will curse the name of the person who exhorted them to use conda and follow best practices.  Which in several cases has been me.  

Please deprecate or remove `conda develop`.

astrojuanlu:
I arrived here after trying to figure out what `conda develop` is because it's mentioned in the SciPy developer documentation. If it's not maintained, the best thing to do is indeed to deprecate and remove it.

> The logical solution is to use whatever is the equivalent of `pip install -e in conda`.

[...]

PR #5380, opened 2024-06-18
rough outline of a "conda develop" using pypa/build

dholth:
### Description

This is how to create and install an editable wheel with pypa/build, as a replacement for "conda develop". At each step we have the opportunity to install dependencies using our preferred installer instead of pip.

#4251 

### Checklist - did you ...


dholth:
This is what [pypa/build's cli does](https://github.com/pypa/build/blob/main/src/build/__main__.py#L125) to build an ordinary wheel in its own environment, providing pip or uv as the available installers. We could add a conda installer.

```
def _build_in_isolated_env(
    srcdir: StrPath,
    outdir: StrPath,
    distribution: Distribution,
    config_settings: ConfigSettings | None,
    installer: _env.Installer,
) -> str:
    with DefaultIsolatedEnv(installer=installer) as env:
        builder = ProjectBuilder.from_isolated_env(env, srcdir)
        # first install the build dependencies
        env.install(builder.build_system_requires)
        # then get the extra required dependencies from the backend (which was installed in the call above :P)
        env.install(builder.get_requires_for_build(distribution, config_settings or {}))
        return builder.build(distribution, outdir, config_settings or {})
```

conda_build/develop.py
```diff
@@ -180,3 +180,106 @@ def execute(
...
+
+    if isinstance(recipe_dirs, str):
+        recipe_dirs = [recipe_dirs]
+
+    for path in recipe_dirs:
+        pkg_path = abspath(expanduser(path))
+
+        if uninstall:
+            raise NotImplementedError("nope")
+            # uninstall then exit - does not do any other operations
```
dholth, line 221: Ask them to use conda uninstall or pip uninstall since these are installed like standard Python packages

dholth: Could we convert the ephemeral wheel into a conda package and install that instead? Or would it be too much trouble to avoid keeping that in conda's package cache?

PR #5387, opened 2024-06-24
ci: update to latest intel osx runner and download 10.15 SDK

beckermr:
### Description

This PR updates the CI to the latest intel osx runner and to download the 10.15 SDK.

I have preserved the older option of directly using the system SDK via xcode. This should make it easier for folks to work locally in some cases without doing their own SDK download.

Resolves https://github.com/conda/conda-build/issues/5378

### Checklist - did you ...

- [x] Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [x] Add / update necessary tests?
- [x] Add / update outdated documentation?

PR #5379, opened 2024-06-18
Create fixture to fetch MacOSX SDKs

kenodegard:
### Description

The `macos-11` GitHub runner is deprecated and slated to be discontinued on June 28th. GitHub appears to have begun their customary brownouts to aid in detecting any remaining usages.

With inspiration from conda-forge's SDK download script (https://github.com/conda-forge/conda-forge-ci-setup-feedstock/blob/f3c9863319df2c75f05d378df47b2154473e46a4/recipe/download_osx_sdk.sh) we have created a pytest fixture to download (and cache) the necessary MacOS SDK.

Resolves #5378

### Checklist - did you ...

- [ ] Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [x] Add / update necessary tests?
- [ ] ~Add / update outdated documentation?~

PR #5390, opened 2024-06-25
Use `conda-package-handling.api.extract` instead of `tarfile.TarFile.extractall`

kenodegard:
### Description

Python 3.12 introduced a deprecation warning in `tarfile` to warn of a future change in Python 3.14 where the default `filter` behavior in `tarfile.TarFile.extract`/`tarfile.TarFile.extractall` will change. ~Instead of allowing all file locations (and types) to be unpacked, the new default behavior is to only unpack the safe files (relative to the current location, no special file types, etc.). This is a reasonable change so here we attempt to adopt this change preemptively (and end up simplifying some of our code too!).~

~The decision to be made is whether we wish to use the strictest filtering (https://docs.python.org/3/library/tarfile.html#tarfile.data_filter) or the more permissive but still safer than before filtering (https://docs.python.org/3/library/tarfile.html#tarfile.tar_filter), I'm assuming we do not want the wild west no filtering of today (https://docs.python.org/3/library/tarfile.html#tarfile.fully_trusted_filter).~

Instead of adjusting how we use `tarfile` we instead opt to use `conda-package-handling` for our extraction needs (it is then up to `conda-package-streaming` & `conda-package-handling` to filter correctly).

Side note, it is odd that this deprecation warning didn't surface earlier.

Xref https://github.com/conda/conda-build/pull/5387 https://github.com/conda/conda-build/pull/5379 https://github.com/conda/conda-build/pull/5281 https://github.com/conda/conda-package-streaming/issues/87

### Checklist - did you ...

### DIFF
--- a/conda_build/convert.py
+++ b/conda_build/convert.py
@@ -132,7 +132,12 @@ def extract_temporary_directory(file_path):
     temporary_directory = tempfile.mkdtemp()
 
     with tarfile.open(file_path) as tar:
-        tar.extractall(temporary_directory, filter="data")
+        # FUTURE: Python 3.12+, remove try-except
+        try:
+            tar.extractall(path=temporary_directory, filter="data")
+        except TypeError:
+            # TypeError: `filter` is unsupported in this Python version
+            tar.extractall(path=temporary_directory)
 
     return temporary_directory
 
--- a/conda_build/convert.py
+++ b/conda_build/convert.py
@@ -131,9 +131,8 @@ def extract_temporary_directory(file_path):
     """
     temporary_directory = tempfile.mkdtemp()
 
-    source = tarfile.open(file_path)
-    source.extractall(temporary_directory)
-    source.close()
+    with tarfile.open(file_path) as tar:
+        tar.extractall(temporary_directory, filter="data")
 
     return temporary_directory
 
--- a/conda_build/convert.py
+++ b/conda_build/convert.py
@@ -18,6 +18,8 @@ import tempfile
 from pathlib import Path
 from typing import TYPE_CHECKING
 
+from conda_package_handling.api import extract
+
 from .utils import ensure_list, filter_info_files, walk
 
 if TYPE_CHECKING:
@@ -129,17 +131,9 @@ def extract_temporary_directory(file_path):
     Positional arguments:
     file_path (str) -- the file path to the source package tar file
     """
-    temporary_directory = tempfile.mkdtemp()
-
-    with tarfile.open(file_path) as tar:
-        # FUTURE: Python 3.12+, remove try-except
-        try:
-            tar.extractall(path=temporary_directory, filter="data")
-        except TypeError:
-            # TypeError: `filter` is unsupported in this Python version
-            tar.extractall(path=temporary_directory)
-
-    return temporary_directory
+    tmp = tempfile.mkdtemp()
+    extract(file_path, dest_dir=tmp)
+    return tmp
 
 
 def update_dependencies(new_dependencies, existing_dependencies):
--- a/conda_build/render.py
+++ b/conda_build/render.py
@@ -934,7 +934,7 @@ def open_recipe(recipe: str | os.PathLike | Path) -> Iterator[Path]:
     elif recipe.suffixes in [[".tar"], [".tar", ".gz"], [".tgz"], [".tar", ".bz2"]]:
         # extract the recipe to a temporary directory
         with TemporaryDirectory() as tmp, tarfile.open(recipe, "r:*") as tar:
-            tar.extractall(path=tmp)
+            tar.extractall(path=tmp, filter="data")
             yield Path(tmp)
     elif recipe.suffix == ".yaml":
         # read the recipe from the parent directory
--- a/conda_build/render.py
+++ b/conda_build/render.py
@@ -31,6 +31,7 @@ from conda.exceptions import NoPackagesFoundError, UnsatisfiableError
 from conda.gateways.disk.create import TemporaryDirectory
 from conda.models.records import PackageRecord
 from conda.models.version import VersionOrder
+from conda_package_handling.api import extract
 
 from . import environ, exceptions, source, utils
 from .exceptions import DependencyNeedsBuildingError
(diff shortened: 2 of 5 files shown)


conda_build/convert.py
```diff
@@ -131,9 +131,8 @@ def extract_temporary_directory(file_path):
     """
     temporary_directory = tempfile.mkdtemp()
 
-    source = tarfile.open(file_path)
-    source.extractall(temporary_directory)
-    source.close()
+    with tarfile.open(file_path) as tar:
+        tar.extractall(temporary_directory, filter="data")
```
kenodegard, line 135: This extraction is used for converting packages to other archs, is it ok for us to strip group and other's write and execute permissions? See https://docs.python.org/3/library/tarfile.html#tarfile.data_filter

conda_build/render.py
```diff
@@ -934,7 +934,7 @@ def open_recipe(recipe: str | os.PathLike | Path) -> Iterator[Path]:
     elif recipe.suffixes in [[".tar"], [".tar", ".gz"], [".tgz"], [".tar", ".bz2"]]:
         # extract the recipe to a temporary directory
         with TemporaryDirectory() as tmp, tarfile.open(recipe, "r:*") as tar:
-            tar.extractall(path=tmp)
+            tar.extractall(path=tmp, filter="data")
```
kenodegard, line 937: This happens when we render from an existing tarball package. The only file we care about in this scenario is the recipe itself so AFAIK removing group and other's write and execute bits is safe to do.

conda_build/utils.py
```diff
@@ -783,32 +783,14 @@ def _tar_xf_fallback(tarball, dir_path, mode="r:*"):
...
-        if not os.path.realpath(member.name).startswith(cwd):
-            member.name = member.name.replace("../", "")
-        if not os.path.realpath(member.name).startswith(cwd):
-            sys.exit("tarball contains unsafe path: " + member.name + " cwd is: " + cwd)
-        members[i] = member
-
-    t.extractall(path=dir_path)
-    t.close()
+    with tarfile.open(tarball, mode) as tar:
+        tar.extractall(path=dir_path, filter="data")
```
kenodegard, line 793: This is used in a variety of locations (like in skeleton's unpacking of PyPI packages). Filtering for absolute paths outside of the destination is valid (the old code also did this with some homegrown logic). The question is whether it is also safe to remove group and other's write and execute bits.

jaimergp: The key ideas are for tar logic are in [this conda-package-streaming subclass](https://github.com/conda/conda-package-streaming/blob/168beb4da1c971e8c99cee2c06c4862e267f8d81/conda_package_streaming/package_streaming.py#L38), I think. Looks like we don't care about owners, but we do want to respect the permissions as long as they conform to the umask.

conda_build/convert.py
kenodegard: The more I think about this I'm thinking this should be changed to the `'tar'` filter

conda_build/utils.py
kenodegard: This should probably also be the `'tar'` filter.

kenodegard:
@zklaus ah I had not caught that it was backported, I guess the easiest in this case would be duck typing

jaimergp:
The deprecation warning is only added in 3.12, so I'd say we only specify `filter` in those versions. The alternative is to reuse conda-package-streaming class for the extraction and handle the filters there?

zklaus:
I suppose the continuation of the discussion about the appropriate filter belongs in cph now, but anyway, I think we might even need a custom filter, at least if we want to maintain the logic including umasks.

conda_build/utils.py
```diff
@@ -783,32 +780,13 @@ def _tar_xf_fallback(tarball, dir_path, mode="r:*"):
...
-        cwd = os.path.realpath(os.getcwd())
-        if not os.path.realpath(member.name).startswith(cwd):
-            member.name = member.name.replace("../", "")
-        if not os.path.realpath(member.name).startswith(cwd):
-            sys.exit("tarball contains unsafe path: " + member.name + " cwd is: " + cwd)
-        members[i] = member
-
-    t.extractall(path=dir_path)
-    t.close()
+    extract(tarball, dest_dir=dir_path)
```
beckermr, line 789: This section of code appears to prevent unpacking of unsafe paths. Do we have tests or are we sure that deferring to cph covers the potential security issues this code accounts for?

kenodegard: CPH delegates to CPS which includes a check for unsafe paths:

https://github.com/conda/conda-package-streaming/blob/168beb4da1c971e8c99cee2c06c4862e267f8d81/conda_package_streaming/extract.py#L32-L51

```python
def is_within_dest_dir(name):
    abs_target = os.path.realpath(os.path.join(dest_dir, name))
    prefix = os.path.commonpath((dest_dir, abs_target))
    return prefix == dest_dir

for tar_file, _ in stream:
    # careful not to seek backwards
    def checked_members():
        # from conda_package_handling
        for member in tar_file:
            if not is_within_dest_dir(member.name):
                raise exceptions.SafetyError(f"contains unsafe path: {member.name}")
            yield member

    try:
        tar_file.extractall(path=dest_dir, members=checked_members())
    except OSError as e:
        if e.errno == ELOOP:
            raise exceptions.CaseInsensitiveFileSystemError() from e
        raise
```

conda_build/convert.py
kenodegard: this discussion is moot since we delegate to CPH/CPS

conda_build/render.py
kenodegard: this thread is moot since we delegate to CPH/CPS

conda_build/utils.py
kenodegard: this thread is moot since we delegate to CPH/CPS

dholth:
We implemented our own filter in conda-package-streaming. Could we use the Python one instead or pass our filter function through that parameter?

conda_build/render.py
```diff
@@ -31,6 +31,7 @@
 from conda.gateways.disk.create import TemporaryDirectory
 from conda.models.records import PackageRecord
 from conda.models.version import VersionOrder
+from conda_package_handling.api import extract
```
dholth, line 34: This API [can only extract `.conda` or `.tar.bz2`,](https://github.com/conda/conda-package-handling/blob/main/src/conda_package_handling/api.py) not any other `.tgz` etc. variant.

PR #5451, opened 2024-08-09
Remove `sys.exit` calls from `convert.py`

beeankha:
### Description

Replacing `sys.exit` calls in `conda_build/convert.py` with the new `CondaBuildUserError` exception for better error handling + updating corresponding unit tests.

The changes in this PR are separated out from work previously done by @kenodegard in #5255
Xref #4209 

### Checklist - did you ...

- [ ] ~~Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release notes?~~
     
- [ ] ~~Add / update outdated documentation?~~

### DIFF
--- a/conda_build/convert.py
+++ b/conda_build/convert.py
@@ -18,6 +18,7 @@ import tempfile
 from pathlib import Path
 from typing import TYPE_CHECKING
 
+from .exceptions import CondaBuildUserError
 from .utils import ensure_list, filter_info_files, walk
 
 if TYPE_CHECKING:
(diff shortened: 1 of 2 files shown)


beeankha:
The `sys.exit` call in [this code block](https://github.com/conda/conda-build/blob/main/conda_build/convert.py#L814-L821) seems to make sense to leave as-is, but I can also replace this one as well if everything should be updated to the `CondaBuildUserError` exception.

```
    if show_imports:
        imports = retrieve_c_extensions(file_path)
        if len(imports) == 0:
            print("No imports found.")
        else:
            for c_extension in imports:
                print(c_extension)
        sys.exit()
```

kenodegard:
@beeankha given where `conda_convert` is called I actually think it should just be a `return` instead, otherwise calling `conda_convert(show_imports=True)` in the Python REPL will also exit the REPL

kenodegard (approved):

PR #5469, opened 2024-08-29
Set strict channel priority

### LINKED ISSUE #5470: ABI conflict in test suite on `osx-64`
### Checklist

### What happened?

Some of the tests are failing due to an ABI conflict on the `osx-64` runners, see https://github.com/conda/conda-build/actions/runs/10602558349/job/29384860323.

This is caused by channel mixing where we end up getting the following installed which causes a conflict:

```
pkgs/main::conda-libmamba-solver
conda-forge::libmambapy
conda-forge::libmamba
...
```

The channel mixing is unintentional as the only package we need from conda-forge is `xdoctest`. There are a few possible solutions here:
1. explicitly require `defaults::libmamba[py]`, see #5468
2. set `channel_priority: strict` for the test environment creation, see #5469
3. use `conda-forge` instead of `defaults`
4. request `xdoctest` be built for `defaults`
5. do we need `xdoctest` in the first place, maybe we just remove it? see #5471

kenodegard:
### Description

For some reason our test environments were installing the following:

```
pkgs/main::conda-libmamba-solver
conda-forge::libmambapy
conda-forge::libmamba
...
```

Mixing `defaults` and `conda-forge` like this results in ABI conflicts and renders the libmamba solver unusable (see https://github.com/conda/conda-build/actions/runs/10602558349/job/29384860323). To correct this we set the channel priority to strict.

Closes #5468
Closes #5470 

### Checklist - did you ...

- [ ] ~Add a file to the `news` directory ([using the template](https://github.com/conda/conda-build/blob/main/news/TEMPLATE)) for the next release's release notes?~
     
### DIFF
--- a/.github/condarc
+++ b/.github/condarc
@@ -4,4 +4,3 @@ notify_outdated_conda: False
 changeps1: False
 always_yes: True
 local_repodata_ttl: 7200
-channel_priority: strict
--- a/.github/condarc
+++ b/.github/condarc
@@ -4,3 +4,4 @@ notify_outdated_conda: False
 changeps1: False
 always_yes: True
 local_repodata_ttl: 7200
+channel_priority: strict
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -126,6 +126,7 @@ jobs:
         run: >
           conda install
           --yes
+          --strict-channel-priority
           --file tests/requirements.txt
           --file tests/requirements-${{ runner.os }}.txt
           --file tests/requirements-ci.txt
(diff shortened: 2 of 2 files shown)


.github/condarc
```diff
@@ -4,3 +4,4 @@ notify_outdated_conda: False
 changeps1: False
 always_yes: True
 local_repodata_ttl: 7200
+channel_priority: strict
```
jaimergp, line 7:
```suggestion
channel_priority: strict
```

This should only apply to the CI step of provisioning `base`. Otherwise it'll affect all tests.

jaimergp:
The canary CI should put the `conda-canary/label/dev` channel first, then followed by `defaults`. Ideally the dev label should be per-project, like `conda_dev`. This would prevent other canary projects in the channel (e.g. menuinst) from leaking in there.

Issue #5455, opened 2024-08-11
BUG: problems new `files:` include/exclude functionality


h-vetinari:
The functionality introduced in https://github.com/conda/conda-build/pull/5216 by @carterbox promises to be very useful on several feedstocks that need to "slice and dice" the result of a build into different outputs, so I started testing this for llvmdev in https://github.com/conda-forge/llvmdev-feedstock/pull/283. This runs into a couple of different problems though.

## Exclusion rule causes glob error

One output has:
```yaml

```yaml
  - name: llvm-tools-{{ major_ver }}
    files:
      include:
        - bin/*-{{ major_ver }}     # [unix]
      exclude:
        # belongs into llvmdev
        - bin/llvm-config*          # [unix]
      [...]

  - name: llvm-tools
    files:
      include:
        # opt-viewer tool is in share
        - bin/*                     # [unix]
        - share/*                   # [unix]
        - Library/bin/*.exe         # [win]
        - Library/share/*           # [win]
      exclude:
        # belongs into llvmdev
        - bin/llvm-config*          # [unix]
        - Library/bin/llvm-config*  # [win]
    requirements:
      host:
        - {{ pin_subpackage("llvm-tools-" ~ major_ver, exact=True) }}   # [not win]
```

h-vetinari:
I guess there are two different use-cases that are at odds with each other. In the case of the bootstrap compilers, we actually _do_ want [everything](https://github.com/conda-forge/clang-compiler-activation-feedstock/blob/76c5b95f3c0c0f0544e474b93dbd49f767eedf30/recipe/meta.yaml#L136-L142) from `host:` to be picked up.

But since we now have a distinction between
```yaml
files:
  - abc
```

and

```yaml
files:
  include:
    - abc
  exclude:
    - def
```
I guess we could keep the former as "take everything" and the latter as "take into account snapshotting"?

carterbox:
> files: does not take into account files already existing in host:

The implicit filtering/ignoring of files added to the prefix from other packages only applies to files installed from packages which were created in separate recipes. It doesn't work when the host dependcies come from another output of the same recipe. You will need to craft your include exclude globs so they do not intersect.

If you want a "everything not already in other outputs" then you would do something like this:

```yaml
- name: A
  files:
    include:
     - A
- name: B
  files:
    include:
     - B
- name: C
  files:
    include:
     - "*"
    exclude:
     - A
     - B
``` 

[...]

PR #5472, opened 2024-08-30
DOC: Demote logging about non-matching glob to warning

carterbox:
The expand_globs function from conda_build.utils logs an ERROR when a glob expression returns no matches, this is overly alarming because the user may now use negative glob expressions which they don't care if it returns empty or the user may want to use the same set of glob expressions for multiple platforms some of which may return empty on some platforms.

https://github.com/conda/conda-build/pull/5216
https://github.com/conda/conda-build/issues/5455

### Checklist - did you ...

~~- [ ] Add / update necessary tests?~~ No relevant tests
~~- [ ] Add / update outdated documentation?~~ No relevant documentation


beeankha (approved):

carterbox:
Thanks @beeankha

PR #5535, opened 2024-11-13
fix: ensure variable finding works properly and add unit tests for variable finding

beckermr:
### Description

Currently, conda-build will say the variant variable `python` is used for a statement like

```yaml
requirements:
  host:
    - python {{ python_min }}
```

This is incorrect and only `python_min` is used in this case AFAIK.

It also says this statement does NOT use `python` when it does

```yaml
requirements:
  run:
    - {{ pin_compatible('python') }}
```

This bug was introduced in https://github.com/conda/conda-build/pull/2838.

The issue appears to be a buggy regex for searching for variants. 

This PR proposes a new regex and adds unit tests that assert the correct behavior.

### Checklist - did you ...

### DIFF
--- a/conda_build/variants.py
+++ b/conda_build/variants.py
@@ -750,7 +750,7 @@ def find_used_variables_in_text(variant, recipe_text, selectors_only=False):
         v_regex = re.escape(v)
         v_req_regex = "[-_]".join(map(re.escape, v.split("_")))
         variant_regex = (
-            rf"\{{\s*(?:pin_[a-z]+\(\s*?['\"])?{v_regex}[^_0-9a-zA-Z].*?\}}}}"
+            rf"\{{\s*(?:pin_[a-z]+\(\s*?['\"])?{v_regex}[^_0-9a-zA-Z].*?\}}\}}"
         )
         selector_regex = rf"^[^#\[]*?\#?\s\[[^\]]*?(?<![_\w\d]){v_regex}[=\s<>!\]]"
         # NOTE: why use a regex instead of the jinja2 parser/AST?
--- a/conda_build/variants.py
+++ b/conda_build/variants.py
@@ -749,7 +749,9 @@ def find_used_variables_in_text(variant, recipe_text, selectors_only=False):
             continue
         v_regex = re.escape(v)
         v_req_regex = "[-_]".join(map(re.escape, v.split("_")))
-        variant_regex = rf"\{{\s*(?:pin_[a-z]+\(\s*?['\"])?{v_regex}[^'\"]*?\}}\}}"
+        variant_regex = (
+            rf"\{{\s*(?:pin_[a-z]+\(\s*?['\"])?{v_regex}[^_0-9a-zA-Z].*?\}}}}"
+        )
         selector_regex = rf"^[^#\[]*?\#?\s\[[^\]]*?(?<![_\w\d]){v_regex}[=\s<>!\]]"
         # NOTE: why use a regex instead of the jinja2 parser/AST?
         # One can ask the jinja2 parser for undefined variables, but conda-build moves whole
--- a/news/5535-fix-variant-finding.rst
+++ b/news/5535-fix-variant-finding.rst
@@ -6,6 +6,7 @@
 
 * Fixed a bug where variants were incorrectly found as being used when they matched a leading substring of
   another variant. (#5535)
+* Fixed a bug where variants were not found when variables were used in ``pin_*`` statements. (#5535)
 
 ### Deprecations
 
--- a/news/5535-fix-variant-finding.rst
+++ b/news/5535-fix-variant-finding.rst
@@ -1,0 +1,20 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Fixed a bug where variants were incorrectly found as being used when they matched a leading substring of
+  another variant. (#5535)
+
+### Deprecations
+
+* <news item>
+
+### Docs
+
+* <news item>
+
+### Other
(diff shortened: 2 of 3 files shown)


beckermr:
@conda/builds-tools @jezdez @kenodegard @beeankha @jaimergp This request is unusual, but can we consider merging this PR before the next conda-build release?

This PR fixes a very non-trivial bug that has been present in the code base for at least five years. conda-build is misdetecting which variants are being used, which is not great. :/

jaimergp:
Hm, I wonder if this is behind https://github.com/conda-forge/conda-smithy/issues/2100 and https://github.com/conda-forge/conda-smithy/issues/2071

beckermr:
Looks like it very likely could be. Currently `conda-build` will confuse the `pin_compatible` statement for `napari-base` with using `napari`.

conda_build/variants.py
```diff
@@ -749,7 +749,9 @@ def find_used_variables_in_text(variant, recipe_text, selectors_only=False):
             continue
         v_regex = re.escape(v)
         v_req_regex = "[-_]".join(map(re.escape, v.split("_")))
-        variant_regex = rf"\{{\s*(?:pin_[a-z]+\(\s*?['\"])?{v_regex}[^'\"]*?\}}\}}"
+        variant_regex = (
+            rf"\{{\s*(?:pin_[a-z]+\(\s*?['\"])?{v_regex}[^_0-9a-zA-Z].*?\}}}}"
```
jaimergp, line 753: Do we need to escape this curly brace too?
```suggestion
            rf"\{{\s*(?:pin_[a-z]+\(\s*?['\"])?{v_regex}[^_0-9a-zA-Z].*?\}}\}}"
```

jaimergp:
And side note, I think these regexes (which have received several contributions lately) would benefit from [`re.VERBOSE`](https://docs.python.org/3/library/re.html#re.VERBOSE). Maybe that helps with the review too.

beckermr:
That is a good idea, but I'd like to defer to another PR. Also,  I am not sure I could actually describe what it is doing. I use [https://regex101.com/](https://regex101.com/) to build and test regex strings because they are so confusing. :/

jaimergp:
> I use [regex101.com](https://regex101.com/) to build and test regex strings

Heh, same here but with regexr.com 😂 

> That is a good idea, but I'd like to defer to another PR.

Let me give it a try, but yes, non-blocking.

jaimergp:
Hm, let's see:


```python
variant_regex = re.compile(
    rf"""
    \{{\{{              # Double curly brace opens Ninja expression
        \s*             # Zero or more spaces
        (?:             # Opens non-capturing group
            pin_[a-z]+  # Matches any lowercase pin_* function name (e.g. pin_compatible)
            \(          # Matches the opening parenthesis of the pin_* function
                \s*?    # Non-greedy glob for any number of spaces (JRG: not sure why we need the non greedy version)
                ['"]    # Single or double quote, required once
        )?              # Closes non-capturing group and makes it optional; i.e. the whole pin_compatible(' part can be ommitted 
        {v_regex}       # The regex from above, should match a variant name
        [^_0-9a-zA-Z]   # One instance of any character that is NOT a number, a letter or an underscore; i.e. not part of the variant name
        .*?             # Any number of any character, in a non-greedy way; aka "whatever comes before the closing curly braces"
    \}}\}}              # Double curly brace closes Ninja expression
    """,
    re.VERBOSE
)
```

beckermr:
@jaimergp my concern is that the conda-build test suite is not robust enough to catch bugs that might be introduced when we change the regex strings to the more verbose style. It'd be better to defer to future PRs that change the strings and introduce more robust tests to ensure nothing breaks.

jaimergp:
> It'd be better to defer to future PRs that change the strings and introduce more robust tests to ensure nothing breaks.

That's ok with me. Just updated my comment because I was already "parsing" the regexes for the review, so might as well write it down.

conda_build/variants.py
```diff
@@ -749,7 +749,9 @@ def find_used_variables_in_text(variant, recipe_text, selectors_only=False):
             continue
         v_regex = re.escape(v)
         v_req_regex = "[-_]".join(map(re.escape, v.split("_")))
-        variant_regex = rf"\{{\s*(?:pin_[a-z]+\(\s*?['\"])?{v_regex}[^'\"]*?\}}\}}"
+        variant_regex = (
+            rf"\{{\s*(?:pin_[a-z]+\(\s*?['\"])?{v_regex}[^_0-9a-zA-Z].*?\}}\}}"
```
jaimergp, line 753: My understanding of the change:

- The previous regex was non-greedy-matching any characters after the variant name that were NOT a single or double quote. That meant that partial matches were possible (e.g. it found `pari` in `pin_compatible('napari')`.
- The new regex tries to non-greedy-match any character after the variant name that is NOT a valid variant name character (underscores, numbers and letters).

beckermr: That's correct. We assert the variable name is terminated by a non-allowed character for variables. Then we get the rest of the characters up until the jinja2 statement is closed, asserting it is closed.

beckermr:
shall we merge @jaimergp? no pressure, but I am contemplating adding a patch to the conda-build feedstock if this PR is going to sit for a while.

jaimergp:
Let's give @beeankha and @ForgottenProgramme a few hours to take a look. The release branch has not been created yet so there's no rush to merge just yet. It does LGTM though.

beeankha (approved):

kenodegard (approved):

PR #5555, opened 2024-11-25
fix: allow incomplete specs during intermediate parsing stages

beckermr:
### Description

This PR allows incomplete specs if they do not parse by skipping them for intermediate parsing stages. This behavior has been in the code for a while if a spec was completely specified by eg jinja2 (eg a list element `- {{ mpi }}`). The PR extends this behavior to any soec that does not parse during intermediate stages. For the final parsing stage, specs that do not parse will raise. 

### Checklist - did you ...

### DIFF
--- a/tests/test_metadata.py
+++ b/tests/test_metadata.py
@@ -655,3 +655,56 @@ build:
         pytest.fail(
             "Undefined variable caused error, even though this build is skipped"
         )
+
+
+@pytest.mark.parametrize("have_variant", [True, False])
+def test_parse_until_resolved_missing_jinja_in_spec(
+    testing_metadata: MetaData,
+    tmp_path: Path,
+    have_variant: bool,
+) -> None:
+    (recipe := tmp_path / (name := "meta.yaml")).write_text(
+        """
+package:
+    name: dummy
+    version: 1.0.0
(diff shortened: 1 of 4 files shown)


tests/test_metadata.py
```diff
@@ -655,3 +655,56 @@ def test_parse_until_resolved_skip_avoids_undefined_jinja(
...
+        delattr(testing_metadata.config, "variant_config_files")
+        delattr(testing_metadata.config, "variants")
+
+    try:
+        testing_metadata.parse_until_resolved()
+        if not have_variant:
+            pytest.fail("Undefined variable did NOT cause spec parsing error!")
+        else:
+            print("parsed OK!")
+    except (Exception, SystemExit):
```
jaimergp, line 704: Do we have a more concrete exception set?

jaimergp (approved):
LGTM, just had a small question.

tests/test_metadata.py
beckermr: Not reliably no. I have already fixed a bug where when the exception handling in one part of the code changed, the test suite broke because a new kind of exception was raised in another part of the code. For now, any exception will work.

Issue #5571, opened 2024-12-19
BUG: corrupted build string in direct violation of its specification


h-vetinari:
### Checklist

- [x] I added a descriptive title
- [x] I searched open reports and couldn't find a duplicate

### What happened?

Our blas metapackage is surely not trivial and ends up pushing conda-build into unusual corners (there are several bugs in addition to this one; perhaps I get around to raising more issues).

This issue is pretty blatant though. As of https://github.com/h-vetinari/blas-feedstock/commit/4a63a0013c00b608a830d7a9e2b486f4ce2b5e1a, the recipe contains:
```yaml
  - name: blas
    version: "{{ blas_major }}.{{ blas_minor }}"
    script: test_blas.sh   # [unix]
    script: test_blas.bat  # [win]
    build:
      string: "{{ blas_impl }}"
```
And yet, when the blis-variant gets built, what we see is (added linebreaks for readability)

```
BUILD START: [
    'libblas-3.9.0-27_hde320c6_blis.conda',
    'libcblas-3.9.0-27_h35a6ab0_blis.conda',
    'blas-devel-3.9.0-27_h35a6ab0_blis.conda',
    'blas-2.127-blish14e4bdd.conda'
                    ^^^^^^^^
                    What?!
]
```

That's despite the fully rendered recipe (in the logs) looking like this:
```yaml
  - build:
      activate_in_script: true
      string: blis
    name: blas
    requirements:
```
Later this obviously falls over when it doesn't match expectations anymore:
```
Could not solve for environment specs
The following packages are incompatible
├─ blas 2.127.* blish14e4bdd is requested and can be installed;
└─ libcblas 3.9.0.* 27_h35a6ab0_blis is not installable because it requires
   └─ blas * blis, which conflicts with any installable versions previously reported.
```

<details>

### Conda Info

```shell
++ conda info

     active environment : base
    active env location : /opt/conda
            shell level : 1
       user config file : /home/conda/.condarc
 populated config files : /opt/conda/.condarc
                          /home/conda/.condarc
          conda version : 24.11.1
    conda-build version : 24.11.2
         python version : 3.12.8.final.0
                 solver : libmamba (default)
       virtual packages : __archspec=1=broadwell
                          __conda=24.11.1=0
                          __glibc=2.34=0
                          __linux=6.5.0=0
                          __unix=0=0
       base environment : /opt/conda  (writable)
      conda av data dir : /opt/conda/etc/conda
  conda av metadata url : None
[...]

PR #5601, opened 2025-01-25
Warn if parsed and raw outputs mismatch

jaimergp:
### Description

See https://github.com/conda/conda-build/issues/5571#issuecomment-2613973480

### Checklist - did you ...

### DIFF
--- a/conda_build/metadata.py
+++ b/conda_build/metadata.py
@@ -2229,11 +2229,14 @@ class MetaData:
             outputs = [{"name": self.name()}]
 
         if len(output_matches) != len(outputs):
+            # See https://github.com/conda/conda-build/issues/5571
             utils.get_logger(__name__).warning(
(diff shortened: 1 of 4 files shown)


beckermr (changes requested):
IMHO this should have a test. That way we have a test case to use to attempt to fix the underlying issue.

jaimergp:
Added.

beckermr (approved):

jaimergp:
I'll open a PR to skip the failing tests for now.

isuruf (commented):
Doesn't this log a warning for all recipes without an `outputs` section?

jaimergp:
True:

```
WARNING: Number of parsed outputs does not match detected raw metadata blocks. Identified output block may be wrong! If you are using Jinja conditionals to include or exclude outputs, consider using `skip: true  # [condition]` instead.
```

I'll open an issue to track.

PR #5617, opened 2025-02-11
Cran skeleton update

danielnachun:
### Description

The R 4.4 + UCRT migration is now complete for all but one CRAN package (see https://github.com/conda-forge/r-volumodel-feedstock/pull/2), and that failure is not related to the migration.  We have a handful of other R packages from GitHub that are not yet migrated, also due to issues unrelated to the migration.  I believe we are now in a state where we should update `conda skeleton cran` so that it generates valid recipes now that this migration is now effectively complete.  The specific changes as can been in my commit are as follows:

- remove the `{% set native = 'm2w64-' if win else '' %}` Jinja line as we no longer use any `m2w64` packages except `m2w64-sysroot`, and that is handled by `conda-smithy`, not in the recipe
- no longer add `m2w64-gcc-libs` to the dependency list for Windows, as we no longer use this package (see previous point)
- remove `merge_build_host: True` on Windows, as this is now irrelevant after the UCRT migration and will actually break the build in many cases
- Use `pkg-config` and `make` on Windows instead of `m2-pkg-config` and `m2-make`

I also removed the `license_family` entry as we no longer use this given that we have SPDX license identifiers now.

I'm not sure if these changes constitute a "significant" change worthy of a news update but I'm happy to add one if needed.

### Checklist - did you ...


dbast:
@conda-bot check

jaimergp:
not stale

@danielnachun, are there any other changes that we should incorporate here, especially after the discussions in Zulip ([#general > v1 recipe for R yet? @ 💬](https://conda-forge.zulipchat.com/#narrow/channel/457337-general/topic/v1.20recipe.20for.20R.20yet.3F/near/572128376), [#general > v1 recipe for R yet? @ 💬](https://conda-forge.zulipchat.com/#narrow/channel/457337-general/topic/v1.20recipe.20for.20R.20yet.3F/near/572128376))?

Also, these are conda-forge specific, maybe, so I'm unsure whether this should go here, or protected behind a flag like `grayskull` does?

Issue #5632, opened 2025-02-27
Release `25.3.x`


kenodegard:
### Summary

Placeholder for `conda-build 25.3.x` release.

| Pilot | @kenodegard |
|---|---|
| Co-pilot | @ForgottenProgramme |

### Tasks

[milestone]: https://github.com/conda/conda-build/milestone/45
[process]: https://github.com/conda/conda-build/blob/main/RELEASE.md
[releases]: https://github.com/conda/conda-build/releases
[main]: https://github.com/AnacondaRecipes/conda-build-feedstock
[conda-forge]: https://github.com/conda-forge/conda-build-feedstock
[ReadTheDocs]: https://readthedocs.com/projects/continuumio-conda-build/
[zulip]: https://conda.zulipchat.com/#narrow/channel/480811-releases

<details>  
<summary><h4>The week before release week</h4></summary>

- [x] Create release branch https://github.com/conda/conda-build/tree/25.3.x
- [x] Ensure release candidates are being successfully built https://anaconda.org/conda-canary/repo/files?type=any&label=rc-conda-build-25.3.x
- [x] [Complete outstanding PRs][milestone]
- [x] Check for deprecated features
- [ ] Test release candidates https://github.com/conda/conda-build/compare/25.1.x...25.3.x
    - [x] Tested on macOS by @kenodegard 
    - [ ] #5580
    - [ ] #5599
    - [x] #5603
    - [ ] #5605
    - [ ] #5622
    - [ ] #5601
    - [ ] #5607
    - [ ] #5623
    - [x] #5600
    - [x] #5456
    - [ ] #5622
    - [ ] #5575
    - [x] #5637
    - [x] #5627
    - [x] #4977

</details>

<details open>  
<summary><h4>Release week</h4></summary>

- [x] Create release PR (see [release process][process]) https://github.com/conda/conda-build/pull/5642
- [x] Create Zulip thread on [#releases][zulip] [#releases > conda-build 25.3.x](https://conda.zulipchat.com/#narrow/channel/480811-releases/topic/conda-build.2025.2E3.2Ex)
    - [x] Announce `25.3.0` in-progress [#releases > conda-build 25.3.x @ 💬](https://conda.zulipchat.com/#narrow/channel/480811-releases/topic/conda-build.2025.2E3.2Ex/near/506598753)
- [x] [Publish release][releases] https://github.com/conda/conda-build/releases/tag/25.3.0
- [x] https://github.com/conda/conda-build/pull/5643
[...]

PR #5652, opened 2025-03-24
Changelog 25.3.1

kenodegard:
## Description

✂️ snip snip ✂️ the making of a new release.

Xref #5632


beckermr (approved):

beckermr:
We'll need to add more retries for these flaky tests or move some of them to the serial jobs maybe. Hopefully a retry of the job will do the trick.

kenodegard:
Resolved all but one of the failures (they were in fact genuine failures)

For the last failure I have not been able to replicate locally yet, looking to see if it is caused by a leaky test (but given the error that would be an odd leakage)

Issue #5078, opened 2023-11-20
Remaing CI test concurrency basic issues


mbargull:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### What happened?

`conda-build`'s CI tests are still a bit flaky.
gh-5068 and gh-5076 lessened the chance of failures, but there are still some underlying issues:

### The `serial` tests are actually not run in series anymore
- Happend since https://github.com/conda/conda-build/pull/4648/files#diff-1db27d93186e46d3b441ece35801b244db8ee144ff1405ca27a163bfe878957fR82
- Reason is that there is a `xxx && 0 || 'auto'` which will always evaluate to `auto`.

I did *not* change/fix that in the recent pull requests (since tests run somewhat fine and otherwise it's a waste of resources).
We actually should run most of those tests in parallel since the reasons they were marked as `serial` are most likely connected to reasons handled by the pull requests above.
There may be some tests that should actually be run in series -- we might want to run them separately (or at least within a single worker, e.g., via [`--dist loadgroup`](https://pytest-xdist.readthedocs.io/en/latest/distribution.html#distribution-modes).

### The tests (and `conda-build` in general) use the same global package cache
With the recent changes, the tests should now use individual working directories, but they still use the same shared package cache.
When the tests are run concurrently, this means they can corrupt indexes (or possibly packages when different channels/subdirs are used).

Ideally we want the package cache separated for the tests too.
However, this could increase resource usage significantly since the same packages would be downloaded/extracted multiple times to different caches then.
Hence, we might want to do this selectively (I haven't thought out a plan for this yet, though).

Another problem related to the shared package cache is that `conda-build` will add the packages it built to the package cache when it builds subpackage builds or runs tests (because that's how `conda` works: to create an environment, the packages are put into `pkgs_dirs` beforehand).
For the CI tests, this means that packages with the same name+version+build will get clobbered, e.g., see [this recent workaround for such a case](https://github.com/conda/conda-build/pull/5076/commits/8f509baac723823add489824d45e13c7a10033f6#diff-2a3c6ee2fa7994b8aeac77119034006e68e707577a8ccec15a48f4088c59668dR24).
**We may want to change that behavior overall -- not just for the CI tests -- and let `conda-build` put its own builds into a separate `pkgs` directory relative to its working directory.**

[...]

PR #5657, opened 2025-03-28
Debugging corrupt package failures

kenodegard:
### Description

Noticing an increase in corrupt package test failures:
- https://github.com/conda/conda-build/actions/runs/14114350528/job/39540950302
- https://github.com/conda/conda-build/actions/runs/14131729595/job/39593520975?pr=5654
- https://github.com/conda/conda-build/actions/runs/14114164270/job/39540208622?pr=5652

### Checklist - did you ...

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -169,7 +169,7 @@ jobs:
   linux-benchmarks:
     # only run test suite if there are code changes
     needs: changes
-    if: needs.changes.outputs.code == 'true'
+    if: false # needs.changes.outputs.code == 'true'
 
     runs-on: ubuntu-latest
     defaults:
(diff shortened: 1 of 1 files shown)


kenodegard:
great now the failure has vanished from this PR...

beckermr:
So the test jobs marked as "serial" are being run with two purest-xdist workers. Is that maybe the issue?

kenodegard:
interestingly all of the jobs are now cache hits

kenodegard:
> So the test jobs marked as "serial" are being run with two purest-xdist workers. Is that maybe the issue?

yeah that's been the case for a while now: https://github.com/conda/conda-build/issues/5078

I'd be surprised if that's the reason, plus we're seeing failures on both the "parallel" (`tests/cli/test_main_build.py::test_slash_in_recipe_arg_keeps_build_id`) and the "serial" (`tests/test_api_build.py::test_recipe_builds[relative_url_download]`) jobs

beckermr:
Maybe add some retry test markers to have the test suite try twice as a stopgap?

kenodegard:
We could add the retry but it is hard to gauge how effective that would be since the corrupt package is in the cache (`/home/runner/conda_pkgs_dir/`) so a rerun would just reuse the corrupt package?

beckermr:
Hrmmm. Can we try things actually in serial as a test?

PR #5690, opened 2025-04-17
fix: allow tuples, lists and dicts in selectors

### LINKED ISSUE #5689: selector failure on tuples / lists
### Checklist

### What happened?

The latest conda-build 25.4.0 throws a validation error when tuples or lists are in selectors.

See this error output

```
conda_build.exceptions.CondaBuildUserError: Invalid selector in meta.yaml line 149:
offending selector:
  [os.environ.get("BUILD_PLATFORM") == "linux-64" and os.environ.get("DEFAULT_LINUX_VERSION", "alma9") in ("alma8", "ubi8")]
exception:
  ValidationException: Node type 'Tuple' is not allowed. (whitelist it manually)
```

### Conda Info

```shell
N/A
```

### Conda Config

```shell
N/A
```

### Conda list

```shell
N/A
```

### Additional Context

N/A

beckermr:
### Description

This PR allows tuples, lists and dicts in selectors as nodes in addition to allowing the creation functions. This let's folks use literals like `("a", "b")` in a selector.

### Checklist - did you ...

closes #5689

### DIFF
--- a/conda_build/metadata.py
+++ b/conda_build/metadata.py
@@ -289,7 +289,7 @@ def parseNameNotFound(error) -> str:
 @cache
 def evalidate_model():
     model = base_eval_model.clone()
-    model.nodes.extend(["Call", "Attribute"])
+    model.nodes.extend(["Call", "Attribute", "Tuple", "List", "Dict"])
     model.allowed_functions += [
         "int",
         "str",
--- a/news/5690-selector-lists.rst
+++ b/news/5690-selector-lists.rst
@@ -1,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Fixed bug where selectors with lists, tuples, or dicts could not be processed. (#5690)
+
+### Deprecations
+
+* <news item>
+
+### Docs
+
+* <news item>
+
+### Other
+
+* <news item>
--- a/tests/test_metadata.py
+++ b/tests/test_metadata.py
@@ -83,11 +83,14 @@ def test_select_lines():
             "test {{ JINJA_VAR[:2] }} # stuff yes [abc]",
             "test {{ JINJA_VAR[:2] }} # [abc] stuff yes",
             '{{ environ["test"] }}  # [abc]',
+            '{{ environ["test"] }}  # [d in ("a", "b")]',
+            '{{ environ["test"] }}  # [d in ["a", "b"]]',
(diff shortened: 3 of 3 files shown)


beckermr:
pre-commit.ci autofix

jaimergp (dismissed):

beckermr:
@jaimergp sorry for the dismissed review - need a bug fix in the tests

Can you take another look?

h-vetinari:
Can we add a test here that fetches the newest pinning CBC and tries to rerender a simple recipe with it? Just that we ensure we can parse the pinning. Or if necessary for CI stability, use the `conda-forge-pinning` package for a given version and extract the CBC from the package payload.

beckermr:
Being able to parse the conda forge linking at any given time is not a formal test requirement for conda build. 

I'll add a static copy as a test, but conda build cannot promise the pinnings file will always work no matter what syntax it has.

h-vetinari:
> but conda build cannot promise the pinnings file will always work no matter what syntax it has.

It's the other way around. Conda-forge can never adopt syntax in the pinning that's incompatible with conda-build (well at least until perhaps we've moved to v1 completely in several years). It's IMO much more practical to accept that tight relationship, and avoid a situation like with 25.4 (i.e. c-b being broken on the pinning requires it to be marked broken immediately).

beckermr:
The bug here is not that conda build failed on the conda forge pinning. That is how we found the bug. 

The bug is that we didn't properly add certain AST nodes to the evalidate config for python syntax we intended to support. 

Had the original syntax of using a tuple worked, and instead conda build failed on some other syntax, we may have chosen instead to adjust the conda forge pinning instead. This would especially be true if the syntax wasn't something we could support securely.

beckermr:
OK @jaimergp let's go ahead with this one

jaimergp (approved):

conda_build/metadata.py
```diff
@@ -289,7 +289,7 @@ def parseNameNotFound(error) -> str:
 @cache
 def evalidate_model():
     model = base_eval_model.clone()
-    model.nodes.extend(["Call", "Attribute"])
+    model.nodes.extend(["Call", "Attribute", "Tuple", "List", "Dict"])
```
jaimergp, line 292: This was an oversight on my end, sorry about the breakage. I thought including them in allowed_functions was enough 😬

PR #5708, opened 2025-05-07
Cleanup news formatting

### LINKED ISSUE #5709: News fragments are unformatted
Rever uses reStructuredText for news fragments. Meanwhile conda-build uses Markdown for its news fragments However the news fragments we have and our template don't use extensions. This can cause confusion when formatting (like how to many backticks (\`) to use)

jakirkham:
Clarify conda-build news fragments' format. Do this by adding extensions to all news files and the template. Use Markdown (`.md`) as that is what conda-build uses.

Fixes https://github.com/conda/conda-build/issues/5709


jezdez (changes requested):
The way we have rever configured is that we're taking news snippets in as markdown, since that's what our changelog is configured as: https://github.com/conda/conda-build/blob/2b1f1a33e09a5481b617d9b523ecb5c25c1c05d6/rever.xsh#L13-L16

We can do the same change like you propose, but append the `md` file extension to fix this.

jakirkham:
Thanks for clarifying Jannis! 🙏

We use reStructuredText in conda-smithy. Rever also uses that by default. So missed the customization that is happening here

Have pushed a fix and corrected this in the issue and PR descriptions

Please let me know if there is anything else

beckermr (approved):

jezdez (approved):

jakirkham:
Thank you for the reviews! 🙏

Yep completely understand using Markdown for this purpose

Think we can rollout this change across Conda projects to standardize things a bit

PR #5748, opened 2025-07-08
Improve build number and string example

nicoddemus:
### Description

The paragraph above says:

> When redefining the default string,
we strongly recommend following the convention of adding the build number at the end of the string, with a preceding underscore.

We should add the recommendation to the example.

### Checklist - did you ...

- [ ] ~~Add a file to the `news` directory ([using the template](https://github.com/conda/conda-build/blob/main/news/TEMPLATE)) for the next release's release notes?~~
     
- [ ] ~~Add / update necessary tests?~~
- [X] Add / update outdated documentation?

### DIFF
--- a/docs/source/resources/define-metadata.rst
+++ b/docs/source/resources/define-metadata.rst
@@ -331,7 +331,7 @@ number at the end of the string, with a preceding underscore.
 
    build:
      number: 1
-     string: abc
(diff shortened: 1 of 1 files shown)


ryanskeith (approved):
Looks good to me.

kenodegard (approved):

PR #5753, opened 2025-07-18
CI Fixes

### LINKED ISSUE #5757: Upgrade Windows test runners in GHA
GitHub is retiring their `windows-2019` GHA images:

- https://github.blog/changelog/2025-04-15-upcoming-breaking-changes-and-releases-for-github-actions/#windows-server-2019-is-closing-down

We currently have three separate pull requests which all attempt to resolve this issue:

- https://github.com/conda/conda-build/pull/5734
- https://github.com/conda/conda-build/pull/5750
- https://github.com/conda/conda-build/pull/5753

The CI for conda-build is effectively broken until we can find a solution for this, so this needs to be resolved ASAP.

ryanskeith:
### Description

Closes: #5757 

This PR was mostly focused on getting the Windows CI working but as a result of changing some code a few other small changes were also done.  These can be broken up if desired but as there were a few issues to deal with to address different parts of the CI on different platforms, I wanted to assure all the changes could co exist.

Summary of changes:
- CI windows image update from `windows-2019` (which was deprecated) -> `windows-latest`. This was the original intent of the change.
- Tests were slightly changed to use the "on_win" versus the sys.platform ==/!= "win32" statement
- Updated cl_versions to include more modern versions of Visual Studio.  Note: CL versions are the compiler versions and not tied to the release year or version. <- A frustrating gotcha.
- It was found that the location of the Microsoft Visual Studio has changed format, so logic was added to `build_vcvarsall_vs_path` to determine where newer versions are located.  Also, some checks were done to make it agnostic to which flavor a developer installs.
- Debug messaging was added for better log messaging to the user if Visual Studio was not found.
- Created a get_json_encoder helper function to mitigate the EntityEncoder deprecation
  - The helper function is required for now as the runner uses an older version of conda that does not have CondaJSONEnoder
  - Also ignored the deprecation warning. 
- Fixed rpath testing recipe for Mac OS.  Mac OS Arm64 is less forgiving so an update was required.

### Checklist - did you ...

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -278,6 +278,9 @@ jobs:
           path: D:\conda_pkgs_dir
           key: cache-${{ env.HASH }}
 
+      - name: Setup MSVC
+        uses: ilammy/msvc-dev-cmd@0b201ec74fa43914dc39ae48a89fd1d8cb592756 # v1.13.0
+
       - name: Setup Miniconda
         uses: conda-incubator/setup-miniconda@505e6394dae86d6a5c7fbb6e3fb8938e3e863830 # v3.1.1
         with:
@@ -285,12 +288,6 @@ jobs:
           run-post: false  # skip post cleanup
           pkgs-dirs: D:\conda_pkgs_dir
           installation-dir: D:\conda
-
-      - name: Setup MSVC
-        uses: ilammy/msvc-dev-cmd@0b201ec74fa43914dc39ae48a89fd1d8cb592756 # v1.13.0
-        with:
-          arch: "x64"
-          vsversion: "2017"
 
       - name: Conda Install
         run: >
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -286,8 +286,8 @@ jobs:
           pkgs-dirs: D:\conda_pkgs_dir
           installation-dir: D:\conda
 
-      - name: Choco Install
-        run: choco install visualstudio2017-workload-vctools
+      - name: Setup MSVC
+        uses: ilammy/msvc-dev-cmd@0b201ec74fa43914dc39ae48a89fd1d8cb592756 # v1.13.0
 
       - name: Conda Install
         run: >
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -287,7 +287,7 @@ jobs:
           installation-dir: D:\conda
 
       - name: Choco Install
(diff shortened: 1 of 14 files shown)


conda_build/build.py
```diff
@@ -27,10 +27,10 @@
 import yaml
 from bs4 import UnicodeDammit
 from conda import __version__ as conda_version
+from conda.auxlib.entity import EntityEncoder
```
kenodegard: `EntityEncoder` has been deprecated (https://github.com/conda/conda/blob/fe42f66c373b5164a7fc751876b236ff603be332/conda/auxlib/entity.py#L975-L981), so we should do a `try-except` import

kenodegard: Oops, managed to comment on a reverted commit

.github/workflows/tests.yml
```diff
@@ -287,6 +287,7 @@ jobs:
           installation-dir: D:\conda
 
       - name: Choco Install
+      # We may need the complete tooling so that cmake tests pass on windows
```
kenodegard:
```suggestion
        # We may need the complete tooling so that cmake tests pass on windows
```

conda_build/utils.py
```diff
@@ -2295,3 +2295,22 @@ def opener(path, flags):
...
+    try:
+        # Try to import the new encoder first
+        from conda.common.serialize.json import CondaJSONEncoder
+
+        return CondaJSONEncoder
+    except ImportError:
+        # Fall back to the old encoder
+        from conda.auxlib.entity import EntityEncoder
+
+        return EntityEncoder
```
kenodegard: We've previously relied on inline try-except imports, e.g.:

https://github.com/conda/conda-build/blob/dec1054628a9d2db6414e776fac10fa44627e6ad/conda_build/index.py#L31-L38

don't think we need to deviate from that here

```python
try:
    from conda.common.serialize.json import CondaJSONEncoder
except ImportError:
    # FUTURE: remove for `conda>=25.9`
    from conda.auxlib.entity import EntityEncoder as CondaJSONEncoder
```

kenodegard (dismissed):
awesome sleuthing!!

travishathaway (approved):

kenodegard (approved):

tests/requirements.txt
```diff
@@ -17,6 +17,7 @@ pip
 pkginfo
 psutil
 py-lief
+pytest-rerunfailures  # for handling flaky tests
```
jezdez, line 20: Yay!

Issue #5759, opened 2025-07-30
conda-build behaves unpredictably with slashes and backslashes if you use pytest command with the --ignore flag and run SP_DIR/tests (only on Winodws)


skupr-anaconda:
### Checklist

- [x] I added a descriptive title
- [x] I searched open reports and couldn't find a duplicate

### What happened?

conda-build behaves unpredictably in the test environment. Look at slashes:
```
    - pytest -v ${SP_DIR}/json_merge_patch/tests.py  # [unix]
    - pytest -v %SP_DIR%/json_merge_patch/tests.py  # [win]
```
They work. See an example https://github.com/AnacondaRecipes/json-merge-patch-feedstock/pull/1

I tried the same for `python-gssapi` with `--ignore=%SP_DIR%/path/to_test` but slashes and backslashes fail on Windows if you try to ignore tests. Only two feedstocks in `aggregate` repo run pytest with `%SP_DIR%` on Windows:
```
rg "pytest.*%SP_DIR%"
pyarrow-18.1.x-feedstock/recipe/meta.yaml
116:    - python -m pytest -v -k "not ({{ tests_to_skip }})" %SP_DIR%/pyarrow  # [win and py<313]

pyarrow-feedstock/recipe/meta.yaml
116:    - python -m pytest -v -k "not ({{ tests_to_skip }})" %SP_DIR%/pyarrow  # [win and py<313]
```
but no one has the `ignore` flag.

As a result, I skipped `python-gssapi`'s tests on Windows https://github.com/AnacondaRecipes/python-gssapi-feedstock/pull/3

### Conda Info

```shell

```

### Conda Config

```shell

```

### Conda list

```shell

```

PR #5765, opened 2025-08-01
Bugfix for issue #5759

ryanskeith:
## Fix path separator handling in pytest commands with --ignore flag and SP_DIR on Windows

Fixes GitHub issue #5759 where conda-build behaved unpredictably with slashes and backslashes when using pytest commands with the `--ignore` flag and `SP_DIR` paths on Windows.

### What was the problem?

When `SP_DIR` environment variable was set with backslashes on Windows (e.g., `C:\path\to\site-packages`), pytest commands using forward slashes (e.g., `pytest -v --ignore=%SP_DIR%/path/to/ignore`) would create mixed path separators when expanded, causing pytest to fail.

### What's the solution?

Added a path normalization function `_normalize_path_separators_in_command()` that ensures consistent backslashes for the entire path after environment variables in test commands on Windows. This handles 13 common environment variables including `SP_DIR`, `PREFIX`, `BUILD_PREFIX`, `SRC_DIR`, etc.

### Key changes:

- **Added**: `_normalize_path_separators_in_command()` function in `conda_build/create_test.py`
- **Modified**: `create_shell_files()` to normalize commands before writing to shell scripts

### Example:

**Before**: `pytest -v --ignore=%SP_DIR%/path/to/ignore` → `pytest -v --ignore=C:\path\to\site-packages/path/to/ignore` (mixed separators)

**After**: `pytest -v --ignore=%SP_DIR%/path/to/ignore` → `pytest -v --ignore=C:\path\to\site-packages\path\to\ignore` (consistent backslashes)

### Checklist - did you ...

- [x] Add a file to the `news` directory for the next release's release notes?

**Note**: This is a bug fix that resolves a Windows-specific issue with path handling in test commands. The fix is backward compatible and doesn't affect Unix systems.

### DIFF
--- a/conda_build/create_test.py
+++ b/conda_build/create_test.py
@@ -65,15 +65,52 @@ def _normalize_path_separators_in_command(cmd: str) -> str:
         envs_separator[env_var] = guess_separator(env_value)
 
     # Process each part of the command to normalize the path separators
-    for part in shlex.split(cmd, posix=False):
+    # shlex.split does not handle quoted strings correctly (see docs), so we need to handle them manually
+    parts = shlex.split(cmd, posix=False)
+
+    # Process each part to assure they are complete and then normalize
+    corrected_parts = []
+    corrected_parts.append(
+        parts[0]
+    )  # Shlex does ok with the first part of a command in windows
+
+    i = 1
+    while i < len(parts):
+        part = parts[i]
+
+        # If part has a quote but does not end with a quote, we need to find the next quote and join the parts
+        # This is the case where we have a --option="value with spaces"
+        if '"' in part and not part.endswith('"'):
+            quoted_part = [part]
+            i += 1
+
+            while i < len(parts) and not parts[i].endswith('"'):
+                quoted_part.append(parts[i])
+                i += 1
+
+            # Added last part ending with a quote
+            if i < len(parts):
+                quoted_part.append(parts[i])
+
+            corrected_parts.append(" ".join(quoted_part))
+            i += 1
+        else:
+            corrected_parts.append(part)
+            i += 1
+
+    new_cmd_parts = []
+    for part in corrected_parts:
         for env_var in cmd_path_env_vars:
-            if env_var in part:
+            if env_var in part or env_var.lower() in part:
                 part_separator = guess_separator(part)
                 if part_separator != envs_separator[env_var]:
                     changed_part = part.replace(part_separator, envs_separator[env_var])
-                    cmd = cmd.replace(part, changed_part)
+                    new_cmd_parts.append(changed_part)
+            else:
+                new_cmd_parts.append(part)
+    new_cmd = " ".join(new_cmd_parts)
 
-    return cmd
+    return new_cmd
 
 
 def create_files(m: MetaData, test_dir: Path) -> bool:
--- a/conda_build/create_test.py
+++ b/conda_build/create_test.py
@@ -23,8 +23,9 @@ def _normalize_path_separators_in_command(cmd: str, is_windows: bool) -> str:
     """
     Normalize path separators in test commands to ensure consistency.
 
-    This function handles the issue where SP_DIR contains backslashes on Windows,
-    but pytest commands may use forward slashes, creating mixed separators.
+    This function handles the issue where environment variables containing paths
+    (like SP_DIR, PREFIX, BUILD_PREFIX, SRC_DIR, STDLIB_DIR) contain backslashes
+    on Windows, but commands may use forward slashes, creating mixed separators.
 
     Args:
         cmd: The command string to normalize
@@ -36,20 +37,35 @@ def _normalize_path_separators_in_command(cmd: str, is_windows: bool) -> str:
     if not is_windows:
         return cmd
 
-    # For Windows, we need to handle cases where SP_DIR is used in paths
-    # SP_DIR contains backslashes, but commands may use forward slashes
-    # We normalize to use backslashes consistently for Windows paths
+    # Common environment variables that contain paths and need normalization
+    # These variables are set with backslashes on Windows but commands may use forward slashes
+    path_env_vars = [
+        "SP_DIR",
+        "PREFIX",
+        "BUILD_PREFIX",
+        "SRC_DIR",
+        "STDLIB_DIR",
+        "RECIPE_DIR",
+        "LIBRARY_PREFIX",
+        "LIBRARY_BIN",
+        "LIBRARY_INC",
+        "LIBRARY_LIB",
+        "SCRIPTS",
+        "SYS_PREFIX",
+        "ROOT",
+    ]
 
-    # Pattern to match SP_DIR followed by a path
-    # This matches %SP_DIR%/path or %SP_DIR%\\path
-    sp_dir_pattern = r"%SP_DIR%([/\\])"
+    # Pattern to match any of these environment variables followed by a path separator
+    # This matches %VAR%/path or %VAR%\\path for any VAR in the list
+    env_pattern = r"(%" + "|%".join(f"{var}%" for var in path_env_vars) + r")([/\\])"
 
-    def replace_sp_dir_path(match):
-        # Always use backslash for Windows paths when SP_DIR is involved
-        return "%SP_DIR%\\"
+    def replace_env_path(match):
+        env_var = match.group(1)
+        # Always use backslash for Windows paths when environment variables are involved
+        return f"{env_var}\\"
 
-    # Replace SP_DIR path separators to use backslashes consistently
-    normalized_cmd = re.sub(sp_dir_pattern, replace_sp_dir_path, cmd)
+    # Replace environment variable path separators to use backslashes consistently
+    normalized_cmd = re.sub(env_pattern, replace_env_path, cmd)
 
     return normalized_cmd
 
--- a/conda_build/create_test.py
+++ b/conda_build/create_test.py
@@ -8,7 +8,6 @@ from __future__ import annotations
 
 import json
 import os
-import re
 from os.path import basename, exists, isfile, join
 from pathlib import Path
 from typing import TYPE_CHECKING
@@ -23,22 +22,10 @@ def _normalize_path_separators_in_command(cmd: str, is_windows: bool) -> str:
     """
     Normalize path separators in test commands to ensure consistency.
 
-    This function handles the issue where environment variables containing paths
-    (like SP_DIR, PREFIX, BUILD_PREFIX, SRC_DIR, STDLIB_DIR) contain backslashes
-    on Windows, but commands may use forward slashes, creating mixed separators.
-
-    Args:
-        cmd: The command string to normalize
-        is_windows: Whether this is running on Windows
-
-    Returns:
-        The command with normalized path separators
+    This function detects the separator used in environment variables and matches
+    the path separators accordingly, rather than assuming Windows = backslashes.
+    Only normalizes the specific parts containing environment variables.
     """
-    if not is_windows:
-        return cmd
-
-    # Common environment variables that contain paths and need normalization
-    # These variables are set with backslashes on Windows but commands may use forward slashes
     path_env_vars = [
         "SP_DIR",
         "PREFIX",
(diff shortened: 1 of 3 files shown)


alempartegonzalezjorge-cmyk:
[Compra_Bizum_Lu.pdf](https://github.com/user-attachments/files/21583820/Compra_Bizum_Lu.pdf)

danyeaw (approved):
Hi @ryanskeith, I reviewed this and I really like this approach!

Some other things I considered during the review:

1. Could there be other edge cases where this unintentionally replaces path separators
Yes, for a case like `pytest --ignore=%SP_DIR%\tests --message='ignore=%SP_DIR%/tests failed'`, it would give the message `ignore=%SP_DIR%\tests failed`. However, I think the current implementation is a good balance of simplicity, and use cases like calling `--message` on the CLI are unlikely and wouldn't make a big impact.

2. Could you use `os` or `Pathlib` to handle the path separators. Potentially, but since there are unexpanded environment variables like `%SP_DIR%/tests`, not actual file paths. I think your targeted approach to focus on swapping the separators just for this use case is good.

I also looked at this macOS test failure and it doesn't seem related to this PR since it is a macOS shared library loading problem during build, and your changes return immediately for test commands if not on Windows.

tests/test_create_test.py
```diff
@@ -149,3 +150,25 @@ def test_create_run_test(
...
+    """Test that path separators are normalized correctly in test commands."""
+    # Test with forward slash environment variable
+    import os
+
+    os.environ["PREFIX"] = "C:/path/to/prefix"
+    cmd1 = "pytest -v --ignore=%PREFIX%\\tests"
+    result1 = _normalize_path_separators_in_command(cmd1, True)
+    assert result1 == "pytest -v --ignore=%PREFIX%/tests"
+
+    # Test with backslash environment variable
```
soapy1: These are all seperate test cases. With this current test setup, if an early test fails the other tests won't run. Since each of these cases don't depend on eachother, it would be better if these ran independently. This can be achieved by either setting this up as a parameterized test case, or by splitting these out into different test functions.

tests/test_create_test.py
```diff
@@ -149,3 +150,25 @@ def test_create_run_test(
...
         test_file = Path(testing_metadata.config.test_dir, "run_test").with_suffix(ext)
         assert not test_file.exists()
+
+
+def test_path_separator_normalization():
+    """Test that path separators are normalized correctly in test commands."""
+    # Test with forward slash environment variable
+    import os
+
+    os.environ["PREFIX"] = "C:/path/to/prefix"
```
soapy1: hmmm, I think this approach of setting the environment variable may have unintended consequences on the tests. Using this approach, the `PREFIX` env var is modified from it's original state and not restored. That means subsequent tests, will have the last set `PREFIX` env var from this test. For example, consider the test running right after `test_path_separator_normalization`:

```
def test_spdir_not_set():
    import os
    assert os.environ["PREFIX"] != "C:\\path\\to\\prefix"
```

Right now, this will fail.  But you could rearrange test_path_separator_normalization is rearranged  so that istarts passing. For example:

```
def test_path_separator_normalization():
    """Test that path separators are normalized correctly in test commands."""
    # Test with forward slash environment variable
    import os

    # Test with backslash environment variable
    os.environ["PREFIX"] = "C:\\path\\to\\prefix"
    cmd2 = "pytest -v --ignore=%PREFIX%/tests"
    result2 = _normalize_path_separators_in_command(cmd2, True)
    assert result2 == "pytest -v --ignore=%PREFIX%\\tests"

    # Test that non-Windows commands are unchanged
    cmd3 = "pytest -v --ignore=${PREFIX}/tests"
    result3 = _normalize_path_separators_in_command(cmd3, False)
    assert result3 == "pytest -v --ignore=${PREFIX}/tests"

    os.environ["PREFIX"] = "C:/path/to/prefix"
    cmd1 = "pytest -v --ignore=%PREFIX%\\tests"
    result1 = _normalize_path_separators_in_command(cmd1, True)
    assert result1 == "pytest -v --ignore=%PREFIX%/tests"
```

I think this is a little bit naughty. But can be avoided by setting the environment variable for just this test. Something like:

```
def test_path_separator_normalization(monkeypatch):
    """Test that path separators are normalized correctly in test commands."""
    # Test with forward slash environment variable

    monkeypatch.setenv("PREFIX", "C:/path/to/prefix")
    cmd1 = "pytest -v --ignore=%PREFIX%\\tests"
    result1 = _normalize_path_separators_in_command(cmd1, True)
    assert result1 == "pytest -v --ignore=%PREFIX%/tests"

    # Test with backslash environment variable
    monkeypatch.setenv("PREFIX", "C:\\path\\to\\prefix")
    cmd2 = "pytest -v --ignore=%PREFIX%/tests"
    result2 = _normalize_path_separators_in_command(cmd2, True)
    assert result2 == "pytest -v --ignore=%PREFIX%\\tests"

    # Test that non-Windows commands are unchanged
    cmd3 = "pytest -v --ignore=${PREFIX}/tests"
    result3 = _normalize_path_separators_in_command(cmd3, False)
    assert result3 == "pytest -v --ignore=${PREFIX}/tests"
```

soapy1 (changes requested):
Thanks for the PR! A couple of things to update in the tests. 

I think I don't really understand enough about how conda-build tests should work in order to confidently say that this is a good solution. There is noting obviously wrong to me, but it would be good to get another opinon here.

news/5759-windows-path-separators.md
```diff
@@ -0,0 +1,3 @@
+### Bug fixes
+
+* Fix path separator handling in test commands with environment variables on Windows. The fix now detects the actual separator used in environment variables and matches path separators accordingly, using efficient string operations to normalize only the specific parts containing environment variables and prevent mixed separator issues that could cause test failures. (#5759)
```
kenodegard:
```suggestion
* Fix path separator handling in test commands with environment variables on Windows. The fix now detects the actual separator used in environment variables and matches path separators accordingly, using efficient string operations to normalize only the specific parts containing environment variables and prevent mixed separator issues that could cause test failures. (#5759 via #5765)
```

conda_build/create_test.py
```diff
@@ -83,7 +136,9 @@ def create_shell_files(m: MetaData, test_dir: os.PathLike) -> list[str]:
                     f.write("set -ex\n\n")
                 f.write("\n\n")
                 for cmd in commands:
-                    f.write(cmd)
+                    # Normalize path separators for consistent handling
+                    normalized_cmd = _normalize_path_separators_in_command(cmd, status)
+                    f.write(normalized_cmd)
```
kenodegard: Instead of passing in `status`/`is_windows` can we just skip the normalization for non-windows?

```suggestion
                    if status:
                        cmd = _normalize_path_separators_in_command(cmd)
                    f.write(cmd)
```

conda_build/create_test.py
```diff
@@ -18,6 +18,59 @@
...
+        "STDLIB_DIR",
+        "RECIPE_DIR",
+        "LIBRARY_PREFIX",
+        "LIBRARY_BIN",
+        "LIBRARY_INC",
+        "LIBRARY_LIB",
+        "SCRIPTS",
+        "SYS_PREFIX",
+        "ROOT",
+    ]
```
kenodegard: Where did this list come from? How comfortable are we with this being hardcoded and not fuzzier matching?

If we are to leave it hardcoded please move it outside function as a global constant.

conda_build/create_test.py
```diff
@@ -18,6 +18,59 @@
...
+
+    if not is_windows:
+        return cmd
+
+    # Find which environment variables are present in the command
+    cmd_path_env_vars = [var for var in path_env_vars if f"%{var}%" in cmd]
```
kenodegard: Should we account for windows case insensitivity?

conda_build/create_test.py
```diff
@@ -18,6 +18,59 @@
...
+    if not cmd_path_env_vars:
+        return cmd
+
+    # Determine the expected separator for each environment variable
+    envs_separator = {}
+    for env_var in cmd_path_env_vars:
+        env_value = os.environ.get(env_var, "")
+        envs_separator[env_var] = (
+            "/" if env_value.count("/") > env_value.count("\\") else "\\"
+        )
```
kenodegard: If the variable is undefined it will always resolve to `\`

We should call that out in a comment so it is clear that is intentional and add a test for this case

conda_build/create_test.py
```diff
@@ -18,6 +18,59 @@
...
+
+    # Process each part of the command to normalize the path separators
+    for part in cmd.split():
+        for env_var in cmd_path_env_vars:
+            if env_var in part:
+                part_separator = "/" if part.count("/") > part.count("\\") else "\\"
```
kenodegard: Would it be better to extract this as a lambda?

```python
guess_sep = lambda part: "/" if part.count("/") > part.count("\\") else "\\"
```

conda_build/create_test.py
```diff
@@ -18,6 +18,59 @@
...
+    # Determine the expected separator for each environment variable
+    envs_separator = {}
+    for env_var in cmd_path_env_vars:
+        env_value = os.environ.get(env_var, "")
+        envs_separator[env_var] = (
+            "/" if env_value.count("/") > env_value.count("\\") else "\\"
+        )
+
+    # Process each part of the command to normalize the path separators
+    for part in cmd.split():
```
kenodegard: hm should we be using `shlex.split` instead? otherwise `--ignore="%PREFIX%/dir with space/trailing/path"` will do odd things

conda_build/create_test.py
ryanskeith: From docs but admittedly other than assuring SP_DIR was there the rest were just viable add ons.  We could be more selective if we want to just address the testing phase for now.

conda_build/create_test.py
ryanskeith: Are you saying `sp_dir` is just as valid as `SP_DIR` ?

conda_build/create_test.py
ryanskeith: We can do a test case but I was only going to target environment variables that we know are created by conda-build.

conda_build/create_test.py
ryanskeith: Oh, I didn't know that was a thing.  YES, we should using that instead.

ryanskeith: On second thought, we may not want to go down this route.  I will look into it more but it seems that shlex has [only some support for Windows shells](https://docs.python.org/3/library/shlex.html).

In a test it seems to be ok with totally paths but not spaces in sub clauses.
```
Command: '"C:\\Program Files\\My App\\app.exe" --config="C:\\Users\\name\\con fig.ini"'
------------------------------------------------------------
POSIX (posix=True):     ['C:\\Program Files\\My App\\app.exe', '--config=C:\\Users\\name\\con fig.ini']
Non-POSIX (posix=False): ['"C:\\Program Files\\My App\\app.exe"', '--config="C:\\Users\\name\\con', 'fig.ini"']
```

conda_build/create_test.py
```diff
@@ -17,6 +18,100 @@
...
+    the path separators accordingly, rather than assuming Windows = backslashes.
+    Only normalizes the specific parts containing environment variables.
+    """
+
+    # Find which environment variables are present in the command
+    cmd_path_env_vars = [
+        var
+        for var in WINDOWS_PATH_ENV_VARS
+        if f"%{var}%" in cmd or f"%{var.lower()}%" in cmd
+    ]
```
kenodegard, line 50: This is probably better as a regex pattern

conda_build/create_test.py
```diff
@@ -17,6 +18,100 @@
...
+    new_cmd_parts = []
+    for part in corrected_parts:
+        for env_var in cmd_path_env_vars:
+            if env_var in part or env_var.lower() in part:
+                part_separator = guess_separator(part)
+                if part_separator != envs_separator[env_var]:
+                    changed_part = part.replace(part_separator, envs_separator[env_var])
+                    new_cmd_parts.append(changed_part)
+            else:
+                new_cmd_parts.append(part)
```
kenodegard, line 110: This potentially causes the part to be added multiple times:

```python
>>> _normalize_path_separators_in_command('pytest --ignore="%PREFIX%/%SP_DIR%"')
'pytest pytest --ignore="%PREFIX%\\%SP_DIR%" --ignore="%PREFIX%\\%SP_DIR%"'
```

Or even not at all:

```python
>>> _normalize_path_separators_in_command('pytest --ignore=%PREFIX%\\a"')
'pytest'
```

conda_build/create_test.py
```diff
@@ -17,6 +18,100 @@
...
+        parts[0]
+    )  # Shlex does ok with the first part of a command in windows
+
+    i = 1
+    while i < len(parts):
+        part = parts[i]
+
+        # If part has a quote but does not end with a quote, we need to find the next quote and join the parts
+        # This is the case where we have a --option="value with spaces"
+        if '"' in part and not part.endswith('"'):
```
kenodegard, line 83: Is checking for an odd number of quotes better?

kenodegard (changes requested):
I have concerns about the approach we're taking here, as it introduces significant complexity.

Revisiting the problem statement:
> When `SP_DIR` environment variable was set with backslashes on Windows (e.g., `C:\path\to\site-packages`), pytest commands using forward slashes (e.g., `pytest -v --ignore=%SP_DIR%/path/to/ignore`) would create mixed path separators when expanded, causing pytest to fail.

I believe responsibility for correct path separator usage should fall to the user, especially when they're explicitly using Windows environment variables like `%SP_DIR%`.

Additionally, it's worth noting that the slash normalization that makes commands like `pytest -v %SP_DIR%/json_merge_patch/tests.py` work is actually a feature of `pytest`, not `conda-build`. This is demonstrated by the following commands, all of which run successfully on Windows outside of `conda-build`:

```cmd
pytest "%CD%/tests/test_create_test.py"
pytest "%CD%/tests\test_create_test.py"
pytest "%CD%\tests/test_create_test.py"
```

ryanskeith:
Close this fix as it doesn't solve the base issue.

Issue #3337, opened 2018-12-31
Run environment is cleared when multiple outputs are present


chrisburr:
The issue appears to be similar to #3310, however the PR that fixed that issue doesn't resolve this.

### Actual Behavior

When building the `5.x` branch of conda-forge's clangdev feedstock, the build fails with `conda-build >= 3.17.0` as the `$PREFIX` directory is emptied. This affects https://github.com/conda-forge/clangdev-feedstock/pull/44.

From my quick attempt at fixing this, the build dependencies (`m.ms_depends('build')`) appear to be "forgotten" when [this line](https://github.com/conda/conda-build/blob/c96c469a886d23200cde6d1e46a537c8d363c4d3/conda_build/build.py#L1409) is executed.

### Expected Behavior

Configure stage of the build is sucessful, as is the case for `conda-build <= 3.16.3`.

### Steps to Reproduce

```
git clone https://github.com/conda-forge/clangdev-feedstock.git -b 5.x
cd clangdev-feedstock
conda install conda-build=3.17.6
conda build -c conda-forge/label/gcc7 recipe/ --variants '{clang_variant: cling}'
```

This works if `3.17.6` is replaced with `3.16.3` and still fails if `3.17.0` is used.

msarahan:
I'm pretty sure this is the same as https://github.com/conda/conda-build/issues/3277#issuecomment-442979390

Because the MetaData object model in conda-build doesn't separate top-level from outputs in meaningful ways (there's just MetaData objects for both), there's a lot of annoying bugs/edge cases when the top-level package name is the same as an output name.  It's simplest to just avoid these by naming the top-level package name any arbitrary name, and keeping the outputs as the name you really want.

@bergtholdt has a PR to raise an error when the names are the same, https://github.com/conda/conda-build/pull/3314.  I agree with that PR, but I suspect that it will break enough things that it should be handled carefully.

While this is probably a bug, I don't think we're going to fix it unless changing the top-level package name just doesn't work.  A longer-term fix of addressing the data model issue would be time better spent, but that's unfortunately more than we have time for right now.  Community contributions would be most welcome and appreciated.

chrisburr:
Thanks for the quick reply! You're correct, this is the same as https://github.com/conda/conda-build/issues/3277#issuecomment-442979390 and renaming the top-level package avoids the bug.

If raising an error in #3314 is risky perhaps a warning could be issued instead? It could then be promoted to an error once you're confident it's safe to do so.

Issue #3313, opened 2018-12-13
Using same name in main package and outputs causes all kinds of problems, e.g. tests won't get run


bergtholdt:
### Actual Behavior

When re-using the name from the package in an output, all kinds of unexpected behavior can be observed. The latest issue I had was, that tests don't get run.

### Expected Behavior

One of two:

1. error out with a message that you cannot use the package name as an output name
2. any section in output overwrites corresponding package section, only the output gets built, not the parent package

My preference is the first. I implemented it in #3314

### Steps to Reproduce

Just create a test with this content

```yaml
package:
  name: pytest
  version: 1.0

outputs:
  - name: pytest
    test:
      imports:
        - re
```

and you'll get the output like

Nothing to test for: /usr/local/data/miniconda3/conda-bld/linux-64/pytest-1.0-0.tar.bz2

If you change pytest in either the package or the outputs section, the test will run.

##### `Output of conda info`

```
          conda version : 4.5.11
    conda-build version : 3.17.3+6.gc8eae496.dirty
         python version : 3.7.1.final.0
```

bergtholdt:
In the appveyor builds of the patch there are many failing tests due to using the same name in package name and outputs, i.e. the patch works.
But since tests use that kind of naming, maybe 2. is preferred. However I remember having read many posts from @msarahan that it is discouraged to do so. Instead of discouraging I'd prefer checking and error-out since any error message for issues, like the one above, are just not helpful for a user who does not know all conda build internals and has just studied the documentation.

msarahan:
I agree that the addition of errors here is a good thing, and we should make that separation forced.  It may break a few things in the meantime, but it will improve the situation with buggy sometimes-working attempts at merging things.  I'll try to help sort out the tests on your PR.  Forcing this error might mean that we have to cut a conda-build 4.0, but it will be worthwhile I think.

keithj:
Hi,
I'm following this discussion with interest as we have a number of packages where the package and one of its subpackages have the same name. I was surprised that this is discouraged because the Conda documentation implies that this is permitted [1]: "

>  If you do this and you do not **define a subpackage name that matches the top-level package/name**, conda build creates a metapackage for you

If this behaviour is to change, I think the documentation would benefit from a clarification.

[1] https://conda.io/docs/user-guide/tasks/build-packages/define-metadata.html#implicit-metapackages

msarahan:
Yes, the docs need to be changed.

PR #5767, opened 2025-08-04
docs: discourage main-subpackage name collisions for multiple `outputs:` in define-metadata.rst

dlaehnemann:
### Description

This is an attempt at correcting a misleading wording and providing good guidance for the naming of subpackages when using multiple `outputs:` in a `meta.yaml` recipe.

For cross-reference, this docs change was noted as being necessary in previous issues:
* https://github.com/conda/conda-build/issues/3313#issuecomment-447370851

Also, the previous wording implied that having the top-level name of a recipe repeated in one of the subpackage names. This is exactly the behavior that is cited to cause the unintended side-effects, for example here:
* https://github.com/conda/conda-build/issues/4172#issuecomment-757555946
* https://github.com/conda/conda-build/issues/3337#issuecomment-450672091
* https://github.com/conda/conda-build/issues/3277#issuecomment-442979390

Feel free to further improve the wording, this is mostly meant to get the ball rolling again. Especially since previous efforts to make this name re-use a hard error in `conda-build` have all gone stale, see for example:
* #3314
* #4172
* #5715

So this is kind of the minimal fix, something like #3314 would be much nicer in the mid-term. And I guess that in the long-term, this should go away if the new recipe format is fully implemented and enforced, see:
* https://github.com/conda/ceps/pull/56
* https://conda.org/learn/ceps/cep-0014/
* 

### Checklist - did you ...


ryanskeith (approved):
Thank you for the addition of this warning.

jezdez (approved):

Issue #5665, opened 2025-04-02
Invalid tag seen with new lief (flaky)


h-vetinari:
### Checklist

- [x] I added a descriptive title
- [x] I searched open reports and couldn't find a duplicate

### What happened?

Ah, lief, the gift that keeps on giving...

On the arrow feedstock, I'm getting
```
  File "/opt/conda/lib/python3.12/site-packages/conda_build/os_utils/liefldd.py", line 381, in get_uniqueness_key
    result = [e.name for e in dynamic_entries if e.tag == ELF_DYNAMIC_TAGS.SONAME]
                                                 ^^^^^
ValueError: 1879047926 is not a valid TAG.
```
but only on linux-aarch64. The other architectures are fine ([logs](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1213078&view=results)).

More importantly, the failures seem to have gone away after a restart. But given the issues with getting compatibility with lief 0.15+ off the ground, I thought I'd raise the issue nevertheless, in case someone else happens to run into it.

### Conda Info

<details>

```shell
++ conda info

     active environment : base
    active env location : /opt/conda
            shell level : 1
       user config file : /home/conda/.condarc
 populated config files : /opt/conda/.condarc
                          /home/conda/.condarc
          conda version : 25.3.0
    conda-build version : 25.3.1
         python version : 3.12.9.final.0
                 solver : libmamba (default)
       virtual packages : __archspec=1=x86_64_v4
                          __conda=25.3.0=0
                          __glibc=2.34=0
                          __linux=6.8.0=0
                          __unix=0=0
       base environment : /opt/conda  (writable)
      conda av data dir : /opt/conda/etc/conda
  conda av metadata url : None
           channel URLs : https://conda.anaconda.org/conda-forge/linux-64
                          https://conda.anaconda.org/conda-forge/noarch
          package cache : /home/conda/feedstock_root/build_artifacts/pkg_cache
                          /opt/conda/pkgs
       envs directories : /opt/conda/envs
                          /home/conda/.conda/envs
               platform : linux-64
             user-agent : conda/25.3.0 requests/2.32.3 CPython/3.12.9 Linux/6.8.0-1021-azure almalinux/9.5 glibc/2.34 solver/libmamba conda-libmamba-solver/25.3.0 libmambapy/2.0.8
                UID:GID : 1001:1001
             netrc file : None
           offline mode : False
```

</details>

### Conda Config

<details>

```shell
[...]

PR #5778, opened 2025-08-27
handle dynamic tags unknown to lief

cbouss:
### Description

Lief dynamic entry tags are defined here:
https://github.com/lief-project/LIEF/blob/main/include/LIEF/ELF/DynamicEntry.hpp#L47

We are seeing linux aarch64 binaries with tag TLSDESC_PLT, which is not defined in lief.

conda build trips on this while looking for tag SONAME, with a `ValueError: 1879047926 is not a valid TAG.`, as described in this issue: https://github.com/conda/conda-build/issues/5665

This is true for both lief 0.16.4 and 0.16.6.

Used workaround have been so far to:
- pin lief to version 0.12.3, which does not have TLSDESC_PLT defined, but does not throw a value error. https://github.com/anaconda/docker-images/pull/682
- use `overlinking_ignore_patterns`. https://github.com/AnacondaRecipes/oracle-instant-client-feedstock/pull/1

This change filters dynamic_entries to only valid instances of ELF_DYNAMIC_TAGS.

### Checklist - did you ...


kenodegard:
While some of the test failures are unrelated, e.g.:
- #5779 
- #5785 

Others are possibly related?

ryanskeith (approved):
These changes look great to me.  Thank you @cbouss and @kenodegard

ryanskeith:
@cbouss let's work together to figure out why the tests are failing.

news/5778.md
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Handle unknown dynanic tags in liefldd. (#5778)
```
h-vetinari:
```suggestion
* Handle unknown dynamic tags in liefldd. (#5778)
```

tkelman:
Don't consider this question blocking at all for this workaround, but would there be any reason upstream lief wouldn't want to add an entry for TLSDESC_PLT to that enum?

PR #5817, opened 2025-10-16
Add debug argument to conda debug

### LINKED ISSUE #5816: Conda debug does not provide a `--debug` option
### Checklist

### What is the idea?

If you want to see a verbose output of the resolution (Etc. etc.), conda-build provides
```
$ conda debug --debug recipe
DEBUG conda.gateways.logging:set_log_level(223): log_level set to 10
usage: conda debug [-h] [-V] [-n] [--python PYTHON] [--perl PERL] [--numpy NUMPY] [--R R_BASE] [--lua LUA] [--append-file APPEND_SECTIONS_FILE] [--clobber-file CLOBBER_SECTIONS_FILE] [-m VARIANT_CONFIG_FILES] [-e EXCLUSIVE_CONFIG_FILES] [--use-channeldata] [--variants VARIANTS] [-c CHANNEL] [--use-local] [--override-channels]
                   [--repodata-fn REPODATA_FNS] [--experimental {jlap,lock}] [--no-lock] [--repodata-use-zst | --no-repodata-use-zst] [-p PATH] [-o OUTPUT_ID] [-a]
                   recipe_or_package_file_path
conda debug: error: unrecognized arguments: --debug
```
as an option. However, if you want the same information for conda debug this does not work (tested on conda-debug 25.7.0)

### Why is this needed?

I periodically see differences in how conda build and conda debug resolve environments. The debug flag allows for more verbose output during build, but it missing during debug means that I cannot compare.

### What should happen?

The `--debug` flag should turn on verbose/debugging information, same as for conda build.

### Additional Context

_No response_

simon-ess:
### Description

Resolves #5816 by adding a `--debug` option to conda debug.

### Checklist - did you ...

- ~[ ] Add / update necessary tests?~
- ~[ ] Add / update outdated documentation?~

### DIFF
--- a/.github/workflows/builds-review.yaml
+++ b/.github/workflows/builds-review.yaml
@@ -26,7 +26,7 @@ jobs:
     runs-on: ${{ matrix.runner }}
     steps:
       - name: Remove build label
-        uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea
+        uses: actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd
         with:
           github-token: ${{ secrets.CANARY_ACTION_TOKEN }}
           script: |
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -113,7 +113,7 @@ jobs:
         run: echo "HASH=${{ runner.os }}-${{ runner.arch }}-Py${{ matrix.python-version }}-${{ matrix.conda-version }}-${{ matrix.test-type }}-$(date -u "+%Y%m")" >> $GITHUB_ENV
 
       - name: Cache Conda
-        uses: actions/cache@0400d5f644dc74513175e3cd8d07132dd4860809 # v4.2.4
+        uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.3.0
         with:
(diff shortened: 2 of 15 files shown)


conda_build/cli/main_debug.py
```diff
@@ -63,6 +63,11 @@ def get_parser() -> ArgumentParser:
         help="Output only the string to the used generated activation script.  Use this for creating envs in scripted "
         "environments.",
     )
+    p.add_argument(
+        "--debug",
```
jaimergp, line 67: Do we want to add `--verbose` too? Looks like the `debug` subcommand parser extends the `render` subcommand parser, but they are adding `--verbose` separately there.

simon-ess: Good question. I note that `conda build` extends the same parser and also does not include `--verbose`as a flag...

However in each of those cases it seems that there is a differenct setting for verbosity:
```py
        activation_string = api.debug(
            parsed.recipe_or_package_file_path,
            verbose=(not parsed.activate_string_only),
            **parsed.__dict__,
        )
```
(see https://github.com/conda/conda-build/blob/5fc121ed3521755ce57208d9c6977febffd2aa4c/conda_build/cli/main_debug.py#L104)

Issue #5733, opened 2025-06-09
`linux_vars` function setting `BUILD` wrong when `cdt_name` is in use


bwingconda:
### Checklist

- [x] I added a descriptive title
- [x] I searched open reports and couldn't find a duplicate

### What happened?

When building with a `conda_build_config.yaml` containing a change to the `cdt_name` variable:
```
cdt_name:  # [linux]
  - el8    # [linux]
```
[the proper updates](https://github.com/conda/conda-build/blob/main/conda_build/jinja_context.py#L559-L568) are made to fill out the Jinja templates with the correct complete package names; however, there is still one environment variable being set using guesses from the host architecture.

I would expect the `BUILD` environment variable to be set using the overridden `cdt_name` when it is provided. Instead, [the code](https://github.com/conda/conda-build/blob/main/conda_build/environ.py#L747-754) looks to only allow for `cos6` or `cos7` purely based on the architecture of the host.

> (build) [root@41bccd68ff6c pkgs]# env
cdt_name=el8
BUILD=x86_64-conda_cos6-linux-gnu
...

### Conda Info

```shell
active environment : build
    active env location : /opt/conda/envs/build
            shell level : 2
       user config file : /root/.condarc
 populated config files : /opt/conda/.condarc
          conda version : 24.1.2
    conda-build version : 24.1.2
         python version : 3.12.4.final.0
                 solver : libmamba (default)
       virtual packages : __archspec=1=icelake
                          __conda=24.1.2=0
                          __glibc=2.28=0
                          __linux=5.10.237=0
                          __unix=0=0
       base environment : /opt/conda  (writable)
      conda av data dir : /opt/conda/etc/conda
  conda av metadata url : None
           channel URLs : https://repo.anaconda.com/pkgs/main/linux-64
                          https://repo.anaconda.com/pkgs/main/noarch
                          https://repo.anaconda.com/pkgs/r/linux-64
                          https://repo.anaconda.com/pkgs/r/noarch
          package cache : /opt/conda/pkgs
                          /root/.conda/pkgs
       envs directories : /opt/conda/envs
                          /root/.conda/envs
               platform : linux-64
             user-agent : conda/24.1.2 requests/2.32.3 CPython/3.12.4 Linux/5.10.237-230.948.amzn2.x86_64 rocky/8.9 glibc/2.28 solver/libmamba conda-libmamba-solver/25.3.0 libmambapy/2.0.5 aau/0.7.0 c/. s/. e/.
[...]

PR #5836, opened 2025-11-04
Bumping ruff target-version to Python 3.10 and fixing CI

ryanskeith:
### Description

- Bumped ruff python target to 3.10.
- Ran precommit autofix for fixes

### Checklist - did you ...


ryanskeith:
This is feeling like it is related to issue #5733 and it is appearing now as the cdt's were changed.  I feel we should address that issue first so we can allow the user to choose the cdt they wish.

tests/test-recipes/metadata/_render_recipe/conda_build_config.yaml
h-vetinari, line 1: I recognize a lot of older iterations of the conda-forge pinning in that file. There's a bunch of things that likely need to update (not least vs2019 -> vs2022, drop CUDA 11, update compiler pins on unix, etc.)

h-vetinari (commented):
My original suggestion was going to be to set
```yaml
cdt_name:  # [linux]
  - conda  # [linux]
```
in that test CBC; however, it doesn't actually seem to govern that specific test, because despite the current value being
https://github.com/conda/conda-build/blob/5fc121ed3521755ce57208d9c6977febffd2aa4c/tests/test-recipes/metadata/_render_recipe/conda_build_config.yaml#L143-L144
and https://github.com/conda/conda-build/blob/main/tests/test-recipes/metadata/cdt_linking/meta.yaml only using vanilla `{{ cdt("foo") }}`, `tests/test_api_build.py::test_recipe_builds[cdt_linking]` ends up using `cos6` CDTs.

From what I can tell from a quick code search (on `cdt_name`), it appears likely that this just falls back to the default here

https://github.com/conda/conda-build/blob/5fc121ed3521755ce57208d9c6977febffd2aa4c/conda_build/jinja_context.py#L559-L562

i.e. the test doesn't pick up the CBC you're updating here.

PS. Obviously that `cos6` default is highly questionable nowadays and should likely be updated in and of itself.

PR #5857, opened 2025-11-26
Fix empty channels being passed to solver handling `_h_env` creation

jaimergp:
### Description

Use `Index.expanded_channels` instead of `LAST_CHANNEL_URLS` hack (which is a global populated by `conda`'s `Index` on initialization; 😷). This fixes a problem with the channel list being passed as empty in some contexts, which results in an error after conda 25.9 because `context.channels` _may_ be empty if no configuration file is present.

Fixes https://github.com/conda/conda-libmamba-solver/issues/790. 

### Checklist - did you ...

### DIFF
--- a/conda_build/environ.py
+++ b/conda_build/environ.py
@@ -1281,7 +1281,7 @@ def install_actions(
         },
         callback=reset_context,
     ):
-        channels: tuple[Channel, ...] = tuple(index.expanded_channels) or None
+        channels: tuple[Channel, ...] | None = tuple(index.expanded_channels) or None
         subdirs: tuple[str, ...] | None = tuple(index._subdirs) or None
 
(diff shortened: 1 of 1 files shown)


conda_build/environ.py
```diff
@@ -1280,25 +1281,8 @@ def install_actions(
...
-                tuple(
-                    dict.fromkeys(
-                        subdir for channel in channels if (subdir := channel.subdir)
-                    )
-                )
-                or context.subdirs
-            )
-        else:
-            channels = subdirs = None
+        channels: tuple[Channel, ...] = tuple(index.expanded_channels) or None
```
isuruf:
```suggestion
        channels: tuple[Channel, ...] | None = tuple(index.expanded_channels) or None
```

PR #5282, opened 2024-04-12
Inline `MetaData` ordering when dumping to yaml

kenodegard:
### Description

Simplified `conda_build.render.output_yaml` to avoid the most costly `MetaData.copy` and to eliminate the need for special customizations of the yaml engine.

### Checklist - did you ...

- [x] Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [ ] ~Add / update necessary tests?~
- [ ] ~Add / update outdated documentation?~

PR #5902, opened 2026-01-26
Use inclusive language: rename whitelist to allowlist

### LINKED ISSUE #4540: Rename variables named `white*` to `allow*`
### Description

Derived from our inclusive language efforts (see https://github.com/conda/conda-build/issues/4494 and [7.1 of our styleguide](https://github.com/conda/infra/blob/main/STYLEGUIDE.md)).

> **@kenodegard said in https://github.com/conda/conda-build/issues/4494#issuecomment-1195559278:**
> There are two `white*` variables in conda-build, both are recipe keywords: `missing_dso_whitelist` and `runpath_whitelist`.
> 
> Both keywords are also in use in `mamaba`/`boa` so we will need to offer similar modifications there.
> 
> Furthermore, since conda-build doesn't have an aliasing system like `conda.base.context.Context` does, this becomes a large enough change that I'm wondering if we should backlog the `allow`/`deny` rewording for now?
> 
> At the very least, the `allow`/`deny` rewording should be spun off from this epic into a separate effort.

jezdez:
## Summary

This PR updates the codebase to use more inclusive language by renaming `whitelist` to `allowlist`.

- Add `missing_dso_allowlist` and `runpath_allowlist` as preferred recipe keys
- Old keys (`missing_dso_whitelist`, `runpath_whitelist`) remain fully supported for backward compatibility
- Rename internal variables and constants
- Update documentation with backward compatibility note

## Backward Compatibility

- Both old and new recipe keys are validated and work
- New keys take precedence when both are specified  
- Existing recipes continue to work unchanged
- Test recipes using old keys verified to still work

Supersedes #5301
Partially resolves #4540

### DIFF
--- a/conda_build/post.py
+++ b/conda_build/post.py
@@ -43,6 +43,7 @@ from conda.misc import walk_prefix
 from conda.models.records import PrefixRecord
 
 from . import utils
+from .deprecations import deprecated
 from .exceptions import (
     CondaBuildException,
     OverDependingError,
@@ -782,7 +783,7 @@ def library_nature(
 
 
 # This is really just a small, fixed sysroot and it is rooted at ''. `libcrypto.0.9.8.dylib` should not be in it IMHO.
-DEFAULT_MAC_WHITELIST = [
+DEFAULT_MAC_ALLOWLIST = [
     "/opt/X11/",
     "/usr/lib/libSystem.B.dylib",
     "/usr/lib/libcrypto.0.9.8.dylib",
@@ -830,7 +831,7 @@ DEFAULT_MAC_WHITELIST = [
 
 # Should contain the System32/SysWOW64 DLLs present on a clean installation of the
 # oldest version of Windows that we support (or are currently) building packages for.
-DEFAULT_WIN_WHITELIST = [
+DEFAULT_WIN_ALLOWLIST = [
     "**/ADVAPI32.dll",
     "**/bcrypt.dll",
     "**/COMCTL32.dll",
@@ -853,6 +854,22 @@ DEFAULT_WIN_WHITELIST = [
     "**/ntdll.dll",
     "**/msvcrt.dll",
 ]
+
+# Deprecated constant aliases for backward compatibility
+deprecated.constant(
+    "26.9",
+    "27.3",
+    "DEFAULT_MAC_WHITELIST",
+    DEFAULT_MAC_ALLOWLIST,
+    addendum="Use 'DEFAULT_MAC_ALLOWLIST' instead.",
+)
+deprecated.constant(
+    "26.9",
+    "27.3",
+    "DEFAULT_WIN_WHITELIST",
+    DEFAULT_WIN_ALLOWLIST,
+    addendum="Use 'DEFAULT_WIN_ALLOWLIST' instead.",
+)
 
 
 def _expand_dsolist(dsolist):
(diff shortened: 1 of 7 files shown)


conda_build/post.py
```diff
@@ -1444,17 +1444,17 @@ def check_overlinking_impl(
...
-            if not whitelist and not denylist:
+            allowlist, denylist = dsolists["allow"], dsolists["deny"]
+            if not allowlist and not denylist:
                 sysroots = ["C:/Windows"]
-                whitelist = DEFAULT_WIN_WHITELIST
+                allowlist = DEFAULT_WIN_ALLOWLIST
             build_is_host = True if on_win else False
 
-    whitelist += missing_dso_whitelist or []
+    allowlist += missing_dso_whitelist or []
```
```suggestion
    allowlist += missing_dso_allowlist or []
```

jezdez: `missing_dso_whitelist` is still passed in

conda_build/post.py
jezdez:
```suggestion
    missing_dso_allowlist,
```

conda_build/post.py
```diff
@@ -1673,8 +1673,13 @@ def check_overlinking(m: MetaData, files, host_prefix=None):
         [req.split(" ")[0] for req in m.get_value("requirements/host", [])],
         host_prefix or m.config.host_prefix,
         m.config.build_prefix,
-        m.get_value("build/missing_dso_whitelist", []),
-        m.get_value("build/runpath_whitelist", []),
+        # Support both old and new recipe keys (new preferred)
+        m.get_value("build/missing_dso_allowlist")
+        if m.get_value("build/missing_dso_allowlist") is not None
+        else m.get_value("build/missing_dso_whitelist", []),
```
isuruf: How about
```suggestion
        m.get_value("build/missing_dso_allowlist", m.get_value("build/missing_dso_whitelist", [])),
```
?

isuruf (approved):

jezdez:
@kenodegard — addressing your "Deprecations?" question. The rebase in [`1952e96d`](https://github.com/conda/conda-build/commit/1952e96d) added the deprecation machinery for all four touchpoints using the `conda_build.deprecations` framework, matching the pattern used in prior deprecation PRs (e.g. #5282, #5726):

- `@deprecated.argument` on `check_overlinking_impl` for the `missing_dso_whitelist` / `runpath_whitelist` kwargs (automatic rename to the new names).
- `deprecated.constant` on the module-level `DEFAULT_MAC_WHITELIST` / `DEFAULT_WIN_WHITELIST` (still resolve to the new constants, with a warning).
- `deprecated.topic` in `check_overlinking()` for the `build/missing_dso_whitelist` / `build/runpath_whitelist` recipe keys, only firing when the legacy key is set and the new key is not.

Versions: pending in `26.9`, removal in `27.3`. Just pushed [`a3a7f75f`](https://github.com/conda/conda-build/commit/a3a7f75f) adding tests in `tests/test_post.py::TestWhitelistDeprecations` covering:

1. Both constants still alias the new values and emit a deprecation warning.
2. Both `check_overlinking_impl` kwargs still work (with warning + rename).
3. Both legacy recipe keys emit a warning via `check_overlinking()`.
4. When both legacy and new keys are set, the new key wins silently (no warning).

Expanded the news file to list all four deprecations individually with their `deprecate_in` / `remove_in` versions so the release notes read cleanly.

jaimergp (approved):

PR #5904, opened 2026-01-27
Add test environment template cloning and CI tuning

jezdez:
### Description

Two things: a `create_env` clone optimization that skips the solver when a pre-built template environment matches, and some CI workflow cleanup.

The bigger matrix/runner changes (Python 3.14, macOS ARM, `linux-benchmarks` job) already landed on `main` via #5941 and #5944 while this sat in review, so the workflow diff is smaller than it used to be.

#### Template cloning (`conda_build/environ.py`, `conda_build/config.py`)

New `test_env_template` config setting. When pointed at a valid prefix, `create_env` runs `conda create --clone` instead of solving from scratch. Falls back to the normal solve when:

- any requested spec doesn't match a template record (exact `name=version=build` for `PackageRecord` inputs, `MatchSpec.match()` for string specs)
- the template has packages the caller didn't ask for
- `config.disable_pip` is set and the template contains pip/setuptools/wheel

Uses `PrefixData.get()` and `PackageRecord.spec` directly, nothing hand-rolled.

#### `warm_package_cache` fixture (`tests/conftest.py`)

Session-scoped, autouse fixture that builds a shared template env (python+pip+setuptools+expat) and feeds it into `testing_config`. Guarded by `filelock.FileLock` so pytest-xdist workers don't race on `~/conda_pkgs_dir`.

#### Workflow tuning (`.github/workflows/tests.yml`)

- Drop Python 3.11 from Linux (3.10/3.12 cover the boundary). PRs run 3.10 + 3.13; main/schedule adds 3.12.
- Exclude benchmark tests from regular jobs (`and not benchmark` in `PYTEST_MARKER`).
- Add `restore-keys` fallbacks on all cache steps.
- Reduce flaky reruns from 3 to 1 on Linux.

#### Test recipe fixes

- xz 5.2.3 -> 5.6 in `30_top_level_finalized` (old version gone from defaults), assertion updated
- libpng 1.6.34 -> 1.6.37 in `always_include_files_glob`
- `test_disable_pip` and `test_add_pip_as_python_dependency_from_condarc_file` set `test_env_template = None` so the template doesn't defeat them

### Checklist - did you ...

### CI impact

Job counts (test matrix only):

| Platform | before | after (main) | after (PRs) |
| -------- | ------ | ------------ | ----------- |
| Linux | 12 + 1 bench | 8 + 1 bench | 6 + 1 bench |
| Windows | 4 | 4 | 4 |
| macOS | 4 | 4 | 4 |


.github/workflows/tests.yml
```diff
@@ -402,9 +417,9 @@ jobs:
             conda-version: canary
             test-type: parallel
     env:
-      # Use conda-forge for release tests since conda 25.11+ is not in defaults for osx-64
       CONDA_CHANNEL_LABEL: ${{ matrix.conda-version == 'canary' && 'conda-canary/label/dev' || 'conda-forge' }}
```
beckermr, line 431:
```suggestion
      CONDA_CHANNEL_LABEL: ${{ matrix.conda-version == 'canary' && 'conda-canary/label/dev' || 'defaults' }}
```

conda_build/environ.py
```diff
@@ -979,6 +980,108 @@ def get_install_actions(
...
+        for spec in specs:
+            if isinstance(spec, str):
+                # Parse "python>=3.10" or "python" format
+                name = (
+                    spec.split()[0]
+                    .split("=")[0]
+                    .split(">")[0]
+                    .split("<")[0]
+                    .split("!")[0]
+                )
```
beckermr: What about `~=`? What about specs with `[]` in them? 

This code block really should use one of the builtin spec parsers. I do not know precisely which one though since there is more than one syntax floating around.

conda_build/environ.py
```diff
@@ -979,6 +980,108 @@ def get_install_actions(
...
+                "Template env missing packages %s, skipping clone",
+                missing,
+            )
+            return False
+
+    except Exception:
+        # If we can't check the template, fall back to normal creation
+        return False
+
+    # Template satisfies our needs, clone it using the fastest available method
```
beckermr: What about prefix replacement? Shouldn't we be using `conda clone` here?

beckermr (changes requested):
I have a few questions and confusions on my part.

conda_build/environ.py
jezdez: yeah, agree

beckermr (approved):
LGTM!

PR #5908, opened 2026-01-30
CI Improvements for stability

### LINKED ISSUE #5078: Remaing CI test concurrency basic issues

ryanskeith:
### Description

This PR seeks to reduce CI flakiness by fixing issue #5078 due to the fact that tests marked by serial were not running serially.  This is done by using the py-test marker to run a slightly different pytest command.  Also, marked a few more tests as serial during testing that were causing issues on parallel.  Went more conservative thinking Stability > Speed.

### Changes
1. Fixed CI workflow (.github/workflows/tests.yml)
- Serial tests now run without pytest-xdist (single process)
- Parallel tests continue using -n auto
- Cross-platform compatible using PYTEST_XDIST env var
2. Marked 5 flaky tests as serial
- test_subpackage_recipes - repodata cache locking
- test_inner_python_loop_with_output - repodata cache locking
- test_remove_workdir_default - package cache corruption
- test_overlinking_detection - package cache corruption + tool crashes
- test_conda_pkg_format - Windows file locking
3. Suppressed deprecation warning (pyproject.toml)
- Ignore CONDA_PACKAGE_EXTENSIONS warning from conda

#### Impact
- ✅ Serial tests now actually run serially
- ✅ Expected significant improvement in CI stability
- ⚠️ Some tests run slower (no parallelism) but reliably (hopefully)

### Checklist - did you ...

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -102,6 +102,8 @@ jobs:
       CONDA_CHANNEL_LABEL: ${{ matrix.conda-version == 'canary' && 'conda-canary/label/dev::' || '' }}
       CONDA_VERSION: ${{ contains('canary|release', matrix.conda-version) && 'conda' || format('conda={0}', matrix.conda-version) }}
       PYTEST_MARKER: ${{ matrix.test-type == 'serial' && 'serial' || 'not serial' }}
+      # Serial tests run without xdist (-n auto) to avoid race conditions
+      PYTEST_XDIST: ${{ matrix.test-type == 'serial' && ' ' || '-n auto' }}
 
     steps:
       - name: Free Disk Space
@@ -180,12 +182,13 @@ jobs:
         run: conda list --show-channel-urls
 
       - name: Run Tests
-        run: |
-          if [ "${{ matrix.test-type }}" = "serial" ]; then
-            pytest --cov=conda_build --reruns 3 --reruns-delay 1 -m "${{ env.PYTEST_MARKER }}"
-          else
-            pytest --cov=conda_build --reruns 3 --reruns-delay 1 -n auto -m "${{ env.PYTEST_MARKER }}"
-          fi
+        run: >
+          pytest
+          --cov=conda_build
+          --reruns 3
+          --reruns-delay 1
+          ${{ env.PYTEST_XDIST }}
+          -m "${{ env.PYTEST_MARKER }}"
 
       - name: Upload Coverage
         uses: codecov/codecov-action@5a1091511ad55cbe89839c7260b706298ca349f7 # v5.5.1
@@ -295,6 +298,8 @@ jobs:
       ErrorActionPreference: Stop  # powershell exit on first error
       CONDA_CHANNEL_LABEL: ${{ matrix.conda-version == 'canary' && 'conda-canary/label/dev' || 'defaults' }}
       PYTEST_MARKER: ${{ matrix.test-type == 'serial' && 'serial' || 'not serial and not slow' }}
+      # Serial tests run without xdist (-n auto) to avoid race conditions
+      PYTEST_XDIST: ${{ matrix.test-type == 'serial' && ' ' || '-n auto' }}
 
     steps:
       - name: Checkout Source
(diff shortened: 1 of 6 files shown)


.github/workflows/tests.yml
```diff
@@ -102,6 +102,8 @@ jobs:
       CONDA_CHANNEL_LABEL: ${{ matrix.conda-version == 'canary' && 'conda-canary/label/dev::' || '' }}
       CONDA_VERSION: ${{ contains('canary|release', matrix.conda-version) && 'conda' || format('conda={0}', matrix.conda-version) }}
       PYTEST_MARKER: ${{ matrix.test-type == 'serial' && 'serial' || 'not serial' }}
+      # Serial tests run without xdist (-n auto) to avoid race conditions
```
dholth, line 105:
```suggestion
      # Serial tests run without xdist worker processes (-n auto) to avoid race conditions
```

Is there a higher-level place that a non-repeated env can go?

beckermr (changes requested):
I am a strong -1 on this PR. My understanding of the original issue is that changes were made so that things that had been marked as serial no longer needed to be serial. Further the conda-build tests run fine after a few retries. 

OTOH this pr is producing CI times in excess of 1.5 hours and the CI is still failing. 

We need to go through each test and determine how it needs to be run, removing serial markers as much as possible. Then we can merge this PR for the remaining tests that actually need to be serial.

.github/workflows/tests.yml
ryanskeith: Sadly, it cannot. Matrix.test-type is on the job level.

beckermr:
@jezdez I suggest we work on and merge your PR #5904 before this one. Effort on that PR will help make further debugging of flaky testing failures faster.

Issue #5910, opened 2026-01-30
Replace deprecated CONDA_PACKAGE_EXTENSIONS before conda 27.3


ryanskeith:
### Checklist

- [x] I added a descriptive title
- [x] I searched open reports and couldn't find a duplicate

### What happened?

conda.base.constants.CONDA_PACKAGE_EXTENSIONS is pending deprecation and will be removed in conda 27.3. conda-build currently imports and uses this constant in several files:
- conda_build/utils.py
- conda_build/build.py
- conda_build/api.py
- conda_build/environ.py
- conda_build/cli/validators.py

The deprecation warning states:
```
PendingDeprecationWarning: conda.base.constants.CONDA_PACKAGE_EXTENSIONS is pending deprecation and will be removed in 27.3. Use `conda.base.context.context.plugin_manager.get_package_extractors()` instead.
```

Current workaround: The warning is suppressed in pyproject.toml filterwarnings to keep CI passing.

Required action: Before conda 27.3, conda-build should migrate from CONDA_PACKAGE_EXTENSIONS to the new plugin-based API context.plugin_manager.get_package_extractors().

### Conda Info

```shell
Found during CI testing.
```

### Conda Config

```shell
Found during CI testing.
```

### Conda list

```shell
Found during CI Testing.
```

### Additional Context

- Discovered during CI testing with conda-canary/label/dev
  - https://github.com/conda/conda-build/actions/runs/21500298577/job/61944958845?pr=5908
- Warning is currently suppressed in pyproject.toml to unblock CI
- Deadline: conda 27.3 release

kenodegard:
conda-build uses `CONDA_PACKAGE_EXTENSIONS` as a list of extensions it supports so it doesn't make sense for it to rely on `context.plugin_manager.get_package_extractors()` since it wont know what to do with non-conda formats.

PR #5917, opened 2026-02-03
Inline `CONDA_PACKAGE_EXTENSIONS`

kenodegard:
### Description

Fix https://github.com/conda/conda-build/issues/5910

### Checklist - did you ...

- [ ] ~Add a file to the `news` directory ([using the template](https://github.com/conda/conda-build/blob/main/news/TEMPLATE)) for the next release's release notes?~
     

kenodegard:
> Hm, this might not be enough actually; we should make a decision if we can switch to get_conda_package_extensions or not

I don't think so.

`get_package_extractors()` returns formats that conda can extract, but conda-build only works with conda package formats (.tar.bz2/.conda). Even if a plugin adds support for, e.g., extracting .whl files, conda-build still only knows how to build/test .tar.bz2/.conda.

Wouldn't using the plugin API here cause conda-build to accept formats it doesn't actually know how to handle? Or am I missing a use case where conda-build should be plugin-aware for package formats?

ryanskeith (approved):
I like the simple fix here.  It looks like the deprecation error is still showing itself in testing so the code may still need to filter out the warning.

PR #5880, opened 2025-12-17
Initial rattler-build integration

jsmolic:
This PR introduces an initial prototype for integrating `rattler-build` with `conda-build`. The integration is done purely using the `rattler-build` CLI, without the python bindings.

It adds `rattler-build` equivalents for `conda-build`, `conda-debug`, `conda-inspect`, `conda-render` and `conda-skeleton` commands.

After merging this PR, implementation can be rewritten using [py-rattler-build](https://github.com/prefix-dev/rattler-build/tree/main/py-rattler-build).

Few need to be added as well as see if there are more shared arguments between `rattler-build` and `conda-*` commands that can be used with rattler calls.

Reference issue: https://github.com/conda/conda-build/issues/5858

PR #5919, opened 2026-02-06
Add support for generating v1 recipes

jsmolic:
This PR adds support for generating v1 recipes using `conda-recipe-manager` to convert recipes generated with `conda-skeleton` to the newer format.

### Checklist - did you ...


beckermr (changes requested):
We should not merge this. Instead we should centralize around grayskull and move all recipe generation to that package.

jaimergp:
This is coming from https://github.com/conda/conda-build/pull/5880. Initially the integration strategy used `rattler-build` directly, but Isuru suggested to convert whatever `skeleton` was creating for `meta.yaml` into `recipe.yaml` via `conda-recipe-manager`.

I don't see it as controversial, but if this is not going in, we should think of deprecating `skeleton` fully.

Issue #5898, opened 2026-01-15
Add network configuration for v1 recipe support


danyeaw:
### Checklist

- [x] I added a descriptive title
- [x] I searched open requests and couldn't find a duplicate

### What is the idea?

Pass networking config (auth tokens, user-agents, proxies) to py-rattler-build.

### Why is this needed?

Private channels need authentication, corporate networks need proxies.

### What should happen?

1. Research networking API in py-rattler-build
2. Pass through auth tokens from `.condarc` or env vars
3. Configure user-agent and proxy settings

### Additional Context

- Auth: `.condarc` or `CONDA_TOKEN_*` env vars
- Proxies: `HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY`
- May require coordinating new py-rattler-build APIs if networking details not currently exposed

jsmolic:
I'll pick this issue up.

jsmolic:
According to my research and the available documentation, `conda-build` itself doesn't do much for private channel authentication. Instead, users are expected to authenticate using other tools such as `conda-auth` or `anaconda-auth` and then use `conda-build` with those channels. Rattler-build does have an [`auth`](https://rattler-build.prefix.dev/latest/reference/cli/rattler-build/auth/) command for this purpose, but it's not currently exposed in the python API.

My suggestion is that we document the current behavior and close the issue for now.

What do you think @danyeaw @jaimergp ?

danyeaw:
Hey @jsmolic, if a user did authenticate with conda-auth, build a v1 recipe, and then tried to upload to a private location, would it work?

jsmolic:
Hey @danyeaw, I did try uploading package built from a v1 recipe in https://github.com/conda/conda-build/pull/5955#issue-4298844994. Since I only have a free account I can't make my testing channel private (I still had to authenticate via `conda-auth` or `.condarc` to upload a package), but I think what you are describing should work as `conda-build` in this workflow would be used to build and upload the package, so as long as authentication is successful I don't see why it would not work.

jaimergp:
Upload to private channels should work as in conda-build because it uses the same machinery, IIRC. Building from private channels may need testing because I'm not sure how well rattler libs will read local configs on their own, or whether we are passing everything already.

jaimergp:
Oops, jinx!

danyeaw:
Ok great, how should we document this?

jsmolic:
[...]

Issue #5896, opened 2026-01-15
Create documentation for v1 recipes


danyeaw:
### Checklist

- [x] I added a descriptive title
- [x] I searched open reports and couldn't find a duplicate

### What happened?

Create user documentation for v1 recipes, so that users can understand how to build v1 recipes and any limitations. This includes:

1. Add `docs/source/user-guide/v1-recipes.rst`
2. Update README.md and CLI docs
3. Create migration guide (v0→v1)
4. Document limitations

### Additional Context

This should be updated incrementally once we have this initial set of docs.

jsmolic:
I'll pick this issue up

Issue #5894, opened 2026-01-15
Support anaconda.org upload after build for v1 recipes


danyeaw:
### Checklist

- [x] I added a descriptive title
- [x] I searched open requests and couldn't find a duplicate

### What is the idea?

Support `--anaconda-upload` flag to upload built v1 packages, and respect `--no-anaconda-upload` flag.

### Why is this needed?

Users expect automatic upload after successful builds.

### What should happen?

1. Check if upload enabled after successful build (`parsed_args.anaconda_upload`)
2. Reuse conda-build's existing upload code
3. Pass through upload options (`--token`, `--user`, `--label`)
4. Handle errors gracefully (warn, don't fail)
5. Add tests and documentation

### Additional Context

- Default is to upload (controlled by `context.binstar_upload`)
- Use `--no-anaconda-upload` to disable

jsmolic:
I'll pick this issue up

PR #5924, opened 2026-02-17
Port v1 recipe support to py-rattler-build

### LINKED ISSUE #5891: Replace conda-build v1 recipe subprocess calls with Python API
### Checklist

### What is the idea?

Replace subprocess calls with py-rattler-build Python API for building and rendering v1 recipes.

### Why is this needed?

Better error handling, progress reporting, and access to build metadata.

### What should happen?

1. Add `py-rattler-build` to dependencies
2. Create private `conda_build/_rattler_integration.py` module
3. Refactor `run_rattler_build()` in `main_build.py` and `main_render.py` to use Python API
4. Pass through CLI arguments to py-rattler-build
5. Add tests and documentation

### Additional Context

Plan right now is to keep py-rattler-build APIs private (not exposed in conda_build.api)

### LINKED ISSUE #5892: Support multi-output v1 recipes
### Checklist

### What is the idea?

Handle v1 recipes that produce multiple packages.

### Why is this needed?

Many recipes produce multiple packages (library + CLI, Python bindings + C library).

### What should happen?

1. Handle `MultiOutputRecipe` and `outputs` list
2. Support staging outputs and inter-output dependencies
3. Add test recipes and documentation

### Additional Context

See pyrattler notebook `2-multi_output_and_staging.py` for examples.

### LINKED ISSUE #5893: Enhance v1 recipe error handling and progress feedback
### Checklist

### What is the idea?

Clear error messages and progress feedback for v1 recipe builds.

### Why is this needed?

Users need to understand failures and see progress.

### What should happen?

1. Catch rattler-build exceptions and re-raise as CondaError with context
2. Add context-specific error messages
3. Show progress (variant X/Y, package name/version)
4. Add tests for error cases

### Additional Context

- At minimum: re-raise all exceptions as CondaErrors
- Use `progress_callback` and `BuildResult.log`

### LINKED ISSUE #5895: Add test suite for v1 recipes
### Checklist

### What is the idea?

Create test recipes covering common v1 patterns.

### Why is this needed?

The current draft PR has only one test recipe, we should expand this to improve confidence that v1 recipes will successfully build.

### What should happen?

1. Create test recipes: simple, variants, multi-output, dependencies, noarch
2. Add integration tests for build, render, errors

### Additional Context

_No response_

### LINKED ISSUE #5897: Handle channel configuration for v1 recipes
### Checklist

### What is the idea?

Map conda-build's channel config to py-rattler-build.

### Why is this needed?

Recipes need packages from specific channels.

### What should happen?

1. Research channel API in py-rattler-build
2. Map `config.channel_urls` preserving priority
3. Handle `local` channel and aliases

### Additional Context

Depends on https://github.com/conda/rattler/issues/1327 for local and multi-channel support.

jsmolic:
Closes: https://github.com/conda/conda-build/issues/5891
Closes: https://github.com/conda/conda-build/issues/5892
Closes: https://github.com/conda/conda-build/issues/5893
Closes: https://github.com/conda/conda-build/issues/5897
Closes: https://github.com/conda/conda-build/issues/5895

### Description

This PR continues the work started in https://github.com/conda/conda-build/pull/5880 replacing the `rattler-build` CLI subprocess calls with the new py-rattler API.

It currently supports rendering and building v1 packages. Support for `conda-debug` was removed due to issues with the debug API; which has since been refactored in py-rattler and will be integrated in a subsequent PR to reduce the complexity in this one.

Py-rattler API documentation is available [here](https://rattler-build.prefix.dev/latest/py-rattler-build/reference/), but this summary describes how the API is used within this PR for easier review:

- The logic for finding rattler-build binary has been removed as we are no longer using the CLI tool.
- Core functionality of the code is in `run_rattler()` and `process_recipe()` functions.
- `run_rattler()` initializes configuration variables based on the values stored in `argparse.Namespace` and `Config` objects. It delegates recipe processing to a separate function and handles the build summary reporting at the end.
- Configuration file discovery logic has remained largely unchanged from the previous version
- Most of the rattler-build general settings are now handled through [`ToolConfiguration`](https://rattler-build.prefix.dev/latest/py-rattler-build/reference/configuration/#rattler_build.ToolConfiguration), while [`PlatformConfig`](https://rattler-build.prefix.dev/latest/py-rattler-build/reference/configuration/#platformconfig) and [`RenderConfig`](https://rattler-build.prefix.dev/latest/py-rattler-build/reference/rendering/#renderconfig) define platform-specific settings and rendering behavior, respectively.
- Configuration files are loaded using `VariantConfig.from_file()` method, respecting the order in which they are stacked.

`process_recipes()` performs loading, rendering, build and testing the recipe. Recipes are loaded with `Stage0Recipe.from_file()`, rendered with `recipe.render()` and build with `variant.run_build()`. If enabled, testing runs after the build to follow v0 recipe behavior in `conda-build`. 
 

#### Progress reporting

Py-rattler uses the [`ProgressCallback`](https://rattler-build.prefix.dev/latest/py-rattler-build/reference/progress/#progress) protocol for progress reporting, with some built-in implementations and options to define custom callbacks. This PR integrates `SimpleProgressCallback` and forwards the event levels and messages to conda’s logging system. This implementation is minimal but additional UI improvements are expected to be added in future versions.

#### Build results and summary

Build results are stored in dataclasses that contain recipe and output info about build status and error messages. Each `process_recipe()` call returns a `RecipeResult` object containing all relevant build summary details. Short summary is displayed upon completion, displaying per-recipe and per-output build results along with any errors if present.

#### Limitations/features currently not implemented:
- As previously mentioned, no`conda-debug` support, will be added in a separate PR
- No multi-output with staging outputs recipe support. It is currently labeled as an experimental feature in rattler-build and depends on the draft CEP: https://github.com/conda/ceps/pull/102 
- ~~No multichannel support, depends on https://github.com/conda/rattler/issues/1327~~ **EDIT**: multichannels are taken care of with 5d34a0c9d841a9fa965c3dbaa53660498c2ddb60
- Package upload and networking-specific configuration are currently not supported, subsequent work is planned in https://github.com/conda/conda-build/issues/5894 and https://github.com/conda/conda-build/issues/5898
- User documentation will be added in a separate PR (https://github.com/conda/conda-build/issues/5896)

### Checklist - did you ...

### DIFF
--- a/conda_build/_rattler_build/compat.py
+++ b/conda_build/_rattler_build/compat.py
@@ -1,53 +1,19 @@
 # Copyright (C) 2014 Anaconda, Inc
 # SPDX-License-Identifier: BSD-3-Clause
 import argparse
-import shlex
-import shutil
-import subprocess
-import sys
 from os.path import join
 from pathlib import Path
 
 from conda.base.context import context
+from rattler_build.progress import RichProgressCallback
+from rattler_build.render import RenderConfig
+from rattler_build.stage0 import Stage0Recipe
+from rattler_build.tool_config import PlatformConfig, ToolConfiguration
+from rattler_build.variant_config import VariantConfig
 
 from ..config import CondaPkgFormat, Config
-from ..utils import on_win
 
 CONFIG_FILES = {"conda_build_config.yaml", "variants.yaml"}
-
-
-def find_rattler_build() -> str:
-    """Find rattler-build executable."""
-    executable_dir = Path(sys.executable).parent
-    resolved_dir = Path(sys.executable).resolve().parent
-
-    if on_win:
-        # on Windows: rattler-build can be in $PREFIX/Library/bin or $PREFIX/Scripts
-        candidate_dirs = [
-            executable_dir / "Library" / "bin",
-            executable_dir / "Scripts",
-            resolved_dir / "Library" / "bin",
-            resolved_dir / "Scripts",
-            executable_dir,
-            resolved_dir,
-        ]
-        executable_name = "rattler-build.exe"
-    else:
-        # On Unix: both python and rattler-build are in $PREFIX/bin
-        candidate_dirs = [
-            executable_dir,
-            resolved_dir,
-        ]
-        executable_name = "rattler-build"
-
-    # Check candidate directories
-    for directory in candidate_dirs:
-        rattler_path = directory / executable_name
-        if rattler_path.exists():
-            return str(rattler_path)
-
-    # Fallback to PATH
-    return shutil.which("rattler-build") or "rattler-build"
 
 
 def check_arguments_rattler(
@@ -123,67 +89,50 @@ def check_arguments_rattler(
 
 def run_rattler(command: str, parsed_args: argparse.Namespace, config: Config) -> int:
     """Run rattler-build for v1 recipes"""
-    if command == "build":
-        cmd = ["rattler-build", "build"] + [
-            f"--recipe={join(recipe_dir, 'recipe.yaml')}"
-            for recipe_dir in parsed_args.recipe
-        ]
-    elif command == "render":
-        cmd = [
-            "rattler-build",
-            "build",
-            "--render-only",
-            f"--recipe={join(parsed_args.recipe, 'recipe.yaml')}",
-        ]
-    elif command == "debug":
-        cmd = [
-            "rattler-build",
-            "debug",
-            "--recipe",
-            parsed_args.recipe_or_package_file_path,
-        ]
-    else:
+    if command not in ("build", "render", "debug"):
         raise ValueError(f"Unrecognized subcommand: {command}")
 
-    # common configuration
+    # Initialize configuration defaults
+    test_strategy = None
+    skip_existing = None
+    noarch_build_platform = None
+    channel_priority = None
+    output_dir = config.croot
+    no_include_recipe = False
+    no_build_id = False
+    package_format = None
+    debug = False
+    channels = []
+    extra_context = {}
+    show_logs = not parsed_args.quiet
+    target_platform = config.variant.get("host_platform", config.subdir)
+    build_platform = config.variant.get("build_platform", config.subdir)
+    host_platform = config.variant.get("target_platform", config.subdir)
+    noarch_build_platform = config.variant.get("noarch_build_platform", config.subdir)
+
+    # TODO: investigate why is config.channel_urls
+    # does not pick up condarc settings, need to use context.channels
+    channels = list(context.channels)
     if config.channel_urls:
         for url in config.channel_urls:
+            # TODO: fix -c local
+            # TypeError: argument 'channels': 'Channel' object cannot be cast as 'str'
             if url in context.custom_multichannels:
-                cmd.extend(
-                    [
-                        f"--channel={local_url}"
-                        for local_url in context.custom_multichannels[url]
-                    ]
+                channels.extend(
+                    [local_url for local_url in context.custom_multichannels[url]]
                 )
             else:
-                cmd.append(f"--channel={url}")
-    cmd.extend(
-        ["--build-platform", config.variant.get("build_platform", config.subdir)]
-    )
-    cmd.extend(
-        [
-            "--host-platform",
-            config.variant.get(
-                "host_platform", config.variant.get("target_platform", config.subdir)
-            ),
-        ]
-    )
-    cmd.extend(
-        [
-            "--target-platform",
-            config.variant.get(
-                "target_platform", config.variant.get("host_platform", config.subdir)
-            ),
-        ]
-    )
+                channels.append(url)
+
     if context.channel_priority == "strict":
-        cmd.extend(["--channel-priority", "strict"])
+        channel_priority = "strict"
     else:
-        cmd.extend(["--channel-priority", "disabled"])
+        channel_priority = "disabled"
 
     if command in ("build", "render"):
-        # Ignore rattler's variant auto-discovery
-        cmd.append("--ignore-recipe-variants")
+        # TODO: --ignore-recipe-variants only available via deprecated
+        # rattler_build.cli_api:build_recipes()
+        # cmd.append("--ignore-recipe-variants")
 
         from ..variants import find_config_files
 
@@ -212,56 +161,113 @@ def run_rattler(command: str, parsed_args: argparse.Namespace, config: Config) -
                 recipe_config_filenames=CONFIG_FILES,
             )
 
-        cmd.extend([f"-m={variant}" for variant in config_files])
+        # if config.verbose:
+        # TODO: find this option in py-rattler-build
+        # cmd.append("--verbose")
 
-        if config.verbose:
-            cmd.append("--verbose")
-
-    if command == "debug":
-        if parsed_args.output_id:
-            cmd.extend(["--output-name", parsed_args.output_id])
+    # if command == "debug":
+    # TODO: output_name does not exist in python bindings
+    # if parsed_args.output_id:
+    # cmd.extend(["--output-name", parsed_args.output_id])
     elif command == "build":
-        cmd.extend(
-            [
-                "--noarch-build-platform",
-                config.variant.get("noarch_build_platform", config.subdir),
-            ]
-        )
         if parsed_args.extra_meta:
-            for k, v in parsed_args.extra_meta.items():
-                cmd.append(f"--extra-meta={k}={v}")
+            extra_context.update(parsed_args.extra_meta)
         if parsed_args.output_folder:
-            cmd.extend(["--output-dir", parsed_args.output_folder])
-        else:
-            cmd.extend(["--output-dir", config.croot])
-        if not parsed_args.set_build_id:
-            cmd.append("--no-build-id")
-        if parsed_args.debug:
-            cmd.append("--debug")
-        if parsed_args.notest:
-            cmd.extend(["--test", "skip"])
-        if parsed_args.quiet:
-            cmd.append("-q")
-        if parsed_args.skip_existing != "none":
-            if parsed_args.skip_existing == "local":
-                cmd.append("--skip-existing=local")
-            elif parsed_args.skip_existing == "all":
-                cmd.append("--skip-existing=all")
-        if not parsed_args.include_recipe:
-            cmd.append("--no-include-recipe")
+            output_dir = parsed_args.output_folder
+        no_build_id = not parsed_args.set_build_id
+        debug = parsed_args.debug
+        test_strategy = "skip" if parsed_args.notest else "native"
+        # if parsed_args.quiet:
+        # TODO: quiet does not exist in py-rattler-build
+        # cmd.append("-q")
+        skip_existing = parsed_args.skip_existing or "none"
+        no_include_recipe = not parsed_args.include_recipe
         if parsed_args.conda_pkg_format == CondaPkgFormat.V2:
-            cmd.extend(
-                ["--package-format", f".conda:{parsed_args.zstd_compression_level}"]
-            )
+            package_format = "conda"
         else:
-            cmd.extend(["--package-format", ".tar.bz2"])
+            package_format = ".tar.bz2"
 
-    try:
-        rattler_cmd = find_rattler_build()
-        cmd = [rattler_cmd, *cmd[1:]]
-        print("Running rattler:", shlex.join(cmd))
-        subprocess.run(cmd, check=True)
+    if command in ("build", "render"):
+        if command == "render":
+            recipe_path = join(parsed_args.recipe, "recipe.yaml")
+            recipes = [recipe_path]
+        else:
+            recipes = [
+                join(recipe_dir, "recipe.yaml") for recipe_dir in parsed_args.recipe
+            ]
+
+        from ..variants import find_config_files
+
+        results = []
+
+        # configure variant
+        for variant in config_files:
+            variant_config = VariantConfig.from_file(variant)
+
+        # coon tool / platform / render configuration
+        tool_config = ToolConfiguration(
+            test_strategy=test_strategy,
+            skip_existing=skip_existing,
+            noarch_build_platform=noarch_build_platform,
+            channel_priority=channel_priority,
+        )
+
+        platform_config = PlatformConfig(
+            target_platform=target_platform,
+            build_platform=build_platform,
+            host_platform=host_platform,
+        )
+
+        render_config = RenderConfig(
+            platform=platform_config,
+            extra_context=extra_context,
+        )
+
+        # iterate over all recipes
+        for recipe_path in recipes:
+            recipe = Stage0Recipe.from_file(Path(recipe_path))
+
+            variant_config = None
+            for variant_file in config_files:
+                variant_config = VariantConfig.from_file(variant_file)
+
+            # render the recipe
+            rendered = recipe.render(variant_config, render_config)
+
+            if command == "render":
+                import json
+
+                data = rendered[0].recipe.to_dict()
+                print(json.dumps(data, indent=2, sort_keys=True))
+                continue
+
+            # build all rendered variants
+            for i, variant in enumerate(rendered, 1):
+                print(
+                    f"\n🔨 Building variant {i}/{len(rendered)} "
+                    f"for recipe {recipe_path}"
+                )
+
+                with RichProgressCallback(show_logs=show_logs) as progress_callback:
+                    result = variant.run_build(
+                        tool_config=tool_config,
+                        output_dir=output_dir,
+                        channels=channels,
+                        progress_callback=progress_callback,
+                        recipe_path=recipe_path,
+                        no_build_id=no_build_id,
+                        package_format=package_format,
+                        no_include_recipe=no_include_recipe,
+                        debug=debug,
+                    )
+                    results.append(result)
+
+                    print("\n" + "=" * 60)
+                    print("Build Result:")
+                    print("=" * 60)
+                    print(f"   Package: {result.name} {result.version}")
+                    print(f"   Build string: {result.build_string}")
+                    print(f"   Platform: {result.platform}")
+                    print(f"   Build time: {result.build_time:.2f}s")
+
         return 0
-    except subprocess.CalledProcessError as e:
-        print(f"rattler failed: {e}", file=sys.stderr)
-        return e.returncode
--- a/conda_build/_rattler_build/compat.py
+++ b/conda_build/_rattler_build/compat.py
@@ -12,6 +12,7 @@ from rattler_build.tool_config import PlatformConfig, ToolConfiguration
 from rattler_build.variant_config import VariantConfig
 
 from ..config import CondaPkgFormat, Config
+from ..exceptions import CondaBuildUserError
 
 CONFIG_FILES = {"conda_build_config.yaml", "variants.yaml"}
 
@@ -87,28 +88,124 @@ def check_arguments_rattler(
         )
 
 
+def process_recipes(
+    recipes: list[str],
+    variant_config: VariantConfig,
+    render_config: RenderConfig,
+    tool_config: ToolConfiguration,
+    command: str,
+    output_dir: str,
+    channels: list[str],
+    show_logs: bool,
+    no_build_id: bool,
+    package_format: str,
+    no_include_recipe: bool,
+    debug: bool,
+) -> int:
+    import yaml
+
+    succeeded: list[str] = []
+    failed: dict[str, str] = {}
+
+    for recipe_path in recipes:
+        recipe_path_str = str(recipe_path)
+        try:
+            # load the recipe file
+            try:
+                recipe = Stage0Recipe.from_file(Path(recipe_path))
+            except Exception as e:
+                raise CondaBuildUserError(
+                    f"Failed to process recipe {recipe_path}: {str(e)}"
+                )
+
+            # render the recipe
+            try:
+                rendered = recipe.render(variant_config, render_config)
+            except Exception as e:
+                raise CondaBuildUserError(
+                    f"Failed to render recipe {recipe_path}: {str(e)}"
+                )
+
+            if command == "render":
+                data = rendered[0].recipe.to_dict()
+                print(yaml.safe_dump(data, indent=2, sort_keys=False))
+                succeeded.append(recipe_path_str)
+                continue
+
+            # build all rendered variants
+            for i, variant in enumerate(rendered, 1):
+                print(
+                    f"\n🔨 Building variant {i}/{len(rendered)} "
+                    f"for recipe {recipe_path}"
+                )
+
+                try:
+                    with RichProgressCallback(show_logs=show_logs) as progress_callback:
+                        variant.run_build(
+                            tool_config=tool_config,
+                            output_dir=output_dir,
+                            channels=channels,
+                            progress_callback=progress_callback,
+                            no_build_id=no_build_id,
+                            package_format=package_format,
+                            no_include_recipe=no_include_recipe,
+                            debug=debug,
+                        )
+                except Exception as e:
+                    print(f"Build failed for recipe {recipe_path}: {str(e)}")
+                    raise
+
+            # if all variants built without raising, mark recipe as succeeded
+            succeeded.append(recipe_path_str)
+
+        except Exception as e:
+            # catch any failure for this recipe and continue with the next
+            failed[recipe_path_str] = str(e)
+            print(f"Skipping remaining work for recipe {recipe_path}.")
+
+    # summary
+    print("\n=== Build summary ===")
+    if succeeded:
+        print("Succeeded:")
+        for path in succeeded:
+            print(f"  - {path}")
+    else:
+        print("Succeeded: none")
+
+    if failed:
+        print("\nFailed:")
+        for path, error in failed.items():
+            print(f"  - {path}: {error}")
+    else:
+        print("\nFailed: none")
+
+    return 1 if failed else 0
+
+
 def run_rattler(command: str, parsed_args: argparse.Namespace, config: Config) -> int:
     """Run rattler-build for v1 recipes"""
     if command not in ("build", "render", "debug"):
         raise ValueError(f"Unrecognized subcommand: {command}")
 
     # Initialize configuration defaults
-    test_strategy = None
-    skip_existing = None
-    noarch_build_platform = None
-    channel_priority = None
-    output_dir = config.croot
-    no_include_recipe = False
-    no_build_id = False
-    package_format = None
-    debug = False
-    channels = []
-    extra_context = {}
-    show_logs = not parsed_args.quiet
-    target_platform = config.variant.get("host_platform", config.subdir)
-    build_platform = config.variant.get("build_platform", config.subdir)
-    host_platform = config.variant.get("target_platform", config.subdir)
-    noarch_build_platform = config.variant.get("noarch_build_platform", config.subdir)
+    test_strategy: str | None = None
+    skip_existing: bool | None = None
+    noarch_build_platform: str | None = None
+    channel_priority: str | None = None
+    output_dir: str = config.croot
+    no_include_recipe: bool = False
+    no_build_id: bool = False
+    package_format: str | None = None
+    debug: bool = False
+    channels: list[str] = []
+    extra_context: dict[str] = {}
+    show_logs: bool = getattr(parsed_args, "quiet", False) is False
+    target_platform: str = config.variant.get("host_platform", config.subdir)
+    build_platform: str = config.variant.get("build_platform", config.subdir)
+    host_platform: str = config.variant.get("target_platform", config.subdir)
+    noarch_build_platform: str = config.variant.get(
+        "noarch_build_platform", config.subdir
+    )
 
     # TODO: investigate why is config.channel_urls
     # does not pick up condarc settings, need to use context.channels
@@ -169,7 +266,7 @@ def run_rattler(command: str, parsed_args: argparse.Namespace, config: Config) -
     # TODO: output_name does not exist in python bindings
     # if parsed_args.output_id:
     # cmd.extend(["--output-name", parsed_args.output_id])
-    elif command == "build":
+    if command == "build":
         if parsed_args.extra_meta:
             extra_context.update(parsed_args.extra_meta)
         if parsed_args.output_folder:
@@ -198,13 +295,11 @@ def run_rattler(command: str, parsed_args: argparse.Namespace, config: Config) -
 
         from ..variants import find_config_files
 
-        results = []
-
         # configure variant
         for variant in config_files:
             variant_config = VariantConfig.from_file(variant)
 
-        # coon tool / platform / render configuration
+        # common tool / platform / render configuration
         tool_config = ToolConfiguration(
             test_strategy=test_strategy,
             skip_existing=skip_existing,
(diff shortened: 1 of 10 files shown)


conda_build/_rattler_build/compat.py
```diff
@@ -212,56 +161,113 @@ def run_rattler(command: str, parsed_args: argparse.Namespace, config: Config) -
...
+                    )
+                    results.append(result)
+
+                    print("\n" + "=" * 60)
+                    print("Build Result:")
+                    print("=" * 60)
+                    print(f"   Package: {result.name} {result.version}")
+                    print(f"   Build string: {result.build_string}")
+                    print(f"   Platform: {result.platform}")
+                    print(f"   Build time: {result.build_time:.2f}s")
```
jaimergp: This should be its own separate helper function, I'd say.

conda_build/_rattler_build/compat.py
```diff
@@ -212,56 +161,113 @@ def run_rattler(command: str, parsed_args: argparse.Namespace, config: Config) -
...
+
+            # build all rendered variants
+            for i, variant in enumerate(rendered, 1):
+                print(
+                    f"\n🔨 Building variant {i}/{len(rendered)} "
+                    f"for recipe {recipe_path}"
+                )
+
+                with RichProgressCallback(show_logs=show_logs) as progress_callback:
+                    result = variant.run_build(
```
jaimergp: What happens if the build fails? I think we expect to continue with the next one til we exhaust all recipes (which may fail if they depend on each other). Then, at the end, provide an overview of what was built and what was not.

conda_build/_rattler_build/compat.py
```diff
@@ -212,56 +161,113 @@ def run_rattler(command: str, parsed_args: argparse.Namespace, config: Config) -
...
+
+        from ..variants import find_config_files
+
+        results = []
+
+        # configure variant
+        for variant in config_files:
+            variant_config = VariantConfig.from_file(variant)
+
+        # coon tool / platform / render configuration
```
jaimergp:
```suggestion
        # common tool / platform / render configuration
```

I think?

conda_build/_rattler_build/compat.py
```diff
@@ -212,56 +161,113 @@ def run_rattler(command: str, parsed_args: argparse.Namespace, config: Config) -
...
+        # cmd.append("--verbose")
 
-    if command == "debug":
-        if parsed_args.output_id:
-            cmd.extend(["--output-name", parsed_args.output_id])
+    # if command == "debug":
+    # TODO: output_name does not exist in python bindings
+    # if parsed_args.output_id:
+    # cmd.extend(["--output-name", parsed_args.output_id])
     elif command == "build":
```
jaimergp: You may need this to be `if` while the block above is commented out.

conda_build/_rattler_build/compat.py
```diff
@@ -212,56 +161,113 @@ def run_rattler(command: str, parsed_args: argparse.Namespace, config: Config) -
...
+                    print(f"   Build string: {result.build_string}")
+                    print(f"   Platform: {result.platform}")
+                    print(f"   Build time: {result.build_time:.2f}s")
 
-    try:
-        rattler_cmd = find_rattler_build()
-        cmd = [rattler_cmd, *cmd[1:]]
-        print("Running rattler:", shlex.join(cmd))
-        subprocess.run(cmd, check=True)
         return 0
```
jaimergp, line 484: Return 1 if there were errors, 0 otherwise.

conda_build/_rattler_build/compat.py
```diff
@@ -212,56 +161,113 @@ def run_rattler(command: str, parsed_args: argparse.Namespace, config: Config) -
...
+                variant_config = VariantConfig.from_file(variant_file)
+
+            # render the recipe
+            rendered = recipe.render(variant_config, render_config)
+
+            if command == "render":
+                import json
+
+                data = rendered[0].recipe.to_dict()
+                print(json.dumps(data, indent=2, sort_keys=True))
```
jaimergp: I'd say we expect this to be YAML, don't we?

jsmolic: rattler-build outputs in json by default, but it's not a problem to switch to yaml if we decide to

conda_build/_rattler_build/compat.py
jsmolic: Yeah, that makes sense. I split recipe processing in a separate function and added this

conda_build/_rattler_build/compat.py
```diff
@@ -121,69 +90,183 @@ def check_arguments_rattler(
...
+    build_platform: str,
+    host_platform: str,
+    experimental: bool,
+    extra_context: dict[str],
+    test_strategy: str,
+    skip_existing: bool,
+    noarch_build_platform: str,
+    channel_priority: str,
+) -> int:
+    import yaml
```
jaimergp: TIL that conda-build prefers PyYAML to `ruamel.yaml`. Anyway, we should import this at the top-level.

conda_build/_rattler_build/compat.py
```diff
@@ -121,69 +90,183 @@ def check_arguments_rattler(
...
+    target_platform: str,
+    build_platform: str,
+    host_platform: str,
+    experimental: bool,
+    extra_context: dict[str],
+    test_strategy: str,
+    skip_existing: bool,
+    noarch_build_platform: str,
+    channel_priority: str,
+) -> int:
```
jaimergp: It would be nice to have a short description of the py-rattler-build API here, like the different stages and what's supposed to happen. High level, not too many details. Also describe the different parameters. I'm, btw, surprised they are all required. Is that a deliberate choice?

conda_build/_rattler_build/compat.py
```diff
@@ -121,69 +90,183 @@ def check_arguments_rattler(
...
+            # load the recipe file
+            try:
+                recipe = Stage0Recipe.from_file(Path(recipe_path))
+            except RattlerBuildError as e:
+                raise CondaBuildUserError(
+                    f"Failed to process recipe {recipe_path}: {str(e)}"
+                )
+
+            if isinstance(recipe, MultiOutputRecipe):
+                for idx, output in enumerate(recipe.outputs, 1):
```
jaimergp: What's the enumeration for?

conda_build/_rattler_build/compat.py
```diff
@@ -121,69 +90,183 @@ def check_arguments_rattler(
...
+
+            render_config = RenderConfig(
+                platform=platform_config,
+                extra_context=extra_context,
+            )
+
+            # render the recipe
+            try:
+                rendered = recipe.render(variant_config, render_config)
+            except Exception as e:
```
jaimergp: Don't they have a more concrete exception here?

conda_build/_rattler_build/compat.py
```diff
@@ -121,69 +90,183 @@ def check_arguments_rattler(
...
+                            output_dir=output_dir,
+                            channels=channels,
+                            progress_callback=progress_callback,
+                            no_build_id=no_build_id,
+                            package_format=package_format,
+                            no_include_recipe=no_include_recipe,
+                            debug=debug,
+                        )
+                except RattlerBuildError as e:
+                    raise CondaError(f"Build failed for recipe {recipe_path}: {str(e)}")
```
jaimergp: I think we have usually continued on error for all recipes, so maybe we need to accumulate all errors first, and then launch an exception group at the end?

jaimergp (commented):
Some comments about docstrings and default values, and I'm also curious about failure modes.

tests/cli/test_main_build.py
```diff
@@ -550,9 +550,23 @@ def test_build_with_empty_channel_fails(empty_channel: Path) -> None:
         )
 
 
-def test_build_with_v1_recipe() -> None:
+def test_build_v1_recipe() -> None:
     """Test building a v1 recipe"""
     recipe = os.path.join(metadata_dir, "..", "variants", "32_v1_recipe")
 
-    args = [recipe]
+    args = [recipe, "-c", "conda-forge"]
```
jaimergp: You probably need to `--override-channels` too, because `defaults` is not really a channel "name", and `rattler` does not support multichannels yet.

jaimergp: Also, seeing this output:

```
  Channels: 
   - file:///tmp/pytest-of-runner/pytest-0/popen-gw0/test_build_v1_recipe_multi_out3/out/
   - conda-forge
   - defaults
   - conda-forge
```

You need to deduplicate the channel list.

conda_build/_rattler_build/compat.py
```diff
@@ -121,69 +90,183 @@ def check_arguments_rattler(
...
+
+            # build all rendered variants
+            for i, variant in enumerate(rendered, 1):
+                print(
+                    f"\n🔨 Building variant {i}/{len(rendered)} "
+                    f"for recipe {recipe_path}"
+                )
+
+                try:
+                    with RichProgressCallback(show_logs=show_logs) as progress_callback:
```
jaimergp: I don't know if we can have better control here, because current aspect of the logs is a bit noisy with many terminal codes messing the output:

Image

Also even when redirecting stdout and stderr to a file I get terminal output. What fd are they using for progress bars?

Image

conda_build/_rattler_build/compat.py
jsmolic: Honestly I took it from the py-rattler-build notebook :smile:  but it's enough to loop through recipe.outputs, I changed in the last commit

conda_build/_rattler_build/compat.py
jsmolic: yes there is `RattlerBuildError`; updated in b52c575ec03ef3b16085361fa2339f0aebf50ee6

conda_build/_rattler_build/compat.py
jsmolic: Done in b52c575ec03ef3b16085361fa2339f0aebf50ee6

tests/cli/test_main_build.py
jsmolic: You're right, though rattler doesn't have this option, but I added another condition to handle this and deduplicate channels to our channel cofiguration in b52c575ec03ef3b16085361fa2339f0aebf50ee6

conda_build/_rattler_build/compat.py
jsmolic: hmm it does not look the best indeed. I think those codes are coming from Rich. For reduced verbosity we can use `--quiet` but there is still some output. What level of logging do we want to be the default?

jsmolic: We can also use the simple progress reporting instead of Rich, @jaimergp see how it looks like for you with the last commit

conda_build/_rattler_build/compat.py
jsmolic: > I'm, btw, surprised they are all required. Is that a deliberate choice?

Could you elaborate this question, I don't quite understand. We do need them if we want to pass those settings to rattler-build

conda_build/_rattler_build/compat.py
```diff
@@ -121,69 +96,193 @@ def check_arguments_rattler(
...
+            try:
+                variant.run_build(
+                    tool_config=tool_config,
+                    output_dir=output_dir,
+                    channels=channels,
+                    progress_callback=SimpleProgressCallback(),
+                    no_build_id=no_build_id,
+                    package_format=package_format,
+                    no_include_recipe=no_include_recipe,
+                    debug=debug,
```
jsmolic: I'm thinking of dropping debug support for now as we can only pass this to enable debug mode, but it will throw an exception before creating the scripts:
```
    rust_result = build_rendered_variant_py(
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^
rattler_build.PyRattlerBuildError: PyRattlerBuildError: Script not executed because debug mode is enable
d
```

conda_build/_rattler_build/compat.py
ryanskeith: I like the idea here of an orchestration function though this function has 19 inputs.  Another way to reduce inputs is to follow the VariantConfig path, it seems like tool , platform and render are all configurations that could be set outside of the orchestration loop and reduce the inputs by 7.

If the code could break it up the different parts of the process, (render, build, test) so in the future we could leverage rattler-build to build and test the other recipe format, I think that would be ideal.

conda_build/_rattler_build/compat.py
```diff
@@ -121,69 +96,193 @@ def check_arguments_rattler(
...
+        try:
+            rendered = recipe.render(variant_config, render_config)
+        except RattlerBuildError as e:
+            err = CondaBuildUserError(
+                f"Failed to render recipe {recipe_path}: {str(e)}"
+            )
+            failed[recipe_path_str] = str(err)
+            continue
+
+        if command == "render":
```
ryanskeith: Here is another opportunity to reduce the input list if the rendering was its own separate function and handled outside this full process orchestration loop.

conda_build/_rattler_build/compat.py
ryanskeith: I think that would be a good choice to get the project off the ground.  Also, we need to double check if debug means the same thing between conda-build and rattler-build.

conda_build/_rattler_build/compat.py
```diff
@@ -212,56 +311,62 @@ def run_rattler(command: str, parsed_args: argparse.Namespace, config: Config) -
...
-        print(f"rattler failed: {e}", file=sys.stderr)
-        return e.returncode
+            recipes = [
+                join(recipe_dir, "recipe.yaml") for recipe_dir in parsed_args.recipe
+            ]
+
+        # configure variant
+        if config_files:
+            for variant in config_files:
+                variant_config = VariantConfig.from_file(variant)
```
ryanskeith: This is overwriting and only using the last config file from the list config_files.

ryanskeith (commented):
Overall, the direction to use the rattler_build library instead of calling the tool is a positive one.  I think there would be some value in splitting up the different build phases.

conda_build/_rattler_build/compat.py
jsmolic: In this case meaning is the same; i.e. debug generates and prints debug scripts that users can then run to debug the build, but it was a bit awkwardly implemented.

After discussing with the prefix folks, current `debug` support has been removed from the bindings in https://github.com/prefix-dev/rattler-build/commit/c607abf0f3f06a3bdee90418230679860bfe77ea and the work for properly exposing `rattler-build debug` in the bindings is currently in progress.

conda_build/_rattler_build/compat.py
jsmolic: Good catch, thanks for pointing that out! It turns out that python bindings currently only support loading a single config file. I've opened https://github.com/prefix-dev/rattler-build/issues/2336 for this, after it's implemented in py-rattler-build we can go back and fix the code here

conda_build/_rattler_build/compat.py
jsmolic: I agree with your point @ryanskeith, initially I kept the tool, platform and render config outside of this function, however for multi-output staging support we need to load the recipe and then inspect the outputs and set the `experimental` flag accordingly and pass it onto `PlatformConfig()` which is why I moved those configs inside of `process_recipes`. 

@jaimergp noted that we might need to provide this as an opt-in behavior given that it's marked as an experimental feature in `rattler-build`. This depends on the in-progress CEP https://github.com/conda/ceps/pull/102.

Another idea is to drop the staging outputs from the first version in conda-build altogether, indicate it in the documentation as a limitation of the current impl and re-add it later once it is supported properly in rattler-build.

conda_build/_rattler_build/compat.py
jaimergp: > I'm thinking of dropping debug support for now as we can only pass this to enable debug mode, but it will throw an exception before creating the scripts:

I mean, you can catch and ignore the exception only in debug mode (perhaps guarded by checking the exception message), instead of dropping support in the current version. Future iterations may not need it and we just drop the try/except block.

jsmolic: > I mean, you can catch and ignore the exception only in debug mode (perhaps guarded by checking the exception message)

That's what it was doing by default but given that the `debug` flag has been [removed](https://github.com/prefix-dev/rattler-build/commit/c607abf0f3f06a3bdee90418230679860bfe77ea) altogether now I'll drop the debug bits from the code and readd using the new API when it lands in rattler-build

tests/requirements.txt
```diff
@@ -1,7 +1,7 @@
 beautifulsoup4
 chardet
 conda >=25.11.0
-conda-forge::rattler-build
+conda-forge::py-rattler-build >=0.58.2 # v1 recipe support
```
cbouss: py-rattler-build 0.58.4 is available on the main channel.

conda_build/_rattler_build/compat.py
ryanskeith: That's good to know.  Any place is a good place to start.

conda_build/_rattler_build/compat.py
ryanskeith: Oh, that is interesting.  I think I need to look at this again as this idea just broke my mental model.  I agree though. We can only work with what we have noting it and moving on is the best we can do for now.

conda_build/_rattler_build/compat.py
jaimergp: I checked the logs now, I have a few thoughts I'll post to the main thread to keep this tidy.

conda_build/_rattler_build/compat.py
```diff
@@ -1,26 +1,45 @@
...
+class CondaProgressCallback(SimpleProgressCallback):
+    def __init__(self, show_logs: bool = True):
+        super().__init__()
+        self.show_logs = show_logs
+
+    def on_log(self, event: LogEvent) -> None:
+        if not self.show_logs:
+            return
+
+        level_name = {"error": "error", "warn": "warning"}.get(event.level, "info")
```
jaimergp, line 71: Are we sure there's no such thing as a critical level of sorts?

conda_build/_rattler_build/compat.py
```diff
@@ -1,26 +1,45 @@
...
+        getattr(log, level_name)(event.message)
+
```
jaimergp, line 73: You can also do `log.log(LEVEL_GOES_HERE, message)` if you prefer doing level mapping instead of method mapping.

conda_build/_rattler_build/compat.py
```diff
@@ -107,86 +124,70 @@ def process_recipes(
...
+        err = CondaBuildUserError(f"Failed to render recipe {recipe_path}: {str(e)}")
+        return False, str(err), None
+
+    if command == "render":
+        for item in rendered:
+            data = item.recipe.to_dict()
+            print(yaml.safe_dump(data, indent=2, sort_keys=False))
+        return True, None, None
+
+    for i, variant in enumerate(rendered, 1):
```
jaimergp, line 183: It would be nice to be able to pick which variant we want to build, I think there's an option for that in `conda build` or `conda debug`. Something something `--output-id` maybe?

conda_build/_rattler_build/compat.py
```diff
@@ -107,86 +124,70 @@ def process_recipes(
...
+        print(
```
jaimergp, line 184: Can you check if we are `print`ing always instead of using a logger for UI output? I'd like to know what the usual pattern is in conda-build and then respect that.

conda_build/_rattler_build/compat.py
```diff
@@ -107,86 +124,70 @@ def process_recipes(
...
-        try:
-            rendered = recipe.render(variant_config, render_config)
         except RattlerBuildError as e:
-            err = CondaBuildUserError(
-                f"Failed to render recipe {recipe_path}: {str(e)}"
+            err = CondaBuildUserError(f"Failed to build recipe {recipe_path}: {str(e)}")
+            return False, str(err), None
+
+        if not parsed_args.notest:
+            built_package_path = build_result.packages[0]
```
jaimergp: How many packages can there be in a single build_result? I'd expect just one for a single variant? 🤔

conda_build/_rattler_build/compat.py
```diff
@@ -301,7 +285,7 @@ def run_rattler(command: str, parsed_args: argparse.Namespace, config: Config) -
 
     # common tool / platform / render configuration
     tool_config = ToolConfiguration(
-        test_strategy=test_strategy,
+        test_strategy=None,
```
jaimergp, line 371: Let's add a comment here mentioning that we are handling this ourselves.

conda_build/_rattler_build/compat.py
```diff
@@ -328,23 +312,63 @@ def run_rattler(command: str, parsed_args: argparse.Namespace, config: Config) -
...
+            for p, e in failed.items():
+                print(f"  - {Path(p).resolve()}")
+
+            if failed_logs and not show_logs:
+                print("\n=== Test logs ===")
+                for p, log in failed_logs.items():
+                    print(f"\n--- {Path(p).resolve()} ---")
+                    print(log, end="" if log.endswith("\n") else "\n")
+
+            msg = "Recipe build failures:\n" + "\n".join(
```
jaimergp: Careful with that string addition. better to do something like:

```
msg = "\n".join(["Recipe build failures", *[f"  - {Path(p).absolute()}: {e}" for p, e in failed.items()]])

File: conda_build/_rattler_build/compat.py [hunk 25]
```diff
@@ -328,23 +312,63 @@ def run_rattler(command: str, parsed_args: argparse.Namespace, config: Config) -
...
+            )
+            if ok:
+                succeeded.append(recipe_path)
+            else:
+                failed[recipe_path] = err or f"Failed recipe {recipe_path}"
+                if test_log:
+                    failed_logs[recipe_path] = test_log
+
+        # summary
+        print("\n=== Build summary ===")
```
jaimergp: Commenting here what I proposed during our last chat, since I think we should be more careful about per-output (and not per-recipe) log results:

```bash
=== Build summary ===
Tried to build 3 recipe files, resulting in 23 outputs.
3 succeeded, 20 failed.

Details:
- ❌ /home/jakov/git-conda/pytest-feedstock/recipe/recipe.yaml
  - ✅ package-multi-output-base: Succeeded in 1m20s
  - ❌ package-multi-output-python: Reason for error (exit code 1)
  - ❌ package-multi-output-bindings: Reason for error (exit code -124667)
- ✅ /home/jakov/git-conda/napari-feedstock/recipe/recipe.yaml
  - ✅ package-single-output: Succeeded in 42m230s.
... etc # you get the point

There were failed outputs. Check logs above for more details.
```

conda_build/_rattler_build/compat.py
jsmolic: Yeah there is `--output-id` option in `conda-debug` if user wants to pick which output to setup debug scripts for. As far as picking what variant to build, I don't think `conda-build` or `rattler-build` have this option

conda_build/_rattler_build/compat.py
jsmolic: Hmm there are a bunch of `print`s sprinkled around the code and even things like `print(f"INFO ` and  `print(f"WARNING ` so I'm not sure if there is some kind of pattern, but I'm fine with whatever we agree on here

conda_build/_rattler_build/compat.py
jsmolic: Yes, but it will store it as a list of package paths so we have to do this to access it, e.g:

```
>>> build_result.packages
[PosixPath('/tmp/rattler_build_rd928nuw/output/noarch/toml-0.10.2-pyh4616a5c_0.conda')]"
```

jaimergp (commented):
I think we are almost here. Let's update the PR description with:

- a good summary of all changes
- all the decisions we made and why
- deliberate omissions that are not implemented yet

Don't forget to add the news either! Thanks!

conda_build/_rattler_build/compat.py
jsmolic: AFAIK no, there are only these 3 levels https://github.com/prefix-dev/rattler-build/blob/5c7f4a958ee2a88a23d49117434b3a0230ad45e5/py-rattler-build/src/rattler_build/progress.py#L171-L179

tests/requirements.txt
```diff
@@ -1,7 +1,7 @@
 beautifulsoup4
 chardet
 conda >=25.11.0
-conda-forge::py-rattler-build >=0.61.1 # v1 recipe support
+conda-forge::py-rattler-build >=0.61.4 # v1 recipe support
```
jsmolic: `py-rattler-build` is still on version 0.58.4 in https://repo.anaconda.com/pkgs/main/ so we have to use conda-forge channel for the time being

danyeaw: Hey @jsmolic, 0.61.4 is now on main 👍

danyeaw:
I tested this with the libadwaita feedstock which is a v1 recipe:

```
git clone https://github.com/conda-forge/libadwaita-feedstock
cd libadwaita-feedstock
conda-build recipe/ -m .ci_support/osx_arm64_python3.13.____cp313.yaml
```

Everything worked great.

conda_build/_rattler_build/compat.py
```diff
@@ -121,70 +137,184 @@ def check_arguments_rattler(
...
-        ]
-    elif command == "debug":
-        cmd = [
-            "rattler-build",
-            "debug",
-            "--recipe",
-            parsed_args.recipe_or_package_file_path,
-        ]
-    else:
+    if command not in ("build", "render"):
```
ryanskeith, line 279: Local testing for debug didn't land here.  Though I am not sure what debug in the rattler sense should do.  I was expecting an Unrecognized subcommand but got:
```Cannot process several recipe versions at the same time!```
If I am understanding the code correctly,  main_debug calls run_rattler and should raise this error.

File: conda_build/_rattler_build/compat.py [hunk 28]
```diff
@@ -121,70 +137,184 @@ def check_arguments_rattler(
...
+    show_logs: bool,
+    no_build_id: bool,
+    package_format: str | None,
+    no_include_recipe: bool,
+    tool_config: ToolConfiguration,
+    render_config: RenderConfig,
+    parsed_args: argparse.Namespace,
+) -> RecipeResult:
+    """
+    Function to parse, render and optionally build or test a conda packlage recipe using the py-rattler-build API.
```
ryanskeith:
```suggestion
    Function to parse, render and optionally build or test a conda package recipe using the py-rattler-build API.
```

conda_build/cli/main_debug.py
ryanskeith, line 1: Is the plan to make this not function for now?  It seems debug support was removed from both `run_rattler` and `check_arguments_rattler`.

ryanskeith (commented):
I like that this is using the api approach. I think there is some work to do though when and how should be discussed.  This could be a first step for beta.

First, it seems that channels are just passed through and we may need to process them more for ratter.  In my .condarc, I have made defaults my channel (this could be also another symptom if rattler_build is using the rattler_solver) and rattler_build doesn't understand this.

```
conda-build 32_v1_recipe

Building variant 1/1 for recipe /Users/rkeith/code/conda-build/tests/test-recipes/variants/32_v1_recipe/recipe.yaml
Starting build of 1 outputs
Fetching source from url: https://pypi.org/packages/source/p/pytest/pytest-9.0.2.tar.gz
Downloading from: https://pypi.org/packages/source/p/pytest/pytest-9.0.2.tar.gz
Copying source from cache: /Users/rkeith/miniconda3/envs/conda-build-dev-rattler/conda-bld/src_cache/1204a59d4e7c526b_extracted to /Users/rkeith/miniconda3/envs/conda-build-dev-rattler/conda-bld/bld/rattler-build_pytest_1776384646/work
Could not find noarch/repodata.json. Creating new one.
Adding 0 packages to subdir noarch.
Successfully added 0 packages to subdir noarch.
Writing repodata to noarch/repodata.json
  Platform: osx-arm64 [__unix=0=0, __osx=15.7.4=0, __archspec=1=m1]
  Channels:
   - file:///Users/rkeith/miniconda3/envs/conda-build-dev-rattler/conda-bld/
   - defaults
  Specs:
   - python >=3.10
   - setuptools >=77
   - setuptools-scm >=6.2.3
   - pip
subdir osx-arm64 of channel file:///Users/rkeith/miniconda3/envs/conda-build-dev-rattler/conda-bld/ was not found, ignoring
sharded repodata seems to be missing for https://conda.anaconda.org/defaults/osx-arm64/, falling back to repodata.json files
sharded repodata seems to be missing for https://conda.anaconda.org/defaults/noarch/, falling back to repodata.json files

   Downloading repodata [00:00:00] [━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━]      1 B @ 2 B/s                                                                                                                         Downloading repodata [00:00:00] [━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━]      1 B @ 2 B/s                                                                                                                         Downloading repodata [00:00:00] [━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━]      1 B @ 7 B/s                                                                                                                         Downloading repodata [00:00:00] [━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━]      1 B @ 8 B/s
=== Build summary ===
Tried to build 1 recipe file, resulting in 1 output.
0 succeeded, 1 failed.

Details:
- ❌ /Users/rkeith/code/conda-build/tests/test-recipes/variants/32_v1_recipe/recipe.yaml
  - ❌ pytest: PyRattlerBuildError: Failed to resolve dependencies: could not find subdir 'noarch' in channel 'https://conda.anaconda.org/defaults/'

```
This didn't work for `main` channel either but this could be that the recipe itself is conda-forge reliant.  We should look into this more.  Tests should generally be main and conda-forge compatible.

Also, I think this should start communicating with the conda-build user that it did detect a v1 recipe so they can better understand which code path it may or may not be taking.

What is our MVP for this round for conda-build?  I am ok with removing debug from the list but we should remove it completely and not by halves.

conda_build/_rattler_build/compat.py
jsmolic: Re debug, this was discussed in the previous thread, see https://github.com/conda/conda-build/pull/5924#discussion_r2918908228 and following comments.

We made the decision at the time to drop  `conda-debug` support from this PR to keep it concise given there were some issues with the API. Since then, https://github.com/conda/conda-build/pull/5950 has been opened to add `conda-debug` support separately and it's expected to land soon following this PR.

I did indeed forgot to remove leftover code in `main_debug`, thanks for pointing that out. It has been addressed in d00659695137ee4464d9e30a8c0b10476f108a1a

conda_build/_rattler_build/compat.py
jsmolic: Thanks, fixed

conda_build/cli/main_debug.py
jsmolic: It was leftover code which I forgot to cleanup. Will be re-added subsequently as part of https://github.com/conda/conda-build/pull/5950

jsmolic:
@ryanskeith Regarding your build failure, the test recipe itself works both with `main` and `conda-forge`, but I think the issue you had was that default channels were not passed properly to py-rattler, causing the build to fail. This is indeed a bug in the code, I've addressed this in 699e46c4ec10167954d6bb0aee1fbb2e8253c01a to include default channels if set in condarc.

jsmolic:
Looking more at multichannels, it seems we can just iterate over them and pass the subchannel URLs. Added that now so https://github.com/conda/conda-build/issues/5897 should be taken care of (more testing welcome as always).

For following `.condarc`:
```
channels:
  - main
  - multi_1
custom_multichannels:
   multi_1:
     - conda-forge
     - bioconda
```
We now get `['main', 'https://conda.anaconda.org/conda-forge', 'https://conda.anaconda.org/bioconda']` as channels and `['file:///home/jakov/miniforge3/conda-bld', 'main', 'https://conda.anaconda.org/conda-forge', 'https://conda.anaconda.org/bioconda']` if we pass `-c local`. Also added a test for it and tried out multiple combinations locally but I think that should be good now.

jaimergp:
> Looking more at multichannels, it seems we can just iterate over them and pass the subchannel URLs. 

Yes, it works in most cases. The only difference is channel priority, which is dealt with differently (channels within a multichannel have the same channel priority, while "exploded" multichannels would result in each item having different priority). This can be amended in the solver side of things but it is not super important for now. Let's add a `FIXME` comment about this caveat so we don't forget about it.

jaimergp (approved):

jsmolic:
Thanks @jaimergp! :rocket:

PR #5962, opened 2026-04-27
Numpy warning removal

### LINKED ISSUE #3170: Irrelevant warning about NumPy
### Actual Behavior

A build that does not involve NumPy emits info/warning messages about NumPy:
```
No numpy version specified in conda_build_config.yaml.  Falling back to default numpy value of 1.11
WARNING:conda_build.metadata:No numpy version specified in conda_build_config.yaml.  Falling back to default numpy value of 1.11
```
### Expected Behavior

No irrelevant NumPy messages. Well, I _think_ they're irrelevant -- I can't think how this information would be useful in this case.

### Steps to Reproduce

In a directory with this `meta.yaml`:
```
package:
  name: aws
  version: 0

requirements:
  run:
    - awscli
```
do `conda build -c conda-forge .`.

##### `Output of conda info`
```
     active environment : None
            shell level : 0
       user config file : /home/pmadden/.condarc
 populated config files : 
          conda version : 4.5.11
    conda-build version : 3.15.1
         python version : 3.7.0.final.0
       base environment : /home/pmadden/miniconda  (writable)
                          https://repo.anaconda.com/pkgs/free/linux-64
                          https://repo.anaconda.com/pkgs/free/noarch
                          https://repo.anaconda.com/pkgs/pro/linux-64
                          https://repo.anaconda.com/pkgs/pro/noarch
          package cache : /home/pmadden/miniconda/pkgs
                          /home/pmadden/.conda/pkgs
       envs directories : /home/pmadden/miniconda/envs
                          /home/pmadden/.conda/envs
               platform : linux-64
             user-agent : conda/4.5.11 requests/2.19.1 CPython/3.7.0 Linux/4.15.0-34-generic ubuntu/18.04 glibc/2.27
                UID:GID : 1000:1000
             netrc file : None
           offline mode : False
```

ryanskeith:
### Description

Fixes #3170.

#### What went wrong
parse_config_file evaluated YAML line selectors with the same selector setup as the full recipe.  As the merged variant is being read at that moment the numpy key isn't available yet (unless --numpy was used in the cli call), so conda-build could emit the verbose “No numpy version specified in conda_build_config.yaml” warning at the very beginning every time.

#### What was changed

- Introduce build_selector_map with two entry points: 
   - cbc_line_selectors (used when reading conda_build_config.yaml)
   - recipe_selectors (used for Jinja + parse + get_recipe_text + FilteredLoader).
- get_selectors is a thin wrapper around recipe_selectors for backward compatibility
- The CBC path does not run the “numpy missing from variant” warning; the recipe path keeps the existing verbose warning when numpy is not in the variant.
- environ wiring: at the end of build_selector_map, apply d.update(environ) then set d["environ"] = environ so a rare env var named environ cannot overwrite the full mapping, while keeping prior behavior for normal keys.
- removal of ns_cfg function which has had a warning of PendingDeprecation for a few years now.

#### User-visible impact
Removes numpy warnings when processing conda_build_config.yaml chain (which is the reason most people would see that in the first line); no intentional change to recipe rendering semantics beyond that.

#### How tested
Selector-related tests in tests/test_metadata.py (and a full local run of the test suite) pass.

### Checklist - did you ...

### DIFF
--- a/conda_build/api.py
+++ b/conda_build/api.py
@@ -52,7 +52,10 @@ def render(
     """Given path to a recipe, return the MetaData object(s) representing that recipe, with jinja2
        templates evaluated.
 
-    Returns a list of (metadata, need_download, need_reparse in env) tuples"""
+    Returns a list of (metadata, need_download, need_reparse in env) tuples.
+
+    See :mod:`conda_build.render`.
+    """
     from .render import render_metadata_tuples, render_recipe
 
     config = get_or_merge_config(config, **kwargs)
--- a/conda_build/build.py
+++ b/conda_build/build.py
@@ -93,9 +93,38 @@ from .utils import (
 )
 from .variants import (
     dict_of_lists_to_list_of_dicts,
+    get_default_variant,
     get_package_variants,
     set_language_env_vars,
 )
+
+
+def _warn_implicit_numpy_variant(m: MetaData) -> None:
+    """Log when build/host need numpy but the matrix numpy pin still matches conda-build's default.
+
+    If ``variant['numpy']`` differs from :func:`~conda_build.variants.get_default_variant`,
+    assume an intentional pin and stay quiet.
+
+    Matching the default can still mean it was set but this should be rare.
+    """
+
+    default_np = get_default_variant(m.config)["numpy"]
+    pinned = m.config.variant.get("numpy")
+    if pinned is not None and pinned != default_np:
+        return
+    if not any(
+        ms.name == "numpy"
+        for section in ("build", "host")
+        for ms in m.ms_depends(section)
+    ):
+        return
+    utils.get_logger(__name__).warning(
+        "numpy is required by this recipe under requirements build/host but the numpy "
+        "variant is still conda-build's default (%s). Prefer setting numpy in "
+        "conda_build_config.yaml (or equivalent) when you depend on numpy at build time.",
+        default_np,
+    )
+
 
 if on_win:
     from . import windows
@@ -2453,6 +2482,8 @@ def build(
                         r"|".join(rf"(?:^{exc}(?:\s|$|\Z))" for exc in excludes)
                     )
             add_upstream_pins(m, False, exclude_pattern, [])
+
+        _warn_implicit_numpy_variant(top_level_pkg)
 
         create_build_envs(top_level_pkg, notest)
 
--- a/conda_build/metadata.py
+++ b/conda_build/metadata.py
@@ -9,11 +9,10 @@ import os
 import re
 import sys
 import time
-import warnings
 from collections import OrderedDict
 from functools import cache, lru_cache
 from os.path import isdir, isfile, join
-from typing import TYPE_CHECKING, NamedTuple, overload
+from typing import TYPE_CHECKING, Any, NamedTuple, overload
 
 import jinja2
 import yaml
@@ -56,6 +55,7 @@ from .variants import (
 )
 
 if TYPE_CHECKING:
+    from collections.abc import Iterable, Mapping
     from typing import Any, Literal, Self
 
     OutputDict = dict[str, Any]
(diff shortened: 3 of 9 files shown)


travishathaway:
@ryanskeith,

I was testing this pull request manually with the original steps to reproduce from the issue:

1. Create a meta.yaml like this in a `test-recipe/` folder:
```yaml
package:
  name: aws
  version: 0

requirements:
  run:
    - awscli
```
2. Run `conda build` from the development environment:
```
python -m conda build test-recipe
```

But, when I test this, I'm still seeing that annoying `numpy` warning. Maybe it would make sense to add the above as an integration test and then use that as the basis for solving this problem?

Warning message:
```
python -m conda build test-recipe
WARNING: No numpy version specified in conda_build_config.yaml.  Falling back to default numpy value of 1.26
WARNING: Number of parsed outputs does not match detected raw metadata blocks. Identified output block may be wrong! If you are using Jinja conditionals to include or exclude outputs, consider using `skip: true  # [condition]` instead.
```

I think those refactors are good to have, but we also need to make sure we're fulfilling the original wish and spirit of the person who made that request.

ryanskeith:
@travishathaway, we could remove the warning completely but it is a useful warning to have.  The PR does get rid of the first error that was immediately triggered:
Before:
```
> conda build ca-certificates-feedstock
WARNING: No numpy version specified in conda_build_config.yaml.  Falling back to default numpy value of 1.26
Adding in variants from /Users/rkeith/code/aggregate/conda_build_config.yaml
WARNING: Number of parsed outputs does not match detected raw metadata blocks. Identified output block may be wrong! If you are using Jinja conditionals to include or exclude outputs, consider using `skip: true  # [condition]` instead.
```

The first warning was running against the CBC yaml files.  With my background from the packaging team, I was overjoyed to see that gone and declared victory a little too soon it seems.  I still think the error is valid but it should not trigger when it is not suppose to.  I will look and see why the code may still by triggering the warning.

ryanskeith:
Update:

So it turns out that this bug was actually in three places.  The first was fixed by creating a specialized function to just run on CBC.yaml (config) files.  The next one that @travishathaway found is due to the fact that the code was running the first recipe parsing that the variant information from the CBC weren't loaded into the Config object.  Thus this change was made:

```
+    # recipe_selectors / get_selectors read config.variant (e.g. numpy from merged CBC)
+    metadata.config.variant = variants[0]
+
     recipe_requirements = metadata.extract_requirements_text()
     recipe_package_and_build_text = metadata.extract_package_and_build_text()
     recipe_text = recipe_package_and_build_text + recipe_requirements
     if hasattr(recipe_text, "decode"):
         recipe_text = recipe_text.decode()

-    metadata.config.variant = variants[0]
     used_variables = metadata.get_used_loop_vars(force_global=False)
     top_loop = metadata.get_reduced_variant_set(used_variables)
```

This fixes the error that @travishathaway is seeing BUT there is another one that occurs while testing.  Also, this doesn't have any effect as far as I have seen in testing the render or metadata tests so I think this is a sound change.
  
```
TEST START: /Users/rkeith/miniconda3/envs/conda-build-dev/conda-bld/osx-arm64/ca-certificates-2025.12.2-hca03da5_0.conda
Adding in variants from /var/folders/bn/54095dcn7kqgt35n7mgtyq1m0000gp/T/tmp3kehwlsb/info/recipe/conda_build_config.yaml
Adding in variants from config.variant
WARNING: No numpy version specified in conda_build_config.yaml.  Falling back to default numpy value of 1.26
```

In testing, the package is tested in a stand alone matter. The build recipe is parsed.  The problem is that because `numpy` isn't used in the recipe that the key from the stored in the config.variant.   So the warning triggers at this point.

So because of the deduplication filters of warnings, the warning was only showing up once at the beginning though in reality it was actually being triggered thrice.

Way forward:

Make the warning aware if the recipe actually requires numpy before issuing the warning.

travishathaway:
@ryanskeith,

I like the way forward, and I think it respects the wish of the initial bug report!

travishathaway (approved):
Thanks for adding the new tests. This change looks good to me.

PR #5975, opened 2026-05-10
ci: Update self install for canary builds

jsmolic:
`conda-pypi` has recently been added as a dependency in [conda](https://github.com/conda/conda/commit/bae25d4c3d6be09130e24627690810853537fc73) which breaks the current self install via `pip` (externally managed env error). Add the `--break-system-packages` flag to pip install to mitigate this.

### Checklist - did you ...

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -177,7 +177,7 @@ jobs:
 
       # TODO: how can we remove this step?
       - name: Install Self
-        run: pip install -e .
+        run: pip install -e . --no-deps --no-build-isolation --break-system-packages
 
       - name: Conda Info
         # view test env info (not base)
@@ -359,7 +359,7 @@ jobs:
 
       # TODO: how can we remove this step?
       - name: Install Self
-        run: pip install -e . --no-deps --no-build-isolation
+        run: pip install -e . --no-deps --no-build-isolation --break-system-packages
 
       - name: Conda Info
         # view test env info (not base)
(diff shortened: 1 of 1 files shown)


jaimergp:
> conda-pypi has recently been added as a dependency in [conda](https://github.com/conda/conda/commit/bae25d4c3d6be09130e24627690810853537fc73) which breaks the current self install via pip.

What type of breakage? The `EXTERNALLY-MANAGED` stuff? I'd rather export the bypass env var in that step instead of adding code we need to maintain later. I guess it's `PIP_BREAK_SYSTEM_PACKAGES=1` or something like that.

jsmolic:
> > conda-pypi has recently been added as a dependency in [conda](https://github.com/conda/conda/commit/bae25d4c3d6be09130e24627690810853537fc73) which breaks the current self install via pip.
> 
> What type of breakage? The `EXTERNALLY-MANAGED` stuff? I'd rather export the bypass env var in that step instead of adding code we need to maintain later. I guess it's `PIP_BREAK_SYSTEM_PACKAGES=1` or something like that.

Yeah, makes sense. In that case it's simpliest to just pass `--break-system-packages` to pip, I changed the PR to do so

dholth:
Should conda-build `conda pypi install --editable <self>`

jsmolic:
> Should conda-build `conda pypi install --editable <self>`

That's what I initially did in 2aecc2417b1ed96e5e5d1bbf8e109a293e9881fa, but @jaimergp suggested the other approach.

jaimergp (approved):

PR #5986, opened 2026-05-21
Remove dependency on rattler-build

jsmolic:
We are not using rattler-build CLI anymore in the code so we can drop the unneded dependency.

### Checklist - did you ...

### DIFF
--- a/news/5986-remove-rattler-build-dependency.md
+++ b/news/5986-remove-rattler-build-dependency.md
@@ -1,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Remove redundant dependency on rattler-build. (#5986)
+
+### Deprecations
+
+* <news item>
+
+### Docs
+
+* <news item>
+
+### Other
+
+* <news item>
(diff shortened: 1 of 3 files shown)


ryanskeith (approved):
Looks good to me.

jsmolic:
Thanks @ryanskeith!

jsmolic:
@beckermr Hi, I got an e-mail notification about your comment here but github is acting up again so it's not visible, I'll quote it here and respond:

> This PR makes no sense to me. The _rattler_build module is still present on main (https://github.com/conda/conda-build/blob/main/conda_build/_rattler_build/compat.py) and not having causes an import error for 26.5.

We are already declaring dependency on `py-rattler-build` which is what the module is using: https://github.com/conda/conda-build/blob/637792380a6cd4bf47dec6e8289b8daf3094d8cb/recipe/meta.yaml#L53 

Previous version used `rattler-build` binary as a subprocess but after the switch to `py-rattler-build` we don't need it anymore. We also have test coverage for this, so any issues would have been caught in CI.

beckermr:
Yes! I figured that out after I posted the comment and then deleted it. Sorry for the noise! Thank you!

jsmolic:
No problem @beckermr!

PR #5990, opened 2026-05-27
Fix SP_DIR resolving to wrong path for free-threading Python builds

### LINKED ISSUE #5563: SP_DIR vs. freethreading builds: wrong path?
If I install
```
conda install -c conda-forge python-freethreading numpy
```
what gets populated is
```
$PREFIX/lib/python3.13t/site-packages/{numpy,...}  # including all the built-in imports
$PREFIX/lib/python3.13/site-packages/pip           # and presumably other noarch packages
```

However, when trying to build something on top (e.g. scipy), `SP_DIR` for the free-threading builds doesn't match the main site-package directory (`python3.13` instead of `python3.13t`) and so the build fails[^1]

[^1]: I'm not talking about the `pip install`, which works fine (and installs into `python3.13t`), but we need to manually [copy something](https://github.com/conda-forge/scipy-feedstock/blob/9fe7f7450dc83b1b366dfafc3a335238103d1951/recipe/build-output.sh#L9) to `$SP_DIR/scipy`, due to the way the tests get split off into their own package.

Not sure what's the best approach here, but my understanding would be that this is simply a bug that `SP_DIR` needs to reflect the `t` for freethreading builds.

CC @isuruf @jjhelmus

jschueller:
- Prefer exact python{py_ver} directory match in get_stdlib_dir() before falling back to the glob pattern, so that python3.13t/ is correctly selected when py_ver is "3.13t"
- Honor is_freethreading variant key in get_py_ver() by appending the "t" suffix to the Python version string when the key is set

Closes #5563

### DIFF
--- a/conda_build/environ.py
+++ b/conda_build/environ.py
@@ -96,7 +96,10 @@ def get_py_ver(config):
     py = config.variant.get("python", get_default_variant(config)["python"])
     if not hasattr(py, "split"):
         py = py[0]
-    return ".".join(py.split(".")[:2])
+    ver = ".".join(py.split(".")[:2])
+    if config.variant.get("is_freethreading", False) and not ver.endswith("t"):
+        ver += "t"
+    return ver
 
 
 def get_r_ver(config):
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -15,6 +15,27 @@ from conda_build.exceptions import BuildLockError
 
 
 @pytest.mark.skipif(
+    utils.on_win, reason="unix-specific path logic"
+)
+def test_get_stdlib_dir_exact_match(testing_workdir):
+    lib_dir = os.path.join(testing_workdir, "lib")
+    for name in ("python3.13t", "python3.13"):
+        (Path(lib_dir) / name).mkdir(parents=True)
+    result = utils.get_stdlib_dir(testing_workdir, "3.13t")
+    assert result == os.path.join(lib_dir, "python3.13t")
+
+
+@pytest.mark.skipif(
+    utils.on_win, reason="unix-specific path logic"
+)
+def test_get_stdlib_dir_fallback(testing_workdir):
+    lib_dir = os.path.join(testing_workdir, "lib")
+    os.makedirs(lib_dir)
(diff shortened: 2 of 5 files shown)


jsmolic (commented):
Thanks for the PR, is there any particular reason why this is marked as draft?

We'd need a test to verify the correct behavior before merging.

Are you willing to finish the work in this PR @jschueller?

tests/test_utils.py
```diff
@@ -14,6 +14,26 @@
 from conda_build.exceptions import BuildLockError
 
 
+@pytest.mark.skipif(
+    utils.on_win, reason="unix-specific path logic"
+)
+def test_get_stdlib_dir_exact_match(testing_workdir):
+    lib_dir = os.path.join(testing_workdir, "lib")
+    os.makedirs(os.path.join(lib_dir, "python3.13t"))
```
jsmolic: I think the test should check that `3.13t` gets picked if both directories are present:
```suggestion
    for name in ("python3.13t", "python3.13"):
        (Path(lib_dir) / name).mkdir(parents=True)
```

jsmolic (commented):
We need a news item in a similar format as https://github.com/conda/conda-build/blob/e4cfe88586634f6adba01adb768668ef563f6f29/news/5987-fix-duplicate-rpaths-macos.md?plain=1, could you please also add this?

jsmolic:
pre-commit.ci autofix

conda_build/environ.py
```diff
@@ -96,7 +96,10 @@ def get_py_ver(config):
     py = config.variant.get("python", get_default_variant(config)["python"])
     if not hasattr(py, "split"):
         py = py[0]
-    return ".".join(py.split(".")[:2])
+    ver = ".".join(py.split(".")[:2])
+    if config.variant.get("is_freethreading", False) and not ver.endswith("t"):
+        ver += "t"
```
isuruf, line 101: I don't think this is a good change. freethreading doesn't have anything to do with the version. Similar to other options like unicode support (back in 2.x days) and other implementations like pypy, we don't include the ABI in this field

jsmolic: @isuruf Do you have any pointers in where should this be fixed in the code?

PR #5992, opened 2026-05-27
fix(utils): length-prefix paths in ``compute_content_hash``; deprecate bare ``content_sha*`` recipe keys

pb01ka:
### Description

Fix a hash-collision bug in `compute_content_hash` reported in [conda/ceps#150](https://github.com/conda/ceps/issues/150) and implement the v2 algorithm specified in the accompanying CEP (supersedes CEP 19).

#### Bug fix - new v2 algorithm

**Root cause.** The hash stream for each directory entry was built by concatenating raw bytes in the order `<path><type><content><separator>`, with no field-length information. Because filenames can contain the same bytes used as type markers (`F`, `D`, `L`) and the entry separator (`-`), two structurally different trees could produce an identical byte stream.

For example:

- Tree 1: a single file named `testFhello-world` with content `www`
- Tree 2: a file `test` (content `hello`) plus a file `world` (content `www`)

Both produced stream `testFhello-worldFwww-`, yielding the same SHA-256 digest.

**Fix.** Each variable-length field is now prefixed with its decimal byte length followed by `:` before being fed to the hasher:

| Field          | Before (CEP-19)  | After (v2)                           |
| -------------- | ---------------- | ------------------------------------ |
| Path           | `<path_bytes>`   | `<len(path_bytes)>:<path_bytes>`     |
| Symlink target | `<target_bytes>` | `<len(target_bytes)>:<target_bytes>` |

This makes field boundaries unambiguous and eliminates the collision.

#### Backwards compatibility and key naming

The algorithm change produces different digests for the same directory contents, so existing stored hashes are not compatible with the new algorithm. Two mechanisms are provided:

1. **Distinct key families.** The new fixed algorithm is exposed under the `content_sha256_v2` / `content_sha384_v2` / `content_sha512_v2` recipe keys. New recipes should use these. The original CEP-19 algorithm is retained under the bare `content_sha256` / `content_sha384` / `content_sha512` keys for backwards compatibility, but **those keys are now deprecated** - using them emits a `DeprecationWarning` at build time directing recipe authors to migrate to the `_v2` keys.

2. **`legacy` parameter.** `compute_content_hash()` accepts a new `legacy=True` keyword argument that reproduces the original CEP-19 byte stream exactly, for any code that computes or verifies hashes programmatically against pre-existing stored values. Passing `legacy=True` also emits a `DeprecationWarning`.

#### Migration guide

| Old key (deprecated)  | New key             |
| --------------------- | ------------------- |
| `content_sha256`      | `content_sha256_v2` |
| `content_sha384`      | `content_sha384_v2` |
| `content_sha512`      | `content_sha512_v2` |

Re-compute your hashes with `compute_content_hash(directory, legacy=False)` (the default) and update the recipe keys accordingly.

### Checklist - did you ...

- ~Add / update outdated documentation?~ (no user-facing docs affected)

### DIFF
--- a/conda_build/metadata.py
+++ b/conda_build/metadata.py
@@ -620,6 +620,11 @@ FIELDS = {
         "content_sha256": None,
         "content_sha384": None,
         "content_sha512": None,
+        # Legacy v1 keys (CEP-19 algorithm, no length-prefixing). Use these when you have
+        # existing stored hashes computed with the original CEP-19 algorithm.
+        "content_sha256_v1": None,
+        "content_sha384_v1": None,
+        "content_sha512_v1": None,
         "content_hash_skip": list,
         "path": str,
         "path_via_symlink": None,
--- a/conda_build/metadata.py
+++ b/conda_build/metadata.py
@@ -617,14 +617,14 @@ FIELDS = {
         "sha256": None,
         "sha384": None,
         "sha512": None,
+        # Deprecated: use the original CEP-19 algorithm (no length-prefixing). Migrate to
+        # content_sha*_v2 keys which use the fixed algorithm.
         "content_sha256": None,
         "content_sha384": None,
         "content_sha512": None,
-        # Legacy v1 keys (CEP-19 algorithm, no length-prefixing). Use these when you have
-        # existing stored hashes computed with the original CEP-19 algorithm.
-        "content_sha256_v1": None,
-        "content_sha384_v1": None,
-        "content_sha512_v1": None,
+        "content_sha256_v2": None,
+        "content_sha384_v2": None,
+        "content_sha512_v2": None,
         "content_hash_skip": list,
         "path": str,
         "path_via_symlink": None,
--- a/conda_build/source.py
+++ b/conda_build/source.py
@@ -1120,6 +1120,7 @@ def provide(metadata):
                     os.makedirs(src_dir)
 
             skip = ensure_list(source_dict.get("content_hash_skip") or ())
+
             def _check_content_hashes(content_hash_keys, legacy=False):
                 for hash_type in content_hash_keys:
                     if hash_type in source_dict:
--- a/conda_build/source.py
+++ b/conda_build/source.py
@@ -49,9 +49,7 @@ git_submod_re = re.compile(r"(?:.+)\.(.+)\.(?:.+)\s(.+)")
 ext_re = re.compile(r"(.*?)(\.(?:tar\.)?[^.]+)$")
 ACCEPTED_HASH_TYPES = ("md5", "sha1", "sha224", "sha256", "sha384", "sha512")
 CONTENT_HASH_KEYS = ("content_sha256", "content_sha384", "content_sha512")
-# Legacy v1 keys use the original CEP-19 algorithm (no length-prefixing).
-# Recipes that stored hashes computed with the old algorithm should use these keys.
-CONTENT_HASH_KEYS_V1 = ("content_sha256_v1", "content_sha384_v1", "content_sha512_v1")
+CONTENT_HASH_KEYS_V2 = ("content_sha256_v2", "content_sha384_v2", "content_sha512_v2")
 
 
 def append_hash_to_fn(fn, hash_value):
@@ -1130,9 +1128,9 @@ def provide(metadata):
                                 f"Empty {hash_type} hash provided for source item #{idx}"
                             )
                         if legacy:
-                            algorithm = hash_type[len("content_") : -len("_v1")]
-                        else:
                             algorithm = hash_type[len("content_") :]
+                        else:
+                            algorithm = hash_type[len("content_") : -len("_v2")]
                         obtained_content_hash = compute_content_hash(
                             src_dir,
                             algorithm,
@@ -1146,10 +1144,9 @@ def provide(metadata):
                                 f"expected '{expected_content_hash}'"
                             )
 
-            _check_content_hashes(CONTENT_HASH_KEYS)
-            # Legacy v1 keys use the CEP-19 algorithm (no length-prefixing) for backwards
-            # compatibility with hashes that were computed before the algorithm was fixed.
-            _check_content_hashes(CONTENT_HASH_KEYS_V1, legacy=True)
+            # Un-versioned keys use the original CEP-19 algorithm (deprecated).
+            _check_content_hashes(CONTENT_HASH_KEYS, legacy=True)
+            _check_content_hashes(CONTENT_HASH_KEYS_V2)
             patches = ensure_list(source_dict.get("patches", []))
             patch_attributes_output = []
             for patch in patches:
--- a/conda_build/source.py
+++ b/conda_build/source.py
@@ -49,6 +49,9 @@ git_submod_re = re.compile(r"(?:.+)\.(.+)\.(?:.+)\s(.+)")
 ext_re = re.compile(r"(.*?)(\.(?:tar\.)?[^.]+)$")
 ACCEPTED_HASH_TYPES = ("md5", "sha1", "sha224", "sha256", "sha384", "sha512")
 CONTENT_HASH_KEYS = ("content_sha256", "content_sha384", "content_sha512")
+# Legacy v1 keys use the original CEP-19 algorithm (no length-prefixing).
+# Recipes that stored hashes computed with the old algorithm should use these keys.
+CONTENT_HASH_KEYS_V1 = ("content_sha256_v1", "content_sha384_v1", "content_sha512_v1")
 
 
 def append_hash_to_fn(fn, hash_value):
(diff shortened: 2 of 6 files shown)


conda_build/utils.py
```diff
@@ -2033,6 +2033,7 @@ def compute_content_hash(
     by their full path. For each entry in the contents table, compute the hash for the concatenated
     bytes of:
 
+    - The decimal UTF-8 byte length of the path, followed by a UTF-8 encoded `:` separator.
```
jaimergp: UTF-8 byte length as in the number of bytes required to encode the path with UTF-8? And why do we need the colon separator?

tests/test_utils.py
```diff
@@ -399,6 +399,122 @@ def test_find_recipe_multipe_base(tmp_path: Path, file: str):
     assert path1.samefile(utils.find_recipe(tmp_path))
 
 
+# ---------------------------------------------------------------------------
+# compute_content_hash
+# ---------------------------------------------------------------------------
+
```
jaimergp:
```suggestion
```

No need for these.

pb01ka:
Seems like the following test is failing,

```
FAILED tests/test_api_build.py::test_recipe_builds[source_url] - RuntimeError: content_sha256 mismatch in source item #0: obtained 'fa18683d70b5b776b017ac0c55b1086a70c7a12584fc1c2fd5166b79f568c687' != expected 'a884ace5aa3a7e7f5a8b5adeb5cbfa7209f2ae88134d362c8bbca9c82ad2bb06'
```

This means the change I have made might be backwards incompatible? Is this a genuine concern?

cc: @jaimergp

conda_build/utils.py
pb01ka: Added an explanation for it.

jaimergp:
> This means the change I have made might be backwards incompatible? Is this a genuine concern?

That's to be decided in the CEP amendment we need to co-submit along with this fix. If it's a problem, we may need to add a `_v2` suffix or similar.

pb01ka:
@jaimergp The PR is passing all the tests. Please let me know if the changes look good here.

conda_build/metadata.py
```diff
@@ -620,6 +620,11 @@ def parse(data, config, path=None):
         "content_sha256": None,
         "content_sha384": None,
         "content_sha512": None,
+        # Legacy v1 keys (CEP-19 algorithm, no length-prefixing). Use these when you have
+        # existing stored hashes computed with the original CEP-19 algorithm.
+        "content_sha256_v1": None,
+        "content_sha384_v1": None,
+        "content_sha512_v1": None,
```
jaimergp: Not sure about the `v1` suffix here. It makes it look like this is actually the successor, not the deprecated version. We need a different term. Maybe just a leading underscore? Or something else?

- `_content_sha256`
- `_legacy_content_sha256`
- `_deprecated_content_sha256`

Or the other way around, _adding_ something to the new one (which is maybe better for bw compat):

- `length_prefixed_content_sha256` (too long?)
- `delimited_content_sha256`
- `normalized_content_sha256` (true in both editions, but is maybe more informative)

conda_build/source.py
```diff
@@ -1116,25 +1119,37 @@ def provide(metadata):
...
+                            raise RuntimeError(
+                                f"{hash_type} mismatch in source item #{idx}: "
+                                f"obtained '{obtained_content_hash}' != "
+                                f"expected '{expected_content_hash}'"
+                            )
+
+            _check_content_hashes(CONTENT_HASH_KEYS)
+            # Legacy v1 keys use the CEP-19 algorithm (no length-prefixing) for backwards
+            # compatibility with hashes that were computed before the algorithm was fixed.
+            _check_content_hashes(CONTENT_HASH_KEYS_V1, legacy=True)
```
jaimergp: Crazy idea: what if we catch the exception of the first call with `CONTENT_HASH_KEYS` and ALSO check whether the algorithm makes it match with some of `CONTENT_HASH_KEYS_V1`. If this one matches, then the error message tells them about the migration path. Too expensive and intricate for little gain, maybe?

conda_build/metadata.py
pb01ka: Good point on the naming. A couple of thoughts:

1. **Leading underscore vs `_v1` suffix** - semantically these are equivalent: both signal "this is the old/legacy variant." The backward-compatibility story is identical either way - users who have stored hashes computed with the original CEP-19 algorithm (no length-prefixing) still need to opt in to the legacy key explicitly, regardless of whether it's called `_content_sha256` or `content_sha256_v1`. I slightly prefer `_v1` because I think it makes the relationship between the two explicit (v1 = original, unversioned = current/improved), but I'm open to switching if the team finds `_content_sha256` or `_legacy_content_sha256` clearer.

2. **Adding something to the new key instead** - this is actually the cleanest option for backward compatibility. If we keep the existing `content_sha256` name for the legacy algorithm and introduce `length_prefixed_content_sha256` (or `delimited_`/`normalized_`) for the improved one, existing recipes and tooling that already reference `content_sha256` continue to work with zero changes and no deprecation warnings needed. The new, more secure algorithm is purely opt-in via the new name. The only downside is that the "plain" name now refers to the weaker algorithm, which could surprise new users - but that's a documentation problem, not a compatibility one. Among the suggestions, `normalized_content_sha256` reads the best to me since length-prefixing is an implementation detail whereas "normalized" conveys intent.

conda_build/source.py
pb01ka: Great idea, and I don't think it's too expensive or too intricate in practice.

The cost concern is largely moot because this check only happens on the error path - a mismatch is already a fatal build failure, so the extra hash computation is a one-time diagnostic cost on top of something that's already broken. And since content hashing is I/O-bound (reading all source files), doing it a second time with the legacy algorithm isn't dramatically more expensive than the first pass.

On the complexity side, the logic is self-contained within `_check_content_hashes`, so it's easy to reason about in isolation. The flow would be something like: catch the `RuntimeError` from the v2 check, recompute with `legacy=True`, and if that matches the provided hash, raise a more specific error pointing the user to rename `content_sha256` -> `content_sha256_v1`/"whatever naming scheme we decide" (or recompute their hash with the new algorithm). Without this, users who have a hash computed with the old algorithm will just get a cryptic mismatch and no indication of what happened or what to do.

I'll go ahead and implement this if you agree with my opinion.

jaimergp: Both algorithms feed on the same IO anyway, right? Is it too expensive to always compute both? Some benchmarks would help here. If it's super cheap, we may as well leave it as is, and at some point remove when `content_hash` (old legacy one).

We should also have warnings for the legacy option so users are encouraged to move to the next one if possible.

conda_build/metadata.py
jaimergp: I think we can find a better adjective than normalized, if it's about intent. Can you propose a few? e.g. I'm thinking of `corrected_content_sha256` but I'm worried there's yet another flaw and then we have to come up with something else 😂

conda_build/source.py
jaimergp: Note that this is not necessary if we only implement the new algo in a different field name, as discussed in https://github.com/conda/conda-build/pull/5992#discussion_r3372236401

conda_build/metadata.py
pb01ka: Fair concern - "corrected" (and "normalized") both imply "this is the fixed version," which ages badly if another issue surfaces later. Here are a few alternatives grouped by strategy:

**Describe the mechanism, not the quality**

- `delimited_content_sha256` - the entries (paths, symlinks) are delimited before hashing, so concatenation collisions can't occur. Specific enough to be meaningful, but makes no claim about overall algorithm quality.
- `prefixed_content_sha256` - refers directly to the length-prefixing. Accurate, but does expose the implementation detail (`length_prefixed_` has similar concerns).

**Describe the property**

- `distinct_content_sha256` - the inputs are kept distinct; no two different path lists can produce the same hash input. Conveys the intent without implying "fully corrected."
- `unambiguous_content_sha256` - similar angle: the serialization is unambiguous. Slightly more self-explanatory but a bit long.

**Just version it**

- `content_sha256_v2` - the most future-proof option. Makes no quality claim at all, just signals "newer algorithm." If a third revision is ever needed, `v3` slots in cleanly. The downside is it's less informative on its own, but a short docstring fixes that.

My recommendation: `delimited_content_sha256` if you want the name to be self-documenting, `content_sha256_v2` if you want it to be future-proof without overloading meaning onto an adjective.

pb01ka:
The failing test doesn't seem relevant. AFAICT, they seem to fail due to Github constraints. Please let me know if I am perceiving this wrongly. 

Other than this the PR is up to date. Please let me know if any other change is needed.

pyproject.toml
kenodegard, line 1: This should be flipped, `PendingDeprecationWarning` should warn and not error in tests, `DeprecationWarning` and `FutureWarning` should continue to error.

pb01ka: Done.

conda_build/metadata.py
jaimergp: Let's go with `_v2` then.

pb01ka:
The failing test is due to network issues I think,

```zsh
FAILED tests/test_post.py::test_pypi_installer_metadata - conda.CondaMultiError: ('Connection broken: IncompleteRead(8124139 bytes read, 4861299 more expected)', IncompleteRead(8124139 bytes read, 4861299 more expected))
```

Please feel free to re-run the failed checks to be sure. Other than this the PR is ready. If you have any reviews please feel free to share.

I will also send a draft PR with the CEP linking this PR. It will supersede CEP-19.

conda_build/metadata.py
pb01ka: Already done. We are in good shape as of now.

conda_build/source.py
pb01ka: I think this thread can be resolved since we already decided upon `_v2` suffix (see https://github.com/conda/conda-build/pull/5992#discussion_r3403719766).

pb01ka:
The CEP PR is raised at https://github.com/conda/ceps/pull/174.

@jaimergp @kenodegard

jaimergp:
Thanks @pb01ka, I'll mark this as draft until the CEP is approved.

PR #6015, opened 2026-06-17
Fix platform config values for v1 recipes

jsmolic:
The code is currently trying to set platform configuration values by checking `config.variant` object which is not valid as we are not loading config files through conda-build's mechanism and any configuration value will not get applied but rather fall back to the default `config.subdir`. Fix this by obtaining the settings from the variant config loaded in the rattler compat code.
While at it, cleanup redundant command check in `run_rattler()`.

### Checklist - did you ...

### DIFF
--- a/conda_build/_rattler_build/compat.py
+++ b/conda_build/_rattler_build/compat.py
@@ -416,16 +416,14 @@ def run_rattler(command: str, parsed_args: argparse.Namespace, config: Config) -
         else:
             package_format = ".tar.bz2"
 
-    if command in ("build", "render"):
-        if command == "render":
-            recipes = [str(Path(parsed_args.recipe) / "recipe.yaml")]
-        else:
-            recipes = [
-                str(Path(recipe_dir) / "recipe.yaml")
-                for recipe_dir in parsed_args.recipe
-            ]
+    if command == "render":
+        recipes = [str(Path(parsed_args.recipe) / "recipe.yaml")]
+    else:
+        recipes = [
+            str(Path(recipe_dir) / "recipe.yaml") for recipe_dir in parsed_args.recipe
+        ]
 
-        recipe_results: list[RecipeResult] = []
+    recipe_results: list[RecipeResult] = []
 
     for recipe_path in recipes:
         recipe_results.append(
--- a/conda_build/_rattler_build/compat.py
+++ b/conda_build/_rattler_build/compat.py
@@ -306,12 +306,6 @@ def run_rattler(command: str, parsed_args: argparse.Namespace, config: Config) -
     channels: list[str] = []
     extra_context: dict[str, str] = {}
     show_logs: bool = getattr(parsed_args, "quiet", False) is False
-    target_platform: str = config.variant.get("host_platform", config.subdir)
-    build_platform: str = config.variant.get("build_platform", config.subdir)
-    host_platform: str = config.variant.get("target_platform", config.subdir)
-    noarch_build_platform: str = config.variant.get(
-        "noarch_build_platform", config.subdir
-    )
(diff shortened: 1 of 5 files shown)


beckermr (commented):
The test suite apparently didn't catch what appears to be a pretty fundamental issue here. I think we should add tests specifically for this error.

jsmolic:
> The test suite apparently didn't catch what appears to be a pretty fundamental issue here. I think we should add tests specifically for this error.

Yes, I'm working on a test for this atm

beckermr:
Ahh sorry. I thought this was ready for review!

conda_build/_rattler_build/compat.py
```diff
@@ -392,6 +385,19 @@ def run_rattler(command: str, parsed_args: argparse.Namespace, config: Config) -
...
+            output_dir = parsed_args.output_folder
+        no_build_id = not parsed_args.set_build_id
+        skip_existing = parsed_args.skip_existing or "none"
+        no_include_recipe = not parsed_args.include_recipe
+        if parsed_args.conda_pkg_format == CondaPkgFormat.V2:
+            package_format = "conda"
+        else:
+            package_format = ".tar.bz2"
+
     if command in ("build", "render"):
```
ForgottenProgramme: This outer `if` statement is not needed, I think.

jsmolic: Good catch, I removed it (hopefully it won't pollute the diff too much :D)

jsmolic: Looking once again it seems tests are failing so I'll revert and do this change as part of https://github.com/conda/conda-build/pull/6020

ForgottenProgramme (approved):
LGTM!

jaimergp:
AFAIK, rattler-build does not read `target_platform` from CBC files. e.g. see https://github.com/conda-forge/conda-forge-bot/pull/6274

Is `VariantConfig` generic enough to parse it, or does it ignore it?

jsmolic:
> AFAIK, rattler-build does not read `target_platform` from CBC files. e.g. see [conda-forge/conda-forge-bot#6274](https://github.com/conda-forge/conda-forge-bot/pull/6274)
> 
> Is `VariantConfig` generic enough to parse it, or does it ignore it?

Yes, I know that the platform config settings will be ignored there if it's attempted with the CLI as it gives a warning.

`VariantConfig` on the other hand can parse the settings from the cbc file and then we just read and forward those values to `PlatformConfig`

conda_build/_rattler_build/compat.py
```diff
@@ -331,46 +325,45 @@ def run_rattler(command: str, parsed_args: argparse.Namespace, config: Config) -
...
+
+    # configure variant
+    # merge config files in the order they are stacked
+    if config_files:
+        for variant in config_files:
+            variant_config = variant_config.merge(VariantConfig.from_file(variant))
+
+    def get_config_value(name):
+        value = variant_config.get(name, config.subdir)
+        return value[0] if isinstance(value, list) else value
```
jaimergp: Are there cases when there's more than one value in the list? What about tuples?

jsmolic: The expected type in the API is a [list](https://rattler-build.prefix.dev/latest/py-rattler-build/reference/rendering/#rattler_build.VariantConfig.__init__). That being said, I'm not sure if there would be a case where we'd have multiple values for a `*_platform` key so as we discussed we can try erroring out if we get multiple values.

1
jaimergp, line 1: What is this file @jsmolic?

jsmolic: vim accidentally saved a file under a new name :shaking_face:

jaimergp (approved):

PR #6036, opened 2026-07-07
Fix `find_used_variables_in_text()` to match variables without spaces

mgorny:
### Description

Fix the regular expression used to find used variables in text to also match space-less variable references such as `{{python_min}}`.

This specific syntax used not to match because the regex expected the variable name to be followed by at least one character that wasn't a valid variable name; presumably to avoid matching a substring while allowing for all varieties of operators followed by anything.  Normally, this would match the space after the variable name.

However, given no space and no operator it caused the regex not to match at all.  To fix it, make that group optional; the regex will still reject substring matching, but it will accept variable name immediately followed by the closing braces.

### Checklist - did you ...

- [x] ~Add / update outdated documentation?~

### DIFF
--- a/conda_build/variants.py
+++ b/conda_build/variants.py
@@ -776,7 +776,7 @@ def find_used_variables_in_text(variant, recipe_text, selectors_only=False):
         v_regex = re.escape(v)
         v_req_regex = "[-_]".join(map(re.escape, v.split("_")))
         variant_regex = (
-            rf"\{{\s*(?:pin_[a-z]+\(\s*?['\"])?{v_regex}[^_0-9a-zA-Z].*?\}}\}}"
+            rf"\{{\s*(?:pin_[a-z]+\(\s*?['\"])?{v_regex}(?:[^_0-9a-zA-Z].*?)?\}}\}}"
         )
         selector_regex = rf"^[^#\[]*?\#?\s\[[^\]]*?(?<![_\w\d]){v_regex}[=\s<>!\]]"
         # NOTE: why use a regex instead of the jinja2 parser/AST?
--- a/news/6036-fix-used-variables-without-spaces.md
+++ b/news/6036-fix-used-variables-without-spaces.md
@@ -1,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Fix finding used variables that do not feature spaces, e.g. `{{python_min}}`. (#6036)
+
+### Deprecations
+
+* <news item>
+
+### Docs
+
+* <news item>
+
+### Other
+
+* <news item>
--- a/tests/test_variants.py
+++ b/tests/test_variants.py
@@ -835,6 +835,11 @@ def test_get_vars():
             "{{ python_min }}",
             {"python_min"},
         ),
+        (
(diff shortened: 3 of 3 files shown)


conda_build/variants.py
```diff
@@ -776,7 +776,7 @@ def find_used_variables_in_text(variant, recipe_text, selectors_only=False):
         v_regex = re.escape(v)
         v_req_regex = "[-_]".join(map(re.escape, v.split("_")))
         variant_regex = (
-            rf"\{{\s*(?:pin_[a-z]+\(\s*?['\"])?{v_regex}[^_0-9a-zA-Z].*?\}}\}}"
+            rf"\{{\s*(?:pin_[a-z]+\(\s*?['\"])?{v_regex}(?:[^_0-9a-zA-Z].*?)?\}}\}}"
```
jaimergp, line 779: Let me see if I can parse it to human language. Before we were matching termination by finding one character that cannot be used to name Python variables, followed by _anything_, followed by `}}`. Now we are allowing that this group is not present because. Adding `?` to `[^_0-9a-zA-Z]` doesn't work on its own. Correct?

This makes me think that these regexes could be expressed with `re.VERBOSE` so we could comment them, but well, not a blocker.

mgorny: The "non-variable character" is followed by `.*`, so if we made it optional, `.*` could technically match "variable characters" (not sure if such an unpredictable regex is valid). So we're matching either:

- variable followed by `}}` immediately
- variable followed by non-variable character (say, `|` or `+`), followed by anything (filter, variable…), followed by `}}`

mgorny: As for verbose, I didn't really want to have to figure out the earlier part of the regex :-).

beckermr:
I am legitimately shocked that this is a bug. Conda-forge would have broken years ago if this was an issue. Have you tried the new tests without the regex changes?

mgorny:
Yes. In fact, I wrote them before finding what was broken in the first place, to make sure it wasn't a smithy problem. And I've hit it in the wild, while testing something else.

CF doesn't hit it naturally because as a stylistic rule, we use `{{ foo }}` with a space. However, per Jinja plain `{{foo}}` is valid as well, and this is quite an annoying pitfall when you're writing a recipe for local/temporary use and don't bother with styling.

jaimergp:
There are [23 active recipes that use spaceless Jinja](https://github.com/search?type=code&q=org%3Aconda-forge+path%3Ameta.yaml+%2F%5C%7B%5C%7B%5CS%2F++NOT+is%3Aarchived), but apparently those were not critical hotspots for rerendering?

mgorny:
> There are [23 active recipes that use spaceless Jinja](https://github.com/search?type=code&q=org%3Aconda-forge+path%3Ameta.yaml+%2F%5C%7B%5C%7B%5CS%2F++NOT+is%3Aarchived), but apparently those were not critical hotspots for rerendering?

I don't think any of them impact pinning.

beckermr:
Wow ok. Conda forge didn't have that linting rule for a long time, maybe it's first five years or so.

jaimergp (approved):

PR #6047, opened 2026-07-16
Wrap native win-arm64 scripts with a `start /machine` launcher

### LINKED ISSUE #6048: `PROCESSOR_ARCHITECTURE` inherited from parent process causes mismatch with non-default `build_platform`
### Checklist

### What happened?

In the `win-arm64` migration at conda-forge, we are trying to bootstrap the base dependencies from a `win-64` environment running on `windows-11-arm` runners. These emulated processes can then create `_build_env`  environments for `win-arm64` that run natively and we don't have to cross-compile anything.

To do so, `conda-build` must execute the ARM64 slice of `cmd.exe`. Also, the env vars set in `build_env_setup.bat` must be filtered so they don't inherit the values of the parent process.

### Conda Info

```shell
N/A
```

### Conda Config

```shell
N/A
```

### Conda list

```shell
N/A
```

### Additional Context

xrefs:

- https://github.com/conda-forge/conda-forge.github.io/issues/1940
- https://github.com/conda-forge/conda-smithy/pull/2622
- https://github.com/jaimergp/windows-11-arm-tests/actions

jaimergp:
### Description

Closes https://github.com/conda/conda-build/issues/6048 by:

- Wrapping the CMD subprocesses with a `start /mahcine` launcher
- Adjusting `PROCESSOR_ARCHITECTURE` and friends so they are not blindly inherited and injected from the parent process

### Checklist - did you ...

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -69,8 +69,7 @@ jobs:
   linux:
     # only run test suite if there are code changes
     needs: changes
-    if: false
-    # github.event_name == 'schedule' || needs.changes.outputs.code == 'true'
+    if: github.event_name == 'schedule' || needs.changes.outputs.code == 'true'
 
     runs-on: ubuntu-latest
     defaults:
@@ -218,8 +217,7 @@ jobs:
   linux-benchmarks:
     # only run test suite if there are code changes
     needs: changes
-    if: false
-    #needs.changes.outputs.code == 'true'
+    if: needs.changes.outputs.code == 'true'
 
     runs-on: ubuntu-latest
     defaults:
@@ -293,29 +291,25 @@ jobs:
     needs: changes
     if: github.event_name == 'schedule' || needs.changes.outputs.code == 'true'
 
-    runs-on: ${{ matrix.runs-on }}
+    runs-on: ${{ matrix.runs-on && matrix.runs-on || 'windows-2022' }}
     strategy:
       fail-fast: false
       matrix:
         include:
-          # # test lower version (w/ stable conda)
-          # - python-version: '3.10'
-          #   conda-version: release
-          #   test-type: serial
-          #   runs-on: windows-2022
-          # - python-version: '3.10'
-          #   conda-version: release
-          #   test-type: parallel
-          #   runs-on: windows-2022
-          # # upper version (w/ canary conda)
-          # - python-version: '3.14'
-          #   conda-version: canary
-          #   test-type: serial
-          #   runs-on: windows-2022
-          # - python-version: '3.14'
-          #   conda-version: canary
-          #   test-type: parallel
-          #   runs-on: windows-2022
+          # test lower version (w/ stable conda)
+          - python-version: '3.10'
+            conda-version: release
+            test-type: serial
+          - python-version: '3.10'
+            conda-version: release
+            test-type: parallel
+          # upper version (w/ canary conda)
+          - python-version: '3.14'
+            conda-version: canary
+            test-type: serial
+          - python-version: '3.14'
+            conda-version: canary
+            test-type: parallel
           - python-version: '3.14'
             conda-version: release
             test-type: custom
@@ -420,8 +414,7 @@ jobs:
   macos:
     # only run test suite if there are code changes
     needs: changes
-    if: false
-    # github.event_name == 'schedule' || needs.changes.outputs.code == 'true'
+    if: github.event_name == 'schedule' || needs.changes.outputs.code == 'true'
 
     runs-on: ${{ matrix.runs-on }}
     defaults:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -356,7 +356,10 @@ jobs:
           run-post: false  # skip post cleanup
           pkgs-dirs: ${{ runner.arch == 'ARM64' && 'C' || 'D'}}:\conda_pkgs_dir
           installation-dir: ${{ runner.arch == 'ARM64' && 'C' || 'D'}}:\conda
-          architecture: x64  # force x64 even on ARM64 to test emulation
+          # force x64 even on ARM64 to test emulation
+          architecture: x64
+        env:
+          CONDA_SUBDIR: win-64
 
       - name: Choco Install
         if: runner.arch != 'ARM64'
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -319,7 +319,7 @@ jobs:
           - python-version: '3.14'
             conda-version: release
             test-type: custom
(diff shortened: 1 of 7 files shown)


conda_build/windows.py
```diff
@@ -347,6 +380,26 @@ def write_build_scripts(m, env, bld_bat):
...
+def build_command_arguments(m, script: str) -> list[str]:
+    if m.config.build_subdir != context._native_subdir():
+        # If conda-build is run from e.f. a win-64 environment on a win-arm64 machine
+        # users may want to build natively by setting build_platform="win-arm64".
+        # In those cases, we need to ensure that the CMD process is native ARM64
+        # via this `start` wrapper. Otherwise Windows picks the AMD64 slice!
+        wrapper = os.path.join(os.path.dirname(script), "_win_native_wrapper.bat")
+        machine = m.config.build_subdir.split("-")[1]
+        if machine == "64":
+            machine = "amd64"
```
baszalmstra: This should also trigger for win-32 probably.

jaimergp: Yes, thanks, adjusted in [fd3f4e9](https://github.com/conda/conda-build/pull/6047/commits/fd3f4e9a39a1afac99b3de18e2e966e59bf49e05)

conda_build/windows.py
```diff
@@ -383,18 +383,22 @@ def write_build_scripts(m, env, bld_bat):
 
 
 def build_command_arguments(m, script: str) -> list[str]:
-    if m.config.build_subdir != context._native_subdir():
+    machine = m.config.build_subdir.split("-")[1]
+    machine = {"64": "AMD64", "32": "x86"}.get(machine, machine).upper()
+    if (
+        machine == get_native_windows_architecture()
```
baszalmstra: What does this check prevent?

jaimergp: We are handling three sources of truth:

1. What the user wants for `build_platform`
2. What the Python interpreter reports via `context._native_subdir()`
3. What Windows reports (the ground truth)

I guess this gets weird if we try to mix win-32, win-64 and win-arm64. Is that combination likely? I don't think so. Does it hurt to double check that 1 and 3 match? I think that's ok.

baszalmstra: I think we want to force the interpreter to run as win-arm64 when building as win-arm64. We know this is not the case when the parent process is x64. I dont think number 3 buys you anything.

jaimergp: Tried a different approach now. We request `/machine` and override `PROCESSOR_ARCHITECTURE` whenever there's a mismatch between interpreter and build_platform. I clear `PROCESSOR_ARCHITEW6432` when build_platform matches Windows-reported arch.

conda_build/windows.py
```diff
@@ -293,6 +336,13 @@ def write_build_scripts(m, env, bld_bat):
         env["PYTHONDONTWRITEBYTECODE"] = True
     import codecs
 
+    if m.config.build_subdir != _running_subdir():
+        # Can't trust the parent environment under these conditions
+        build_arch = _build_arch(m)
+        env["PROCESSOR_ARCHITECTURE"] = build_arch
+        if build_arch == get_native_windows_architecture():
+            env.pop("PROCESSOR_ARCHITEW6432", None)
```
jaimergp, line 344: Note these are being set to whatever `os.environ` carries at https://github.com/conda/conda-build/blob/f2b66175825335667ce3137e3823b3f85b10bb0a/conda_build/environ.py#L685-L686

so we need to undo that here

conda_build/windows.py
```diff
@@ -74,6 +85,38 @@ def fix_staged_scripts(scripts_dir, config):
...
+
+    For prior versions, query the registry.
+
+    Returns None if it value cannot be obtained.
+    """
+    if sys.platform != "win32":
+        raise OSError("This function is only supported on Windows.")
+
+    if sys.version_info >= (3, 12):
+        return platform.machine() or None
```
isuruf, line 102: This will change if https://github.com/python/cpython/issues/98962 is fixed and I want to push for that.

isuruf (approved):

conda_build/_rattler_build/compat.py
```diff
@@ -450,9 +450,9 @@ def get_config_value(name):
             return value[0]
         return value
 
-    build_platform = get_config_value("build_platform")
-    host_platform = get_config_value("host_platform")
-    target_platform = get_config_value("target_platform")
+    build_platform = config.build_subdir  # Overridable via CONDA_SUBDIR env var
+    host_platform = get_config_value("target_platform", config.host_subdir)
+    target_platform = get_config_value("target_platform", config.target_subdir)
     noarch_build_platform = get_config_value("noarch_build_platform")
```
jaimergp, line 456: @jsmolic, I updated this because `build_platform` and `host_platform` are not variant settings recognized by `conda-build`. Sure you can put it there but they don't have any effect on `meta.yaml`.

Instead, if the user wants to change `--build-platform`, it needs to happen via `CONDA_SUBDIR`. `--host-platform` is tied to `target_platform`. `noarch_build_platform` doesn't exist in conda-build either and I doubt anyone will set it up (quite obscure as one can read in https://github.com/prefix-dev/rattler-build/pull/1192), so I'll leave it there but the important part is that the default now is `None`.

.github/workflows/tests.yml
```diff
@@ -380,7 +390,8 @@ jobs:
           --basetemp=${{ runner.temp }}
           --durations-path=durations\${{ runner.os }}.json
           -n auto
-          -m "${{ env.PYTEST_MARKER }}"
+          ${{ env.PYTEST_MARKER && '-m "' || ''}}${{ env.PYTEST_MARKER }}${{ env.PYTEST_MARKER && '"' || ''}}
+          ${{ matrix.pytest-expression && '-k "' || ''}}${{ matrix.pytest-expression }}${{ matrix.pytest-expression && '"' || ''}}
```
travishathaway, line 394: Maybe these lines are sufficiently complex enough to warrant a comment? 😅 I'm struggling to understand exactly what this is doing. Especially because of the doubled `env.PYTEST_MARKER` and `matrix.pytest-expression` usage.

conda_build/_rattler_build/compat.py
jsmolic: Got it, thanks for fixing that up @jaimergp

.github/workflows/tests.yml
jaimergp: These are ternary operators in JavaScript, adapted to Github Actions. Equivalent to `'-m "' if env.PYTEST_MARKER else ''`. So if PYTEST_MARKER is set, then we write `-m "{{ env.PYTEST_MARKER }}"`.

PR #6072, opened 2026-07-30
Support `--variants` when using `--output`

JeanChristopheMorinPerso:
### Description

Honor `--variants` when `conda-build --output` calculates package paths. The option was accepted by the CLI but was not forwarded to `get_output_file_paths()`, so `--output` always returned the full variant matrix.

The regression test covers both the full matrix and a matrix restricted with `--variants "something: abc"`.

### Checklist

- [x] Added a news entry.
- [x] Added regression coverage.
- [x] No documentation changes are needed.

### DIFF
--- a/tests/cli/test_main_build.py
+++ b/tests/cli/test_main_build.py
@@ -124,19 +124,18 @@ def test_build_output_build_path_variants(testing_config, capfd):
     testing_config.debug = False
 
     # Test that without passing --variants, we get all variants
-    args = ["--output", os.path.join(variants_dir, "12_variant_versions")]
+    args = ["--output", os.path.join(variants_dir, "11_variant_output_names")]
     main_build.execute(args)
+
+    names = [
+        "some_output_using_abc_ghi-1.0-hd82c8f6_0.conda",
+        "some_output_using_abc_jkl-1.0-h64b44fd_0.conda",
+        "some_output_using_def_ghi-1.0-h95087dd_0.conda",
+        "some_output_using_def_jkl-1.0-h085f5b8_0.conda",
+    ]
     test_paths = [
-        os.path.join(
-            testing_config.croot,
-            testing_config.host_subdir,
-            "my_package-470.470-h65f20af_0.conda",
-        ),
-        os.path.join(
-            testing_config.croot,
-            testing_config.host_subdir,
-            "my_package-480.480-h1f30878_0.conda",
-        ),
+        os.path.join(testing_config.croot, testing_config.host_subdir, name)
+        for name in names
     ]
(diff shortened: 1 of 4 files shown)


tests/cli/test_main_build.py
```diff
@@ -119,6 +120,56 @@ def test_build_output_build_path(
...
 
+def test_build_output_build_path_variants(testing_config, capfd):
+    testing_config.verbose = False
+    testing_config.debug = False
+
+    # Test that without passing --variants, we get all variants
+    args = ["--output", os.path.join(variants_dir, "11_variant_output_names")]
+    main_build.execute(args)
+
+    # Build hash (h...) is not stable between local and CUI builds.
```
jaimergp: What's CUI? CI?

tests/cli/test_main_build.py
```diff
@@ -119,6 +120,56 @@ def test_build_output_build_path(
     assert error == ""
 
 
+def test_build_output_build_path_variants(testing_config, capfd):
```
jaimergp: This test is a good candidate for `pytest.subtest`.

tests/cli/test_main_build.py
jaimergp:
```suggestion
    # Build hash (h...) is not stable between local and CI builds.
```

jaimergp: Not sure why this is the case, though. Should be stable to recipe inputs, regardless the system environment 🤔

tests/cli/test_main_build.py
JeanChristopheMorinPerso: Good point. I used `pytest.parametrize` instead.

tests/cli/test_main_build.py
JeanChristopheMorinPerso: Yeah, I'm not really sure why it's not stable in this case...

JeanChristopheMorinPerso:
@jezdez Why did you remove the change to `conda render` in https://github.com/conda/conda-build/pull/6072/changes/5e3e19a3f119eabd3a59f2f9edd10646e5bef534?

jezdez:
I removed the extra argument because `conda render` already passes `parsed.variants` to `api.render()` before calculating output paths. At that point, `get_output_file_paths()` receives rendered metadata and doesn’t use its `variants` argument.

`conda build --output` passes a recipe path instead, so the missing forwarding matters there. I should have explained that distinction in the commit.

JeanChristopheMorinPerso:
Ah, thanks for the explanation. It look like I missed that. Thanks for fixing that.

jezdez (approved):

PR #5545, opened 2024-11-18
Add unsigned, MSVC-build `win-arm64` launchers

jaimergp:
### Description

Taken directly from [win-arm64/conda-launchers-24.7.1-h2722cbf_1.tar.bz2](https://anaconda.org/conda-canary/conda-launchers/24.7.1/download/win-arm64/conda-launchers-24.7.1-h2722cbf_1.tar.bz2) at [conda-canary/conda-launchers](https://anaconda.org/conda-canary/conda-launchers/files?version=24.7.1), which were built in [this CI run](https://github.com/conda/conda-launchers/actions/runs/11898535180/job/33155248271) for [this commit](https://github.com/conda/conda-launchers/commit/d23e5578fb780bff2255c9180a7e5346b200a6d6).

Needs to be signed with Anaconda certificates.

### Checklist - did you ...

- [ ] Add a file to the `news` directory ([using the template](https://github.com/conda/conda-build/blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [ ] Add / update necessary tests?
- [ ] Add / update outdated documentation?

PR #6075, opened 2026-08-03
Use published Windows launchers

jaimergp:
### Description

Use Windows entry point executables from the published `conda-launchers >=24.7.1` package. Select the executable for the target architecture from `share/conda-launchers`, including native `win-arm64`, and verify its SHA256 against the installed package metadata before copying it. Report missing native launchers instead of falling back to x64.

Retain the existing signed 32-bit executables as a transitional fallback when the installed package does not own those paths, since defaults currently provides only x64 and ARM64 launchers. Verify the bundled files against their checked-in SHA256 values. An owned package file that is missing or fails verification remains an error.

Legacy noarch packages include the available Windows launchers and select the ARM64 executable when installed with native Windows ARM64 Python. The package can use either conda-forge's NumFOCUS-signed launchers or defaults' Anaconda-signed launchers.

Refs conda/conda#16293 and conda/conda#15448.
Closes #5545.

### Checklist - did you ...

- [x] Add a file to the `news` directory ([using the template](https://github.com/conda/conda-build/blob/main/news/TEMPLATE.md)) for the next release's release notes?


tests/requirements.txt
```diff
@@ -2,6 +2,7 @@ beautifulsoup4
 chardet
 conda >=25.11.0
 conda-index >=0.4.0
+conda-launchers
```
jaimergp:
```suggestion
conda-forge::conda-launchers
```

jaimergp:
pre-commit.ci autofix

jezdez:
pre-commit.ci autofix

jezdez (approved):
This looks good to me. Thanks for working on this, @jaimergp!

I made the following commits on this PR:

- 6e2bf998: Use published native Windows launchers
- bf3c892e: Preserve win32 launchers without changing test channels
- 3f63751c: Merge branch main into conda-launchers
- 987e3064: Merge main and require conda 26.1
- 3b5ac364: Use portable command in v1 package tests
- 4f19cff4: Preserve integer host architectures when selecting launchers

Could you please review my changes as a whole, including the launcher selection and checksum verification, the win32 fallback, the minimum conda version, and the test updates? I would appreciate your second look before merging.

conda_build/utils.py
```diff
@@ -1104,6 +1104,50 @@ def iter_entry_points(items):
         yield m.groups()
 
 
+def locate_conda_launcher(
+    arch: str, *, launcher_type: Literal["cli", "gui"] = "cli"
```
kenodegard: Isn't arch also a literal?


```suggestion
    arch: Literal["32", "64", "arm64"], *, launcher_type: Literal["cli", "gui"] = "cli"
```

kenodegard (approved):

conda_build/_link.py
```diff
@@ -111,8 +112,13 @@ def create_script(fn):
     if sys.platform == "win32":
         shutil.copy2(src, dst + "-script.py")
         FILES.append(f"Scripts/{fn}-script.py")
+        arch = (
+            "arm64"
+            if sysconfig.get_platform() == "win-arm64"
```
jaimergp, line 117: A rare case of conda and Python agreeing! Had to double check to make sure, but yes, that's [the returned value](https://docs.python.org/3/library/sysconfig.html#sysconfig.get_platform).

conda_build/noarch_python.py
```diff
@@ -144,9 +144,12 @@ def transform(m, files, prefix):
...
     if d["python-scripts"]:
-        for fn in "cli-32.exe", "cli-64.exe":
-            # FIXME: Update once win-arm64 native launcher is available
-            shutil.copyfile(join(this_dir, fn), join(prefix, fn))
+        for arch in "32", "64", "arm64":
+            try:
+                source = locate_conda_launcher(arch)
+            except FileNotFoundError:
+                continue
+            shutil.copyfile(source, join(prefix, f"cli-{arch}.exe"))
```
jaimergp, line 149: What if no launcher is found? Should that be an error?

conda_build/utils.py
```diff
@@ -1104,6 +1104,50 @@ def iter_entry_points(items):
...
+        )
+    path_data = next(
+        (path for path in record.paths_data.paths if path.path == short_path), None
+    )
+    if path_data is None:
+        if arch != "32":
+            raise FileNotFoundError(
+                f"The installed conda-launchers package does not provide {short_path}."
+            )
+        # defaults does not yet publish 32-bit launchers.
```
jaimergp:
```suggestion
        # defaults does not yet publish 32-bit launchers.
        # use the bundled ones for now
```

pyproject.toml
```diff
@@ -81,7 +82,7 @@ sort = "Miss"
 relative_files = true
 
 [tool.hatch.build]
-include = ["conda_build", "conda_build/templates/*", "conda_build/cli-*.exe"]
+include = ["conda_build", "conda_build/templates/*"]
```
jaimergp, line 85: Don't we need to keep the `*-32.exe` ones?

tests/test_utils.py
```diff
@@ -634,15 +634,3 @@ def test_max_cmd_line_length_default():
         assert utils.MAX_CHUNK_SIZE == 8190
     else:
         assert utils.MAX_CHUNK_SIZE == 32760
-
-
-@pytest.mark.parametrize("arch", ["32", "64", "arm64"])
-def test_create_entry_point(monkeypatch, tmp_path, testing_config, arch):
```
jaimergp, line 640: Note to reviewers: moved to `test_launchers.py`.

conda_build/noarch_python.py
jezdez: Yes. The 32-bit fallback prevents the zero-launcher case, but skipping a missing x64 or ARM64 launcher still produces a package that fails to install on that platform. Fixed in 05195867 by letting the error propagate, with regression coverage for both missing-launcher cases.

conda_build/_link.py
jezdez: Yes, and using Python’s platform keeps x64 Python running under emulation on the x64 launcher.

pyproject.toml
jezdez: We do, and both files remain included. Hatch’s `conda_build` include pattern covers the directory recursively, and neither executable is excluded by `.gitignore`. The explicit `cli-*.exe` pattern was redundant.

jaimergp:
Can't approve my own PR, so have this instead




(LGTM!)

PR #6083, opened 2026-08-07
Omit REQUESTS_CA_BUNDLE from build env defaults

### LINKED ISSUE #6063: botocore error on empty string for environment variables
https://github.com/conda/conda-build/blob/e7ad913b79114ea03c34e02a7f5e0e1c65ef80a7/conda_build/environ.py#L416

is causing 

https://github.com/boto/botocore/pull/3751/changes#diff-fe9f2a328fcc761f9d5cf26bf3e438716ffc27eb35da763a1f40ef8365662977R490

to trigger an error on environment that use boto core.

Could we change the syntax to simply not define `REQUESTS_CA_BUNDLE` in case it isn't present in the environment?

ryanskeith:
## Description

Fixes [#6063](https://github.com/conda/conda-build/issues/6063): stop injecting `REQUESTS_CA_BUNDLE=""` into the build environment when the variable is unset.

### Changes
- Pass it through via `get_default`mechanism so it is only present when set in the host environment.
- Updated Document the inherited behavior; empty CA paths break TLS clients such as recent botocore ([boto/botocore#3751](https://github.com/boto/botocore/pull/3751)).

### Notes for review
- Left `HTTP_PROXY` / `HTTPS_PROXY` on the older `os.getenv(..., "")` path for now; they have the same pattern and could follow in a follow-up.
- `REQUESTS_CA_BUNDLE` is in `os_vars` so it applies on all platforms (including Windows `requests`/botocore).

## Testing
- I was undecided if this should be made test.

Here is a LLM generated test used for local testing:
```
python - <<'PY'
import os, tempfile
from pathlib import Path
from conda_build.config import Config
from conda_build.metadata import MetaData
from conda_build.environ import get_dict

workdir = tempfile.mkdtemp()
recipe = Path(workdir) / "recipe"
recipe.mkdir()
(recipe / "meta.yaml").write_text(
    "package:\n  name: probe\n  version: 0.0.1\n"
    "build:\n  number: 0\n  noarch: generic\n"
    "about:\n  home: https://example.com\n  license: BSD-3-Clause\n  summary: probe\n"
)
m = MetaData(str(recipe), config=Config(croot=workdir, verbose=False, anaconda_upload=False))

os.environ.pop("REQUESTS_CA_BUNDLE", None)
assert "REQUESTS_CA_BUNDLE" not in get_dict(m=m), "unset should omit"

os.environ["REQUESTS_CA_BUNDLE"] = "/tmp/dummy-ca.pem"
assert get_dict(m=m)["REQUESTS_CA_BUNDLE"] == "/tmp/dummy-ca.pem", "set should pass through"
print("ok")
PY
```


danyeaw (approved):
Hi @ryanskeith, this looks good, thanks! I agree on deferring `HTTP(S)_PROXY`. Optional: instead of your manual script, it would be nice to add a small os_vars/get_dict test in `tests/test_environ.py` for unset vs set so this doesn’t regress.

kenodegard (approved):

PR #6092, opened 2026-08-13
Expect InvalidVersion from packaging if version is not string

### LINKED ISSUE #6091: Incompatibility with the latest packaging
This piece of code is no longer compatible with the latest packaging:

https://github.com/conda/conda-build/blob/c7a17ef0fcad9e5a1fd24bbf1d31792a6e69f667/conda_build/deprecations.py#L76-L80

because packaging now raises `InvalidVersion` instead of `TypeError` when `Version` is given a non-string. See: https://packaging.pypa.io/en/stable/changelog.html#id1

IIUIC, the solution here is to catch both exceptions.

frenzymadness:
Fixes: https://github.com/conda/conda-build/issues/6091

### DIFF
--- a/conda_build/deprecations.py
+++ b/conda_build/deprecations.py
@@ -8,6 +8,7 @@ import sys
 import warnings
 from argparse import Action
 from functools import wraps
+from packaging.version import InvalidVersion
 from types import ModuleType
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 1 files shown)


ryanskeith:
@frenzymadness, thank you so much for your contribution.  Would you mind signing the CLA?  Also, I think a quick news blurb in the news folder would be a good call here.  In this repository, most PR's have a news item that then get combined into the changelog when we release a new version.

frenzymadness:
I have to ask my employer (Red Hat) for permission to sign the CLA first.

jezdez:
Thanks for identifying the fix. Closing this as superseded by #6105, which provides an independent fix for #6091.

PR #6094, opened 2026-08-14
feat: Use cloud mirror of CRAN for performance

matthewfeickert:
### Description

* https://cloud.r-project.org/ is a cloud mirror of CRAN that auto-redirects to the optimal server based on where the user is located. Switch to using this over 'https://cran.r-project.org' in all cases.
   - c.f. https://github.com/conda-forge/r-tinkr-feedstock/pull/9#discussion_r3771229903
* Parse both CRAN mirror directory listing formats
   - https://cran.r-project.org serves Apache fancy-table directory listings while https://cloud.r-project.org serves plain pre-formatted ones, so the CRAN skeleton's `<td>`-based regexes matched nothing against the https://cloud.r-project.org mirror ("Package data.table not found" in CI). Make the listing regexes format-agnostic, matching the existing `get_available_binaries` pattern, and continue with a warning when fetching the full CRAN archive listing fails (e.g. the intermittent CloudFront `504` on the ~27k-entry `/src/contrib/Archive/` listing), which only affects archived-only packages.
* Cover CRAN directory listing parsing for both mirror formats
   - Add offline unit tests for `get_cran_index` and `get_cran_archive_versions` exercising the Apache fancy-table (https://cran.r-project.org) and plain pre-formatted (https://cloud.r-project.org) listing formats, the warn-and-continue path when the full archive listing is unavailable, and the 404 path for packages with no archive directory.
* Add corresponding news item.

### Checklist - did you ...

- [N/A] Add / update outdated documentation?

### DIFF
--- a/conda_build/variants.py
+++ b/conda_build/variants.py
@@ -54,7 +54,7 @@ DEFAULT_VARIANTS = {
         "ignore_build_only_deps",
         "extend_keys",
     ],
-    "cran_mirror": "https://cran.r-project.org",
+    "cran_mirror": "https://cloud.r-project.org",
 }
 
 # map python version to default compiler on windows, to match upstream python
--- a/conda_build/skeletons/cran.py
+++ b/conda_build/skeletons/cran.py
@@ -675,6 +675,38 @@ def get_session(output_dir, verbose=True):
     return session
 
 
+# CRAN mirrors serve directory listings in several flavours: Apache fancy
+# tables (cran.r-project.org), Apache plain pre-formatted text
+# (cloud.r-project.org) and nginx autoindex, so match the anchors alone
+# rather than relying on the surrounding <td> markup.
+LISTING_FILE = re.compile(r'<a href="([^"]+)">\1</a>')
+LISTING_DIR = re.compile(r'<a href="([^"]+)/">\1/</a>')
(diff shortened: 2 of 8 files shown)


matthewfeickert:
cc @danielnachun @mfansler for feedback from the conda-forge/r team.

conda_build/variants.py
```diff
@@ -54,7 +54,7 @@
         "ignore_build_only_deps",
         "extend_keys",
     ],
-    "cran_mirror": "https://cran.r-project.org",
+    "cran_mirror": "https://cloud.r-project.org",
```
matthewfeickert, line 57: As `cran_mirror` exists in the global pinning now, `cran_mirror` should be set only when it is not yet defined.

isuruf: These are global defaults and are set at the very beginning. Downstream users (like conda-forge) can change it in their pinnings files.

matthewfeickert: @isuruf ah so `conda-build` is in the opposite situation as `rattler-build-conda-compat` (given your comment https://github.com/conda-forge/rattler-build-conda-compat/pull/138#pullrequestreview-4958331386), correct? Which "makes sense" as `conda-build` is a tool that doesn't have any dependence on conda-forge existing, while https://github.com/conda-forge/rattler-build-conda-compat is a "rattler-build module which offer compatibility with conda-smithy".

PR #6102, opened 2026-08-21
Error out properly if v1 and v0 recipes are present

jsmolic:
Ensure conda-build will properly error out in case user tries to build a v1 recipe and has a v0 recipe file present in the same directory. Previously, the `utils.is_v1_recipe()` check would return True only if there is no v0 recipe file which coupled with the check in `main_build.py` would not error out and lead to unexpected behavior.

### Checklist - did you ...


conda_build/cli/main_build.py
```diff
@@ -574,17 +574,18 @@ def execute(args: Sequence[str] | None = None) -> int:
...
+    if n_v1_recipes > 0 and n_v0_recipes == 0:  # all are v1, proceed with rattler-build
         # check cli arguments
         parser, parsed_only_recipe = parse_args(parsed.recipe)
         command = parser.prog.split()[-1]
         check_arguments_rattler(command, parsed, parsed_only_recipe)
         # run rattler command
         return run_rattler(command, parsed, config)
 
-    if n_v1_recipes > 0:  # mixed recipe formats, error out
+    if n_v1_recipes > 0 and n_v0_recipes > 0:  # mixed recipe formats, error out
```
jaimergp:
```suggestion
    elif n_v1_recipes > 0 and n_v0_recipes > 0:  # mixed recipe formats, error out
```

jaimergp: And you may want to add an `else` to check if there were none

jsmolic: Done.

jsmolic: > And you may want to add an `else` to check if there were none

That change seems to have broken several tests so I'll revert it

conda_build/cli/main_build.py
```diff
@@ -574,22 +574,27 @@ def execute(args: Sequence[str] | None = None) -> int:
...
+    elif n_v1_recipes > 0 and n_v0_recipes > 0:  # mixed recipe formats, error out
         print(
             "Cannot process several recipe versions at the same time!", file=sys.stderr
         )
         return 1
 
+    # elif n_v1_recipes == 0 and n_v0_recipes == 0:
+    #    print("No valid recipes found!", file=sys.stderr)
+    #    return 1
+
```
schuylermartin45: Is this meant to be included/uncommented?

conda_build/utils.py
```diff
@@ -1365,9 +1365,15 @@ def find_recipe(path: str) -> str:
 def is_v1_recipe(recipe_dir: Path) -> bool:
     """Check if recipe.yaml exists"""
     recipe_dir = Path(recipe_dir)
-    return (recipe_dir / "recipe.yaml").exists() and not any(
-        (recipe_dir / meta).exists() for meta in VALID_METAS
-    )
+    return (recipe_dir / "recipe.yaml").exists()
```
schuylermartin45: `.is_file()` I think is _slightly_ safer than `exists()` in that it detects that the file exists AND it is a normal file.

I'm not that familiar what the implications are on Windows/if the check is as safe on Windows as it is with POSIX systems, so take this with a grain of salt.

schuylermartin45: In case this is ever useful, CRM has a [is_v1_recipe()](https://conda.github.io/conda-recipe-manager/_autosummary/conda_recipe_manager.parser.v0_recipe_formatter.html#conda_recipe_manager.parser.v0_recipe_formatter.V0RecipeFormatter.is_v0_recipe) function that does a little more than check the file name exists in the path. It requires parsing of the recipe file and then looks at some key differentiating structural differences between V0 and V1.

It isn't a perfect check and the parser isn't perfect, but it might be a nice thing to use in other applications. Unfortunately CRM depends on `conda-build` for a few key types/definitions so importing CRM just to do this check would cause a nasty circular dependency nobody wants to deal with.

conda_build/utils.py
```diff
@@ -1365,9 +1365,15 @@ def find_recipe(path: str) -> str:
...
+
+
+def is_v0_recipe(recipe_dir: Path) -> bool:
```
schuylermartin45, line 1371:
```suggestion
def is_v0_recipe(recipe_dir: Path) -> bool:
    """Check if meta.yaml exists"
```
I don't know what the current feelings on docstrings are in this project but this at least matches the sister function.

conda_build/cli/main_build.py
```diff
@@ -574,22 +574,27 @@ def execute(args: Sequence[str] | None = None) -> int:
     config.channel_urls = get_channel_urls(parsed.__dict__)
 
     config.verbose = not parsed.quiet or parsed.debug
-
+    n_v0_recipes = sum(1 for recipe in parsed.recipe if is_v0_recipe(recipe))
```
schuylermartin45, line 577: I'm surprised you have to handle multiple recipe files. Is this looking at the rendered recipe variants? Is that what "parsed" means in this context?

tests/cli/test_main_build.py
```diff
@@ -673,3 +673,23 @@ def test_build_v1_no_anaconda_upload(capsys) -> None:
     output = captured.out + captured.err
 
     assert "# Automatic uploading is disabled" in output
+
+
+def test_error_on_mixed_v0_v1_recipes(testing_workdir: str, capsys) -> None:
```
schuylermartin45, line 678: If the commented `elif` I mentioned is added back, we should add a test for "error if no recipe files are present".

schuylermartin45 (commented):
I have not spent a ton of time with `conda-build` and I'm not familiar with the current style requirements or current practices, but I hope these comments and thoughts are useful!

conda_build/cli/main_build.py
jsmolic: It was originally included but broke several other tests in `conda-build` so I decided to drop it, I'll remove the accidental leftover

conda_build/utils.py
schuylermartin45: I may be misremembering and CRM may only be importing `MatchSpec` from `conda` not `conda-build`. In either case, I don't think it would be a popular opinion to include CRM in conda-build. At lest not without serious planning and discussion.

jsmolic: Yeah I think I even looked at some point at using some of the functions from CRM but abandoned the idea of adding another dependency for a single or small number of functions.
That said, yes, the fact that the check is based purely on the file name isn't great. conda-build already does that in `utils.find_recipe()` so I decided to keep the same approach, but it would be nicer to have something more robust.

conda_build/cli/main_build.py
jsmolic: EDIT: scratch the comment, was looking at a different branch  ~~Hmm actually this code is not even present in my tree, not sure what is happening here, could be a bug in the GH UI?~~

conda_build/cli/main_build.py
jsmolic: This is looking at the `ArgumentParser` object after the command is invoked

tests/cli/test_main_build.py
jsmolic: As mentioned in https://github.com/conda/conda-build/pull/6102#discussion_r3876100782 this won't be included so no need to add a test

conda_build/utils.py
jsmolic: Thanks, added

conda_build/utils.py
jsmolic: Switched to `.is_file()` usage

conda_build/cli/main_build.py
schuylermartin45: I marked this as unresolved as it looks like it is still in the diff for this PR and it sounds like we'd prefer to remove the commented code(?)

jsmolic: You're absolutely right, I had forgotten to remove the commented part in my branch before updating the PR

jsmolic: This is done in f1097bc26afa2b6d98eaa4baa8e038d32aab1429

schuylermartin45 (approved):
I'm not an expert in conda-build and somebody more familiar should also approve this. But my feedback has been addressed, thanks for fixing this!

conda_build/utils.py
```diff
@@ -1363,11 +1363,18 @@ def find_recipe(path: str) -> str:
...
+    """Check if recipe.yaml exists."""
     recipe_dir = Path(recipe_dir)
-    return (recipe_dir / "recipe.yaml").exists() and not any(
-        (recipe_dir / meta).exists() for meta in VALID_METAS
-    )
+    return (recipe_dir / "recipe.yaml").is_file()
+
+
+def is_v0_recipe(recipe_dir: Path) -> bool:
+    """Check if meta.yaml exists."""
```
jaimergp: Not to be pedantic but _technically_ this will also return true if `conda.yaml` exists (or their `.yml` counterparts), as per `find_recipe`, which relies on `VALID_METAS`:

```python
VALID_METAS = ("meta.yaml", "meta.yml", "conda.yaml", "conda.yml")
```

So maybe let's rethink this docstring:

```suggestion
    """Check if meta.yaml or any of the valid v0 recipe filenames exist."""
```

Tune it to your liking, but you get the idea.

tests/cli/test_main_build.py
jaimergp, line 1: Do we have a test case for no recipe files present?

jsmolic: Added in test_build_no_recipe_files

jezdez (approved):

PR #6105, opened 2026-08-26
Handle InvalidVersion in deprecation version fallback

### LINKED ISSUE #6091: Incompatibility with the latest packaging

jezdez:
### Description

Fixes #6091.

`packaging` 26.3 raises `InvalidVersion` instead of `TypeError` when package version metadata is unavailable. Catch both exceptions while keeping the `packaging.version` import inside the fallback path used by deprecation checks.

### Checklist - did you ...

- [x] ~~Add / update outdated documentation?~~

### DIFF
--- a/conda_build/deprecations.py
+++ b/conda_build/deprecations.py
@@ -70,13 +70,13 @@ class DeprecationHandler:
         # If self._version or version could not be represented by a simple
         # tuple[int, ...], do a more elaborate version parsing and comparison.
         # Avoid this import otherwise to reduce import time for conda activate.
-        from packaging.version import parse
+        from packaging.version import InvalidVersion, parse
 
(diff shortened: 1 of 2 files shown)


jsmolic (approved):
Thanks, just happened to encounter this failure on some of my PRs, good to get a fix in!

PR #6124, opened 2026-09-07
Add `--exclude-newer` to conda-build

### LINKED ISSUE #6119: Implement `--exclude-newer` just like in conda
### Checklist

### What is the idea?

https://github.com/conda/conda/pull/15761 was recently merged in conda (but not yet released I think?). This implements [CEP 47](https://github.com/conda/ceps/pull/154).

### Why is this needed?

`--exclude-newer` is going to be very helpful in conda-build. It will allow us to rebuild a package with only packages from a given date in the past. This allows to more easily reproduce a build.

In theory, I think it will be able to set it via `.condarc` or via `CONDA_EXCLUDE_NEWER`, but I would be nice to have it has a native CLI argument too.

### What should happen?

_No response_

### Additional Context

_No response_

jezdez:
### Description

Add `--exclude-newer` to the shared build, render, and debug CLI. Apply conda's resolved policy to build, host, and test dependencies for both `meta.yaml` and `recipe.yaml` recipes, including package and channel overrides.

For `meta.yaml`, preserve the policy across context resets and configuration copies, filter the index supplied to the classic solver, and prevent solved-record caches and template environments from bypassing it. Keep unrestricted solver indexes lazy and include virtual packages when filtering an active cutoff. For `recipe.yaml`, pass the same cutoffs through rattler-build's build, build-time package test, standalone package test, and debug APIs. Keep each build's output channel available for new outputs while preserving explicit package override precedence.

Cutoff policies require conda 26.9.0 or newer, which has not yet been released. The v1 backend requires a release containing the Python API additions in prefix-dev/rattler-build#2801. Libmamba support for `meta.yaml` depends on a release containing conda/conda-libmamba-solver#905 and a fix for conda's generated sharded solver subclass losing the policy capability flags, tracked in conda/conda#15759. Older conda or rattler-build bindings report an error when an unsupported cutoff is requested.

Rattler-build currently filters on `timestamp`, while conda prefers CEP 47 `indexed_timestamp` when available. Both backends keep packages without timestamps through this integration.

Fixes #6119.
Part of conda/conda#15759.

### Checklist - did you ...

- [x] Add a file to the `news` directory ([using the template](https://github.com/conda/conda-build/blob/main/news/TEMPLATE.md)) for the next release's release notes?

### DIFF
--- a/conda_build/config.py
+++ b/conda_build/config.py
@@ -25,6 +25,7 @@ from conda.base.constants import (
 from conda.base.context import context
 from conda.utils import url_path
 
+from .exceptions import CondaBuildUserError
 from .utils import (
     get_build_folders,
     get_conda_operation_locks,
@@ -37,6 +38,8 @@ from .variants import get_default_variant
 if TYPE_CHECKING:
     from pathlib import Path
     from typing import Any, TypeVar
+
+    from conda.core.exclude_newer import ExcludeNewerPolicy
 
     T = TypeVar("T")
 
@@ -129,6 +132,7 @@ def _get_default_settings():
         Setting("anaconda_upload", context.binstar_upload),
         Setting("force_upload", True),
         Setting("channel_urls", []),
+        Setting("exclude_newer", getattr(context, "exclude_newer", None)),
         Setting("dirty", False),
         Setting("include_recipe", True),
         Setting("no_download_source", False),
@@ -300,6 +304,7 @@ class Config:
 
     def __init__(self, variant=None, **kwargs):
         super().__init__()
+        self._exclude_newer_now = time.time()
         # default variant is set in render's distribute_variants
         self.variant = variant or {}
         self.set_keys(**kwargs)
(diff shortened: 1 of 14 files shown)


jezdez:
`py-rattler-build 0.73.0` still lacks the cutoff APIs this PR needs.

Suggested path:

1. Test a commit-pinned build of prefix-dev/rattler-build#2801 with canary conda. Fail on missing APIs instead of skipping, and include an unmocked standalone package test.
2. Get the upstream change approved, merged, and released.
3. Bump `py-rattler-build` in `recipe/meta.yaml` and `tests/requirements.txt` once compatible packages reach our CI channels.

Before bumping, check `defaults` availability and Python 3.10 support. Current conda-forge 0.73.0 packages require newer Python. The separate conda/libmamba prerequisites remain documented in the PR description.

conda_build/config.py
```diff
@@ -530,6 +538,29 @@ def output_folder(self):
...
+        return ExcludeNewerPolicy.from_values(
+            self.exclude_newer,
+            context.exclude_newer_package,
+            (
+                *context.channel_settings,
+                # Newly built outputs must remain available to dependent outputs
+                # and package tests. Other local channels still obey the cutoff.
+                {"channel": url_path(self.output_folder), "exclude_newer": False},
+            ),
+            now=self._exclude_newer_now,
```
ryanskeith: I don't think the time deltas are enough to warrant a special save.  We could just pass it here.

Issue #6079, opened 2026-08-05
Discovery: Python 3.15 support


danyeaw:
### What is the idea?

Run a discovery spike to determine what is required for conda-build to support Python 3.15, and produce a short work breakdown for follow-up issues.

### Why is this needed?

Python 3.15 is in the release-candidate phase, with final release planned for October 2026. Users and packagers will expect conda to run and be tested on 3.15 soon after (ideally before the final release).

### What should happen?

- Assess conda-build on Python 3.15 (starting from available pre-releases / channel builds) and record failures or warnings that block support.
- Check readiness of the pieces we depend on: Python 3.15 packages in defaults and conda-forge, and whether key dependencies (including the solver stack) install and work under 3.15
- Coordinate timing with conda’s own Python 3.15 support work, since conda-build depends on conda. See https://github.com/conda/conda/issues/16496
- Turn findings into follow-up issues (fix failures / declare support / extend CI), reusing the shape of the 3.14 rollout.

### Additional Context

- Schedule: [PEP 790 – Python 3.15 Release Schedule](https://peps.python.org/pep-0790/)
- Prior art: Prior art: [#5912](https://github.com/conda/conda-build/issues/5912) (3.14 tarfile filters), [#5843](https://github.com/conda/conda-build/issues/5843) (free-threaded cp314t selection when targeting 3.14)

PR #6143, opened 2026-09-17
Read site-packages location from the installed python's record

### LINKED ISSUE #5563: SP_DIR vs. freethreading builds: wrong path?

chrisburr:
### Description

`get_site_packages()` derived the path from the python version and the platform, hardcoding `Lib/site-packages` on Windows. In python 3.15 conda-forge started using `lib/python/site-packages` on Windows, so `SP_DIR` pointed at a directory that does not exist and recipes relying on it silently produced empty packages.

See also https://github.com/conda-forge/cfep/blob/main/cfep-27.md

Closes #5563
See also #6079

### Checklist - did you ...


conda_build/utils.py
```diff
@@ -1006,7 +1007,29 @@ def get_stdlib_dir(prefix, py_ver):
...
+    return getattr(record, "python_site_packages_path", None) or None
+
+
 def get_site_packages(prefix, py_ver):
+    # python declares where its site-packages lives, which is the only way to get
+    # this right for layouts that are not derivable from the version alone -- e.g.
+    # free-threaded builds (lib/python3.14t/site-packages) and, per CFEP-27,
+    # python >=3.15 on Windows (lib/python/site-packages, not Lib/site-packages).
+    # https://github.com/conda-forge/cfep/blob/main/cfep-27.md
+    if site_packages_path := _get_python_site_packages_path(prefix):
```
beckermr, line 1031: It's a walrus!

jaimergp (approved):

jezdez (approved):

